PR #10, opened 2023-11-15
ci: testing...

kaspar030:


kaspar030:
found the issue, fixed in #11.

PR #13, opened 2023-11-17
ci: make matrix size more flexible

kaspar030:


kaspar030:
hmno, doesn't look better.

PR #22, opened 2023-12-07
boards: Add microbit-v2

chrysn:
Works out of the box once the minimal declarations are added, and is an easy-access board until I figure out how to do boards with bootloaders :-)


kaspar030:
Thanks! :+1:

PR #23, opened 2023-12-11
boards: Add particle-xenon (for full-image and riotboot)

chrysn:
This adds yet another board, but this time it's tricky because that board doesn't have an on-board debugger. It is commonly used in either of three modes:

* with the upstream bootloader (which uses dfu-util and an elaborate startup sequence)
* with riotboot-dfu (which uses 16KiB bootloader followed by two slots, which is a weird default for a USB based bootloader, but convenient when you also want to do OTA)
* with an external programmer (using a 2x5 pin ~50mil IDC connector)

On the long run I hope we can support the upstream bootloader, but riotboot is a) easier (no checksumming outside the header, and I can flash it and debug step through the loading process), and b) more widely spread in our ecosystem. This PR includes riotboot support.

Riotboot support is implemented using cargo flags that tell a build.rs which partitioning is used. On the long run, that code will need to be more universal; right now it's sitting in nrf52/build.rs for further experimentation.

I've successfully used this to flash a board through riotboot-dfu, but so far that's a manual process, because it involves:

```
$ laze -C examples/embassy-net-udp build --verbose -b particle-xenon-riotboot-dfu-slot0 -s probe-rs-run
$ objcopy build/bin/particle-xenon-riotboot-dfu-slot0/embassy-net-udp/embassy-net-udp.elf -Obinary out.bin
$ ${RIOT}/dist/tools/riotboot_gen_hdr/bin/genhdr generate out.bin 1 16640 256 header.bin
$ cat header.bin out.bin > total.bin
$ dfu-util -a 0 -D total.bin
```

I'm confident that either laze or probe-rs (or whatever the Rust ecosystem uses for flashing when no probe is attached) will be helpful in automating this, but first steps first and I'd appreciate a rough yes/no on the direction before continuing.

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -118,6 +118,7 @@ contexts:
       flash-riotboot:
         required_vars:
           - FLASH_OFFSET
+          - FLASH_SLOT_OFFSET
         cmd:
           - riotboot-genhdr generate ${out} $$(date +%s) $(${FLASH_OFFSET} + ${FLASH_SLOT_OFFSET}) ${FLASH_SLOT_OFFSET} - > ${out}.hdr.bin
           - ${OBJCOPY} -Obinary ${out} ${out}.bin
@@ -520,6 +521,17 @@ modules:
         FEATURES:
           - riot-rs/wifi-esp
 
+  - name: runner-riotboot-dfu
+    required_vars:
+      - FLASH_OFFSET
+      - FLASH_SLOT_OFFSET
+      - FLASH_SIZE_TOTAL
+      - FLASH_SLOT
+      - FLASH_NUM_SLOTS
+    env:
+      global:
+        CARGO_RUNNER: "${SCRIPTS}/runner-riotboot.sh\\ ${FLASH_OFFSET}\\ ${FLASH_SLOT_OFFSET}\\ ${FLASH_SIZE_TOTAL}\\ ${FLASH_SLOT}\\ ${FLASH_NUM_SLOTS}"
+
 builders:
   # host builder (for housekeeping tasks)
   - name: host
@@ -593,6 +605,38 @@ builders:
 
   - name: particle-xenon
     parent: nrf52840
+
+  - name: _particle-xenon-riotboot-dfu
+    parent: particle-xenon
+    env:
+      # Bootloader size. This is why the board is riotboot-dfu: The version
+      # without DFU needs less flash for the bootloader.
+      FLASH_OFFSET: "0x4000"
+      # Size of image header
+      FLASH_SLOT_OFFSET: "0x400"
+      # This is default for nrf52/build.rs anyway, but runner-riotbuild needs
+      # the information explicitly.
+      FLASH_NUM_SLOTS: "2"
+      # FIXME: This information ("The nrf52840 has 1024KiB of flash") is
+      # duplicated between here and src/riot-rs-boards/nrf52/build.rs.
+      FLASH_SIZE_TOTAL: "1048576"
+      CARGO_ENV:
+        - NRF52_FLASH_OFFSET=${FLASH_OFFSET}
+        - NRF52_FLASH_SLOT_OFFSET=${FLASH_SLOT_OFFSET}
+        - NRF52_FLASH_SLOT=${FLASH_SLOT}
+        - NRF52_FLASH_NUM_SLOTS=${FLASH_NUM_SLOTS}
+    selects:
+      - runner-riotboot-dfu
+
+  - name: particle-xenon-riotboot-dfu-slot0
+    parent: _particle-xenon-riotboot-dfu
+    env:
+      FLASH_SLOT: "0"
+
+  - name: particle-xenon-riotboot-dfu-slot1
+    parent: _particle-xenon-riotboot-dfu
+    env:
+      FLASH_SLOT: "1"
 
   - name: rpi-pico
     parent: rp2040
--- a/scripts/runner-riotboot.sh
+++ b/scripts/runner-riotboot.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+# Take an image built merely with the right offsets (and possibly higher leve
+# riotboot functionality such as reboot-into-bootloader), send it through
+# riotboot-genhdr, and flash it using dfu-util.
+#
+# It may be practical to split this into a build-image-from-elf and a
+# flash-image step using laze, but right now this works.
+
+set -e
+
+FLASH_OFFSET="$1"
+FLASH_SLOT_OFFSET="$2"
+FLASH_SIZE_TOTAL="$3"
+FLASH_SLOT="$4"
+FLASH_NUM_SLOTS="$5"
+BUILT_ELF="$6"
+
+IMAGE="${BUILT_ELF}.img"
+VERSION=$(date +%s)
+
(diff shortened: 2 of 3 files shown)


chrysn:
@kaspar Can we feed build.rs effects back into laze, or would that be doing things bent over backwards?

I'd guess not, then we have two paths for building riotboot images:
* Do it outside of cargo. Then all the pre-calculation ((firmware size - bootloader) / slots)) needs to happen in laze, laze will need to set those numbers to be available to build.rs, and that has the simple task of just creating a linker script of that. The header building then runs using the existing riotboot-genhdr tool, which receives the values from laze.
* Do it in cargo. laze will not do much except pass on the board selection (as I'm doing it now), and build.rs will get the extra task of also creating some binary that represents the boot loader header, and includes it through the linker setup.

I'd prefer the latter, because it stands chances of being usable also for other Rust based systems, but I'll still need to verify whether some deep-down crate can still do things to the final product (because for bootloaders such as the particle-xenon native one, we'll need to include some check sum of the application).

chrysn:
The plan I think I'd prefer is this:
* Do all the relevant calculations in the build.rs (passing on information to the compiler as needed), to ensure that the ELF file contains all information later image builders need.
* Then we can use laze to do either of two things:
  * Use a runner that makes use of all the information there (but gets little else from laze) to build an image and flash it.
  * Have some target (note to self: learn laze) to build the image for later flashing.

kaspar030:
Nice! I think some of the bootloader logic is already covered by ld-memory.

src/riot-rs-boards/nrf52/build.rs
```diff
@@ -1,5 +1,61 @@
...
+                panic!("riotboot-slot selected but no riotboot-partitioning set");
+            }
+            // FIXME: We should make this explicit
+            Partition::FullDevice
+        }
+    }
+
+    /// Given the chip's total flash size, return flash start and end of the usable flash region in
+    /// the given partition
+    fn get_rom(&self, total_rom: u64) -> (u64, u64) {
```
kaspar030, line 43: I think this is duplicating what ld-memory already provides (though via environment variables): https://github.com/kaspar030/ld-memory/blob/ee1fd420c9d0071b3b3bbcf4853a6dcf17e0b27a/lib/ld-memory/src/lib.rs#L154-L182.

There's `NRF52_FLASH_OFFSET` that offsets the whole flash area (set to bootloader size), `NRF52_FLASH_SLOTS` to set to the total number of slots, `NRF52_FLASH_SLOT` which when unset, disables the slotting, of when set to a slot number (starting at 0), calculates for that slot. There's then `NRF52_FLASH_SLOT_OFFSET` which adds another offset within the slot (set to 256 or hdr_len).

Shouldn't that cover it?

kaspar030:
> I'd guess not, then we have two paths for building riotboot images:
guessed right. :)

>     * Do it outside of cargo. Then all the pre-calculation ((firmware size - bootloader) / slots)) needs to happen in laze, laze will need to set those numbers to be available to build.rs, and that has the simple task of just creating a linker script of that. The header building then runs using the existing riotboot-genhdr tool, which receives the values from laze.

>     * Do it in cargo. laze will not do much except pass on the board selection (as I'm doing it now), and build.rs will get the extra task of also creating some binary that represents the boot loader header, and includes it through the linker setup.
> 
> 
> I'd prefer the latter, because it stands chances of being usable also for other Rust based systems, but I'll still need to verify whether some deep-down crate can still do things to the final product (because for bootloaders such as the particle-xenon native one, we'll need to include some check sum of the application).

I'm generally in favor or using cargo for everything that it actually can do without getting too hacky.

In this case, bending cargo to do more than building a .elf is getting too hacky. We could wrap the linker, or the runner, to build a binary including e.g., the boot loader header or checksum. But IMO that's just a step *after* cargo built the .elf.
Laze should have all the info (whether the bootloader should be used, which slot to build, ...), so it can pass it down to cargo and still use it later for flashing or running.

That said, laze is a bit clumsy here. For starters, there needs to be some way to add post processing recipes. Currently, the simplest way to create some binary that includes the bootloader would be a flashing task (script), but that is re-run on every invocation.
And `tasks` can only be disabled by making them depend on a variable. It should maybe possible to define them in modules (vs. only in contexts), so they go through the usual dependency&conflicts machinery.

kaspar030:
[here](https://github.com/future-proof-iot/RIOT-rs/commit/08aa0da0f71a9b1f9f7cea7a891a03a4cb3c5574) is how the existing ldmemory infra can be used to do the slotting.

chrysn:
Yes, thanks! I've had to do do some adjustments here and there (eg. calculate NRF52_FLASH_OFFSET from FLASH_SLOT_OFFSET and FLASH_OFFSET b/c SLOT_OFFSET is not known to ld-memory ... guess where https://github.com/kaspar030/ld-memory/pull/1 came from ;-) ), but I've just successfully run `laze -C examples/embassy-net-udp build -vv -b particle-xenon-riotboot-dfu-slot0 run` and the embassy network interface came up.

chrysn:
Now updated to a working version, which squashes the back-and-forth between my original version and your commit which you referenced.

I've been using it like this:

```
$ PATH=~/git/RIOT/dist/tools/riotboot_gen_hdr/bin:$PATH laze -C examples/embassy-net-udp build -vv -b particle-xenon-riotboot-dfu-slot0 run
```

Open FIXMEs (which we may address now or merge it and leave them for later):
* Can we get any better story for how to find riotboot's genhdr than asking the user to put it in the PATH?
* To do the address calculations, genhdr needs the full memory story, which was missing the total flash size (so that "shift by OFFSET, divide the rest by NUM_SLOTS, shift into the right slot and then add SLOT_OFFSET" can be calculated). Right now this is passed in as FLASH_SIZE_TOTAL, but it's duplication against the 1024 in nrf52's build.rs script; we may manage to do better.
* I'm rather naive in what can be done in laze, so I did the easiest thing and passed a few variables to the CARGO_RUNNER with a healthy serving of trial-and-error escaping. (There is **way** too much "we define this here and later text-paste it there, where it gets evaluated by the shell into a variable, which is better used with quotes because it still contains whitespace" in all of this for my taste). There may well be a clean way of prep'ing all the data in laze. Such a path would likely not go through `cargo run` any more, and I don't have an opinion yet on whether that'd be good or bad. (As long as we go through `cargo run`, we do need one script/executable that gets us all the way from "here's the elf file" to "and it's running". Given we flash and don't run, and in particular don't capture the stdout, maybe run is not the right term).

chrysn:
Concrete plan now that I need this working to work on other parts:

* I'm rebasing this onto master (without any testing).
* I'm squashing things into a separate branch that will be a new PR for programmer-only particle-xenon
* We can figure out how to do bootloader stuff best later, and apply it to more chips than just this one.

chrysn:
This is now based on https://github.com/future-proof-iot/RIOT-rs/pull/224

emmanuelsearch:
this PR is relatively "old" for this repo ;) Is closure still planned?

emmanuelsearch:
@chrysn is this PR still relevant at this stage?

emmanuelsearch:
@chrysn is this PR still relevant at this stage, or should we close it?

ROMemories:
Closing as support for the board itself itself was added in #224 and we dropped support for riotboot in #670. Feel free to reopen if/when needed.

PR #25, opened 2023-12-12
feat: port the Embassy USB HID keyboard example

ROMemories:
*This is experimental, for documentation only; should not be merged.*

This ports the [Embassy USB HID keyboard example](https://github.com/embassy-rs/embassy/blob/14f41a71b6ea9dedb4ee5b9c741fe10575772c7d/examples/nrf52840/src/bin/usb_hid_keyboard.rs) to RIOT-rs.

It highlights some remaining challenges, in particular:

- How should we expose peripherals to tasks, after some of them have been moved out of the [`Peripherals`](https://docs.embassy.dev/embassy-nrf/git/nrf52840/struct.Peripherals.html) struct during the initialization phase?
- How can we allow configuring these drivers from within user tasks?
- Can we make sure at compile time that drivers that can be used only once are actually taken by no more than one user task?


kaspar030:
This got merged with #87.

PR #30, opened 2023-12-18
refactor(riot-rs-embassy): make net feature gates more targeted

ROMemories:
We want to be able to have multiple network stacks at the same time in the future.
We will need to rename the `stack` field of `TaskArgs` to be more specific, but this is a breaking change left for a future change.

### DIFF
--- a/src/riot-rs-embassy/src/lib.rs
+++ b/src/riot-rs-embassy/src/lib.rs
@@ -15,7 +15,7 @@ pub type Task = fn(Spawner, TaskArgs);
 
 #[derive(Copy, Clone)]
 pub struct TaskArgs {
-    #[cfg(feature = "net")]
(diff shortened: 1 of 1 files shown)


src/riot-rs-embassy/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub type Task = fn(Spawner, TaskArgs);
 
 #[derive(Copy, Clone)]
 pub struct TaskArgs {
-    #[cfg(feature = "net")]
+    #[cfg(feature = "usb_ethernet")]
```
kaspar030, line 18: Hm. The idea here is that network applications can just use "stack", independent of the underlying ethernet, wifi or 6lowpan.

kaspar030: offline discussion: this is correct atm

PR #34, opened 2023-12-21
ci: increase build concurrency

kaspar030:


kaspar030:
So the setup time of each build dwarfs actual build time. I'll postpone this until the matrix gets a bit larger.

kaspar030:
Now it seems to save a couple of minutes.

PR #43, opened 2024-01-10
refactor(arch): move arch modules to their own files

ROMemories:


src/riot-rs-embassy/src/lib.rs
```diff
@@ -4,6 +4,12 @@
 
 pub mod assign_resources;
 
+#[cfg(context = "nrf52")]
```
kaspar030, line 7: We can replace this with:

```
#[cfg_attr(context = "nrf52", path = "arch/nrf52.rs")]
#[cfg_attr(context = "rp2040", path = "arch/rp2040.rs")]
pub mod arch;
```

... and then drop the others. That way we only have one line per arch plus one for the "pub mod arch", instead of four per arch.

PR #45, opened 2024-01-11
feat(assign_resources): remove the `split_resources` macro

ROMemories:
Replace `split_resources!` with a method, so that it becomes possible to use `assign_resources!` multiple times at once.


src/riot-rs-embassy/src/assign_resources.rs
```diff
@@ -57,19 +57,20 @@ macro_rules! assign_resources {
             pub type $resource_alias = peripherals::$resource_field;
         )?)*)*
 
-        #[macro_export]
-        /// `split_resources!` macro
-        macro_rules! split_resources (
-            ($p:ident) => {
-                $resources {
+        impl $resources {
+            pub fn extract_from(
```
kaspar030, line 61:
```suggestion
            pub fn take_from(
```

PR #63, opened 2024-02-01
examples/embassy-net-tcp: document tcp echo server access via telnet

bergzand:
This extends the README.md of the embassy-net-tcp example with a line or two about the running telnet server, which seems to be *the feature* of this example.

### DIFF
--- a/examples/embassy-net-tcp/README.md
+++ b/examples/embassy-net-tcp/README.md
@@ -12,4 +12,8 @@ In this folder, run
     laze build -b nrf52840dk run
 
 With the device USB cable connected, a USB ethernet device should pop up.
-RIOT-rs will reply to ping requests on 10.42.0.61.
+RIOT-rs will reply to ping requests on 10.42.0.61 and host a tcp service on
+port 1234 that will echo the input back to the client. It can be accessed with
+e.g., `telnet`:
+
+    telnet 10.42.0.61 1234


examples/embassy-net-tcp/README.md
```diff
@@ -12,4 +12,7 @@ In this folder, run
     laze build -b nrf52840dk run
 
 With the device USB cable connected, a USB ethernet device should pop up.
-RIOT-rs will reply to ping requests on 10.42.0.61.
+RIOT-rs will reply to ping requests on 10.42.0.61 and hosts a telnet server on
```
kaspar030, line 15:
```suggestion
RIOT-rs will reply to ping requests on 10.42.0.61 and host a tcp service on
```

examples/embassy-net-tcp/README.md
```diff
@@ -12,4 +12,7 @@ In this folder, run
...
+port 1234 that will echo the input back to the client:
```
kaspar030, line 16:
```suggestion
port 1234 that will echo the input back to the client. It can be accessed with e.g., `telnet`:
```

examples/embassy-net-tcp/README.md
bergzand: Thanks, should be fixed

kaspar030:
Thanks! :)

PR #64, opened 2024-02-01
riot-rs-embassy: Use locally admin MAC addresses

bergzand:
We're not Sillicon Laboratories, don't [claim](https://maclookup.app/search/result?mac=CC:CC:CC) to be a device from them


kaspar030:
Thanks!

PR #67, opened 2024-02-08
feat(nrf52-embassy): allow accessing the internal temp sensor

ROMemories:


emmanuelsearch:

PR #71, opened 2024-02-09
lib/rbi: use wrapping_sub to calculate available slots

elenaf9:
Small bug I noticed while viewing the code.

Currently, the `rbi::RingBufferIndex` panics if the more than `u8::MAX`-times elements have been added and removed again. 
The reason for this is that in `RingBufferIndex::available` the `writes` counter has wrapped over, but the `reads` counter hasn't and thus an overflow occurs in the subtraction.
The `counter_overflow_size*` tests didn't catch this because it only occurs if the `writes` counter is already 1 "ahead" of the `reads` counter and then `available` is called (either directly, of by `put`ting another element into the buffer).

This is fixed by simply using `wrapping_sub` as it is done in the other methods.

I added this check to the test, which panicked without this fix. The panic would alternatively also be visible if in said test per iteration always two items would be put and removed again to the rbi.

Lmk if I am missing something.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -820,6 +820,7 @@ dependencies = [
  "embassy-net",
  "embassy-time",
  "embedded-io-async",
+ "heapless 0.8.0",
  "riot-rs",
  "riot-rs-boards",
 ]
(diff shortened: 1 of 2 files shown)


kaspar030:
Thanks! :smile:

PR #75, opened 2024-02-14
feat(embassy): rework and simplify the way applications are defined

ROMemories:


kaspar030:
`Application` trait got kicked out ...

PR #78, opened 2024-02-14
bump embassy

kaspar030:
This bumps the embassy fork to today's upstream (14th Feb 2024). The "path" dependencies of modified crates are patched out now.


chrysn (approved):
Awesome, this makes working from own patch branches of "unrelated" embassy crates much easier.

PR #80, opened 2024-02-19
Refactor/embassy wifi cyw43

ROMemories:


kaspar030:
I'm not too happy with having two cyw43.rs files in riot-rs-embassy. Would you mind moving the board specific part back into the other (that is now in wifi)? The way it is now, it's not clear how to add support of e.g., another board with different pinout. And, there's two modules named `cyw43` (confusing to me).

Or, as a suggestion, move the board specifics into wifi/cyw43/rpi-pico-w.rs, and dispatch in wifi/cyw43.rs as we do [here](https://github.com/future-proof-iot/RIOT-rs/blob/e4470f7caa1ecd556f1929be709874f9c4f8ab31/src/riot-rs-embassy/src/lib.rs#L9) (call the module sth like cyw43::params, use rpi-pico-w.rs when builder==rpi-pico-w, etc).

kaspar030:
> it still feels weird to have board-specific code in the system; we could revisit that later.

agreed ...

kaspar030:
please squash!

PR #82, opened 2024-02-19
refactor(embassy): refactor `init_task` to ease future refactorings

ROMemories:


src/riot-rs-embassy/src/lib.rs
```diff
@@ -85,6 +75,16 @@ async fn init_task(mut peripherals: arch::OptionalPeripherals) {
 
     #[cfg(feature = "usb")]
     let mut usb_builder = {
+        #[cfg(context = "nrf52")]
```
kaspar030, line 78: can we keep this before starting the EMBASSY_TASKS?

kaspar030:
This is good to go, please squash!

PR #103, opened 2024-02-22
ci: integrate clippy

ROMemories:


.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ on:
   merge_group:
 
 jobs:
-  CI:
+  CI-build:
```
kaspar030, line 13: if we rename this here, we need to update the "needs:" in line 90 and 93

.github/workflows/main.yml
```diff
@@ -60,6 +60,29 @@ jobs:
 
         laze build --partition hash:${{ matrix.partition }} --builders microbit-v2,nrf52840dk,rpi-pico,rpi-pico-w -g
 
+  CI-lint:
+    runs-on: ubuntu-latest
+    container:
```
kaspar030, line 65: this "container:" is not needed. (it is gone in "main" already).

.github/workflows/main.yml
```diff
@@ -60,6 +60,29 @@ jobs:
...
+      image: ghcr.io/kaspar030/riot.rs-build:v0.0.2
+
+    env:
+      SCCACHE_GHA_ENABLED: "true"
+
+    steps:
+    - name: Install toolchain
+      uses: dtolnay/rust-toolchain@nightly
+      with:
+        # targets: thumbv6m-none-eabi,thumbv7m-none-eabi,thumbv7em-none-eabi
```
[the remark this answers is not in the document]
kaspar030: I don't think so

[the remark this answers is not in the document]
kaspar030: AFAIU clippy would use the default target to build, which is x86_64 on the github runners.

.github/workflows/main.yml
kaspar030: fixed

.github/workflows/main.yml
kaspar030: rebased

PR #114, opened 2024-02-23
chore: make sure all feature names are kebab case

ROMemories:
Even though there is no official convention yet, we need consistent casing and kebab case seems to be the consensus in the community.

See the following for official conventions regarding casing: https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case


kaspar030 (approved):
LGTM.

PR #115, opened 2024-02-23
ci(rustfmt): enforce rustfmt formatting

ROMemories:


kaspar030:
Re-based on #117 in #118.

PR #116, opened 2024-02-23
ci(cargo-sort): check formatting and item order of Cargo.toml files

ROMemories:


kaspar030:
I've moved this into the shared "lint" job.

PR #119, opened 2024-02-24
ci: cache PR build results based on git tree hash

kaspar030:
So currently, a branch gets built at lest three times: first on the PR, second in the merge queue and third for the master push event.

This PR adds result caching to the "CI build step" that caches positive build results.


kaspar030:
So this is hard to test I think. I'll try to re-run.

kaspar030:
So if the cache doesn't 429, it seems to work at least on PR run reruns.

PR #126, opened 2024-02-28
feat(ci/hax): extract `runqueue` to F*, lax check it

W95Psp:
As discussed earlier today, here is an initial PR that adds extraction to F* for the crate `riot-rs-runqueue` and lax-checking (which was almost for free!).

This PR requires https://github.com/hacspec/hax/pull/538 to be merged on hax' side.


src/riot-rs-runqueue/proofs/fstar/extraction/Riot_rs_runqueue.Runqueue.Clist.fst
```diff
@@ -0,0 +1,216 @@
+module Riot_rs_runqueue.Runqueue.Clist
```
kaspar030, line 1: these two .fst files, they are generated, right? they don't have to get checked in?

W95Psp: yes, you're right, those are generated, I can add them to gitignore

kaspar030:
(I've squashed out the "temp" and its revert.)

Thanks a bunch! :smile:

PR #133, opened 2024-03-04
(WIP) docs: improve docs for `riot-rs-threads`

elenaf9:
Improve API documentation for `riot-rs-threads` and `riot-rs-runqueue`. 
I'll work on next to extract the essence about scheduling in riot-rs into a markdown file in the book (most likely as _User Guide_ and _Developer Guide_?) but I thought that it might make sense to already open this as a draft PR now.

Open Questions:

- Should the user/developer guide be in the `book/` or `docs/` folder?
- I don't fully understand `Threads::flag_wait_one` - @kaspar030 could you help me here?

### DIFF
--- a/src/riot-rs-rt/src/threading.rs
+++ b/src/riot-rs-rt/src/threading.rs
@@ -3,6 +3,10 @@ use riot_rs_threads::{start_threading, THREAD_FNS};
 /// # Safety
 ///
 /// The caller must ensure that this function is only called once.
+///
+/// # Panics
+///
+/// Panics if no thread exists.
 pub unsafe fn start() -> ! {
     for thread_fn in THREAD_FNS {
         thread_fn();
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -4,7 +4,56 @@
 
 # User Guide
 
+## Multi-threading
+
+Riot-rs implements a scheduler based on fixed priorities and preemption.
+Within the same priority level, threads are scheduled cooperatively.
+
+Threading can be enabled with the `threading` feature.
+If the feature is enabled, at least one thread **must** be defined.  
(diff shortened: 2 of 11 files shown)


kaspar030:
> * I don't fully understand `Threads::flag_wait_one` - @kaspar030 could you help me here?

`Threads::flag_wait_any(mask)` returns when any of `mask` are set, and returns all of those (`set & mask`). `Threads::flag_wait_one(mask)` returns when any of `mask` are set, and returns exactly one of those (`ffs (set & mask)`), leaving the not returned active.

This is (maybe) useful if the handler would rather handle the flags one-by-one.

kaspar030:
> * Should the user/developer guide be in the `book/` or `docs/` folder?

Let's have them in the book. That's markdown after all. There are sections for user/developer docs already.

src/riot-rs-rt/src/threading.rs
```diff
@@ -3,6 +3,10 @@ use riot_rs_threads::{start_threading, THREAD_FNS};
 /// # Safety
 ///
 /// The caller must ensure that this function is only called once.
+///
+/// # Panics
+///
+/// Panics if no thread exists.
```
kaspar030, line 9: This is a bug btw. :)

kaspar030:
Maybe it makes sense to split this PR into smaller chunks - maybe one per crate and one for the book?

src/riot-rs-rt/src/threading.rs
elenaf9: Extracted into #148 :)

elenaf9:
> Maybe it makes sense to split this PR into smaller chunks - maybe one per crate and one for the book?

Good idea.

Closing this PR in favor of #149, #150 and #152

PR #134, opened 2024-03-04
threads: fix `flag_wait_all`

elenaf9:
I believe the line should check for all bits in the mask? :slightly_smiling_face:

### DIFF
--- a/src/riot-rs-threads/src/thread_flags.rs
+++ b/src/riot-rs-threads/src/thread_flags.rs
@@ -71,10 +71,9 @@ impl Threads {
 
     fn flag_wait_all(&mut self, mask: ThreadFlags) -> Option<ThreadFlags> {
         let thread = self.current().unwrap();
-        if thread.flags & mask != 0 {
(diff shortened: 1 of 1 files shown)


src/riot-rs-threads/src/thread_flags.rs
```diff
@@ -71,7 +71,7 @@ impl Threads {
 
     fn flag_wait_all(&mut self, mask: ThreadFlags) -> Option<ThreadFlags> {
         let thread = self.current().unwrap();
-        if thread.flags & mask != 0 {
+        if thread.flags & mask == mask {
             let result = thread.flags & mask;
             thread.flags &= !mask;
             Some(result)
```
kaspar030, line 77:
```suggestion
        if thread.flags & mask == mask {
            thread.flags &= !mask;
            Some(mask)
```

kaspar030:
> I believe the line should check for all bits in the mask? 🙂

Yup. :+1:

kaspar030:
LGTM, please squash!

PR #146, opened 2024-03-05
examples/embassy-http-server: some fixes for rpi-pico-w

kaspar030:
- increase stack size
- add an extra barrier so `network_stack()` doesn't panic if the stack is not done configurint (e.g., b/c device initialization is await'ing something)

marking as draft as both changes should maybe guarded.


kaspar030:
#708 fixed the last issue here.

PR #152, opened 2024-03-06
book: add sections about threading

elenaf9:
Extracted from #133.

Add sections about multi-threading / `riot-rs-threads` to user and developer guide in the book.

Couple of open questions (marked as _TODOs_) within the text. 

Happy to add additional sections/ details if you think anything is missing, or to remove stuff that might be too deep or still subject to change.


elenaf9:
Friendly ping @kaspar030, mind giving a review here? :)
Everyone else is also welcome to leave comments/ suggestions.

book/src/SUMMARY.md
```diff
@@ -5,7 +5,112 @@
...
+
+The `riot_rs_threading` crate supports basic synchronization primitives:
+- a synchronous (blocking) channel for sending data between threads (the data must implement `Copy`)
+- a lock that can be used to implement mutually exclusive access to a resource
+- a thread-flags implementation that enables threads to wait for specific flags
+  (*TODO: should the thread\_flags even be mentioned here/ are they meant to be used by users?*)
+
+### Example
+
+```rs
```
kaspar030, line 27: As asked elsewhere: this would be useful as an example. maybe just copy `examples/threading` to `examples/threading_channel`, put his code there, then use here via mdbook include (with anchor)?

(We can postpone that to a follow up)

book/src/SUMMARY.md
```diff
@@ -5,7 +5,112 @@
 
 # User Guide
 
+## Multi-threading
```
kaspar030, line 8: SUMMARY.md is supposed to contain a TOC with links. please move the content to per-chapter md files.

book/src/SUMMARY.md
```diff
@@ -5,7 +5,112 @@
...
+
+Riot-rs implements a scheduler based on fixed priorities and preemption.
```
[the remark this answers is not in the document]
elenaf9: Done in 402434e3bd066081be82364ac4e75f6a1b797adb.

book/src/SUMMARY.md
elenaf9: Done with 4726e45fe247e5ff9527acbcfa5ee6483e2829ff.

book/src/SUMMARY.md
elenaf9: Implemented in #223.

elenaf9:
> It looks to me that most of this could become rustdoc of the `riot-rs-threads` crate. Keeping the docs closer to the impl would make them more likely to get updated when the implementation changes, and allow to link to items, again preventing the docs from lagging behind in case items are removed/renamed (broken internal links are already tested in CI).

I agree. Most of these docs are extracted from the rustdocs anyway, but they could be moved into library/ module level docs in `riot-rs-threads`. However, from how I understood @emmanuelsearch and @kaspar030, their idea was to also have these separate user- and developer guides in the book.

elenaf9:
Closing this PR. 
We agreed in an out-of-band discussion that these kind of docs should be rustdocs inside the crate instead of book sections. 
I've extracted the user-guide docs in #449. 
The developer-guide docs were in fact simply a summary of our existing docs/ code-comments, so nothing new really.

Issue #157, opened 2024-03-07
[Tracking Issue] `riot-rs-threads`: add RISC-V support


elenaf9:
As part of #7, RISC-V support should be added to the `riot_rs_threads` crate.

This mostly concerns the arch specific interrupt handling and context switching logic.

# TODO 

(more items might be added while I am working on this)

- [ ] Document existing scheduler/ multi-threading logic: #152 
- [x] Make `riot_rs_threads::arch::cortex_m::sched` arch-independent: #159 
- [x] #156
- [x] Implement `threads::arch::riscv` module: #182 
    - [x] Wire up the <s>exception</s> interrupt handler
    - [x] Implement context switch
    - [x] Implement some sort of WFI if no thread is ready (see https://github.com/future-proof-iot/RIOT-rs/pull/159#discussion_r1516992568)
    - [x] Implement `setup_stack` and `start_threading`

kaspar030:
#182 closes this. Great work @elenaf9!

PR #159, opened 2024-03-07
threads: extract arch-independent scheduling logic

elenaf9:
Part of #157.

This PR:
- introduces an `Arch` trait for arch-specific logic in the scheduler 
- extracts the arch-independent scheduling logic out of the `cortex_m` module.

Details of the new `Arch` trait will probably still change once a second arch is added, but I think this is already a useful first step.
Wdyt?

Reviewing by commit might be easier; I tried to split it up a bit.

### DIFF
--- a/src/riot-rs-threads/src/arch/cortex_m.rs
+++ b/src/riot-rs-threads/src/arch/cortex_m.rs
@@ -1,3 +1,4 @@
+use super::Arch;
 use core::arch::asm;
 use core::ptr::write_volatile;
 use cortex_m::peripheral::SCB;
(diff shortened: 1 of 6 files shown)


src/riot-rs-threads/src/arch/mod.rs
```diff
@@ -1,17 +1,65 @@
+/// Arch-specific implementations for the scheduler.
+pub trait Arch {
+    const DEFAULT_THREAD_DATA: Self::ThreadData;
```
kaspar030, line 3: I guess this is not using `Default` b/c we need to be able to use this in `const` context?

kaspar030:
I think this goes into the right direction.

BTW, we'll need benchmarks. I used to test using the RIOT-c benchmarks, but we can't use them at the moment.
But the `riot-rs::rt::benchmark()` function can be used to clock how fast the scheduler is.
E.g. with sth like this: https://gist.github.com/kaspar030/bde630c06dd6f20babcaf0668d3f7e12
I was surprised, this PR is as fast as main. Zero Cost Abstractions FTW! :)

src/riot-rs-threads/src/arch/mod.rs
```diff
@@ -1,17 +1,65 @@
...
+    /// will be executed after the thread function returned.
+    fn setup_stack(stack: &mut [u8], func: usize, arg: usize) -> usize;
+
+    /// Trigger a context switch.
+    fn schedule();
+
+    /// Setup and initiate the first context switch.
+    fn start_threading(next_sp: usize);
+
+    /// Manually write the pointers to thread data into the registers, as ABI would return the values via stack.
```
kaspar030, line 24: While the functions above all make sense to have them in the trait, I'm not sure about the ones below.

`return_data_in_regs()` and `wfi()` are *very* architecture specific. (`wfi()` also overloads `cortexm::asm::wfi()`).
I know they're what's needed to keep `sched()` platform independent.
But, it might make more sense to just describe the invariants of the context switching process that's externally visible and depended on, like these two:

```rust
threads.threads[current_pid as usize].sp = old_s;
threads.current_thread = Some(next_pid);
```

... and leave the exact `sched()` machinery to the architecture.
But, let's see how this evolves once we've added RISC-V and Xtensa. :)

src/riot-rs-threads/src/scheduler.rs
```diff
@@ -0,0 +1,58 @@
+use critical_section::CriticalSection;
```
kaspar030, line 1: nit: filename doesn't fit IMO, "the scheduler" is either the whole crate, or just `RunQueue`.

src/riot-rs-threads/src/scheduler.rs
```diff
@@ -0,0 +1,58 @@
...
+            current_high_regs = None;
+        }
+
+        let next = &threads.threads[next_pid as usize];
+        let next_sp = next.sp;
+        let next_high_regs = &next.data;
+
+        // Return pointers to the old and new thread data in arch specific registers
+        // so that the exception handler save and restore the state.
+        Cpu::return_data_in_regs(current_high_regs, next_high_regs);
```
kaspar030, line 54: So this is (as it was and as it is now) obviously hacky as hell.
On the Cortex-M, it is possible to return two or four usize encoded in one u64 or one u128. [gotbolt](https://rust.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:2,endLineNumber:21,positionColumn:2,positionLineNumber:21,selectionStartColumn:2,selectionStartLineNumber:21,startColumn:2,startLineNumber:21),source:'%23!!%5Bno_std%5D%0A%0Apub+fn+fourints_tuple()+-%3E+(u32,+u32,+u32,+u32)+%7B%0A++++(0,1,2,3)%0A%7D%0A%0Apub+fn+fourints_u128()+-%3E+u128+%7B%0A++++3+%3C%3C+96+%7C+2+%3C%3C+64+%7C+1+%3C%3C+32+%7C+0%0A%7D%0A%0Apub+fn+twoints_u64()+-%3E+u64+%7B%0A++++2+%3C%3C+32+%7C+1%0A%7D%0A%0Apub+fn+threeints_u64u32()+-%3E+(u64,+u32)+%7B%0A++++(2+%3C%3C+32+%7C+1,+0)%0A%7D%0A%0Apub+fn+threeints_u32()+-%3E+(u32,+u32,+u32)+%7B%0A++++(2,+1,+0)%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1760,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'--target+thumbv7m-none-eabi',overrides:!((name:edition,value:'2021')),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.76.0+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'rustc+1.66.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.76.0+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)
Unfortunately, it looks different on riscv32.

src/riot-rs-threads/src/arch/mod.rs
elenaf9: Unfortunately yes.

src/riot-rs-threads/src/arch/mod.rs
elenaf9: Concerning `wfi`: won't we always need some way of sleeping if no thread is ready to run?

And concerning `return_data_in_regs()`: I agree that it is platform specific. I wonder if there is a cleaner way to do this if we also decide for RISC-V that we save the registers in the struct. I'll think about it a bit more, also considering what you wrote in https://github.com/future-proof-iot/RIOT-rs/pull/159#discussion_r1516868257. 
Or if it turns out to be too complicated or hacky, we still have the alternative of using the stack, which would then obsolete this method.

Either way, I'd argue that the majority of the logic in `sched` is platform independent and therefore it make sense to have it extract out of the architecture-specific stuff. Even if the essence of it are just the two lines that you wrote above.

src/riot-rs-threads/src/scheduler.rs
elenaf9: With this PR I was trying to keep all logic/ implementation details the way that they currently are and just add some abstractions.
But I agree that it is hacky, and this new method definitely doesn't help with that.

kaspar030: > With this PR I was trying to keep all logic/ implementation details the way that they currently are and just add some abstractions.

Which is really great, that way we see the impact of this abstraction (almost none).

src/riot-rs-threads/src/arch/mod.rs
kaspar030: > Concerning `wfi`: won't we always need some way of sleeping if no thread is ready to run?

Yup. But, this is version force-enabling the interrupts for a bit, within a critical section, relying on nested interrupts to preempt PendSV. Not sure any other platform will do it this way. :)

kaspar030: I'm just saying, the logic in `sched()` is both simple and quite specific. It might make sense to just copy & change for the next platform, and that would be OK, at least for a while. If the abstraction gets too messy, it might not be worth it.

elenaf9: > I"ll drop the last commit.

For reference: https://github.com/elenaf9/RIOT-rs/commit/ccfacfa4fc028d4a98ecfc992ee913af66008470 was the commit that implemented this changes and that I've now dropped again.

src/riot-rs-threads/src/scheduler.rs
elenaf9: Marking this as resolved since I reverted the change that introduced `return_data_in_regs` (see https://github.com/future-proof-iot/RIOT-rs/pull/159#discussion_r1517414361).

src/riot-rs-threads/src/threadlist.rs
```diff
@@ -37,7 +37,7 @@ impl ThreadList {
             let old_state = THREADS.with_mut_cs(cs, |mut threads| {
                 self.head = threads.thread_blocklist[head as usize].take();
                 let old_state = threads.set_state(head, ThreadState::Running);
-                arch::schedule();
+                super::schedule();
```
kaspar030, line 40: and here?

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -1,78 +1,76 @@
...
-/// |   PSR   |
-/// +---------+
-///
-/// This function sets up the stack so when the context is switched to this thread,
-/// it starts executing `func` with argument `arg`.
-/// Furthermore, it sets up the link-register with the [`crate::cleanup`] function that
-/// will be executed after the thread function returned.
-pub(crate) fn setup_stack(stack: &mut [u8], func: usize, arg: usize) -> usize {
-    let stack_start = stack.as_ptr() as usize;
+pub struct Cpu;
```
kaspar030, line 9: bikeshedding: `Cpu` ... does anyone have another idea how to name this?

kaspar030:
I think this is good to go. If possible fix the "super" vs "crate" import.

src/riot-rs-threads/src/threadlist.rs
```diff
@@ -22,7 +22,7 @@ impl ThreadList {
             threads.thread_blocklist[thread_id as usize] = self.head;
             self.head = Some(thread_id);
             threads.set_state(thread_id, state);
-            arch::schedule();
+            super::schedule();
```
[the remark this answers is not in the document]
kaspar030: (let's postpone this)

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: Agreed that CPU isn't ideal. I had `Arch` initially, but that conflicted with the trait name...

src/riot-rs-threads/src/threadlist.rs
elenaf9: done

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: alright, let's sit this one out. :)

kaspar030: > Is this per-core or per CPU? If this is per-core, then it could be (Cpu|Processor)Core (not very familiar with that part of the codebase yet, this may not make sense).

It is just an empty struct that the Arch trait is implemented on.

PR #182, opened 2024-03-14
threads: add support for RISC-V (esp32 only)

elenaf9:
Initial support for threading on RISCV.

For now this only supports the `esp32c3` and `esp32c6` boards since we depend on specific features from `esp-hal`.
Adding support for other esp boards can be done in follow up PRs. 

Most of the changes are only within the new `riscv` module, but one general change I'd like to mention:
- Setting `threads.current_id` to the very first running thread is now done within `sched` when the thread actually starts running. Before, this was done already in `start_threading`, but it caused some problems in the RISC-V implementation because then we couldn't differentiate between the thread running for the first time and `current_id == next_id`. I've tested on both arch'es and it still works with this change, so I think it should be alright?

Note: I have not tested using threading and embassy tasks together. But we also don't have any examples that tested this for cortex-m, so I'd argue that it's out of scope for this PR. Wdyt?

Part of #157.

### DIFF
--- a/src/riot-rs-threads/src/arch/riscv.rs
+++ b/src/riot-rs-threads/src/arch/riscv.rs
@@ -1,8 +1,11 @@
 use crate::{cleanup, Arch, Thread, THREADS};
-use critical_section::CriticalSection;
+#[cfg(context = "esp32c6")]
+use esp_hal::peripherals::INTPRI as SYSTEM;
+#[cfg(context = "esp32c3")]
+use esp_hal::peripherals::SYSTEM;
 use esp_hal::{
     interrupt::{self, TrapFrame},
-    peripherals::{Interrupt, SYSTEM},
+    peripherals::Interrupt,
     prelude::*,
     riscv, Cpu as EspHalCpu,
 };
--- a/src/riot-rs-threads/src/arch/riscv.rs
+++ b/src/riot-rs-threads/src/arch/riscv.rs
@@ -1,0 +1,142 @@
+use crate::{cleanup, Arch, Thread, THREADS};
+use critical_section::CriticalSection;
+use esp_hal::{
+    interrupt::{self, TrapFrame},
+    peripherals::{Interrupt, SYSTEM},
+    prelude::*,
+    riscv, Cpu as EspHalCpu,
+};
+
+pub struct Cpu;
+
+impl Arch for Cpu {
+    type ThreadData = TrapFrame;
+    const DEFAULT_THREAD_DATA: Self::ThreadData = default_trap_frame();
+
+    /// Triggers software interrupt for the context switch.
+    fn schedule() {
+        unsafe {
+            (&*SYSTEM::PTR)
+                .cpu_intr_from_cpu_3()
+                .modify(|_, w| w.cpu_intr_from_cpu_3().set_bit());
+        }
+    }
+
+    /// On RISC-V (ESP32), the stack doesn't need to be set up with any register values since
+    /// they are restored from the stored [`TrapFrame`].
+    fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: usize, arg: usize) {
+        let stack_start = stack.as_ptr() as usize;
+        // 16 byte alignment.
+        let stack_pos = (stack_start + stack.len()) & 0xFFFFFFE0;
+        // Set up PC, SP, RA and first argument for function.
+        thread.sp = stack_pos;
+        thread.data.sp = stack_pos;
+        thread.data.a0 = arg;
+        thread.data.ra = cleanup as usize;
+        thread.data.pc = func;
+    }
+
+    /// Enable and trigger the appropriate software interrupt.
+    fn start_threading() {
+        interrupt::disable(EspHalCpu::ProCpu, Interrupt::FROM_CPU_INTR3);
+        Self::schedule();
+        // TODO: handle unwrap error?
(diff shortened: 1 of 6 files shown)


src/riot-rs-threads/src/arch/riscv.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+unsafe fn sched(trap_frame: &mut TrapFrame) {
+    unsafe {
+        let cs = CriticalSection::new();
+        let next_pid = loop {
+            if let Some(pid) = (&*THREADS.as_ptr(cs)).runqueue.get_next() {
+                break pid;
+            }
+            riscv::asm::wfi();
```
kaspar030, line 121: So with the interrupt priority of this ISR set to `min()` (at enable time, line 44), interrupts with higher priority preempt this one?

If no, this `wfi` will stop the CPU until an interrupt occurs, but not actually serve it (b/c this interrupt cannot be preempted).
If yes (which is the case IIUC), you'll actually need to wrap `sched()` in a critical section.
Take a look at the [arm assembly for PendSV](https://github.com/future-proof-iot/RIOT-rs/pull/182/files#diff-92b737e75a3c97cb0e16ed3a87c5649dee59536c7a9dc3a182a16540235315b3R158-R160), it wraps the call to `sched()` in `cpsid i` / `cpsie i`, and that's why it does the (hacky) enable/disable irq around `wfi`.

kaspar030:
This looks already very nice!

src/riot-rs-threads/src/arch/riscv.rs
```diff
@@ -0,0 +1,135 @@
...
+        tp: 0,
+        sp: 0,
+        pc: 0,
+        mstatus: 0,
+        mcause: 0,
+        mtval: 0,
+    }
+}
+
+/// Copies the register state from `stc` to `dst`.
```
kaspar030, line 88:
```suggestion
/// Copies the register state from `src` to `dst`.
```

src/riot-rs-threads/src/arch/riscv.rs
kaspar030: (if `sched()` gets preempted by an isr, as soon as that ISR touches THREADS, it breaks the single &mut guarantee.)

src/riot-rs-threads/src/arch/riscv.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// It copies state from the `TrapFrame` except for CSR registers
+/// `mstatus`, `mcuase` and `mtval`.
```
kaspar030, line 91:
```suggestion
/// `mstatus`, `mcause` and `mtval`.
```

src/riot-rs-threads/src/arch/riscv.rs
```diff
@@ -0,0 +1,135 @@
...
+        // clear FROM_CPU_INTR3
+        (&*SYSTEM::PTR)
+            .cpu_intr_from_cpu_3()
+            .modify(|_, w| w.cpu_intr_from_cpu_3().clear_bit());
+
+        sched(trap_frame)
+    }
+}
+
+unsafe fn sched(trap_frame: &mut TrapFrame) {
```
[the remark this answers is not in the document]
elenaf9: Done in c948701.

src/riot-rs-threads/src/arch/riscv.rs
elenaf9: Isn't the critical section (line 116) sufficient here because the `riscv` `critical-section-single-hart` implementation disables all interrupts? I know the cortex-m implementation has both, the `cpsid`/`cpsie` calls and the critical section, but I didn't fully understand why.

Either way, I do need the enable/disable the interrupts around the `wfi`, good catch!

elenaf9: Wdyt of 7036f86? 

It's a bit different than the cortex-m implementation because I am manually releasing & acquiring the `critical_section` around the WFI instead of enabling/ disabling any interrupts. 
This makes the code slightly more complex, but has the advantage that we don't make any assumptions about the `critical_section` implementation, because we don't actually know what interrupts they are enabling/ disabling.

kaspar030: nice idea to use `acquire()` / `release()` directly. Didn't think of that when doing #187.

But here, doesn't `THREADS.with_mut()` acquire a critical section itself, causing a deadlock?

elenaf9: > nice idea to use `acquire()` / `release()` directly. Didn't think of that when doing #187.

I actually prefer your version moving the critical section inside the loop. I stole that idea now :grin:: 2c47bcf.

> But here, doesn't `THREADS.with_mut()` acquire a critical section itself, causing a deadlock?

Nested critical sections are allowed (implemented as NOP) ([ref](https://docs.rs/critical-section/latest/critical_section/fn.with.html)). I only used it because the `EnsureOnce` API requires a cs.

elenaf9:
This also supports the esp32c6 board now. 

I've merged the master without a rebase for now to include #181 but not destroy the history. I'll rebase properly before merge.

kaspar030:
Please squash & rebase!

elenaf9:
> Please squash & rebase!

Done. Sorry for the many force-pushes, I accidentally merged to commits during rebase that shouldn't have been merged.

kaspar030:
Thanks!

PR #183, opened 2024-03-14
riot-rs-threads: fix channel returning wrong values

### LINKED ISSUE #136: threads: wrong value received with `riot_rs_threads::channel`
I tried to test the usage of `riot_rs_threads::channel::Channel` in `examples/threading`:

```rs
static CHANNEL: Channel<u8> = riot_rs::thread::channel::ChanneChannel::new();

#[riot_rs::thread]
fn thread0() {
    println!("Hello from thread 0");
    CHANNEL.send(&42);
}

#[riot_rs::thread(stacksize = 4096, priority = 2)]
fn thread1() {
    println!("Hello from thread 1");
    let recv = CHANNEL.recv();
    println!(
        "The answer to the Ultimate Question of Life, the Universe, and Everything is {}.",
        recv
    );
}
```

The above code works. 

However, when I tried printing the actual thread id using `riot_rs::thread::current_pid()`, or when I removed the printing line itself, thread1 suddenly receives the value `0` through the channel.
Any idea what the reason for this could be?

kaspar030:
1. use `ptr.write(*something)` instead of `*ptr = *something`. This prevents dropping `*(ptr)`
~~2. wrap `recv()` and `try_recv()` result `MaybeUninit` in `UnsafeCell`~~
3. add `fence()` after `recv()` and `try_recv()` critical sections

Fixes #136.

### DIFF
--- a/src/riot-rs-threads/src/channel.rs
+++ b/src/riot-rs-threads/src/channel.rs
@@ -48,7 +48,7 @@ impl<T: Copy + Send> Channel<T> {
                         }
                         if let ThreadState::ChannelRxBlocked(ptr) = head_state {
                             // copy over `something`
-                            unsafe { *(ptr as *mut T) = *something };
+                            unsafe { (ptr as *mut T).write(*something) };
                         } else {
                             unreachable!("unexpected thread state");
                         }
@@ -77,7 +77,7 @@ impl<T: Copy + Send> Channel<T> {
                         }
                         if let ThreadState::ChannelRxBlocked(ptr) = head_state {
                             // copy over `something`
-                            unsafe { *(ptr as *mut T) = *something };
+                            unsafe { (ptr as *mut T).write(*something) };
                         } else {
                             unreachable!("unexpected thread state");
                         }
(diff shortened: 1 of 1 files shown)


kaspar030:
@elenaf9 could you give this a shot?

src/riot-rs-threads/src/channel.rs
```diff
@@ -92,33 +92,29 @@ impl<T: Copy + Send> Channel<T> {
     }
 
     pub fn recv(&self) -> T {
-        let mut res = MaybeUninit::uninit();
+        let uc = UnsafeCell::<MaybeUninit<T>>::new(MaybeUninit::uninit());
```
elenaf9, line 95: Why is the `UnsafeCell` needed?

elenaf9:
> @elenaf9 could you give this a shot?

Tested with variations of priorities, number of threads and sends, and I always got the right result!

Side note: should we generally add the channel usage to the `examples/threading`? I already have the changes locally and could just PR them.

src/riot-rs-threads/src/channel.rs
kaspar030: I'm not sure it is. My thinking was that I'm passing around a reference to `&res`, and I don't want the compiler to optimize around the assumption that I'm not actually modifying it in that function.

elenaf9: Based on the [docs](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html) I don't the compiler does any optimization for `MaybeUninit`:
> MaybeUninit<T> serves to enable unsafe code to deal with uninitialized data. It is a signal to the compiler indicating that the data here might not be initialized
>
> The compiler then knows to not make any incorrect assumptions or optimizations on this code.

kaspar030: I dropped the "UnsafeCell" wrapping.

PR #187, opened 2024-03-15
riot-rs-threads: cortex_m: make `sched()` less hacky

kaspar030:
This moves the previusly manual critical section into `sched()`, and uses a "u128" to pass our values to the assembly.
(this adds an instruction or two, but on nrf52840dk, this is barely measurable).

### DIFF
--- a/src/riot-rs-threads/src/arch/cortex_m.rs
+++ b/src/riot-rs-threads/src/arch/cortex_m.rs
@@ -2,7 +2,6 @@ use super::Arch;
 use core::arch::asm;
 use core::ptr::write_volatile;
 use cortex_m::peripheral::SCB;
-use critical_section::CriticalSection;
 
 use crate::{cleanup, THREADS};
 
@@ -75,9 +74,7 @@ unsafe extern "C" fn PendSV() {
     unsafe {
         asm!(
             "
-            cpsid i
             bl {sched}
-            cpsie i
             cmp r0, #0
             /* label rules:
              * - number only
@@ -108,9 +105,7 @@ unsafe extern "C" fn PendSV() {
     unsafe {
         asm!(
             "
-            cpsid i
             bl sched
-            cpsie i
             cmp r0, #0
             beq 99f
 
@@ -168,55 +163,49 @@ unsafe extern "C" fn PendSV() {
 /// This function is called in PendSV.
 // TODO: make arch independent, or move to arch
 #[no_mangle]
-unsafe fn sched() -> usize {
-    // SAFETY: interrupts are disabled by caller
-    let cs = unsafe { CriticalSection::new() };
-    let next_pid;
-
+unsafe fn sched() -> u128 {
     loop {
-        {
-            if let Some(pid) = (unsafe { &*THREADS.as_ptr(cs) }).runqueue.get_next() {
-                next_pid = pid;
-                break;
-            }
+        if let Some(res) = critical_section::with(|cs| {
+            let threads = unsafe { &mut *THREADS.as_ptr(cs) };
+            let next_pid = match threads.runqueue.get_next() {
+                Some(pid) => pid,
+                None => {
+                    cortex_m::asm::wfi();
+                    return None;
+                }
+            };
+
+            let current_high_regs;
+            if let Some(current_pid) = threads.current_pid() {
+                if next_pid == current_pid {
+                    return Some(0);
+                }
+
+                threads.threads[current_pid as usize].sp = cortex_m::register::psp::read() as usize;
+                threads.current_thread = Some(next_pid);
+
+                current_high_regs = threads.threads[current_pid as usize].data.as_ptr();
+            } else {
+                threads.current_thread = Some(next_pid);
+                current_high_regs = core::ptr::null();
+            };
+
+            let next = &threads.threads[next_pid as usize];
+            let next_sp = next.sp as usize;
+            let next_high_regs = next.data.as_ptr() as usize;
+
+            // PendSV expects these three pointers in r0, r1 and r2:
+            // r0 = &next.sp
+            // r1 = &current.high_regs
+            // r2 = &next.high_regs
+            // On Cortex-M, a u128 as return value is passed in registers r0-r3.
+            // So let's use that.
+            let res: u128 =
+                //  (r0)                     (r1)                        (r2)
+                (next_sp as u128) |  ((current_high_regs as u128) << 32) | ((next_high_regs as u128) << 64);
+            Some(res)
+        }) {
+            break res;
         }
-        //pm_set_lowest();
-        cortex_m::asm::wfi();
-        unsafe { cortex_m::interrupt::enable() };
-        cortex_m::asm::isb();
-        // pending interrupts would now get to run their ISRs
-        cortex_m::interrupt::disable();
     }
-
-    let threads = unsafe { &mut *THREADS.as_ptr(cs) };
-    let current_high_regs;
-
-    if let Some(current_pid) = threads.current_pid() {
-        if next_pid == current_pid {
-            return 0;
-        }
-        //println!("current: {} next: {}", current_pid, next_pid);
-        threads.threads[current_pid as usize].sp = cortex_m::register::psp::read() as usize;
-        threads.current_thread = Some(next_pid);
-        current_high_regs = threads.threads[current_pid as usize].data.as_ptr();
-    } else {
-        threads.current_thread = Some(next_pid);
-        current_high_regs = core::ptr::null();
-    }
-
-    let next = &threads.threads[next_pid as usize];
-    let next_sp = next.sp;
-    let next_high_regs = next.data.as_ptr();
-
-    //println!("old_sp: {:x} next.sp: {:x}", old_sp, next_sp);
-
-    // PendSV expects these three pointers in r0, r1 and r2:
-    // r1= &current.high_regs
-    // r2= &next.high_regs
-    // r0 = &next.sp (implicitly done here via return value)
-    //
-    // write to registers manually, as ABI would return the values via stack
-    unsafe { asm!("", in("r1") current_high_regs, in("r2") next_high_regs) };
-
-    next_sp
 }


src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -216,53 +212,43 @@ unsafe extern "C" fn PendSV() {
...
-    let next_pid;
-
+unsafe fn sched(old_sp: usize) -> u128 {
     loop {
-        {
-            if let Some(pid) = (&*THREADS.as_ptr(cs)).runqueue.get_next() {
-                next_pid = pid;
-                break;
+        if let Some(res) = critical_section::with(|cs| {
+            let threads = &mut *THREADS.as_ptr(cs);
```
elenaf9, line 218: Why not use `THREADS.with_mut(|mut threads| { ... } )` here?

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -216,53 +212,43 @@ unsafe extern "C" fn PendSV() {
...
+            if let Some(next_pid) = threads.runqueue.get_next() {
+                let current_high_regs = if let Some(current_pid) = threads.current_pid() {
+                    if next_pid == current_pid {
+                        return Some(0);
+                    }
+                    //println!("current: {} next: {}", current_pid, next_pid);
```
elenaf9:
```suggestion
```

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -216,53 +212,43 @@ unsafe extern "C" fn PendSV() {
...
+                    threads.threads[current_pid as usize].sp = old_sp;
+                    threads.current_thread = Some(next_pid);
+                    threads.threads[current_pid as usize].data.as_ptr()
+                } else {
+                    core::ptr::null()
```
elenaf9, line 228: I know that this is old code, but could we add a check in the asm code if r2 is `0` and skip the store commands in that case?

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -216,53 +212,43 @@ unsafe extern "C" fn PendSV() {
...
+                // r1= &current.high_regs
+                // r2= &next.high_regs
+                // On Cortex-M, a u128 as return value is passed in registers r0-r3.
+                // So let's use that.
+                let res: u128 =
+                //  (r0)                     (r1)                        (r2)
+                    (next_sp as u128) |  ((current_high_regs as u128) << 32) | ((next_high_regs as u128) << 64);
+                Some(res)
+            } else {
+                cortex_m::asm::wfi();
```
elenaf9, line 246: Shouldn't this be outside of the critical section (end of the loop, line 253), so that interrupts are enabled again?

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: Because `THREADS.with_mut()` checks for unique access within a CS, which we bypass here (if we are inside PendSV, there can be no other critical section active).

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: that's already happening, if there's not "next_sp" (r0 == 0), the save/restore is skipped.

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: It doesn't have to. "wfi" will always stall the CPU until an ISR is *pending*. This code here would then continue, end the critical section, which would then reenable ISRs and jump into them.

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -216,53 +212,42 @@ unsafe extern "C" fn PendSV() {
...
-
+unsafe fn sched(old_sp: usize) -> u128 {
     loop {
-        {
-            if let Some(pid) = (&*THREADS.as_ptr(cs)).runqueue.get_next() {
-                next_pid = pid;
-                break;
+        if let Some(res) = critical_section::with(|cs| {
+            let threads = &mut *THREADS.as_ptr(cs);
+            if let Some(next_pid) = threads.runqueue.get_next() {
```
elenaf9, line 219: Nit: You could use match here and use an early return in case of `None` to make the code more readable.

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: Ahh okay. I'd still argue that the code is easier to understand if the wfi happens below, but I don't feel strongly about this :).

kaspar030: That might actually cause a race. "critical section ends -> isr -> wfi", the wfi could block until the next isr comes in.

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: `next_sp` is 0 if the next thread equals the current thread.
I was talking about the case when `r2` is `0` because there was no current thread before.

But I just noticed that right now we never have the case that `current_thread` is `None` for cortex-m, because we set it already in `start_threading`, before the first interrupt.
But as part of #182 I am proposing to change this.

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: Doesn't `THREADS.with_mut()` do the exact thing that we are doing here as well? I.e. create a critical section, borrow the mutex and then execute the closure on it?
What am I missing?

kaspar030: Ah, I was missing something, yeah, that sounds good actually.

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: Ahh good point, I didn't think of that scenario.

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: for some reason [that](https://gist.github.com/kaspar030/0a74d4867ae9d48cfe81d56516aa265c)'s 15% slower.

elenaf9: That's weird. The only difference that I see is using a raw pointer vs mutable reference to the data in the ref cell...

If that always impacts the performance this strongly, maybe we should directly change the `EnsureOnce::with_mut` implementation to use a raw pointer?

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: how do you mean?

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: yeah. I suggest we get some measurement & testing in place before going with too many changes.

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: Like this: 
```rs
   let next_id = match threads.runqueue.get_next() {
        Some(pid) => pid,
        None => {
            cortex_m::asm::wfi();
            return None

        }
    };

    let current_high_regs = ...;

    ...

    Some(res)
```

Not a huge diff, but would get rid of one nesting layer.

kaspar030: yup

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -216,53 +212,44 @@ unsafe extern "C" fn PendSV() {
...
+            let threads = &mut *THREADS.as_ptr(cs);
+            let next_pid = match threads.runqueue.get_next() {
+                Some(pid) => pid,
+                None => {
+                    cortex_m::asm::wfi();
+                    return None;
+                }
+            };
+
+            let current_pid = threads.current_pid().unwrap();
```
elenaf9, line 227: This will panic once #182 lands. 

As described in #182, it's necessary for the RISC-V implementation that we can differentiate the case when the very first thread starts running. We can only do that if we don't set `current_pid` in the `start_threading` function and instead below once we actually switch to the thread. With that change, `current_pid` would be `None` the very first time this function runs.

Is there any reason why we shouldn't do it that way?

kaspar030: > Is there any reason why we shouldn't do it that way?

Performance. :)

Let's get measurement in place so any arguments have number. In general, spending even one instruction extra on *every* context switch, to work around needs of the *first* context switch, should be avoided.

Heck, we carried along an idle thread for years in RIOT-c to avoid testing if there's actually a thread. :)

elenaf9: But `unwrap` internally also simply uses a match to check if it's `Some` or `None`: https://doc.rust-lang.org/src/core/option.rs.html#928.

elenaf9:
Friendly ping @kaspar030. 

Is anything blocking this PR? I am currently working on an alternative implementation of #241, where these changes are relevant.

Happy to take over the PR and do the rebasing/ solving of merge conflicts.

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: fixed in e3966d2864a57a49628b9742e31ac3a9ad8f218c.

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: fixed in e3966d2864a57a49628b9742e31ac3a9ad8f218c.

kaspar030:
> Is anything blocking this PR? I am currently working on an alternative implementation of #241, where these changes are relevant.

I've rebased the PR and updated it to support the `current_thread==None` case. IMO it's good to go.
I did a quick benchmark, on nrf52840dk this uses some cycles less (230 -> 227). With only one thread, some cycles more (80 -> 86). I'd be fine with those numbers as we'll be touching / optimizing this anyways.

src/riot-rs-threads/src/arch/cortex_m.rs
kaspar030: ... and, postpone this?

elenaf9 (approved):
LGTM. Thanks!

src/riot-rs-threads/src/arch/cortex_m.rs
elenaf9: > yeah. I suggest we get some measurement & testing in place before going with too many changes.

> ... and, postpone this?

Yes, sounds good.

PR #190, opened 2024-03-15
docs(riot-rs): document Cargo features aimed at users

ROMemories:
It currently looks like this on the root page of `riot-rs`:

Screenshot_20240315_164042


kaspar030:
LGTM!

PR #191, opened 2024-03-15
fix(embassy): automatically enable `time` when using `net`

ROMemories:


kaspar030 (approved):

kaspar030:
This needs a rebase so it includes the toolchain pin.

PR #202, opened 2024-03-19
doc: a note about networking in `examples/README.md`

kaspar030:
Closes https://github.com/future-proof-iot/RIOT-rs/issues/199.

*edit* not all examples are listed, those that are not will be moved to `tests/` once #189 is in.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,45 @@
+# Examples
+
+## Overview
+
+This directory contains example applications that showcase how to use RIOT-rs.
+
+- [benchmark/](./benchmark): how to use `benchmark()`
+- [embassy-net-udp/](./embassy-net-udp): UDP echo example
+- [embassy-net-tcp/](./embassy-net-tcp): TCP echo example
+- [embassy-http-server/](./embassy-http-server): HTTP server example
+- [embassy-usb-keyboard/](./embassy-usb-keyboard): USB HID example
+- [hello-world/](./hello-world): a classic
+- [minimal/](./minimal): minimized to the max RIOT-rs config
+- [threading/](./threading): how to start and use preemptively scheduled threads
+
+## Networking
+
+Some examples configure networking. By default, they will listen on a static
+IPv4 address. Be sure to configure the host computer end of the network
+accordingly.
+
+The default IPv4 address the test examples will configure is `10.42.0.61`.
+You can configure the other end like this:
+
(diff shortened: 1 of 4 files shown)


examples/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Networking
+
+Some examples configure networking. By default, they will listen on a static
+IPv4 address. Be sure to configure the host computer end of the network
+accordingly.
+
+The default IPv4 address the test examples will configure is `10.42.0.61`.
+You can configure the other end like this:
+
+    $ ip address add 10.42.0.1/24 dev <DEVICE>
```
emmanuelsearch, line 25: ip addr ?

examples/README.md
```diff
@@ -0,0 +1,40 @@
...
+    $ ip link set up dev <DEVICE>
+
+Replace `<DEVICE>` with the name of the used network device.
```
emmanuelsearch, line 28: for real beginners, maybe add here something like 

- "to find out the name of your interface you can use a command such as ifconfig". 
- "to double-check that the address has indeed be added, you can use: ip addr show dev <DEVICE>"

examples/embassy-net-udp/README.md
```diff
@@ -13,3 +13,5 @@ In this folder, run
 
 With the device USB cable connected, a USB ethernet device should pop up.
 RIOT-rs will reply to ping requests on 10.42.0.61.
+
```
emmanuelsearch, line 16: not sure if this is absolutely needed but is the ping -I <DEV> option needed here ? (If so add this precision for beginners?)

emmanuelsearch (approved):

examples/README.md
kaspar030: 'address' is actually the normal subcommand name. do you mean we should use a shorter version? The shortest would be just `ip a a ...`.

examples/embassy-net-udp/README.md
kaspar030: `-I` shouldn't be needed?

examples/README.md
kaspar030: Added. And replaced "DEVICE" with "INTERFACE" where appropriate.

emmanuelsearch: perfect

examples/README.md
emmanuelsearch: OK, let's keep it like that ;)

PR #205, opened 2024-03-21
feat: provide Default impls where relevant

ROMemories:
This makes clippy much happier.

*edited to trigger PR dep check*


kaspar030 (changes requested):
ok, lets "allow" the missing runqueue Default for now

kaspar030 (approved):

PR #212, opened 2024-03-25
build(workspace): don't include all directories in `src`

ROMemories:
Workspace crates should be listed explicitly instead.


kaspar030:
I guess this changes Cargo.lock and you'll have to check that in. Nice, I wasn't aware that file gets checked now. :smile:

PR #215, opened 2024-03-25
ci: name the dependency check workflow

ROMemories:


.github/workflows/dependencies-action.yml
```diff
@@ -1,3 +1,5 @@
+name: Dependency check
```
kaspar030, line 1: With the job below named "Check Dependencies", wouldn't this show up as "Dependency check: Check Dependencies"?

How about we call the workflow "PR dependencies" and the job just "Check" ?

kaspar030:
LGTM. (this probably needs updating the branch protection)

PR #216, opened 2024-03-26
riot-rs-rt: silence "dead_code" lints when building dummy

kaspar030:


kaspar030:
(I think the "benchmark" warning is fixed with #208)

PR #224, opened 2024-04-08
boards: add particle-xenon support

chrysn:
https://github.com/future-proof-iot/RIOT-rs/pull/23 was slowed down by the attempt to get the bootloader support in properly together with the board.

Let's have the board in first, which might be easy enough for a quickdraw even, and do bootloaders later.


laze-project.yml
```diff
@@ -590,6 +591,11 @@ builders:
   - name: microbit-v2
     parent: nrf52840
 
+  - name: particle-xenon
+    parent: nrf52840
+    env:
+      BOARD: particle-xenon
```
kaspar030, line 597: This is not necessary, it is set in the `riot-rs` context for all builders.

kaspar030:
Apart from the redundant 'BOARD' setting, LGTM.
(the PR shows the amount of boilerplate currently needed for supporting a board. :/ )

laze-project.yml
chrysn: Thanks, forcepushed. (I think it sneaked in as part of fixing the missing function in the riot-ros-boards/src/lib.rs dispatch.)

kaspar030:
locally tested a full build with `laze build --global --builder particle-xenon`.

PR #227, opened 2024-04-08
docs(threads): document threading-relating panics

ROMemories:
Document possible panics related to invalid `ThreadId`s provided from outside the crate.
Disable indexing lints in `riot-rs-runqueue` and `riot-rs-threads` to reduce noise and because panic-freedom is expected to be proven by hax or possible panics are documented.

depends on #219

### DIFF
--- a/src/riot-rs-threads/src/lib.rs
+++ b/src/riot-rs-threads/src/lib.rs
@@ -2,6 +2,9 @@
 #![feature(inline_const)]
 #![feature(naked_functions)]
 #![feature(used_with_arg)]
+// Disable indexing lints for now, possible panics are documented or rely on internally-enforced
+// invariants
+#![allow(clippy::indexing_slicing)]
 
 use critical_section::CriticalSection;
 
@@ -140,6 +143,10 @@ impl Threads {
     ///
     /// This function handles adding/ removing the thread to the Runqueue depending
     /// on its previous or new state.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `pid` is >= [`THREADS_NUMOF`].
     pub(crate) fn set_state(&mut self, pid: ThreadId, state: ThreadState) -> ThreadState {
         let thread = &mut self.threads[pid as usize];
         let old_state = thread.state;
@@ -153,6 +160,7 @@ impl Threads {
         old_state
     }
 
+    /// Returns the state of a thread.
     pub fn get_state(&self, thread_id: ThreadId) -> Option<ThreadState> {
         if self.is_valid_pid(thread_id) {
             Some(self.threads[thread_id as usize].state)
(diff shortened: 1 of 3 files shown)


src/riot-rs-threads/src/lib.rs
```diff
@@ -128,6 +131,10 @@ impl Threads {
     }
 
     /// Checks if a thread with valid state exists for this `thread_id`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `thread_id` is >= [`THREADS_NUMOF`].
```
elenaf9, line 137: Doesn't it return `false` in that case, without panicking?

src/riot-rs-threads/src/lib.rs
```diff
@@ -153,6 +164,11 @@ impl Threads {
         old_state
     }
 
+    /// Returns the state of a thread.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `thread_id` is >= [`THREADS_NUMOF`].
```
elenaf9, line 171: See above. Because we check here `is_valid_pid`, I don't think a panic can happen.

src/riot-rs-threads/src/lib.rs
```diff
@@ -259,6 +279,10 @@ pub fn current_pid() -> Option<ThreadId> {
 }
 
 /// Checks if a given [`ThreadId`] is valid
+///
+/// # Panics
+///
+/// Panics if `thread_id` is >= [`THREADS_NUMOF`].
```
elenaf9, line 285: Same as above.

src/riot-rs-threads/src/lib.rs
```diff
@@ -246,6 +262,10 @@ pub unsafe fn thread_create_raw(func: usize, arg: usize, stack: &mut [u8], prio:
 ///
 /// Returns `None` if `thread_id` is out of bound or no thread with
 /// valid state exists.
+///
+/// # Panics
+///
+/// Panics if `thread_id` is >= [`THREADS_NUMOF`].
```
elenaf9, line 268: Same here. `get_state` can't panic.

src/riot-rs-threads/src/lib.rs
```diff
@@ -300,6 +324,10 @@ pub fn sleep() {
 /// Wakes up a thread and adds it to the runqueue.
 ///
 /// Returns `false` if no paused thread exists for `thread_id`.
+///
+/// # Panics
+///
+/// Panics if `thread_id` is >= [`THREADS_NUMOF`].
```
elenaf9, line 330: See comment above on `get_state`.  
The call to `set_state` could panic, but because we only do that if `get_state` returned a valid value, `wakeup` itself won't panic.

elenaf9 (changes requested):
I think a most of these functions can't panic, because `is_valid_pid` doesn't panic.
Lmk if I am missing something.

elenaf9 (approved):

kaspar030:
LGTM

PR #240, opened 2024-04-09
ci(rustdoc): lint all crates for rustdoc errors

ROMemories:
`cargo rustdoc` only generates docs for the root crate, not its dependencies, `cargo doc` does.

depends on #257

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -138,8 +138,8 @@ jobs:
         with:
           args: --verbose --locked --features no-boards -p riot-rs -p riot-rs-boards -p riot-rs-buildinfo -p riot-rs-chips -p riot-rs-debug -p riot-rs-embassy -p riot-rs-macros -p riot-rs-random -p riot-rs-rt -p riot-rs-utils
 
-      - name: "rustdoc"
-        run: cargo rustdoc -p riot-rs --features no-boards,bench,threading,random,csprng,hwrng -- -D warnings
+      - name: rustdoc
+        run: RUSTDOCFLAGS='-D warnings' cargo doc -p riot-rs --features no-boards,bench,threading,random,csprng,hwrng
 
       - name: rustfmt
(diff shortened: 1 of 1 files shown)


chrysn:
Unless we can have a `--all-features` that actually works, there will be broken links in some feature configurations. If we merge this, our options are to
* just have no links in some places where there could be some
*  `#[cfgattr(feature = "...", doc("Some doc with [link]"))]` / `#[cfgattr(not(feature = "...", doc("Some doc without link))]` all links (and that doesn't even work everywhere, esp. not in feature documentation)

I suggest we either delay that, or tell it to accept broken links, or ensure that `--all-features` works and apply that.

chrysn:
(Manually moving a comment from @ROMemories over from https://github.com/future-proof-iot/RIOT-rs/pull/251#issuecomment-2051951335)

> I'd say it makes sense to enable all features that are not mutually exclusive, it's just annoying because it's not factored out, but we can live with that for now. Also they do also need to be enabled when actually building the docs (not just when checking for broken links), otherwise the links will actually be broken in the rendered docs.

It's Rust features, they are supposed to not be mutually exclusive :-)

They don't have to typecheck and build without warnings, but they can doc without warnings. This example is not pretty (wouldn't want to do it precisely like that), but it shows that not all warnings that happen at check time with --all-features also trigger the rustdoc checks:

```rust
#[cfg(feature = "a")]
mod af {
    pub fn f() {}
}

#[cfg(feature = "b")]
mod bf {
    pub fn f() {}
}

// At check time, these warn.
pub use af::*;
pub use bf::*;
```

kaspar030:
> So our policy regarding links in rustdoc is to always add links when items are in scope, enable the relevant features when
> generating the documentation to have as many items in scope as possible, and, when it is not possible, to not link to
> items.

Sounds reasonable to me.

LGTM.

PR #249, opened 2024-04-11
bump embassy to for-riot-rs-100424

kaspar030:
This includes stm32 `OptionalPeripherals` re-export.


kaspar030:
Offline ACK from Romain.

PR #251, opened 2024-04-11
Add random number generator infrastructure

chrysn:
This is <del>a first jab at</del> supporting OS level RNG, <del>based on the CoAP branch</del> because that needs it.

<del>For a first coarse review please consider the latest two commits; if that first stage is passed, I'll rebase it to be usable without CoAP.</del>

Closes: https://github.com/future-proof-iot/RIOT-rs/issues/231

### DIFF
--- a/examples/random/src/main.rs
+++ b/examples/random/src/main.rs
@@ -8,7 +8,7 @@ use riot_rs::debug::println;
 #[riot_rs::thread(autostart)]
 fn main() {
     use rand::Rng as _;
-    let mut rng = riot_rs::random::get_rng();
+    let mut rng = riot_rs::random::fast_rng();
 
     let value = rng.gen_range(1..=6);
 
--- a/examples/random/src/main.rs
+++ b/examples/random/src/main.rs
@@ -1,0 +1,16 @@
+#![no_main]
+#![no_std]
+#![feature(type_alias_impl_trait)]
+#![feature(used_with_arg)]
+
+use riot_rs::debug::println;
+
+#[riot_rs::thread(autostart)]
+fn main() {
+    use rand::Rng as _;
+    let mut rng = riot_rs_random::get_rng();
+
+    let value = rng.gen_range(1..=6);
+
+    println!("The random value of this round is {}.", value);
+}
--- a/examples/random/src/main.rs
+++ b/examples/random/src/main.rs
@@ -8,7 +8,7 @@ use riot_rs::debug::println;
 #[riot_rs::thread(autostart)]
 fn main() {
     use rand::Rng as _;
-    let mut rng = riot_rs_random::get_rng();
+    let mut rng = riot_rs::random::get_rng();
 
     let value = rng.gen_range(1..=6);
 
--- a/src/riot-rs-random/Cargo.toml
+++ b/src/riot-rs-random/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "riot-rs-random"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+rand_core = "0.6.4"
+# Not making it optional during development b/c we don't have a small or fast one
+rand_chacha = { version = "0.3.1", default-features = false }
+embassy-sync.workspace = true
+
+[lints]
+workspace = true
+
+[features]
+## If set, the provided main RNG is also a cryptographically secure pseudo
+## random number generator (CSPRNG) and thusly annotated by the
+## rand_core::CryptoRngCore trait.
+main-is-csprng = []
--- a/src/riot-rs-random/Cargo.toml
+++ b/src/riot-rs-random/Cargo.toml
@@ -7,15 +7,16 @@ repository.workspace = true
 
 [dependencies]
 rand_core = "0.6.4"
-# Not making it optional during development b/c we don't have a small or fast one
-rand_chacha = { version = "0.3.1", default-features = false }
+
 embassy-sync.workspace = true
+
+rand_pcg = "0.3.1"
+rand_chacha = { version = "0.3.1", default-features = false, optional = true }
 
 [lints]
 workspace = true
 
 [features]
-## If set, the provided main RNG is also a cryptographically secure pseudo
-## random number generator (CSPRNG) and thusly annotated by the
-## rand_core::CryptoRngCore trait.
-main-is-csprng = []
+## If set, the one global RNG is also a cryptographically secure pseudo
+## random number generator (CSPRNG), and thus, a `CryptoRng` can be produced.
+csprng = ["dep:rand_chacha"]
--- a/src/riot-rs-random/src/lib.rs
+++ b/src/riot-rs-random/src/lib.rs
@@ -52,7 +52,10 @@ pub(crate) type SelectedRng = rand_chacha::ChaCha20Rng;
 /// depend on a known fast RNG and seed it from the OS's RNG, but preferably there should be a
 /// single configured implementation used across applications for ROM optimization, even if it
 /// turns out to be beneficial to have a thread or task local RNG around in RAM for speed).
-pub struct Rng;
+pub struct Rng {
+    // Make the type not Send
+    _private: core::marker::PhantomData<*const ()>,
+}
 
 impl Rng {
     fn with<R>(&mut self, action: impl FnOnce(&mut SelectedRng) -> R) -> R {
@@ -112,5 +115,7 @@ pub fn construct_rng(hwrng: impl RngCore) {
 /// macros.
 #[inline]
 pub fn get_rng() -> Rng {
-    Rng
+    Rng {
+        _private: Default::default(),
+    }
 }
--- a/src/riot-rs-random/src/lib.rs
+++ b/src/riot-rs-random/src/lib.rs
@@ -1,0 +1,116 @@
+//! Provides a seeded random number generator depending on RIOT-rs's configuration.
+//!
+//! The module provides a single function for use by applications, [`get_rng()`], produces an owned
+//! RNG of type [`Rng`] to the application.
+//!
+//! The crate abstracts over multiple aspects of RNGs:
+//! * Where do we take a valid seed for the RNG from?
+//! * What's the type of RNG that we take along?
+//! * Can some of those be shared?
+//!   (This may need an API change, but non-Send tasks could very well use an executor-local RNG
+//!   behind a shared reference to not all carry a full RNG state for each user)
+//!
+//! No matter the choices taken (eventually through the application's setup), all is hidden behind
+//! a main [`Rng`] type, which depending on the feature `main-is-csprng` also implements
+//! [`rand_core::CryptoRng`]
+//!
+//! Before accessing the RNG, it needs to be initialized through the [`construct_rng()`'] function.
+//! This is taken care of by the `riot-rs-embassy` initialization functions. Applications can
+//! ensure that this has happened by depending on the laze feature `rng`.
+//!
+//! ---
+//!
+//! Currently, this provides very little choice, and little fanciness: It (more or less
+//! arbitrarily) uses the [rand_chacha::ChaCha20Rng] generator, seeds uses different copies of that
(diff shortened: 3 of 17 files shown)


src/riot-rs-random/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn try_fill_bytes(&mut self, buf: &mut [u8]) -> Result<(), rand_core::Error> {
+        self.inner.try_fill_bytes(buf)
+    }
+}
+
+#[cfg(feature = "main-is-csprng")]
+mod main_is_csprng {
+    impl rand_core::CryptoRng for super::Rng {}
+
+    fn generate_inner_type() -> super::SelectedRng {
```
kaspar030, line 83: I don't really understand what these two functions do.

kaspar030:
This looks pretty good already!

I'm not sure if I agree on `get_rng()` returning either an csprng or noncs-prng, the former enabled globally through a feature.
I'd expect the crypto rngs to have much larger state, so just enabling the feature might cause components that don't need it carry around that extra state.

I'd rather go with `get_rng() -> impl RngCore` and `get_crypto_rng() -> impl CryptoRngCore`.

If I understand the PR correctly, it suggests to make applications that don't require CryptoRngCore to take the Rng that falls out of `get_rng()` and use that to initialize a "fast" rng? Shouldn't that be unified / be provided?

That said, I actually don't know the state size of e.g., the used chacha rng. The insecure but fast RNGs use an uint32 or two as state.

PR wise, a simple example application just printing a random number, showcasing how to use riot-rs-random, would be great. :)

src/riot-rs-random/src/lib.rs
chrysn: Will be documented: These typecheck that indeed SelectedRng implements CryptoRng. (If we had something like `#[derive(CryptoRng)]`, that'd be checked automatically, but as we just add an impl, I want this typechecked).

To be improved in the docs.

chrysn:
Thanks for the feedback. As all that is on the Rust code size, I'll take it the rest of the execution (own crate, interaction with riot-rs-embassy) is good enough, and start factoring that out.

---

ad get_rng / crypto_rng, this may need a split along another axis as well: So far, we both keep an Rng globally, and hand out Rng instances.

For the global one, if we want to hand out CSPRNGs, we'll need the one-global-that-has-been-well-seeded to be CSPRNG as well (or attackers will observe our organically breathing LED, unwind its random state to when the CSPRNG seed was extracted, and boom). So that one will need to be as-good-as-needed.

For the handouts, we might even experiment a bit more, starting with something simple. (Finally we can just do that, and attach a `#[deprecated]` to something and run a semver check). The option space is not just "hand out a CSPRNG or hand out a fast RNG", but also "hand out a ZST that locks the global RNG and pulls from there", with various granularities of "global". Having a ZST RNG would be quite convenient, eg. for EDHOC where it's copied over into every state. This gives us a different kind of slow -- locking and contention overhead, independent of the per-byte cost.

chrysn:
f851313419941b6c369f8c80d412bf06ae1e2a36 is precisely the state that was in when you reviewed (just rebased onto main); fixes addressing your comments will come on top of that.

The simple ones are addressed, left over are:

* [ ] Provide an example application (that'd also make building easy -- so far, latest changes are not even build tested)
* [ ] Make a concrete suggestion for the get_crypto_rng topic

kaspar030:
So I did some simple benchmarking to get an idea of where the cycles go.

```Rust
fn main() {
    use rand::Rng as _;

    match riot_rs::bench::benchmark(1000, || {
        let mut rng = riot_rs_random::get_rng();
        //        let value = rng.gen_range(1..=6);
        core::hint::black_box(rng);
        //        core::hint::black_box(value);
    }) {
// ...
}
```

On nrf52840dk, the `get_rng()` alone takes 7 cycles, `get_rng()` and `gen_range(1..=6)` take 366 cycles.
I'll now try to get some numbers of a "fast" rng, and some insight on the size of the states of our rng options.

chrysn:
Given that in the latest version get_rng is a no-op (returns a ZST), 7 cycles is a lot already :-)

In the meantime, I'll fix the remaining complains of CI.

chrysn:
Summarizing @kaspar030 from chat: Let's not have one that is floppy. Fast and CSPRNG. For flash impact and cycles, let's measure.

kaspar030:
> Given that in the latest version get_rng is a no-op (returns a ZST), 7 cycles is a lot already :-)

Even a single 'nop' takes more than one cycle. :) with those low numbers, the iteration overhead becomes visible I guess.

chrysn:
ad nop / inline:  should be a complete no-op (not even emitting nop instructions). I've added an `#[inline]` on the current implementation, but getting the RNG should not be on the hot path, so it's more of a code size thing.

examples/random/src/main.rs
```diff
@@ -0,0 +1,16 @@
...
+#![no_std]
+#![feature(type_alias_impl_trait)]
+#![feature(used_with_arg)]
+
+use riot_rs::debug::println;
+
+#[riot_rs::thread(autostart)]
+fn main() {
+    use rand::Rng as _;
+    let mut rng = riot_rs_random::get_rng();
```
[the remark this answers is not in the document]
chrysn: Just that I wasn't aware that this is what we do with all the riot_rs_* modules.

Should the riot_rs crate then have a feature "random", and build-fail if that was requested but not provided for through laze enabled backends?

kaspar030:
> Does adding `#[inline]` on the relevant functions changes the benchmarks? Functions are not inline across crate boundaries by default, AFAIK.

LTO changes this, with LTO enabled (which it is in riot-rs) everything gets inlined if the compiler thinks its worth it.

kaspar030:
> Summarizing @kaspar030 from chat: Let's not have one that is floppy. Fast and CSPRNG. For flash impact and cycles, let's measure.

I lost some time trying to get oorandom into RngCore, not seeing the obvious Pcg32 from the rand crates family. Well, using the latter is a one line change.

The same cycle benchmark as above but using `Pcg32` takes 83 cycles (vs. ~350 with ChaCha20).

`size` output, first is Pcg32, second is ChaCha20:
```
  28168     908    3460   32536    7f18 build/bin/nrf52840dk/random/random.elf
  31988     908    3756   36652    8f2c build/bin/nrf52840dk/random/random.elf
```

I'll think about conclusions over lunch. :)

examples/random/src/main.rs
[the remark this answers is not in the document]
chrysn: Reexported now as a module, conditional on the presence of a laze module that now shares the name.

I've tried to find a good middle ground between duplication in Rust, Laze and Cargo: If the dependencies in Laze are ignored, things probably still build, but fails at runtime for lack of proper initialization. (Sounds like a good path to me -- of course I could add a check that breaks it build-time, but that'd be duplicated code that may not age well.)

src/riot-rs/Cargo.toml
```diff
@@ -44,6 +45,10 @@ threading = [
 ## Enables support for timeouts in the internal executor---required to use
 ## `embassy_time::Timer`.
 time = ["riot-rs-embassy/time"]
+## Enables the riot-rs-random module.
```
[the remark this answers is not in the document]
chrysn: Should have been riot-rs::random, but more generally, see other comment on where laze-initialized features are documented.

src/riot-rs/Cargo.toml
```diff
@@ -44,6 +45,10 @@ threading = [
 ## Enables support for timeouts in the internal executor---required to use
 ## `embassy_time::Timer`.
 time = ["riot-rs-embassy/time"]
+## Enables the riot-rs::random module.
```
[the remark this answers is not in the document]
chrysn: Pushed. Let's talk details of how to deduplicate between here and laze (or how to generally document features that are laze enabled) after this is in.

chrysn:
@kaspar030, is there any verdict on the RNGs you'd like to have in? (I've looked at the oorandom and pcg32 crates you mentioned, neither implements rand_core, so they're tedious.)

If not, can we start with this and refine the API as we go?

kaspar030:
> @kaspar030, is there any verdict on the RNGs you'd like to have in? (I've looked at the oorandom and pcg32 crates you mentioned, neither implements rand_core, so they're tedious.)

`Pcg32` from the `rand_pcg` crate implements the rand_core traits I think. I could literally just plug it in as `SelectedRng`. Let's use that as "fast and small but insecure" for now.

chrysn:
Latest addition has both a "fast" and a "secure" RNG, both through constructors; the fast is currently owned (and has state), the CSPRNG is shared (and locking is surely way faster than the actual crypto operations).

The example/random now flashes 24KiB, whereas building with csprng it flashes 28KiB (because unless *something* requests the CSPRNG, the global one is also a fast one).

I'm not fully happy with the outcome, as I like having a "give me anything that's cheap on flash" option, and would trust users to make their requirements clear (there *can* be a "give me fast" API even when the "whatever" RNG promises neither), for if we don't accept that we'll wind up with spacebar heating -- but it works for me. Once the inevitable CI nits are fixed, I think this can be ready.

src/riot-rs-random/src/lib.rs
```diff
@@ -1,91 +1,127 @@
 //! Provides a seeded random number generator depending on RIOT-rs's configuration.
 //!
-//! The module provides a single function for use by applications, [`get_rng()`], produces an owned
-//! RNG of type [`Rng`] to the application.
+//! The module provides a functions for use by applications, [`fast_rng()`] and [`crypto_rng()`],
+//! which porduce owned types that provide the [`rand_core::RngCore`] and
```
kaspar030, line 4:
```suggestion
//! which produce owned types that provide the [`rand_core::RngCore`] and
```

chrysn: Thx, absorbed.

src/riot-rs-random/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+rand_pcg = "0.3.1"
+rand_chacha = { version = "0.3.1", default-features = false, optional = true }
+
+[lints]
+workspace = true
+
+[features]
+## If set, the one global RNG is also a cryptographically secure pseudo
+## random number generator (CSPRNG), and thus, a `CryptoRng` can be produced.
+csprng = ["rand_chacha"]
```
kaspar030, line 22:
```suggestion
csprng = ["dep:rand_chacha"]
```

src/riot-rs-random/src/lib.rs
```diff
@@ -0,0 +1,160 @@
+//! Provides a seeded random number generator depending on RIOT-rs's configuration.
+//!
+//! The module provides a functions for use by applications, [`fast_rng()`] and [`crypto_rng()`],
```
kaspar030, line 3:
```suggestion
//! The module provides functions for use by applications, [`fast_rng()`] and [`crypto_rng()`],
```

chrysn:
Just preparing a few doc fixes (nothing links related so far). Can't have links that are just conditional on features easily (I don't think we want to go the `#[cfgattr(feature = "...", doc("Some doc with [link]"))]` / `#[cfgattr(not(feature = "...", doc("Some doc without link))]` route), so rather just no links, or can we just ensure the right features are on when we build with warnings-as-errors?

src/riot-rs/Cargo.toml
```diff
@@ -44,6 +45,10 @@ threading = [
 ## Enables support for timeouts in the internal executor---required to use
 ## `embassy_time::Timer`.
 time = ["riot-rs-embassy/time"]
+## Enables the riot-rs-random module.
+random = ["riot-rs-random"]
+## Enables seeding the random number generator from hardware.
```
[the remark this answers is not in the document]
kaspar030: Down the line I guess we'll teach laze to read Cargo.tomls, and extract features from there. But not in this PR...

kaspar030:
> so rather just no links, or can we just ensure the right features are on when we build with warnings-as-errors?

So many things to think about. :/ I think for now, drop the linking.

kaspar030:
LGTM.

PR #252, opened 2024-04-11
fix(macro): fix the `group_peripherals!()` macro

ROMemories:


kaspar030:
I see an added comma, but this compiled before w/o error on the only user we have. Could you say a word or two on what this fixes?

PR #255, opened 2024-04-15
add defmt support

kaspar030:
This PR adds initial defmt support.

On Cortex-M, this currently co-exists with `rtt-target` through `defmt-rtt-target`. Output is line-wise intermingled, seems like probe-rs gets data from the buffers in practically arbitrary order.

Depends on #294.

TODO:

- [x] make `rtt-target` and `defmt` coexist => use (updated) `defmt-rtt-target`
- [x] figure out why `DEFMT_LOG` changes don't propagate => `sccache` seems to be the issue
- [x] add example (in book)
- [x] wire up defmt (or log macros) for esp
- [x] add some documentation (in book)

~~This is in *very* early stages.~~

~~Currently, I'm trying to figure out if the previous `println!()` can be supported together with defmt, as defmt-rtt conflicts with rtt-target.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -276,9 +276,9 @@ dependencies = [
 
 [[package]]
 name = "bytemuck"
-version = "1.16.0"
+version = "1.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78834c15cb5d5efe3452d58b1e8ba890dd62d21907f867f383358198e56ebca5"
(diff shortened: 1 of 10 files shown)


elenaf9:
I just tested the command that was described in the tooling chapter of the book, and didn't get any output. 
What could I be doing wrong?
I am testing on the `espressif-esp32-c6-devkitc-1`, exact command: 
```sh
$ DEFMT_LOG=info laze build -C examples/hello-world-async/ -b espressif-esp32-c6-devkitc-1 --select defmt run
```

PR #256, opened 2024-04-15
refactor(rng): move hwrng to the nrf arch module

ROMemories:
This allows to enable doc generation for the `hwrng` Cargo feature, to fix broken links in the docs.

depends on #258

PR #257, opened 2024-04-15
ci(random): update the CI for the `riot-rs-random` crate

ROMemories:
Enable required features to fix broken links in the docs, and run clippy on the crate.

depends on #256

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Build rustdoc docs
         run: |
-          cargo doc -p riot-rs --features no-boards,bench,threading,random
+          cargo doc -p riot-rs --features no-boards,bench,threading,random,csprng,hwrng
           echo "<meta http-equiv=\"refresh\" content=\"0; url=riot_rs\">" > target/doc/index.html
           mkdir -p ./_site/dev/docs/api && mv target/doc/* ./_site/dev/docs/api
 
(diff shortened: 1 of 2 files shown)


src/riot-rs-embassy/src/arch/nrf52.rs
```diff
@@ -59,9 +59,30 @@ pub mod usb {
 
 #[cfg(feature = "hwrng")]
 pub mod hwrng {
+    use crate::arch;
+
     embassy_nrf::bind_interrupts!(pub struct Irqs {
         RNG => embassy_nrf::rng::InterruptHandler<embassy_nrf::peripherals::RNG>;
     });
+
```
chrysn, line 67: I think that with this change the Irqs struct can become non-pub.

src/riot-rs-embassy/src/arch/nrf52.rs
```diff
@@ -59,9 +59,30 @@ pub mod usb {
...
+                    peripherals
+                        .RNG
+                        // We don't even have to take it out, just use it to seed the RNG
+                        .as_mut()
+                        .expect("RNG has not been previously used"),
+                    arch::hwrng::Irqs,
+                );
+
+                riot_rs_random::construct_rng(rng);
+            } else if #[cfg(context = "riot-rs")] {
```
chrysn, line 81: Why would this ever not be the case?

kaspar030: when the build is not initiated by laze, e.g., just doing workspace-level `cargo  test ...`.

chrysn (approved):

PR #258, opened 2024-04-15
refactor(random): rewrite a static assert checking for an impl

ROMemories:
This verifies that `SelectedRng` is indeed a CSRNG using a shorter, arguably easier to read construct. It is also easier to make clippy happy about it.


kaspar030:
lgtm

PR #260, opened 2024-04-15
docs(random): fix a few typos and minor docs issues

ROMemories:
The second parameter of `*fill*()` functions has to be named `dest` because the documentation of these is provided on the trait and names that parameter that.

depends on #257


chrysn:
I take it the PR is only about the latest commit, and builds on the other current PRs on the same files?

chrysn (approved):

PR #261, opened 2024-04-16
fix(threads): require that references passed to a thread be 'static

ROMemories:
References passed to a thread must be valid for its entire lifetime.


kaspar030:
yup, lgtm.

PR #262, opened 2024-04-16
riot-rs-threads: always set `threads.current_thread` in `sched()`

### LINKED ISSUE #148: rt: panic if the `threading` feature is enabled but no threads exists
If the `threading` feature is enabled but the user didn't define any threads with the `#[thread]` macro (or manually), the startup panics.

**Details**

With the feature enabled, the runtime  starts all `THREAD_FNS` (where threads are created and added to the runqueue) and then the threading itself:
https://github.com/future-proof-iot/RIOT-rs/blob/af8d78dd2c31ad95cfb11446e8d1f6bdd8ec60bc/src/riot-rs-rt/src/threading.rs#L6-L14

`start_threading` panics if there is no thread in the runqueue:
https://github.com/future-proof-iot/RIOT-rs/blob/af8d78dd2c31ad95cfb11446e8d1f6bdd8ec60bc/src/riot-rs-threads/src/lib.rs#L150-L154

Came up in https://github.com/future-proof-iot/RIOT-rs/pull/133#discussion_r1511728662. Lmk if I am missing something.

kaspar030:
This is taken from #182: 

```
Setting threads.current_id to the very first running thread is now done within sched when the thread actually starts running. Before, this was done already in start_threading, but it caused some problems in the RISC-V implementation because then we couldn't differentiate between the thread running for the first time and current_id == next_id. I've tested on both arch'es and it still works with this change, so I think it should be alright?
```

This changes makes the number of cycles needed for one context switch go up from 254 to 259 (according to tests/benchmarks/bench_sched_yield on nrf52840dk). With the added line in `sched()` *after* the loop (instead of putting it in both cases), it'd go up to 261 cycles.

Fixes #148.

### DIFF
--- a/src/riot-rs-threads/src/arch/cortex_m.rs
+++ b/src/riot-rs-threads/src/arch/cortex_m.rs
@@ -259,6 +259,7 @@ unsafe fn sched(old_sp: usize) -> usize {
         threads.current_thread = Some(next_pid);
         current_high_regs = threads.threads[current_pid as usize].data.as_ptr();
     } else {
+        threads.current_thread = Some(next_pid);
         current_high_regs = core::ptr::null();
     }
 
--- a/src/riot-rs-threads/src/lib.rs
+++ b/src/riot-rs-threads/src/lib.rs
@@ -187,7 +187,6 @@ pub unsafe fn start_threading() {
     let cs = unsafe { CriticalSection::new() };
     let next_sp = THREADS.with_mut_cs(cs, |mut threads| {
         let next_pid = threads.runqueue.get_next().unwrap();
-        threads.current_thread = Some(next_pid);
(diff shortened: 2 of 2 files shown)


elenaf9 (approved):
Seems like the compiler previously optimized the `else` path away, which is now not possible anymore. I played around a bit, and when doing (on main) `current_high_regs = core::hint::black_box(core::ptr::null())` in the `else` path it also takes 259 cycles.

Either way,  as you also said in our out-of-band discussion @kaspar030, with 2% I think this is acceptable.

>  With the added line in sched() after the loop (instead of putting it in both cases), it'd go up to 261 cycles.

I kinda dislike that we now set it in both branches, but agree that performance should take priority over readability.

kaspar030:
Thanks for taking a look!

PR #264, opened 2024-04-18
refactor(arch): rename the nrf52 module to nrf

ROMemories:
This module actually relates to the entire nRF family, including nRF51, nRF52, and nRF53.


bergzand:
Bikeshedding here: How about nrf5x? Maybe the nrf61 will be risc-v based :smile:

kaspar030:
> Bikeshedding here: How about nrf5x? Maybe the nrf61 will be risc-v based 😄

We're aligning with `embassy-nrf` here ...

kaspar030:
LGTM.

PR #267, opened 2024-04-18
refactor(arch): split the rp2040 arch into multiple submodule files

ROMemories:
Split the `rp2040` arch to make it more manageable when it keeps growing.

depends on #266


kaspar030:
LGTM.

PR #268, opened 2024-04-18
refactor(arch): split the dummy arch into multiple submodule files

ROMemories:
Split the `dummy` arch to make it more manageable when it keeps growing.

depends on #267


kaspar030:
LGTM.

PR #269, opened 2024-04-19
cleanup: drop use of patched nrf-pacs

kaspar030:
Not needed anymore since the switch to embassy.


kaspar030:
I remember this being a compile error issue, and this still compiles now without the patched pacs. So I expect this to not cause more issue. I did run a couple of tests locally anyways (hello-world and random on nrf52840dk).

PR #270, opened 2024-04-19
feat(laze): expose `builder` as cfg context

ROMemories:
As builders are contexts, it makes sense to expose them as such as well.

We'll need this for the board setup file, so we don't have to treat the builder (i.e., the board name) separately.


kaspar030:
LGTM.

PR #272, opened 2024-04-24
feat(arch): re-export Embassy gpio modules

ROMemories:
The gpio modules need to be re-exported under `arch`, so that we can access the relevant one when codegening sensor initialization.


kaspar030:
LGTM.

PR #274, opened 2024-04-24
examples/usb-keyboard: Add nrf5340dk support

chrysn:
Pinout taken from the board's back side.

---

I appreciate the choice of keystrokes sent. I may submit a patch later to s/t/u/ -- after all, we're in constrained environments here.


kaspar030:
Tested, working. Thanks!

PR #278, opened 2024-04-30
stdio/rtt: Don't silently drop println statements

chrysn:
This still silently truncates prints, but during printf debugging, those are still understood more easily than skipped lines.

---

We're probably migrating to something more defmt anyway, but this fixes a repeated issue I've been running into when I thought my code is in weird branches just because the printf("Stage 1 {:?}", ...) silently didn't do anything at all.

(At some point we'll probably have similar high-level stdio discussions as in riot-c regarding whether to block and what to do in ISRs, but this is just a small step here and shouldn't touch the controversial topics).

### DIFF
--- a/src/riot-rs-debug/src/lib.rs
+++ b/src/riot-rs-debug/src/lib.rs
@@ -30,7 +30,7 @@ mod backend {
     pub use rtt_target::rprint as print;
     pub use rtt_target::rprintln as println;
     pub fn init() {
-        rtt_target::rtt_init_print!();
+        rtt_target::rtt_init_print!(NoBlockTrim);
     }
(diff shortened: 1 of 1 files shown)


kaspar030:
So this makes the output always blocking? would that stall applications when there's no debugger attached?

kaspar030:
> So this makes the output always blocking? would that stall applications when there's no debugger attached?

ah no, that would be `BlockIfFull`.

Somehow I was under the impression that probe-rs actually switches to blocking mode if it is attached, but I can't find a reference anymore.

anyhow, trimming is better than fully skipping. so, lgtm.

PR #279, opened 2024-04-30
refactor: enable `clippy::pedantic` at workspace-level

ROMemories:
Make the required changes to make clippy happy.

Depends on #886.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,4 +101,6 @@ unused_must_use = "deny"
 indexing_slicing = "warn"
 missing_safety_doc = { level = "warn" }
 pedantic = "deny"
-module_name_repetitions = "allow"       # Part of pedantic
+# Part of pedantic but not relevant to us, mostly because we sometimes erase
+# module names when re-exporting types
+module_name_repetitions = "allow"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,3 +100,5 @@ unused_must_use = "deny"
 # Clippy lints are documented here: https://rust-lang.github.io/rust-clippy/master/index.html
 indexing_slicing = "warn"
 missing_safety_doc = { level = "warn" }
+pedantic = "deny"
+module_name_repetitions = "allow" # Part of pedantic
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,4 +101,4 @@ unused_must_use = "deny"
 indexing_slicing = "warn"
 missing_safety_doc = { level = "warn" }
 pedantic = "deny"
-module_name_repetitions = "allow" # Part of pedantic
+module_name_repetitions = "allow"       # Part of pedantic
--- a/src/riot-rs-embassy/src/arch/dummy/executor.rs
+++ b/src/riot-rs-embassy/src/arch/dummy/executor.rs
@@ -6,6 +6,7 @@ pub struct Executor;
 
 impl Executor {
     #[allow(clippy::new_without_default)]
+    #[must_use]
     pub const fn new() -> Self {
         // Actually return a value instead of marking it unimplemented like other dummy
         // functions, because this function is const and is thus run during compilation
@@ -16,6 +17,7 @@ impl Executor {
         unimplemented!();
     }
 
+    #[must_use]
     pub fn spawner(&self) -> Spawner {
         unimplemented!();
     }
@@ -24,9 +26,15 @@ impl Executor {
 pub struct Spawner;
 
 impl Spawner {
-    #[allow(clippy::result_unit_err)]
+    #[allow(
+        clippy::result_unit_err,
+        clippy::needless_pass_by_value,
+        clippy::missing_errors_doc,
+        reason = "dummy implementation"
+    )]
     pub fn spawn<S>(&self, _token: SpawnToken<S>) -> Result<(), ()> {
         unimplemented!();
     }
+    #[allow(clippy::needless_pass_by_value, reason = "dummy implementation")]
     pub fn must_spawn<S>(&self, _token: SpawnToken<S>) {}
 }
--- a/src/riot-rs-embassy/src/arch/dummy/mod.rs
+++ b/src/riot-rs-embassy/src/arch/dummy/mod.rs
@@ -28,6 +28,8 @@ impl From<Peripherals> for OptionalPeripherals {
 #[derive(Default)]
 pub struct Config;
 
+#[allow(clippy::needless_pass_by_value, reason = "dummy implementation")]
+#[must_use]
 pub fn init(_config: Config) -> OptionalPeripherals {
     unimplemented!();
 }
--- a/src/riot-rs-embassy/src/delegate.rs
+++ b/src/riot-rs-embassy/src/delegate.rs
@@ -31,7 +31,7 @@ use crate::sendcell::SendCell;
 /// }
 /// ```
 ///
-/// TODO: this is a PoC implementation.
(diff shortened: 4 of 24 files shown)


Cargo.toml
```diff
@@ -100,3 +100,5 @@ unused_must_use = "deny"
 # Clippy lints are documented here: https://rust-lang.github.io/rust-clippy/master/index.html
 indexing_slicing = "warn"
 missing_safety_doc = { level = "warn" }
+pedantic = "deny"
+module_name_repetitions = "allow" # Part of pedantic
```
chrysn: Is this temporary because of too large a fall-out or deliberately excluded?

src/riot-rs-embassy/src/arch/dummy/executor.rs
```diff
@@ -24,9 +26,15 @@ impl Executor {
 pub struct Spawner;
 
 impl Spawner {
-    #[allow(clippy::result_unit_err)]
+    #[allow(
+        clippy::result_unit_err,
+        clippy::needless_pass_by_value,
```
chrysn, line 31: This is also Clippy subtly telling us that these should really be implementing a trait, even if only one type implementing the trait is ever available on a system. (But disabling the lints is a reasonable step at this point).

src/riot-rs-rt/src/lib.rs
```diff
@@ -53,7 +53,9 @@ static ISR_STACK: [u8; ISR_STACKSIZE] = [0u8; ISR_STACKSIZE];
 fn panic(_info: &core::panic::PanicInfo) -> ! {
     #[cfg(not(feature = "silent-panic"))]
     {
-        println!("panic: {}\n", _info);
+        #[allow(clippy::used_underscore_binding)]
```
chrysn, line 56: I think the better resolution here is to s/_info/info and allow the unused variable (eg. by `let _ =`'ing it)

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -70,7 +71,12 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
     }
 
     fn ffs(val: usize) -> u32 {
-        USIZE_BITS as u32 - val.leading_zeros()
+        #[allow(
+            clippy::cast_possible_truncation,
+            reason = "all targeted platforms have 32-bit wide pointers at most"
+        )]
+        let ffs = USIZE_BITS as u32 - val.leading_zeros();
```
chrysn, line 78: "All targeted platforms' bit widths fit in a u32" (they'd even fit in an u8, because 32 < 255)

Also, why not `const { USIZE_BITS.try_into().expect("All targeted platforms' bit widths fit in a u32) }`? (OK, that doesn't work *yet* without some nightly-only features ... we'd need it to be const so that clippy doesn't complain about a run-time exception).

src/riot-rs-threads/src/lib.rs
```diff
@@ -118,11 +122,16 @@ impl Threads {
...
 
-    /// Returns an unused ThreadId / Thread slot.
+    /// Returns an unused [`ThreadId`] / [`Thread`] slot.
     fn get_unused(&mut self) -> Option<(&mut Thread, ThreadId)> {
         for i in 0..THREADS_NUMOF {
             if self.threads[i].state == ThreadState::Invalid {
-                return Some((&mut self.threads[i], i as ThreadId));
+                #[allow(
+                    clippy::cast_possible_truncation,
+                    reason = "all iteration indices are valid `ThreadId`s"
```
chrysn, line 131: This might be a nice opportunity for refactoring threads to provide an iterator with typed `.enumerate()`.

Cargo.toml
chrysn: Fine with me.

Cargo.toml
```diff
@@ -100,3 +100,5 @@ unused_must_use = "deny"
 # Clippy lints are documented here: https://rust-lang.github.io/rust-clippy/master/index.html
 indexing_slicing = "warn"
 missing_safety_doc = { level = "warn" }
+pedantic = "deny"
+module_name_repetitions = "allow"       # Part of pedantic
```
chrysn, line 104:
```suggestion
module_name_repetitions = "allow"       # Part of pedantic, but due to our frequent reexports it is not quite applicable
```

Cargo.toml
```diff
@@ -100,3 +100,7 @@ unused_must_use = "deny"
 # Clippy lints are documented here: https://rust-lang.github.io/rust-clippy/master/index.html
 indexing_slicing = "warn"
 missing_safety_doc = { level = "warn" }
+pedantic = "deny"
+# Part of pedantic but not relevant to us, mostly because we sometimes erase
+# module names when re-exporting types
+module_name_repetitions = "allow"
```
elenaf9, line 106: We could reexport them with an alias that includes the module-name again. 
But not a strong opinion.

src/riot-rs-embassy/src/arch/dummy/mod.rs
```diff
@@ -28,6 +28,8 @@ impl From<Peripherals> for OptionalPeripherals {
 #[derive(Default)]
 pub struct Config;
 
+#[allow(clippy::needless_pass_by_value, reason = "dummy implementation")]
+#[must_use]
```
elenaf9, line 32: What do you think about adding `clippy::must_use_candidate` to the allowed list in the workspace? 
- according to the lint description: "_this lint just shows places where you could add the attribute._"  and it has many false positives[^1]
- the `#[must_use]` annotation creates a lot of noise in the code and makes it difficult to distinguish the cases where it is actually important that a value is used

[^1]: https://rust-lang.github.io/rust-clippy/master/index.html#/must_use_candidate

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -129,6 +140,7 @@ mod clist {
             self.tail[rq as usize] == Self::sentinel()
         }
 
+        #[allow(clippy::missing_panics_doc, reason = "only used internally")]
```
elenaf9, line 143: I don't think that fact that something is only used internally is a good reason to not add panic documentation. 
We also have them on other internal methods. Was there a specific reason for not adding something here?

elenaf9:
```suggestion
        /// # Panics
        /// 
        /// Panics if `n` >= `N_THREADS` or `rq` >= `N_QUEUES`. 
```

Should probably also be added on `Runqueue::add`, which is public.

> especially if that would help document `sentinel()`

`sentinel` is the largest value that a `u8` can represent (=255). Right now we simply rely on the fact that `N_THREADS` < 255 throughout the whole module (independently of this method here). This is included in the _Assumptions_ list in the `RunQueue` docs.

src/riot-rs-embassy/src/delegate.rs
```diff
@@ -76,6 +78,11 @@ impl<T> Delegate<T> {
         //   This function waits for the `self.send` signal, uses the dereferenced only inside the
         //   closure, then signals `self.reply`
         //   => the mutable reference is never used more than once
+        // NOTE(no-panic):
```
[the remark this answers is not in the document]
kaspar030: I think it is apart from the TODO. :/

emmanuelsearch:

ROMemories:
Superseded by #948.

PR #285, opened 2024-05-03
riot-rs-embassy: improve `SWI` selection

kaspar030:
This PR tries to make the incoming wave of different SWI choices more managable.

- add `executor_swi!(NAME)`-macro (reduces code duplication, see e.g., `arch/nrf/mod.rs`)

- adds logic to `build.rs` that, if set, generates a `swi.rs` that basically just contains `executor_swi!(NAME)` with `NAME` taken from the env variable `CONFIG_SWI`. This will be used in #237.

### DIFF
--- a/src/riot-rs-embassy/build.rs
+++ b/src/riot-rs-embassy/build.rs
@@ -1,0 +1,27 @@
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+
+    // handle CONFIG_SWI
+    {
+        let dest_path = Path::new(&out_dir).join("swi.rs");
+        if let Ok(var) = env::var("CONFIG_SWI") {
+            fs::write(
+                &dest_path,
+                format!("crate::executor_swi!({});\n", var).as_bytes(),
+            )
+            .expect("write failed");
+        } else {
+            fs::write(
+                &dest_path,
+                b"compile_error!(\"swi.rs included but CONFIG_SWI not set!\");\n",
+            )
+            .expect("write failed");
+        }
+
+        println!("cargo::rerun-if-env-changed=CONFIG_SWI");
+    }
+}
--- a/src/riot-rs-embassy/src/executor_swi.rs
+++ b/src/riot-rs-embassy/src/executor_swi.rs
@@ -1,0 +1,42 @@
+//! Use and create executor ISR from string name.
+//!
+//! This module provides [`executor_swi`], a macro that generates the interrupt
+//! that polls the executor.
+//!
+
+/// Create swi interrupt routine and import from string interrupt name.
+///
+/// The macro turns this:
+///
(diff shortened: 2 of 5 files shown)


kaspar030:
> generates a `swi.rs` that basically just contains `executor_swi!(NAME)` with `NAME` taken from the env variable `CONFIG_SWI`.

I didn't find a way to do this directly in Rust. There might be one using proc macros, but I dug quite far, IMO the build.rs solution is preferable. The difficulty lies in `env!()` spitting out a string literal, but we want to use the value in a `use foo;` statement.

src/riot-rs-embassy/src/executor_swi.rs
```diff
@@ -0,0 +1,36 @@
...
+///
+/// Note: this expects the `interrupt` to be present (e.g., "used") and that it contains the ISR
+/// type.
+#[macro_export]
+macro_rules! executor_swi {
+    ($swi:ident) => {
+        pub use interrupt::$swi as SWI;
+        #[interrupt]
+        unsafe fn $swi() {
+            unsafe { crate::EXECUTOR.on_interrupt() }
```
chrysn, line 33: This could take one of the SAFETY comments that are removed in the later commits.

Suggestion, extended from nrf:
```
    // SAFETY:
    // - As required, it is called from an ISR
    // - The interrupt is enabled by start(), thus this is not called before start.
    //   (This macro just adds "only enable it after starting the executor" to the
    //   requirements of the unsafe interrupt starting; the safe start() function
    //    trusts the user to pass the right number.)
```

chrysn (approved):
Looks like a good simplification; see comment inline.

I'm not super happy with how (as I've learned as part of reviewing this) InterruptExecutor::start as a safe function trusts the user to pass in the right interrupt, but that's nothing to fix in this PR.

src/riot-rs-embassy/src/executor_swi.rs
kaspar030: Ok! I was debating (with myself :smile:) whether such a comment would be needed in generated code.
Squashed right in.

src/riot-rs-embassy/build.rs
```diff
@@ -0,0 +1,27 @@
...
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+
+    // handle CONFIG_SWI
+    {
+        let dest_path = Path::new(&out_dir).join("swi.rs");
+        if let Ok(var) = env::var("CONFIG_SWI") {
+            fs::write(
+                &dest_path,
+                format!("crate::executor_swi!({});\n", var).as_bytes(),
```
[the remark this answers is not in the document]
chrysn: Generally, an attacker in control of the execution environment already owns the build process and host machine.

If we want to introduce a check, [syn::Ident](https://docs.rs/syn/latest/syn/struct.Ident.html) can be parsed from a string without making arbitrary validity choices.

PR #286, opened 2024-05-03
threads: restrict visibility

elenaf9:
Make internal types in `riot-rs-threads` private.
Especially `THREADS`/ `Threads` doesn't need to be public

### DIFF
--- a/src/riot-rs-threads/src/lib.rs
+++ b/src/riot-rs-threads/src/lib.rs
@@ -310,6 +310,12 @@ pub fn wakeup(thread_id: ThreadId) -> bool {
     })
 }
 
+/// Returns the size of the internal structure that holds the
+/// a thread's data.
+pub fn thread_struct_size() -> usize {
+    core::mem::size_of::<Thread>()
+}
+
 #[cfg(test)]
 mod tests {
     #[test]
--- a/src/riot-rs-threads/src/lib.rs
+++ b/src/riot-rs-threads/src/lib.rs
@@ -6,9 +6,6 @@
 // invariants
 #![allow(clippy::indexing_slicing)]
 
-use riot_rs_runqueue::RunQueue;
-pub use riot_rs_runqueue::{RunqueueId, ThreadId};
-
 mod arch;
 mod ensure_once;
 mod thread;
@@ -18,13 +15,13 @@ pub mod channel;
 pub mod lock;
 pub mod thread_flags;
 
-pub use arch::schedule;
-pub use thread::{Thread, ThreadState};
+pub use riot_rs_runqueue::{RunqueueId, ThreadId};
 pub use thread_flags as flags;
-pub use threadlist::ThreadList;
 
-use arch::{Arch, Cpu, ThreadData};
+use arch::{schedule, Arch, Cpu, ThreadData};
 use ensure_once::EnsureOnce;
+use riot_rs_runqueue::RunQueue;
+use thread::{Thread, ThreadState};
 
 /// a global defining the number of possible priority levels
 pub const SCHED_PRIO_LEVELS: usize = 12;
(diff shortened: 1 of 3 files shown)


src/riot-rs-threads/src/lib.rs
```diff
@@ -247,14 +244,6 @@ pub unsafe fn thread_create_raw(
...
 }
 
-/// Returns the [`ThreadState`] for this `thread_id`.
-///
-/// Returns `None` if `thread_id` is out of bound or no thread with
-/// valid state exists.
-pub fn get_state(thread_id: ThreadId) -> Option<ThreadState> {
-    THREADS.with(|threads| threads.get_state(thread_id))
-}
-
```
elenaf9, line 257: Removed because it's not safe to expose `ThreadState`. The variants `ChannelRxBlocked(usize)` and `ChannelTxBlocked(usize)` include a pointer to the location from/ to where read/ write, which a user shouldn't access.

elenaf9:
`examples/core-sizes` fails with this because it tries to print the size of `Thread`.
 Do we really need this example/ does the user need to know how big `Thread` is? If so, I guess we could make `Threads` itself public again, but keep all methods private. But imo that would be confusing for the user.

kaspar030:
> `examples/core-sizes` fails with this because it tries to print the size of `Thread`. Do we really need this example/ does the user need to know how big `Thread` is? If so, I guess we could make `Threads` itself public again, but keep all methods private. But imo that would be confusing for the user.

That's used for keeping track, and comparing these sizes to other implementations. IMO it does make sense to somehow get the data out. I'd be fine with some hack (e.g., in riot-rs-threads, a `pub fn()` that returns the size. Dropping it completely will just make us hack it back in when we're writing the next paper/thesis...

kaspar030:
I'm a bit surprised that basically nothing of the API stays public. Might that be because we're not really using the scheduler (only for those three simple applications)?

elenaf9:
> I'm a bit surprised that basically nothing of the API stays public. Might that be because we're not really using the scheduler (only for those three simple applications)?

We still have the public functions:
- `thread_create`, `thread_create_noarg`, `thread_create_raw`
- `current_pid`, `is_valid_pid`
- `yield_same`, `sleep`, `wakeup`

They internally use the methods from `Threads` that are now private. The only _functionality_ that is now not available anymore is:
- `get_state`: see my review comment above
- `schedule`: simply triggering the scheduler doesn't make sense. Only if something in the runqueue changed, which is done through the above public methods.

elenaf9:
> > `examples/core-sizes` fails with this because it tries to print the size of `Thread`. Do we really need this example/ does the user need to know how big `Thread` is? If so, I guess we could make `Threads` itself public again, but keep all methods private. But imo that would be confusing for the user.
> 
> That's used for keeping track, and comparing these sizes to other implementations. IMO it does make sense to somehow get the data out. I'd be fine with some hack (e.g., in riot-rs-threads, a `pub fn()` that returns the size). Dropping it completely will just make us hack it back in when we're writing the next paper/thesis...

:+1: done in 4277371.

src/riot-rs-threads/src/lib.rs
```diff
@@ -321,6 +321,12 @@ pub fn wakeup(thread_id: ThreadId) -> bool {
     })
 }
 
+/// Returns the size of the internal structure that holds the
+/// a thread's data.
+pub fn thread_struct_size() -> usize {
```
elenaf9, line 326: Name this `tcb_size`? `Thread` is basically the TCB, right?

kaspar030: (maybe keep this out of the docs so people won't rely on it. :smile:)

kaspar030:
LGTM. (feel free to rename the struct size function, I don't mind either way.)

PR #287, opened 2024-05-06
laze: change defaults around `run`

### LINKED ISSUE #200: expressif-esp32-c6-devkitc-1: first flash issue
after installing the toolchain as suggested by an initial buidling error with: 

rustup component add rust-src --toolchain nightly-2024-03-11-x86_64-unknown-linux-gnu

I run into the below error:

Running `probe-rs run --chip '${PROBE_RS_CHIP}' ../../build/bin/expressif-esp32-c6-devkitc-1/hello-world/cargo/riscv32imac-unknown-none-elf/release/hello-world`
Error: The chip '${PROBE_RS_CHIP}' was not found in the database.

Caused by:
    The requested chip '${PROBE_RS_CHIP}' was not found in the list of known targets.

kaspar030:
This PR intents to unify `laze run` across our devices.

it:

- removes the `silent-panic` dep from `release` (so panics are now shown by default)
- makes nrf & rp select `?probe-rs-run` (making it the default)
- renames `probe-rs-run` to just `probe-rs`, keeping an alias to `probe-rs-run`
- updates the docs to drop `-s probe-rs-run`
- *edit* also default to using the debug-console on esp32

*edit* I expect this to get another overhaul when [laze supports tasks defined by modules](https://github.com/kaspar030/laze/issues/420).

Fixes #200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ example](https://github.com/future-proof-iot/RIOT-rs/tree/main/examples/hello-wo
 
 1. Compile, flash and the hello-world example using `probe-rs run`
 
-        laze -C examples/hello-world build -b nrf52840dk -s probe-rs-run run
+        laze -C examples/hello-world build -b nrf52840dk run
 
 ![Example](./doc/hello-world_render.svg)
 
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -261,6 +261,8 @@ contexts:
 
   - name: esp
     parent: riot-rs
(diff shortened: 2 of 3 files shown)


chrysn (approved):
Haven't done any testing, but it looks very convenient; looks like I might soon start running laze by entering the command rather than cargo culting the ^R completion.

elenaf9 (approved):
Nice!

PR #291, opened 2024-05-14
doc: update add_board_support.md

kaspar030:
This updates the board support doc. Could be more exhaustive but at least it is not horribly outdated anymore.

### DIFF
--- a/doc/add_board_support.md
+++ b/doc/add_board_support.md
@@ -1,30 +1,100 @@
-# How to add a board
+# RIOT-rs adding hardware support
 
-This assumes the board is already supported by RIOT(-c).
-Currently, each board needs special support in RIOT-rs, as not RIOT(-c)'s
-bringup and linking code is used, but RIOT-rs Rust code.
+This document serves as a guide as to what's currently needed for adding support
+for a board/device to RIOT-rs.
 
-## nrf528xx based
+Feel free to report anything that's unclear!
 
-This is fairly straight forward. Currently, RIOT-rs doesn't do anything with
-peripherals itself (only through the riot-wrappers), so all nrf528xx based
-should work with the same code, apart from flashing.
+## Adding a board
 
-1. choose a suitable base board (nrf52dk for nrf52832 based board, nrf52840dk
-   for nrf52840)
-1. copy `src/riot-rs-boards/<base board>` to `src/riot-rs-boards/<board>`
-1. replace all base board names in string literals with the new board name
-1. add the new board to the features at src/riot-rs-boards/Cargo.toml, e.g.,
+The more similar a board is to one that is already supported, the easier.
+It is usually best to copy & adapt an existing one.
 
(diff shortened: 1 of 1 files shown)


doc/add_board_support.md
```diff
@@ -1,30 +1,92 @@
...
 
-        new_board = { optional=true, path="new_board" }
+- ensure [embassy architecture](#adding-an-embassy-architecture) is supported
+- in laze-project.yml:
+  - add "context" for the mcu (if it doesn't exist, yet)
+    - parent: the closest embassy arch
+      - selects: a [rustc-target](#adding-a-rustc-target-module) module
+      - if the architecture does not have a dedicated SWI, choose one now
+        and set CONFIG_SWI
+      - if the mcu is supported by probe-rs, add PROBE_RS_CHIP and PROBE_RS_PROTOCOL
```
elenaf9, line 20: What if it isn't supported by probe-rs?

doc/add_board_support.md
```diff
@@ -1,30 +1,92 @@
...
-1. choose a suitable base board (nrf52dk for nrf52832 based board, nrf52840dk
-   for nrf52840)
-1. copy `src/riot-rs-boards/<base board>` to `src/riot-rs-boards/<board>`
-1. replace all base board names in string literals with the new board name
-1. add the new board to the features at src/riot-rs-boards/Cargo.toml, e.g.,
+The more similar a board is to one that is already supported, the easier.
+It is usually best to copy & adapt an existing one.
 
-        new_board = { optional=true, path="new_board" }
+- ensure [embassy architecture](#adding-an-embassy-architecture) is supported
```
elenaf9, line 13:
```suggestion
- ensure that the architecture [is supported in riot-rs-embassy](#adding-an-embassy-architecture)
```

doc/add_board_support.md
```diff
@@ -1,30 +1,92 @@
...
+```
+
+Important are the variable RUSTC_TARGET and CARGO_TARGET_PREFIX.
+Add `--cfg foo` as needed.
+
+Chances are that if you need to add this, you'll also have to add support for the architecture to `riot-rs-threads`.
+
+## Adding an embassy architecture
+
+RIOT-rs currently supports nrf, stm32, rp and the externally (from embassy) maintained esp-rs.
```
elenaf9, line 85: That sentence is a bit confusing. Do you mean embassy supports it?

kaspar030: I tried to clarify this

doc/add_board_support.md
kaspar030: hm. esp based boards would inherit use of "espflash".

kaspar030: added some more choices

kaspar030:
Thanks @ROMemories and @elenaf9. I've committed and squashed all suggestions.

elenaf9 (approved):

PR #306, opened 2024-05-31
docs(_dev-doc): introduce the `_dev-doc` feature

ROMemories:
This introduces the `_dev-doc` Cargo feature, which allows us to hide some items from users by default and only show them to developers when this feature is enabled.

---

All items that should be hidden to users may not be part of this PR, you can propose them here or we'll progressively add them; this PR mostly aims to lay the groundwork for easily adding them in the future.

---

We will soon build another version of the docs, aimed at users, with the `_dev-doc` feature disabled.

### DIFF
--- a/src/riot-rs/Cargo.toml
+++ b/src/riot-rs/Cargo.toml
@@ -90,6 +90,10 @@ silent-panic = ["riot-rs-rt/silent-panic"]
 no-boards = ["riot-rs-boards/no-boards"]
 
 # Enables developer docs, shows items that are otherwise hidden from users.
-_dev-doc = ["riot-rs-embassy/_dev-doc", "riot-rs-rt/_dev-doc", "riot-rs-threads/_dev-doc"]
+_dev-doc = [
+  "riot-rs-embassy/_dev-doc",
+  "riot-rs-rt/_dev-doc",
+  "riot-rs-threads/_dev-doc",
+]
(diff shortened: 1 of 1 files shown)


elenaf9:
If I understand correctly, this features is used for functions/ modules that users don't need to care about if they are using our macros.
I haven't seen this pattern/ design before. In what scenario would it be useful? So far I only used `#[docs(hidden)]` for restricting what is showed on `docs.rs`, but aren't there all features enabled anyway?

I think I would prefer not to hide the items. Instead I think we should add docs explaining their purpose and when/how they should be used.

elenaf9:
> If we did that, these items would become part of our API and of our (future) SemVer commitment, which we want to avoid for what's only cross-crates implementation details.

You're right, I didn't think of that. But if we decide that these items only for internal usage, I would be in favor of just making it `hidden`.

> which allows developers to use the rustdoc docs to have a better grasp of the actual API surface of each workspace crate, even when some items are hidden from users.

I'd argue that this is a rather niche use-case. Developers could also just look at the rustdocs docs, and then look what additional items there are that are `doc(hidden)`. I also think the number of items in this category is limited.

---

I see the pro's of this features, but I still think it might be more confusing that helpful. But I don't feel strongly about it.

emmanuelsearch:

PR #309, opened 2024-06-03
fix(wifi): make Wi-Fi-related env vars mandatory when required

ROMemories:
Introduce a new macro that makes it easy to access environment variables for configuration, when it makes no sense to have a default value when the environment variable is not provided by the user.


kaspar030:
offline discussion: we'll need to export defaults in CI & update the networking README to nudge devs to also set those values locally when compiling for a board that deafults to using WiFi.

PR #312, opened 2024-06-06
riot-rs-threads: cortex-m: add compiler fence around `wfi` dance

kaspar030:
Fixes Cortex-M `sched()` not seeing changes to the runqueue.

See https://github.com/future-proof-iot/RIOT-rs/issues/310#issuecomment-2150039448.

As to why the fence is necessary, I cannot explain. The critical section should actually imply a fence. In this case, the runqueue accesses are heavily inlined, so maybe something is off with the optimizer. Or, we're doing something wrong with how we access `THREADS`, which is a lot more likely than a compiler bug. :smirk:

### DIFF
--- a/src/riot-rs-threads/src/arch/cortex_m.rs
+++ b/src/riot-rs-threads/src/arch/cortex_m.rs
@@ -172,6 +172,8 @@ unsafe fn sched() -> u128 {
                 Some(pid) => pid,
                 None => {
                     cortex_m::asm::wfi();
+                    // this fence seems necessary, see #310.
+                    core::sync::atomic::fence(core::sync::atomic::Ordering::Acquire);
                     return None;
                 }
             };


elenaf9 (approved):
> Or, we're doing something wrong with how we access THREADS, which is a lot more likely than a compiler bug.

```diff
 loop {
-        if let Some(res) = critical_section::with(|cs| {
-            let threads = unsafe { &mut *THREADS.as_ptr(cs) };
+        if let Some(res) = THREADS.with_mut(|mut threads| {
             let next_pid = match threads.runqueue.get_next() {
```
Would also fixes this issue. But as discussed in https://github.com/future-proof-iot/RIOT-rs/pull/187#discussion_r1526097316, this negatively impacts performance. My guess is that because it internally borrows the value instead of operating on the pointer, some optimizations are prevented?

Also, I tested using `compiler_fence` instead of `fence`, which didn't fix the issue. So I guess it's on the hardware side rather than the compiler?

Either way, I tested you fix and it works for me :+1:.

kaspar030:
> Would also fixes this issue. But as discussed in [#187 (comment)](https://github.com/future-proof-iot/RIOT-rs/pull/187#discussion_r1526097316), this negatively impacts performance. My guess is that because it internally borrows the value instead of operating on the pointer, some optimizations are prevented?

Could be ... We'll need to get back to this at some point.

Thanks for testing! :+1:

PR #313, opened 2024-06-06
riot-rs-threads: introduce `THREAD_FLAG_WAKEUP`, use in `block_on()`

kaspar030:
This PR introduces `THREAD_FLAG_WAKEUP`, which will be used to kick thread loops like `block_on()` and the executor from #304.

### DIFF
--- a/src/riot-rs-threads/src/thread_flags.rs
+++ b/src/riot-rs-threads/src/thread_flags.rs
@@ -1,8 +1,16 @@
 //! Thread flags.
+//!
+//! Every thread in RIOT-rs has a set of 16 flags that can be set and waited on.
+//!
+//! Note: The highest 4 bits (`1<<15, 1<<14, 1<<13, 1<<12`) are used inside
+//!       RIOT-rs and should not be used for applications.
 use crate::{ThreadId, ThreadState, Threads, THREADS};
 
 /// Bitmask that represent the flags that are set for a thread.
 pub type ThreadFlags = u16;
+
+/// Thread flag used to kick thread loops.
+pub const THREAD_FLAG_WAKEUP: ThreadFlags = 1 << 15;
 
 /// Possible waiting modes for [`ThreadFlags`].
 #[derive(Copy, Clone, PartialEq, Debug)]
--- a/src/riot-rs-embassy/src/blocker.rs
+++ b/src/riot-rs-embassy/src/blocker.rs
@@ -1,14 +1,12 @@
 use core::future::Future;
 use core::pin::Pin;
 use core::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
-use riot_rs_threads::{current_pid, flags, flags::ThreadFlags, ThreadId};
-
-const THREAD_FLAG_WAKER: ThreadFlags = 1; // TODO: find more appropriate value
+use riot_rs_threads::{current_pid, flags, flags::THREAD_FLAG_WAKEUP, ThreadId};
 
 fn wake(ptr: *const ()) {
     // wake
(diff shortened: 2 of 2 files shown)


src/riot-rs-threads/src/thread_flags.rs
```diff
@@ -1,9 +1,17 @@
 //! Thread flags.
+//!
+//! Every thread in RIOT-rs has a set of 16 flags that can be set and waited on.
+//!
+//! Note: The highest 4 bits (`1<<15, 1<<14, 1<<13, 1<<12`) are used inside
+//!       RIOT-rs and should not be used for applications.
 use crate::{ThreadId, ThreadState, Threads, THREADS};
 
 /// Bitmask that represent the flags that are set for a thread.
 pub type ThreadFlags = u16;
```
elenaf9, line 11: We could enforce this through the API:

```suggestion
//!
//! Every thread in RIOT-rs has a set of 16 flags that can be set and waited on.
use crate::{ThreadId, ThreadState, Threads, THREADS};

/// Bitmask that represent the flags that are set for a thread.
pub struct ThreadFlags(u16);

/// Thread flag used to kick thread loops.
pub const THREAD_FLAG_WAKEUP: ThreadFlags = 1 << 15;

impl ThreadFlags {
    /// Create new [`ThreadFlags`].
    /// 
    /// The highest 4 bits (`1<<15, 1<<14, 1<<13, 1<<12`) are reserved
    /// for internal usage.
    /// If they are set in `flags`, `None` will be returned.
    pub fn new(flags: u16) -> Option<Self> {
        if flags >= 1 << 12 {
            return None;
        }
        return Some(ThreadFlags(flags))
    }

    /// Flag that signals a thread to wakup.
    pub(crate) fn wakeup_flag() -> Self {
        ThreadFlags(THREAD_FLAG_WAKEUP)
    }
}

```

kaspar030:
Hmmm. After thinking on this another day, I'm not sure this makes sense at all. I have the impression we're creating a contract where there doesn't need to be one. Both uses that this is "unifying" don't care about the actual flag used or if any other flag also gets sent.

elenaf9:
> Hmmm. After thinking on this another day, I'm not sure this makes sense at all. I have the impression we're creating a contract where there doesn't need to be one. Both uses that this is "unifying" don't care about the actual flag used or if any other flag also gets sent.

It matters when a user uses the same flag for their own signaling and the flag is suddenly set in the background by `riot-rs`.

kaspar030:
> It matters when a user uses the same flag for their own signaling and the flag is suddenly set in the background by `riot-rs`.

I'm just wondering if this would happen. In the two cases at hand (`block_on()` and the thread executor), the flag is only used internally. `riot-rs` wouldn't send those to arbitrary threads.

> The specific flag doesn't matter, but we do need some "reserved" flags so the user knows not to use them.

I'm wondering if that's actually the case. I think we don't. What's a concrete case where we need "reserved" flags?

elenaf9:
> > It matters when a user uses the same flag for their own signaling and the flag is suddenly set in the background by `riot-rs`.
> 
> I'm just wondering if this would happen. In the two cases at hand (`block_on()` and the thread executor), the flag is only used internally. `riot-rs` wouldn't send those to arbitrary threads.
> 
> > The specific flag doesn't matter, but we do need some "reserved" flags so the user knows not to use them.
> 
> I'm wondering if that's actually the case. I think we don't. What's a concrete case where we need "reserved" flags?

:smile:  just had the same thought, see my edit on the comment above.

I can't think of a scenario right now. We can re-open this if one comes up.

kaspar030:
Closing this for now.

PR #314, opened 2024-06-07
feat(thread-macro): accept Expr as macro parameters

ROMemories:


kaspar030:
Tested a plain const and one that was parsed from env (using `usize_from_env_or!()`).

(The error messages when suppling a non-usize parseable are a bit weird)

LGTM.

examples/threading/src/main.rs
```diff
@@ -5,12 +5,16 @@
 
 use riot_rs::debug::println;
 
+const STACK_SIZE_THREAD1: usize = 4096;
```
kaspar030, line 8: Would you remove the changes here? Let's have a proper PR introducing defaults and using them.

src/riot-rs-macros/src/thread.rs
```diff
@@ -22,7 +22,11 @@
 /// This starts a thread with a stack size of 1024 bytes and a priority of 2:
 ///
 /// ```ignore
-/// #[riot_rs::thread(stacksize = 1024, priority = 2)]
+/// const STACK_SIZE: usize = 1024;
```
kaspar030, line 25: There's no need for the constants here, so let's have the example succinct. Maybe either remove changes? Alternatively, change only one of the literals to an expression and add an inline comment like `expressions can also be used`?

kaspar030: great, thanks!

kaspar030:
LGTM, let's squash&merge this!

PR #315, opened 2024-06-10
.github: Add templates for PRs and Issues

elenaf9:
Add Issue and PR templates to help contributors.

I've used the [beta issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) for the issue templates, but I can change it back to traditional markdown if you prefer that.

I drafted the templates based on what I saw in other projects, feel free to suggestion changes.

### DIFF
--- a/.github/workflows/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/workflows/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,42 @@
+name: Bug Report
+description: Create a bug report for RIOT-rs
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for filing a bug report!
+  - type: textarea
+    attributes:
+      label: Summary
+      description: Please provide a short summary of the bug, along with any information you feel relevant to replicate the bug.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Relevant log output
+      description: Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
+      render: shell
+    validations:
+      required: false
+  - type: textarea
+    attributes:
+      label: Hardware/ Setup
+      description: Any relevant information about the used hardware, setup, RIOT-rs version, etc.
+    validations:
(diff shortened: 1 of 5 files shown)


bergzand:
Is there a way to preview the rendered templates before merging this?

.github/workflows/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,42 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for filing a bug report!
+  - type: textarea
+    attributes:
+      label: Summary
+      description: Please provide a short summary of the bug, along with any information you feel relevant to replicate the bug.
+    validations:
+      required: true
```
bergzand, line 14: To me it makes sense to split this into a question asking for expected behaviour and actual behaviour.

If you want to be exhaustive we could even ask for a minimal example/steps to reproduce the issue.

What do you think?

elenaf9: I initially had it in the template, but then decided to remove it. From my own experience, "expected behavior" and "actual behavior" often had a large overlap with the summary section. But I am okay with adding it.

> If you want to be exhaustive we could even ask for a minimal example/steps to reproduce the issue.

:+1: I'm in favor of that.

bergzand: > I initially had it in the template, but then decided to remove it. From my own experience, "expected behavior" and "actual behavior" often had a large overlap with the summary section. But I am okay with adding it.

Fine with leaving it out with me. We can always add it later when we notice it gets omitted in initial reports.

elenaf9: (Note to myself: I should stop writing comments and then directly editing them, because people seems to faster at replying then I am at my edit :D)

bergzand (approved):

PR #317, opened 2024-06-12
feat(embassy-arch): provide peripheral-independent I2C and SPI types

ROMemories:
# Description

This leverages `embassy-embedded-hal` to provide buses instantiated once and then shared between tasks.

This is the first step in landing the WIP sensor abstraction piece by piece.

## Issues/PRs references

Depends on #322 

## Open Questions

The `dummy` module in the `arch` module would need to be updated, but it's unclear what we should provide in it, as the interface of other archs is not exactly the same because of different requirements (e.g., RP2040 require DMA channels for SPI, and nRF GPIO outputs require an [`OutputDriver` parameter](https://docs.embassy.dev/embassy-nrf/git/nrf52840/gpio/struct.Output.html#method.new)).

## TODO

- [x] SPI and I2C for RP2040
- [x] SPI and I2C for nRF52840 and nRF5340
- [x] I2C for ESP32
- [ ] SPI for ESP32 (almost supported; the driver is currently not `Send` because of the underlying DMA linked-list, and there is an unresolved SPI mode-related issue)
- [x] SPI and I2C for STM32s
- [ ] resolve the [licensing issue of `stm32-data`](https://github.com/embassy-rs/stm32-data/issues/514) and decide how to distribute the JSON data files required for SMT32 
- [ ] decide how to handle mutually exclusive peripherals on nRF (because they share the same interrupt)
- [x] introduce timeouts to avoid infinitely blocking if the I2C device is not connected
- [ ] revisit the API we expose

## Change checklist

### DIFF
--- a/src/riot-rs-embassy/src/arch/esp/spi.rs
+++ b/src/riot-rs-embassy/src/arch/esp/spi.rs
@@ -26,7 +26,7 @@ pub struct Config {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            frequency: Frequency::M1, // FIXME
+            frequency: Frequency::_1M,
             mode: Mode::Mode0,
             bit_order: BitOrder::default(),
         }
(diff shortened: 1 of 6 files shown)


maribu:
Cool!

PR #318, opened 2024-06-14
build(deps): bump to esp-hal 0.18+ monorepo

kaspar030:
Upstream has changed their repo setup.

- `esp-wifi` is now part of `esp-hal`
- embassy support has been factored out into `esp-hal-embassy`
- some API changes: https://github.com/esp-rs/esp-hal/releases/tag/v0.18.0


kaspar030:
tested `examples/embassy-http-server` on `expressif-esp32-c6-devkitc-1`

PR #324, opened 2024-06-21
docs(book): add an initial version of the support matrix

ROMemories:
# Description

This introduces a structured data file encoding current hardware and functionality support across supported chips and boards, along with a [Cargo script](https://github.com/rust-lang/cargo/issues/12207) generating HTML from this structured data. The generated HTML is written to a file which is imported into the book, providing comprehensive support information for potential users. That HTML file must be re-generated when the structured data file is updated, but a CI lint check makes sure that this HTML file is always up to date.

This support table covers hardware support as well as software/configuration support.

## Limitations

This is mostly meant to bootstrap this compatibility table and support information definitely incomplete. It must still reflect the current state of support at the time of merge.

@kaspar030 Can you check the current state of debug output and logging, and the current ESP32 support?
@kaspar030 @elenaf9 If threading needs to be added to this table, please do so in a follow-up PR.
@chrysn Could you add HWRNG to this table in a follow-up PR?

---

- The structured data file must currently be manually updated. Ideally we would leverage support information already encoded in the laze configuration to validate/invalidate that file.
- This could also benefit from later introduced an intermediate level between chips and boards, e.g., `chip_families`.

## Issues/PRs references

N/A

## Open Questions

None

## Change checklist

### DIFF
--- a/book/src/hardware_functionality_support.md
+++ b/book/src/hardware_functionality_support.md
@@ -1,0 +1,5 @@
+# Hardware & Functionality Support
+
+> The table below indicates whether we support using the piece of functionality in a portable manner, trough an abstraction layer and platform-aware configuration.
+
+{{#include support_matrix.html }}
--- a/doc/gen_support_matrix_html.rs
+++ b/doc/gen_support_matrix_html.rs
@@ -57,7 +57,7 @@ const KEY_TEMPLATE: &str =
 r##"<p>Key:</p>
 
 <ul>
-  {%- for _, support_key in matrix.support_keys|items %}
+  {%- for support_key in matrix.support_keys %}
   <li class="no-marker">{{ support_key.icon }}&nbsp;&nbsp;{{ support_key.description }}</li>
   {%- endfor %}
 </ul>
@@ -228,7 +228,8 @@ fn render_html(matrix: &schema::Matrix) -> Result<String, Error> {
                 let support_key = if let Some(support_info) = board_info.support.get(name) {
                     let status = support_info.status();
                     matrix.support_keys
-                        .get(status)
+                        .iter()
+                        .find(|s| s.name == status)
                         .ok_or(Error::InvalidSupportKeyNameBoard {
                             found: status.to_owned(),
                             functionality: name.to_owned(),
@@ -250,7 +251,8 @@ fn render_html(matrix: &schema::Matrix) -> Result<String, Error> {
                         })?;
                     let status = support_info.status();
                     matrix.support_keys
-                        .get(status)
+                        .iter()
+                        .find(|s| s.name == status)
                         .ok_or(Error::InvalidSupportKeyNameChip {
                             found: status.to_owned(),
                             functionality: name.to_owned(),
@@ -367,7 +369,7 @@ mod schema {
     #[derive(Debug, Serialize, Deserialize)]
     #[serde(deny_unknown_fields)]
     pub struct Matrix {
-        pub support_keys: HashMap<String, SupportKeyInfo>,
+        pub support_keys: Vec<SupportKeyInfo>,
         pub functionalities: Vec<FunctionalityInfo>,
         pub chips: HashMap<String, ChipInfo>,
         pub boards: HashMap<String, BoardInfo>,
@@ -376,6 +378,7 @@ mod schema {
     #[derive(Debug, Serialize, Deserialize)]
     #[serde(deny_unknown_fields)]
     pub struct SupportKeyInfo {
+        pub name: String,
         pub icon: String,
         pub description: String,
     }
--- a/doc/gen_support_matrix_html.rs
+++ b/doc/gen_support_matrix_html.rs
@@ -56,14 +56,16 @@ r##"<!-- This table is auto-generated. Do not edit manually. -->
 const KEY_TEMPLATE: &str =
 r##"<p>Key:</p>
 
-<ul>
+<dl>
   {%- for support_key in matrix.support_keys %}
-  <li class="no-marker">{{ support_key.icon }}&nbsp;&nbsp;{{ support_key.description }}</li>
+  <div>
+    <dt>{{ support_key.icon }}</dt><dd>{{ support_key.description }}</dd>
+  </div>
   {%- endfor %}
-</ul>
+</dl>
 <style>
-.no-marker::marker {
-  content: '';
+dt, dd {
+  display: inline;
 }
 </style>
 "##;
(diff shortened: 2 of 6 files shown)


kaspar030:
I think this is a great start! I hope we can make this auto-generated at some point - laze should have all info.

kaspar030:
> We use an HTML table as the GitHub Tables extension does not support cells spanning on multiple rows or columns.

Hm, I personally don't support hand-editing raw html tables. I'd prefer using a markdown table (and duplicating e.g., the `rp2040` line)... wdyt?

book/src/hardware_functionality_support.md
```diff
@@ -0,0 +1,93 @@
+# Hardware & Functionality Support
+
+> RIOT-rs always allows to obtain raw peripherals when using the `riot_rs::spawner` and `riot_rs::task` attributes.
```
kaspar030, line 3: This sentence doesn't make sense here I think.

doc/gen_support_matrix_html.rs
```diff
@@ -0,0 +1,426 @@
+#!/usr/bin/env -S cargo -Zscript
```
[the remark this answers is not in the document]
chrysn: Whoever files the first PR after the nightlies learn to format / check that earns the honor of running `cargo fmt` on it :-)

doc/gen_support_matrix_html.rs
```diff
@@ -0,0 +1,426 @@
...
+.support-cell {
+  text-align: center;
+}
+</style>
+"##;
+
+const KEY_TEMPLATE: &str =
+r##"<p>Key:</p>
+
+<ul>
```
chrysn, line 59: Bonus style points for using dl/dt/dd (really, leave it as is unless you want to boast ultra accessible HTML)

doc/gen_support_matrix_html.rs
```diff
@@ -0,0 +1,426 @@
...
+      {%- endfor %}
+    </tr>
+  </thead>
+  <tbody>
+    {%- for board in boards %}
+    <tr>
+      <td>{{ board.chip }}</td>
+      <td>{{ board.name }}</td>
+      {%- for functionality in board.functionalities %}
+      <td class="support-cell">{{ functionality.icon }}</td>
```
chrysn, line 43:
```suggestion
      <td class="support-cell" title="{{ functionality.description }}">{{ functionality.icon }}</td>
```

if that information is available there

doc/support_matrix.yml
```diff
@@ -0,0 +1,130 @@
...
+    description: needs testing; support may be present but we do not currently test it
+  being_reworked:
+    icon: ♻️
+    description: rework in progress, expect breaking changes
+  wip:
+    icon: 🚧
+    description: work in progress
+  not_currently_supported:
+    icon: ❌
+    description: not currently supported by RIOT-rs
```
chrysn, line 22:
```suggestion
    description: present in the hardware, but not currently supported by RIOT-rs
```

doc/gen_support_matrix_html.rs
```diff
@@ -0,0 +1,426 @@
...
+
+mod schema {
+    use std::collections::HashMap;
+
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Serialize, Deserialize)]
+    #[serde(deny_unknown_fields)]
+    pub struct Matrix {
+        pub support_keys: HashMap<String, SupportKeyInfo>,
```
kaspar030, line 368: It's one dependency more, but may I suggest using `IndexMap` here? That way the user has some say about the order.

doc/gen_support_matrix_html.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    #[derive(Debug, Serialize, Deserialize)]
+    #[serde(deny_unknown_fields)]
+    #[serde(untagged)]
+    pub enum SupportInfo {
+        StatusOnly(String),
+        Details {
+            status: String,
+            comments: Option<String>,
+            github_pr: Option<u32>,
```
chrysn, line 414: In case you rework that for some reason or other, maybe just use a URI String here? The values are copy-pasted over from a browser anyway (I for one do not trust myself with copying over 3-digit numbers correctly), and that'd open up for having an issue, a PR, milestone item or even blog post in there, and it's one fewer hardcoded tie to GitHub.

doc/support_matrix.yml
```diff
@@ -0,0 +1,130 @@
...
+    title: Logging
+    description:
+  - name: user_usb
+    title: User USB
+    description:
+  - name: wifi
+    title: Wi-Fi
+    description:
+  - name: ethernet_over_usb
+    title: Ethernet over USB
```
chrysn, line 45: How is this a property of the hardware? I'd expect this to be a functional property of the USB stack available, and possibly of some scalar hardware parameters like available non-control endpoints. Or do we just not have those details available yet, so "Ethernet over USB" is a baked version of those parameters?

doc/support_matrix.yml
```diff
@@ -0,0 +1,130 @@
...
+    support:
+      user_usb: supported
+      ethernet_over_usb: supported
+
+  nrf5340-dk:
+    name: nRF5340-DK
+    chip: nrf53xx
+    support:
+      user_usb: supported
+      wifi: not_available
```
chrysn, line 105:
```suggestion
```

needless override

chrysn (commented):
I like the structure. Makes me want to consider whether we can pioneer https://github.com/RIOT-OS/RIOT/blob/master/doc/memos/rdm0003.md in here.

doc/support_matrix.yml
```diff
@@ -0,0 +1,130 @@
+---
```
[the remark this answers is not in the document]
kaspar030: for now definitely!

kaspar030:
I like this a lot!

doc/support_matrix.yml
kaspar030: I'm having some issues on stm32h7, where usb-serial works fine but usb-ethernet does not. Not sure how fine-grained we want this matrix, but just because something should work doesn't mean it actually does, and that reality can become very fine grained.

doc/gen_support_matrix_html.rs
chrysn: Just remember the [first rule of ARIA use](https://www.w3.org/TR/using-aria/#rule1) ;-)

kaspar030:
Some offline discussion notes:
- too many "Key" options. reduce to "Supported ✅" , "(Supported) ☑️", "Not supported by RIOT-rs ❌", "Not supported by hardware (-)"

kaspar030:
LGTM. please squash!

PR #328, opened 2024-06-27
feat(examples): add `usb-serial` example

kaspar030:
# Description

This adds a usb serial example.
It is basically a port of one of Embassy's [examples](https://github.com/embassy-rs/embassy/blob/main/examples/stm32h7/src/bin/usb_serial.rs).

I tested that it works, with Linux as host, on nrf52840dk, rpi-pico and two stm32 (which are not in main yet).

## Change checklist

### DIFF
--- a/examples/usb-serial/src/main.rs
+++ b/examples/usb-serial/src/main.rs
@@ -1,0 +1,75 @@
+#![no_std]
+#![no_main]
+#![feature(type_alias_impl_trait)]
+#![feature(used_with_arg)]
+
+use riot_rs::{
+    debug::log::info,
+    embassy::{
+        embassy_usb, make_static,
+        usb::{UsbBuilderHook, UsbDriver},
+    },
+};
+
+use embassy_usb::{
+    class::cdc_acm::{CdcAcmClass, State},
+    driver::EndpointError,
+};
+
+#[riot_rs::config(usb)]
+fn usb_config() -> embassy_usb::Config<'static> {
+    let mut config = embassy_usb::Config::new(0xc0de, 0xcafe);
+    config.manufacturer = Some("RIOT-rs");
+    config.product = Some("USB serial example");
+    config.serial_number = Some("12345678");
+    config.max_power = 100;
+    config.max_packet_size_0 = 64;
+
+    // Required for Windows support.
+    config.composite_with_iads = true;
+    config.device_class = 0xEF;
+    config.device_sub_class = 0x02;
+    config.device_protocol = 0x01;
+    config
+}
+
+#[riot_rs::task(autostart, usb_builder_hook)]
+async fn main() {
+    info!("Hello World!");
+
+    let state = make_static!(State::new());
+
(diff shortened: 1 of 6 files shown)


examples/usb-serial/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+};
+
+use embassy_usb::{
+    class::cdc_acm::{CdcAcmClass, State},
+    driver::EndpointError,
+};
+
+#[riot_rs::config(usb)]
+fn usb_config() -> embassy_usb::Config<'static> {
+    let mut config = embassy_usb::Config::new(0xc0de, 0xcafe);
```
chrysn, line 21: There are codes for use with test devices, and while I don't know off my head where to look these up, but they're probably made up.

Good candidates are [1209](https://pid.codes/1209/) / 7D01 (RIOT Test PID) or 1298 / 000[0-f] (generic test PIDs)

examples/usb-serial/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+    config.product = Some("USB serial example");
+    config.serial_number = Some("12345678");
+    config.max_power = 100;
+    config.max_packet_size_0 = 64;
+
+    // Required for Windows support.
+    config.composite_with_iads = true;
+    config.device_class = 0xEF;
+    config.device_sub_class = 0x02;
+    config.device_protocol = 0x01;
```
chrysn, line 32: This implies that we're talking AT commands, which is not only incorrect, but it is also an invitation for NetworkManager and similar to start probing the apparent modem, sending garbage before the device is made available to the user. RIOT OS picks 0 (USB_CDC_PROTOCOL_NONE); 0xFF (vendor-specific) might be a sensible choice as well.

kaspar030:
> gives me `8000 bauds`.

I digged a bit, IMO that can just be ignored.

[upstream code](https://github.com/kaspar030/embassy/blob/584940fc7ea02fe184f3ae346ecaa3d00008c95e/embassy-usb/src/class/cdc_acm.rs#L503-L506) says so. Also, the baud rate can only be set from the host.

examples/usb-serial/src/main.rs
kaspar030: "code cafe" is certainly made up. :) This is copied from upstream. I'll open an issue for default USB ids.

kaspar030: https://github.com/future-proof-iot/RIOT-rs/issues/339

examples/usb-serial/src/main.rs
kaspar030: This is copied from upstream. the subdescriptor is configures with `USB_CDC_PROTOCOL_NONE` iirc: https://github.com/kaspar030/embassy/blob/584940fc7ea02fe184f3ae346ecaa3d00008c95e/embassy-usb/src/class/cdc_acm.rs#L219

PR #329, opened 2024-06-27
feat(net): wrap network stack in a `OnceLock`

kaspar030:
# Description

This PR wraps the network stack in a `OnceCell`. Without this, if tasks call `network_stack()` before `STACK` has been initialized will just panic, e.g., on `rpi-pico-w` where the wifi initialization is not finished before the first tasks get to run.

The PR also bumps the used version of `embassy-sync` to 0.6 from our fork. We were previously using the crates-io released 05, which does not yet include `OnceLock`.

## Issues/PRs references

First encountered in #146.

## Change checklist

- [ ] I have followed the [Coding Conventions](../book/src/coding-conventions.md).

### DIFF
--- a/src/riot-rs-embassy/src/network.rs
+++ b/src/riot-rs-embassy/src/network.rs
@@ -1,10 +1,7 @@
 //! To provide a custom network configuration, use the `riot_rs::config` attribute macro.
 
-use core::cell::OnceCell;
-
-use embassy_executor::Spawner;
 use embassy_net::Stack;
-use embassy_sync::blocking_mutex::CriticalSectionMutex;
+use embassy_sync::once_lock::OnceLock;
 
 use crate::sendcell::SendCell;
 use crate::NetworkDevice;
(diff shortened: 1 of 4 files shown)


src/riot-rs-embassy/src/network.rs
```diff
@@ -14,12 +11,10 @@ pub const ETHERNET_MTU: usize = 1514;
 
 pub type NetworkStack = Stack<NetworkDevice>;
 
-pub(crate) static STACK: CriticalSectionMutex<OnceCell<SendCell<&'static NetworkStack>>> =
-    CriticalSectionMutex::new(OnceCell::new());
+pub(crate) static STACK: OnceLock<SendCell<&'static NetworkStack>> = OnceLock::new();
```
elenaf9, line 14: I think we don't need the `SendCell` here anymore, because `OnceLock` implements `Sync`.

kaspar030: That doesn't help. We're using the `SendCell` to prevent the stack from being accessed from other executors. The `Sync` is necessary here to access the global static in the first place. But IIUC `Sync` on `OnceLock` doesn't make it safe to send `NetworkStack` to another thread / executor.

kaspar030: I think there are actually some trait bounds missing on `OnceLock`'s `Sync` impl.

elenaf9: Hmm, maybe you're right. But in that case I think it's problematic that `OnceLock` blindly implements `Sync`, because it would have allowed our code to compile without the `SendCell`.

elenaf9 (approved):
LGTM!

PR #337, opened 2024-07-01
fix(embassy-init): move out peripherals required for drivers earlier

ROMemories:
# Description

Previously, it would have been possible for user tasks to take peripherals out of `OptionalPeripherals` *before* the required peripherals had been extracted for drivers.
This now takes the peripherals before starting the user tasks, making this impossible for tasks.

## Issues/PRs references

Depends on #533.

## Open Questions

None.

## Change checklist

### DIFF
--- a/src/riot-rs-embassy/src/lib.rs
+++ b/src/riot-rs-embassy/src/lib.rs
@@ -143,12 +143,22 @@ fn init() {
 
 #[embassy_executor::task]
 async fn init_task(mut peripherals: arch::OptionalPeripherals) {
+    use crate::define_peripherals::TakePeripherals;
+
     println!("riot-rs-embassy::init_task()");
 
     #[cfg(feature = "hwrng")]
     arch::hwrng::construct_rng(&mut peripherals);
     // Clock startup and entropy collection may lend themselves to parallelization, provided that
     // doesn't impact runtime RAM or flash use.
+
+    // Move out the peripherals required for drivers, so that tasks cannot mistakenly take them.
+    #[cfg(feature = "usb")]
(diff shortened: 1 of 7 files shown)


src/riot-rs-embassy/src/lib.rs
```diff
@@ -143,13 +143,23 @@ fn init() {
 
 #[embassy_executor::task]
 async fn init_task(mut peripherals: arch::OptionalPeripherals) {
+    use crate::define_peripherals::TakePeripherals;
```
elenaf9, line 146: Style: move imports to top of file.

kaspar030:
I'm not sure I agree with this direction.

1. this makes some parts of the system special ("ours"), demoting user code. IMO this shouldn't be the case, we should rather try harder to build our components on what we offer to users. 
2. this is, imo, clumsier to use. Now there's an extra `define_peripherals!()` for a lot of cases where there is only one (or even no) peripheral used.

kaspar030:
> So, I'm not convinced this is desirable.

What I think might be preferable would be a way to track where peripherals were taken (e.g., modify "take()" to `trace!()` where it was called if nudged to to so by, e.g., sth like `CONFIG_PERIPH_TRACE=PA15,USBD`).

kaspar030:
> user tasks

we could also have two task lists, one for those that need to hook into the regular init, and one "default" that is only started after all regular init has taken place.

kaspar030:
Looking at this again, I think it is a good idea. Please rebase!

kaspar030:
Thanks for the nudge @stappersg! :slightly_smiling_face:

PR #348, opened 2024-07-02
Move CoAP server setup into module

chrysn:
# Description

This should eventually move all the required shared setup for CoAP into coapcore.

So far, it only moves everything after the socket setup (see open questions)

## Issues/PRs references

Depends on https://github.com/future-proof-iot/RIOT-rs/pull/346.
Contributes-To: https://github.com/future-proof-iot/RIOT-rs/issues/340

## Open Questions

Where is the socket setup best moved? My current approach would be to have a new riot-rs-coapcore crate that does the integration (might alsmost be embassy-net-coapcore, but we'll also pull in the RNG in that step).

Is there any better way to do what is effectively an async callback than by introducing a trait in lib.rs? (I've spent about 1h fighting the borrow checker, concluding that I'd probably need a generic associated type like `where for<'a> ClientOps: Fn(...) -> ClientFuture<'a>, ClientFuture<'a>: Future`, but Rust can't do that).

## Change checklist


chrysn:
Closing in favor of https://github.com/future-proof-iot/RIOT-rs/pull/493

PR #349, opened 2024-07-04
feat(threads, runqueue): derive `defmt::Format` for some types

elenaf9:
# Description

Derive `defmt::Format` for `ThreadId`, `RunqueueId` and `ThreadState` if the `defmt` feature is enabled.

Useful for debugging.

## Open Questions

@kaspar030 why are we not using `defmt::println` when `defmt` is enabled? 
I noticed that it results in some chronologically incorrect logs when `println` from other backends is combined with `defmt`. If I remember correctly we had a discussion around this and decided to "all in" for defmt if the feature is enabled. Am I missing something? If not, I would do a follow up PR that switches to `defmt::println`.

## Change checklist

### DIFF
--- a/src/riot-rs/Cargo.toml
+++ b/src/riot-rs/Cargo.toml
@@ -72,7 +72,11 @@ wifi-esp = ["riot-rs-embassy/wifi-esp"]
 ## [`println!`](riot_rs_debug::println).
 debug-console = ["riot-rs-rt/debug-console"]
 ## Enables logging support through `defmt`.
-defmt = ["riot-rs-debug/defmt", "riot-rs-embassy/defmt"]
+defmt = [
+  "riot-rs-debug/defmt",
+  "riot-rs-embassy/defmt",
+  "riot-rs-threads?/defmt",
+]
 ## Enables benchmarking facilities.
 bench = ["dep:riot-rs-bench"]
 ## Prints nothing in case of panics (may help reduce binary size).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,9 +80,9 @@ checksum = "6b2d54853319fd101b8dd81de382bcbf3e03410a64d8928bbee85a3e7dcde483"
 
 [[package]]
 name = "anyhow"
-version = "1.0.82"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
+checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
 name = "arrayvec"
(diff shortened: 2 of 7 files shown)


kaspar030:
> @kaspar030 why are we not using `defmt::println` when `defmt` is enabled?

We don't do that because the `defmt::println` doesn't support printing `Debug` or `Display` instances, only `defmt::Format`.
That in turn is not available when writing to any non-defmt output. So users of `riot_rs::debug::println!()` would be limited to printing types that happen to implement both, or compilation fails ...

> I noticed that it results in some chronologically incorrect logs when `println` from other backends is combined with `defmt`. 

Yep, that happens, on Cortex-M we're using `rtt_target`'s `println` on one rtt channel and defmt on another, and they get polled by probe-rs in arbitrary order (IIUC).

> If I remember correctly we had a discussion around this and decided to "all in" for defmt if the feature is enabled.

We decided to use the log macros (`info!()` et al) for all output and not `println!()`, by default. There's a similar issue around `defmt::Format` if we want to also support the regular log crate, though.

kaspar030:
LGTM!

src/riot-rs/Cargo.toml
```diff
@@ -72,7 +72,7 @@ wifi-esp = ["riot-rs-embassy/wifi-esp"]
 ## [`println!`](riot_rs_debug::println).
 debug-console = ["riot-rs-rt/debug-console"]
 ## Enables logging support through `defmt`.
-defmt = ["riot-rs-debug/defmt", "riot-rs-embassy/defmt"]
+defmt = ["riot-rs-debug/defmt", "riot-rs-embassy/defmt", "riot-rs-threads/defmt"]
```
kaspar030, line 75:
```suggestion
defmt = ["riot-rs-debug/defmt", "riot-rs-embassy/defmt", "riot-rs-threads?/defmt"]
```

elenaf9:
> We decided to use the log macros (info!() et al) for all output and not println!(), by default. 

Right now in `riot-rs-embassy` we are still using `println!`. Would you be okay with changing these logs to `log::debug!`?

elenaf9:
CI somehow has an issue with my `Cargo.lock`, even though I am up-to-date with main and compiled it once before pushing. Any idea what the issue is/ could one of you @kaspar030 @ROMemories check this branch out and see if you `Cargo.lock` differs after building?

kaspar030:
> CI somehow has an issue with my `Cargo.lock`, even though I am up-to-date with main and compiled it once before pushing. Any idea what the issue is/ could one of you @kaspar030 @ROMemories check this branch out and see if you `Cargo.lock` differs after building?

yes, https://gist.github.com/kaspar030/ad95b38af84f919d7895b6d9ec1bc356.

kaspar030:
> Right now in `riot-rs-embassy` we are still using `println!`. Would you be okay with changing these logs to `log::debug!`?

feel free! let the bikeshedding begin on what is "info" and "debug" :)

PR #350, opened 2024-07-04
CoAP storage demo

chrysn:
# Description

This is a first experiment with CoAP server access to the storage.

The README has an added section on how to GET and PUT values through encrypted CoAP

## Issues/PRs references

Depends on https://github.com/future-proof-iot/RIOT-rs/pull/277 (CC'ing @kaspar030)
Depends on https://github.com/future-proof-iot/RIOT-rs/pull/348

## Open Questions

* When will I update the CoAP server handlers to be properly async so I don't need to rely on our backend not really being async? (If it were, things would just 5.00 Internal Server Error)
* Will we need some storage multiplexing or run-time acquisition of the storage as soon as there are multiple handlers, either of which may access storage (or even if there is just one or it multiplexes internally, something non-CoAP and CoAP accesing it)?
* Do we want/need to enumerate and type-introspect items? (Introspection would be limited to types known ahead of time, unless introspection gives us dyn and they are type safe so probably not, plus then it'd be *really* dangerous to persist this across firmware updates)
* Is loading them from postcard internally, serde unpacking them and then serde serializing them for transport as CBOR a sensible thing to do? (Probably: This is a demo, after all)

## Change checklist

- [ ] Anything was done properly at all

  No: This is a very coarse thing right now, and is probably best rewritten from scratch after its dependencies have matured


chrysn:
Replaced by https://github.com/future-proof-iot/RIOT-rs/pull/498

PR #356, opened 2024-07-09
feat(CI): add some missing package checks

elenaf9:
# Description

I noticed that `riot-rs-runqueue` is missing in the CI crate tests, and `riot-rs-runqueue` in the CI clippy check.
This PR adds them.

## Open Questions

Was there are reason why the packages weren't added to the list in these two cases?


kaspar030:
LGTM.

> Was there are reason why the packages weren't added to the list in these two cases?

I don't remember any. Maybe the crate wasn't stubbed out enough to actually pass tests?

PR #361, opened 2024-07-11
test: add `embedded-test` integration

### LINKED ISSUE #331: integrate `embedded-test`

kaspar030:
# Description

This PR intents to provide a seamless integration of `embedded-test`.

~~This is very WIP, basically I'm in the experimentation phase.~~
~~This is taking shape. Still waiting on a probe-rs release, though.~~

This can be tested **with probe-rs-tools >=0.25.0** as follows:

    #  laze -Cexamples/testing build -v -b nrf52840dk -s executor-thread test

## Porting existing tests

The following steps can be followed to port existing tests from the `tests` directory to use `embedded-test`:

- Add `embedded-test` in `dev-dependencies`
- Add a `test` target (keep the `test` name for every test):

```toml
[[test]]
name = "test"
path = "src/test.rs"
harness = false
```

- Rename the existing `src/main.rs` to `src/test.rs`
- In the `laze.yml` file, add the following:

```yaml
    selects:
      - embedded-test
```

## Issues/PRs references

Fixes #331.
Depends on #506, #509, #549.

## Open Questions

- [x] needs probe-rs from git due to https://github.com/probe-rs/probe-rs/pull/2604 => probe-rs 0.25.0 released
- [x] embedded-test allows both async and sync tests. this integration starts a thread, and inside the thread, a thread executor. that means the test cannot use non-Send stuff from the system executor (e.g., network stack).
    so, we need a way to re-use the system executor.
- [x] integration is hacked into embedded-test fork, needs cleanup (& upstream PR)
- [x] how to handle test-only "apps" in laze. no .elf so "laze build" fails on them. *this also makes CI completely fail atm*. => `embedded-test-only` makes `laze build` a noop
- [x] ariel-os-identity test needs to be fixed to expect the right thing depending on whether there should be a device id
- [ ] document ...

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,7 +353,6 @@ dependencies = [
 name = "ariel-os-rt"
 version = "0.1.0"
 dependencies = [
- "ariel-os-boards",
  "ariel-os-debug",
  "ariel-os-threads",
  "ariel-os-utils",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -279,8 +279,11 @@ dependencies = [
 name = "ariel-os-identity"
 version = "0.1.0"
 dependencies = [
+ "ariel-os",
+ "ariel-os-boards",
  "ariel-os-embassy",
  "ariel-os-embassy-common",
+ "embedded-test",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


kaspar030:
> The `ariel-os-rt` `laze.yml` still selects a `disabled_until_test_framework_fixed` laze module; should this be updated?

it should, let's do that as follow-up. `ariel-os-rt` doesn't need its test support anymore, I'll just delete it from laze for now.

PR #370, opened 2024-07-31
feat(threads): dynamic thread priorities

### LINKED ISSUE #320: feat: provide API to change thread priority
Some users have voiced the need for changing a thread's priority at runtime.

elenaf9:
# Description

Allow users to dynamically change thread priorities at runtime.

## Issues/PRs references

Fixes #320 
Depends on #437

## Change checklist

### DIFF
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -84,8 +84,8 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
     /// # Panics
     ///
     /// Panics if `n` is not the queue's head.
-    /// This is fine, RIOT-rs only ever calls `del()` for the current thread.
-    pub fn del(&mut self, n: ThreadId, rq: RunqueueId) {
+    /// This is fine, RIOT-rs only ever calls `pop_head()` for the current thread.
+    pub fn pop_head(&mut self, n: ThreadId, rq: RunqueueId) {
         debug_assert!(usize::from(n) < N_THREADS);
         debug_assert!(usize::from(rq) < N_QUEUES);
         let popped = self.queues.pop_head(rq.0);
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -79,6 +79,11 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
         self.queues.push(n.0, rq.0);
     }
 
+    /// Returns the head of the runqueue without removing it.
+    pub fn peek_head(&self, rq: RunqueueId) -> Option<ThreadId> {
+        self.queues.peek_head(rq.0).map(ThreadId::new)
+    }
+
     /// Removes thread with pid `n` from runqueue number `rq`.
     ///
     /// # Panics
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -96,6 +96,13 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
         }
     }
 
+    /// Removes thread with pid `n`.
+    pub fn del(&mut self, n: ThreadId) {
+        if let Some(empty_runqueue) = self.queues.del(n.0) {
+            self.bitcache &= !(1 << empty_runqueue);
+        }
+    }
+
     fn ffs(val: usize) -> u32 {
         USIZE_BITS as u32 - val.leading_zeros()
     }
@@ -177,6 +184,31 @@ mod clist {
                 self.tail[rq as usize] = n;
                 self.next_idxs[n as usize] = n;
             }
+        }
+
+        /// Removes a thread from the list.
+        ///
+        /// If the thread was the only thread in its runqueue, `Some` is returned
+        /// with the ID of the now empty runqueue.
+        pub fn del(&mut self, n: u8) -> Option<u8> {
+            let mut empty_runqueue = None;
+
+            // Find previous thread in circular runqueue.
+            let prev = self.next_idxs.iter().position(|&next| next == n)?;
+
+            // Handle if thread is tail of a runqueue.
+            if let Some(rq) = self.tail.iter().position(|&tail| tail == n) {
+                if prev == n as usize {
+                    // Runqueue is empty now.
+                    self.tail[rq] = Self::sentinel();
+                    empty_runqueue = Some(rq as u8);
+                } else {
+                    self.tail[rq] = prev as u8;
+                }
+            }
+            self.next_idxs[prev] = self.next_idxs[n as usize];
+            self.next_idxs[n as usize] = Self::sentinel();
+            empty_runqueue
         }
 
         pub fn pop_head(&mut self, rq: u8) -> Option<u8> {
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -199,10 +199,10 @@ mod clist {
             let mut empty_runqueue = None;
 
             // Find previous thread in circular runqueue.
-            let prev = self.next_idxs.iter().position(|&next| next == n)?;
+            let prev = position(&self.next_idxs, n)?;
 
             // Handle if thread is tail of a runqueue.
-            if let Some(rq) = self.tail.iter().position(|&tail| tail == n) {
+            if let Some(rq) = position(&self.tail, n) {
                 if prev == n as usize {
                     // Runqueue is empty now.
                     self.tail[rq] = Self::sentinel();
@@ -248,6 +248,19 @@ mod clist {
             if let Some(head) = self.peek_head(rq) {
                 self.tail[rq as usize] = head;
             }
+        }
+    }
+
+    /// Helper function that is needed because hax doesn't support `Iterator::position` yet.
+    fn position<const N: usize>(slice: &[u8; N], search_item: u8) -> Option<usize> {
+        let mut i = 0;
+        while i < N && slice[i] != search_item {
+            i += 1;
+        }
+        if i < N {
+            Some(i)
+        } else {
+            None
         }
     }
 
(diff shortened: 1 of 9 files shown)


elenaf9:
> Seems like the hax issue in CI (lack of support for `let ... else` statements) is now [fixed upstream](https://github.com/hacspec/hax/pull/779). This is not picked up because we are currently pinning hax to an earlier version:
> 
> https://github.com/future-proof-iot/RIOT-rs/blob/2e84cce6f833a3ff0e0a2319a03804eb5d2c668b/.github/workflows/hax.yml#L14-L18

Thank you for pointing this out! I opened #378 to update our hax version.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -98,6 +105,21 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
         }
     }
 
+    /// Removes thread with pid `n` from runqueue number `rq`.
+    ///
+    /// Warning: will cause undefined behavior if the thread is not in
+    /// this runqueue.
```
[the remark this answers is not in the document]
elenaf9: Hmm, I am hesitant to make this unsafe. None of the "[unsafe superpowers](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#unsafe-superpowers)" are involved or anything like that. Instead, there are just certain pre-conditions, that we definitely meet in `riot-rs-threads`.
I could add a panic or early return if the thread isn't in the runqueue. That would require extra logic/ effort, but I'd prefer that over making it `unsafe`.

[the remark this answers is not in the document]
kaspar030: why does this cause UB?

Anyhow, "causes UB if user does not uphold invariants" is the definition of unsafe, no?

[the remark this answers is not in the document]
elenaf9: Hmm, I thought about it a bit more and changed it to completely work without the `RunqueueId` parameter: 38d4856d1b5c6b979df3019611f9ab610a9beac2.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -175,6 +194,27 @@ mod clist {
             }
         }
 
+        pub fn del(&mut self, n: u8) -> Option<u8> {
```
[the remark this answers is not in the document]
elenaf9: It returns `Some` with a runqueue ID to inform the caller that the runqueue is empty now. That way the bitcache can be updated in `RunQueue`. I added a comment.

src/riot-rs-threads/src/lib.rs
```diff
@@ -322,6 +360,24 @@ pub fn wakeup(thread_id: ThreadId) -> bool {
...
+///
+/// This might trigger a context switch.
+pub fn set_priority(thread_id: ThreadId, prio: u8) {
+    THREADS.with_mut(|mut threads| {
+        if threads.set_priority(thread_id, RunqueueId::new(prio)) {
+            schedule();
+        }
+    })
+}
+
```
elenaf9, line 380: I used `u8` instead of `RunqueueId` in the API here to match the `u8` used in the `thread_create*` functions, but maybe we should use `RunqueueId` in our API instead>

elenaf9:
Still blocked by https://github.com/hacspec/hax/issues/603.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -172,6 +191,31 @@ mod clist {
             }
         }
 
+        /// Removes a thread from the list.
+        ///
+        /// If the thread was the only thread in its runqueue, `Some` is returned
+        /// with the ID of the now empty runqueue.
+        pub fn del(&mut self, n: u8) -> Option<u8> {
```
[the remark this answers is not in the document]
elenaf9: The `CList` submodule is internal and only uses u8 instead of `RunqueueId`/ `ThreadId` because it would otherwise cause issues with hax. See https://github.com/future-proof-iot/RIOT-rs/pull/283#issuecomment-2096316464.

src/riot-rs-threads/src/lib.rs
```diff
@@ -172,6 +179,37 @@ impl Threads {
...
+    fn get_priority(&self, thread_id: ThreadId) -> Option<RunqueueId> {
+        self.is_valid_pid(thread_id)
+            .then(|| self.get_unchecked(thread_id).prio)
+    }
+
+    /// Changes the priority of a thread.
+    ///
+    /// Returns true if the thread is in the runqueue and the order might have
+    /// changed.
```
elenaf9, line 191:
```suggestion
    /// Returns the information if the scheduler should be invoked because the runqueue order 
    /// might have changed.
    /// `false` if the thread isn't in the runqueue (in which case the priority is still changed) 
    /// or if the new priority equals the current one.
```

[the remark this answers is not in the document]
elenaf9: Wdyt of https://github.com/future-proof-iot/RIOT-rs/pull/370#discussion_r1768128158?

src/riot-rs-threads/src/lib.rs
elenaf9: > Are all possible values of `RunqueueId` valid priorities?

Unfortunately not. `RunqueueId` is an u8 and thus can be up to 255, but:
1. `riot_rs_threads::SCHED_PRIO_LEVELS` can be lower than that, which is the max # of prio levels we support
2. "runqueue numbers fit in usize bits (supporting max 32 priority levels)": see https://github.com/future-proof-iot/RIOT-rs/pull/414#discussion_r1758503064

elenaf9:
Rebased & addressed reviews. I also refactored the `threads-dynamic-prios` example into a test instead, because I think it's a simple API that doesn't require a separate example, but it's useful to have a test for the logic.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -214,6 +258,17 @@ mod clist {
         }
     }
 
+    /// Helper function that is needed because hax doesn't support `Iterator::position` yet.
+    fn position<const N: usize>(slice: &[u8; N], search_item: u8) -> Option<usize> {
+        let mut index = None;
+        for i in 0..N {
+            if slice[i] == search_item {
+                index = Some(i);
```
kaspar030, line 266:
```suggestion
                index = Some(i);
                break;
```

kaspar030: (... or just `return Some(i)`, drop the index variable, and have a `None` in line 269. preference.)

src/riot-rs-threads/src/lib.rs
```diff
@@ -108,9 +108,16 @@ impl Threads {
...
-    //     &self.threads[thread_id as usize]
-    // }
+    /// Returns access to any thread data.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `thread_id` is >= [`THREADS_NUMOF`].
+    /// If the thread for this `thread_id` is in an invalid state, the
+    /// data in the returned [`Thread`] is undefined, i.e. empty or outdated.
+    fn get_unchecked(&self, thread_id: ThreadId) -> &Thread {
```
kaspar030, line 118: should this be unsafe? (is the std lib `get_unchecked()` usually unsafe?)

tests/threading-dynamic-prios/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Elena Frank <elena.frank@proton.me>"]
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+riot-rs = { path = "../../src/riot-rs", features = ["threading"] }
+riot-rs-boards = { path = "../../src/riot-rs-boards" }
+portable-atomic = "1.6.0"
```
kaspar030, line 12:
```suggestion
portable-atomic.workspace = true
```

kaspar030 (changes requested):

src/riot-rs-runqueue/src/runqueue.rs
elenaf9: Early returns from loops are unfortunately not supported by hax yet, that's why this work-around is needed :/

src/riot-rs-threads/src/lib.rs
elenaf9: It's an non-mutable version of the existing `get_unchecked_mut` function. 
I'm, not sure if it has to be made unsafe because we explicitly documented that it can panic. `core::ops::Index::index` is also not unsafe. 
On the other hand, it could cause undefined behavior if the thread_id is not out-of-bound, but the thread in an invalid state.
I guess the correct way would be to make it `unsafe` and then bubble up the unsafe to the user (e.g. so that `thread_flags::set` is also unsafe, since this internally calls `get_unchecked_mut`)  and/ or add checks everywhere that the thread id is valid. But I assume that would impact performance quite a bit.

kaspar030: Ok, so get_unchecked(...) panics if thread_id is out-of-bounds. That doesn't make it unsafe.
The thread object reference is always valid memory, so no unsafe there, either.
The object might be in a "weird" state, but that doesn't make it unsafe, either. I don't see how this could lead to "Undefined Behavior". It might lead to a thread in Invalid state to be added to a runqueue, but that's not "Undefined Behavior".

=> IMO, no unsafe necessary.

Down the line, we might want to explore if it is worth the effort to introduce a "ValidThreadId" type to make API misuse less possible, or if we can contain our reasoning to internals.

elenaf9:
Rebased & addressed the reviews. 
Per https://github.com/future-proof-iot/RIOT-rs/pull/370#discussion_r1758778745, `get_priority` and `set_priority` now use `RunqueueId`s instead of `u8`.
The `thread_create*` functions should probably also use the `RunqueueId`, but that's out of scope for this PR, especially because it also involves changes in our macros.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -24,6 +24,13 @@ impl From<RunqueueId> for usize {
     }
 }
 
+impl Deref for RunqueueId {
+    type Target = u8;
+    fn deref(&self) -> &Self::Target {
```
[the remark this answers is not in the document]
elenaf9: Obsolete because we are now using `RunqueueId` in our priority API directly, so the `deref`/ `get` for the `u8` isn't needed anymore.

src/riot-rs-runqueue/src/runqueue.rs
elenaf9: Done in 3fcff18.

kaspar030 (approved):

kaspar030:
LGTM!

PR #374, opened 2024-08-05
ci: set up cargo-deny checking

ROMemories:
# Description

cargo-deny allows to restrict allowed crates based on different criteria, including licenses and sources.
It also warns about security advisories and duplicated dependencies (which could reduce binary size if de-duplicated).

## Issues/PRs references

Depends on #373

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -132,6 +132,23 @@ jobs:
       - name: rust cache
         uses: Swatinem/rust-cache@v2
 
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-binstall
+
+      # Should run as soon as possible to avoid using unwanted crates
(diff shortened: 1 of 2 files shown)


kaspar030:
offline discussion: find way to handle cargo-deny default config and our project specific settings.

kaspar030 (changes requested):
- handle project specifics in a more maintainable way

kaspar030:
> > * handle project specifics in a more maintainable way
> 
> I modified the `deny.toml` configuration to only keep the keys with non-default values as we discussed. I also kept the comments provided by `cargo-deny` because the effect of some keys may otherwise be quite cryptic.

Perfect!

kaspar030 (approved):

kaspar030:
pls squash & merge

PR #381, opened 2024-08-14
feat(threads): sort `ThreadList` by priority

elenaf9:
# Description

Insert new threads in `ThreadList` based on priorities, so that the highest priority thread is the head.
Within a priority the threads are ordered in FIFO.

This reduces priority inversions, and will facilitate proper priority inheritance in the future.

The drawback of this change is that inserting a new Id in the `ThreadList` is now more complex. Not sure if we can avoid this in any way.

## Issues/PRs references

Alternative to #371, which implements the `ThreadList` as pure FIFO.

## Change checklist

### DIFF
--- a/src/riot-rs-threads/src/threadlist.rs
+++ b/src/riot-rs-threads/src/threadlist.rs
@@ -1,6 +1,6 @@
 use critical_section::CriticalSection;
 
-use crate::{ThreadId, ThreadState, THREADS};
+use crate::{thread::Thread, ThreadId, ThreadState, THREADS};
 
 /// Manages blocked [`super::Thread`]s for a resource, and triggering the scheduler when needed.
 #[derive(Debug, Default)]
@@ -18,10 +18,22 @@ impl ThreadList {
     /// Puts the current (blocked) thread into this [`ThreadList`] and triggers the scheduler.
     pub fn put_current(&mut self, cs: CriticalSection, state: ThreadState) {
         THREADS.with_mut_cs(cs, |mut threads| {
-            let thread_id = threads.current_thread.unwrap();
-            threads.thread_blocklist[usize::from(thread_id)] = self.head;
-            self.head = Some(thread_id);
-            threads.set_state(thread_id, state);
+            let &mut Thread { pid, prio, .. } = threads.current().unwrap();
+            let mut curr = None;
+            let mut next = self.head;
+            while let Some(n) = next {
+                if threads.get_unchecked_mut(n).prio < prio {
+                    break;
+                }
+                curr = next;
+                next = threads.thread_blocklist[usize::from(n)];
+            }
+            threads.thread_blocklist[usize::from(pid)] = next;
+            match curr {
+                Some(curr) => threads.thread_blocklist[usize::from(curr)] = Some(pid),
+                _ => self.head = Some(pid),
+            }
+            threads.set_state(pid, state);
             crate::schedule();
         });
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4440,6 +4440,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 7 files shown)


src/riot-rs-threads/src/threadlist.rs
```diff
@@ -18,10 +18,22 @@ impl ThreadList {
     /// Puts the current (blocked) thread into this [`ThreadList`] and triggers the scheduler.
     pub fn put_current(&mut self, cs: CriticalSection, state: ThreadState) {
         THREADS.with_mut_cs(cs, |mut threads| {
-            let thread_id = threads.current_thread.unwrap();
-            threads.thread_blocklist[usize::from(thread_id)] = self.head;
-            self.head = Some(thread_id);
-            threads.set_state(thread_id, state);
+            let &mut Thread { pid, prio, .. } = threads.current().unwrap();
```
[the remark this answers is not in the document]
elenaf9: Did a separate PR because this is happening in multiple places in `riot-rs-threads`: #382.

src/riot-rs-threads/src/threadlist.rs
```diff
@@ -18,10 +18,22 @@ impl ThreadList {
...
+                curr = next;
+                next = threads.thread_blocklist[usize::from(n)];
+            }
+            threads.thread_blocklist[usize::from(pid)] = next;
+            match curr {
+                Some(curr) => threads.thread_blocklist[usize::from(curr)] = Some(pid),
+                _ => self.head = Some(pid),
+            }
+            threads.set_state(pid, state);
             crate::schedule();
```
kaspar030, line 37: only schedule if the thread became the new head?

examples/threading-lock/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use embassy_time::{Duration, Timer};
+use riot_rs::{debug::log::*, thread::lock::Lock, thread::ThreadId};
+
+static LOCK: Lock = Lock::new();
+
+#[riot_rs::task(autostart)]
+async fn task_with_timer() {
+    info!("Task waits for lock");
+    LOCK.acquire();
```
kaspar030, line 14: This lock would put the current thread to sleep. So when called from a thread executor, this blocks the whole executor. if called in an interrupt executor, it violates the huge `**NOTE**` in the `acquire()`docs...

src/riot-rs-threads/src/threadlist.rs
elenaf9: I think we always have to schedule, because the current thread, that was previously running, is now blocked.

kaspar030:
Nice! Changes to `riot-rs-threads` look fine.

Can we turn the example into a test (in `tests/`), and add some logic after each `acquire()` that e.g., checks the value of an atomic uint if it contains the right value, so we can test the order programatically? (e.g., first thread expects a zero, then sets bit zero. second expects bit zero, then ORs in bit one. third one expects bits zero and one, ...)

If you feel changing that now is out of scope, I'd suggest dropping the example and we add tests later.

elenaf9:
> Can we turn the example into a test (in `tests/`), and add some logic after each `acquire()` that e.g., checks the value of an atomic uint if it contains the right value, so we can test the order programatically? (e.g., first thread expects a zero, then sets bit zero. second expects bit zero, then ORs in bit one. third one expects bits zero and one, ...)

Wdyt of 67a8e9c3729d7d899a1c5772d2d5d6fd9f2dc4c9?

examples/threading-lock/src/main.rs
elenaf9: Good point. The idea was to give the other threads a chance to run, while holding the lock, but I solved it differently now.

kaspar030:
LGTM, please squash!

PR #382, opened 2024-08-19
refactor(threads): improve panic handling and docs

elenaf9:
# Description

Add some missing documentation about possible panics in `riot-rs-threads` when a function is expected to be called only inside a thread context.
Avoid panics where possible through early returns. 

Also adds some documentation for `Channel`.

## Issues/PRs references

Extracted from https://github.com/future-proof-iot/RIOT-rs/pull/381#discussion_r1721324000.

## Open Questions

- @ROMemories you proposed in https://github.com/future-proof-iot/RIOT-rs/pull/381#discussion_r1721324000 to use `NOTE(no-panic)`-style comments to document why an `unwrap` can't panic. I've instead used `expect`, because we a) don't know that it will never panic (it's the callers responsibility) and b) it prints the comment as part of the panic message. Wdyt?
 
- > Avoid panics where possible through early returns. 

This concerns the functions `yield_same` and `sleep`. If called from outside a thread context, the operations are now no-ops, instead of causing panics. Do you agree that we should prefer no-ops over panics as long as it doesn't cause any undefined behavior?

## Change checklist

### DIFF
--- a/src/riot-rs-threads/src/channel.rs
+++ b/src/riot-rs-threads/src/channel.rs
@@ -14,12 +14,13 @@ enum ChannelState {
     ReceiversWaiting(ThreadList),
 }
 
-pub struct Channel<T: Copy + Send> {
+/// Blocking channel for sending data between threads.
+pub struct Channel<T> {
     state: UnsafeCell<ChannelState>,
     phantom: core::marker::PhantomData<T>,
 }
 
-unsafe impl<T: Copy + Send> Sync for Channel<T> {}
+unsafe impl<T> Sync for Channel<T> {}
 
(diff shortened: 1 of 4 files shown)


elenaf9:
> maybe it would make sense to add debug_assert!s with a developer-friendly message, and unwrap otherwise

What would be the advantage of this compared to using `expect`?

According to the [Rust docs about Converting Errors into Panics](https://doc.rust-lang.org/core/error/index.html#converting-errors-into-panics):
>  Of the two, `expect` is generally preferred since its `msg` field allows you to convey your intent and assumptions which makes tracking down the source of a panic easier.

kaspar030:
lgtm!

PR #388, opened 2024-08-21
refactor!: remove the embassy module and all its uses

ROMemories:
# Description

The first commits keeps the `embassy` module for backward compatibility, but hides it from documentation.
The second one removes all uses and removes the `embassy` module entirely.

## Open Questions

~~Should we rewrite every uses of the `embassy` modules as part of this PR, or shall we do this gradually?~~ There were fewer uses than I anticipated, so I just went with it and removed them all.

## Change checklist


kaspar030:
LGTM.

Issue #243, opened 2024-04-10
tracking: SMP Support 


elenaf9:
# Summary

Add support for symmetric multiprocessing (SMP), i.e. platforms with multiple symmetric cores. 
An example for such a platform is the Raspberry RP2040, with Dual ARM Cortex-M0+ cores.

## Subtasks for basic a implementation 

A first implementation for the RP2040 is drafted in #241. There is quite a lot still missing there.
The below list includes some central tasks, but will grow while I am working on this.

- [x] ~#244~
- [x] Multicore support in the scheduler: #397 
  - [x] Schedule `n` highest-priority threads on the `n` cores
  - [x] Trigger scheduler of other cores
  - [x] Idle thread for each core
  - [x] Affinity Masks 
  - [x] Optimize scheduler invocations to reduce context switches and priority inversion
  - [x] Supported chips:
    - [x] RP2040 
    - [x] ESP32S3 
- [ ] Optimize locking primitives (spinlocks for shared resources, scheduler lock, ...)
  - Test different levels of locking (fine-grained vs one global lock)
  - Use independent internal critical section 

Additional features (to be discussed):
- [ ] Priority inheritance (for threads holding a shared resource). Started in #398. (_Is this orthogonal?_) 

## Benchmarks

The implementation should be tested with benchmarks to compare the performance with the single-core version.
Off the top of my head, some interesting metrics would be:
- Response time
- Makespan
- Priority inversions
- ...

_Edit: a set of benchmarks is currently implemented outside of this repo: [elenaf9/RIOT-rs-application](https://github.com/elenaf9/RIOT-rs-applications). Selected benchmarks will be moved in this repo eventually_).

kaspar030:
I consider this done, #389 is orthogonal.

PR #363, opened 2024-07-11
fix(atomics): clarify how we select portable-atomic's atomic implementations

ROMemories:
# Description

Tries to clarify how and why we select [`portable-atomic`](https://docs.rs/portable-atomic/latest/portable_atomic/)'s atomic implementations via its Cargo features.

## Issues/PRs references

N/A

## Open Questions

None

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](../book/src/coding-conventions.md).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #397, opened 2024-09-05
feat(threads): multicore support 

elenaf9:
# Description

This PR adds support for symmetric multiprocessing (SMP) for `riot-rs-threads`. It's a refactor and continuation of #241.
Right now, only the RP2040 is supported. I am planning to also add support for esp32s3 in a follow-up PR.

The implementation follows a global scheduling approach where the highest _n_ ready threads are scheduled on the _n_ available cores. 
The main change in the scheduler logic is that it now removes the current thread from the runqueue, which is necessary so that a thread doesn't get picked twice. This also means that the thread has to be added to the runqueue again each time the scheduler is triggered (if it's still running). 
This introduces some overhead if the scheduler is triggered unnecessarily. The PR therefore also includes optimizations that make sure that the scheduler is only triggered when a context switch is actually needed.
The implementation furthermore also supports affinity masks, which allow restricting a thread to certain cores.

All multicore logic is feature-gated, so that the single-core implementation largely remains the same. I still notice some difference in preliminary benchmark data, which I am still investigating.

## Issues/PRs references

Tracking issue: #243 

Depends on:
- #455
- #456
- #457
- #458
- #459
- #460

### Idle threads

If there are no ready threads in the runqueue, the current (= on main) `sched` implementation `WFI`s in a loop. The context of the previous thread is only saved **after** a new thread is ready and the context is actually switched.
This causes some issues on multicore:
1. there is a race condition where it can happen that the still "unsaved" thread becomes ready again, and scheduled on the other core. That core would then pick up an outdated state.
2. we can't WFI inside the critical section because it would also block the other core

By adding idle threads, both of these issues are fixed and we furthermore avoid footguns related to interrupt priorities (e.g. right now we require that all embassy interrupts have a larger priority than `PendSV` so that they can preempt our scheduler when it's stuck in WFI). Wdyt?

## Open TODOS

- [x] Drop commits that have later been reverted (I kept them in the history for now for transparency)
- [ ] Post preliminary benchmark data 
   - compare: main vs this PR (`multicore` not enabled) vs this PR (`multicore` enabled) 
   - boards: RP2040 and nrf52840
   - benchmarks: `bench_sched_yield`, `bench_sched_flags` (new)

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,11 +73,11 @@ embassy-usb = { version = "0.3", default-features = false }
 embedded-hal = { version = "1.0.0", default-features = false }
 embedded-hal-async = { version = "1.0.0", default-features = false }
 
-esp-alloc = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1", default-features = false }
-esp-hal = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1", default-features = false }
-esp-hal-embassy = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1", default-features = false }
+esp-alloc = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1+pr2377", default-features = false }
+esp-hal = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1+pr2377", default-features = false }
+esp-hal-embassy = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1+pr2377", default-features = false }
 esp-println = "0.11.0"
-esp-wifi = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1", default-features = false }
+esp-wifi = { git = "https://github.com/kaspar030/esp-hal", branch = "for-riot-rs-2024-10-16-0.21.1+pr2377", default-features = false }
 
 linkme = { version = "0.3.21", features = ["used_linker"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,7 @@ once_cell = { version = "=1.19.0", default-features = false, features = [
   "critical-section",
 ] }
 paste = { version = "1.0" }
+rp-pac = { version = "6.0", default-features = false }
 static_cell = { version = "2.0.0", features = ["nightly"] }
 
 [profile.dev]
--- a/examples/threading-multicore/Cargo.toml
+++ b/examples/threading-multicore/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "threading-multicore"
+version = "0.1.0"
+authors = ["Elena Frank <elena.frank@proton.me>"]
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+riot-rs = { path = "../../src/riot-rs", features = ["threading"] }
+riot-rs-boards = { path = "../../src/riot-rs-boards" }
--- a/examples/threading-multicore/Cargo.toml
+++ b/examples/threading-multicore/Cargo.toml
@@ -7,5 +7,8 @@ license.workspace = true
 publish = false
 
 [dependencies]
-riot-rs = { path = "../../src/riot-rs", features = ["threading"] }
+riot-rs = { path = "../../src/riot-rs", features = [
+  "threading",
+  "core-affinity",
+] }
 riot-rs-boards = { path = "../../src/riot-rs-boards" }
--- a/examples/threading-multicore/Cargo.toml
+++ b/examples/threading-multicore/Cargo.toml
@@ -7,8 +7,5 @@ license.workspace = true
 publish = false
 
 [dependencies]
-riot-rs = { path = "../../src/riot-rs", features = [
-  "threading",
-  "core-affinity",
-] }
+riot-rs = { path = "../../src/riot-rs", features = ["core-affinity"] }
 riot-rs-boards = { path = "../../src/riot-rs-boards" }
--- a/examples/threading-multicore/README.md
+++ b/examples/threading-multicore/README.md
@@ -1,0 +1,15 @@
+# threading multicore
+
+## About
+
+This application demonstrates basic threading on multicore.
+
+## How to run
+
+In this folder, run
+
+    laze build -b rpi-pico run
+
+The application will start two threads that each print their ID and the core that they are running on, before
+entering a busy loop.
+The thread that is started first in pinned to Core 1 using core affinities.
--- a/src/riot-rs-rt/Cargo.toml
+++ b/src/riot-rs-rt/Cargo.toml
@@ -17,16 +17,20 @@ riot-rs-utils = { workspace = true }
 rtt-target = { version = "0.5.0", optional = true }
 
 [target.'cfg(context = "cortex-m")'.dependencies]
-cortex-m = { workspace = true, features = ["critical-section-single-core"] }
+cortex-m = { workspace = true }
 cortex-m-rt = { workspace = true }
 
 [target.'cfg(context = "rp2040")'.dependencies]
 # Use critical-section to provide an atomic CAS implementation.
 portable-atomic = { workspace = true, features = ["critical-section"] }
+embassy-rp = { workspace = true, features = ["critical-section-impl"] }
 
 [target.'cfg(context = "esp")'.dependencies]
 esp-hal = { workspace = true, default-features = false }
 portable-atomic = { workspace = true }
+
+[target.'cfg(context = "nrf")'.dependencies]
+cortex-m = { workspace = true, features = ["critical-section-single-core"] }
 
 [features]
 #default = ["threading"]
--- a/src/riot-rs-rt/Cargo.toml
+++ b/src/riot-rs-rt/Cargo.toml
@@ -21,9 +21,8 @@ cortex-m = { workspace = true }
 cortex-m-rt = { workspace = true }
 
 [target.'cfg(context = "rp2040")'.dependencies]
-# Use critical-section to provide an atomic CAS implementation.
+embassy-rp = { workspace = true, optional = true }
 portable-atomic = { workspace = true, features = ["critical-section"] }
-embassy-rp = { workspace = true, features = ["critical-section-impl"] }
 
 [target.'cfg(context = "esp")'.dependencies]
 esp-hal = { workspace = true, default-features = false }
@@ -32,14 +31,18 @@ portable-atomic = { workspace = true }
 [target.'cfg(context = "nrf")'.dependencies]
 cortex-m = { workspace = true, features = ["critical-section-single-core"] }
 
+[target.'cfg(context = "stm32")'.dependencies]
+cortex-m = { workspace = true, features = ["critical-section-single-core"] }
+
 [features]
-#default = ["threading"]
 threading = ["dep:riot-rs-threads"]
 
 debug-console = ["riot-rs-debug/debug-console"]
 executor-single-thread = []
 silent-panic = []
 _panic-handler = []
+single-core = ["cortex-m/critical-section-single-core"]
+multicore = ["embassy-rp/critical-section-impl"]
 
 # internal
 # These features are used by `build.rs`, which doesn't "see" context
--- a/src/riot-rs-rt/Cargo.toml
+++ b/src/riot-rs-rt/Cargo.toml
@@ -42,7 +42,7 @@ executor-single-thread = []
 silent-panic = []
 _panic-handler = []
 single-core = ["cortex-m/critical-section-single-core"]
-multicore = ["embassy-rp/critical-section-impl"]
+multi-core = ["embassy-rp/critical-section-impl"]
 
 # internal
 # These features are used by `build.rs`, which doesn't "see" context
--- a/src/riot-rs-rt/src/cortexm.rs
+++ b/src/riot-rs-rt/src/cortexm.rs
@@ -198,11 +198,4 @@ pub fn init() {
             .vtor
             .write(&__RESET_VECTOR as *const _ as u32 - 4)
     };
-
-    // make sure PendSV has a low priority
-    unsafe {
-        use cortex_m::peripheral::scb::SystemHandler;
-        let mut p = cortex_m::Peripherals::take().unwrap();
-        p.SCB.set_priority(SystemHandler::PendSV, 0xFF);
-    }
 }
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -140,6 +140,23 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
         self.iter_from(next, prio).find(predicate)
     }
 
+    /// Pop the thread that should run next.
+    ///
+    /// Pops the next runnable thread of
+    /// the runqueue with the highest index.
+    pub fn pop_next(&mut self) -> Option<ThreadId> {
+        let rq_ffs = ffs(self.bitcache);
+        if rq_ffs == 0 {
+            return None;
+        }
+        let rq = (rq_ffs - 1) as u8;
+        let head = self.queues.pop_head(rq).map(ThreadId::new);
+        if self.queues.is_empty(rq) {
+            self.bitcache &= !(1 << rq);
+        }
+        head
+    }
+
     /// Advances runqueue number `rq`.
     ///
     /// This is used to "yield" to another thread of *the same* priority.
@@ -149,6 +166,12 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
     pub fn advance(&mut self, rq: RunqueueId) -> bool {
         debug_assert!((usize::from(rq)) < N_QUEUES);
         self.queues.advance(rq.0)
+    }
+
+    /// Checks if a runqueue is empty.
+    pub fn is_empty(&mut self, rq: RunqueueId) -> bool {
+        debug_assert!((rq.0 as usize) < N_QUEUES);
+        self.queues.is_empty(rq.0)
     }
 
     /// Returns an iterator over the [`RunQueue`], starting after thread `start` in runqueue `rq`.
--- a/src/riot-rs-threads/src/arch/cortex_m.rs
+++ b/src/riot-rs-threads/src/arch/cortex_m.rs
@@ -1,10 +1,7 @@
-use super::Arch;
-use crate::Thread;
+use crate::{cleanup, Arch, Thread, THREADS};
 use core::arch::asm;
 use core::ptr::write_volatile;
-use cortex_m::peripheral::SCB;
-
-use crate::{cleanup, THREADS};
+use cortex_m::peripheral::{scb::SystemHandler, SCB};
 
 #[cfg(not(any(armv6m, armv7m, armv8m)))]
 compile_error!("no supported ARM variant selected");
@@ -63,7 +60,20 @@ impl Arch for Cpu {
 
     #[inline(always)]
     fn start_threading() {
+        unsafe {
+            // Make sure PendSV has a low priority.
+            let mut p = cortex_m::Peripherals::steal();
+            p.SCB.set_priority(SystemHandler::PendSV, 0xFF);
+        }
         Self::schedule();
+    }
+
+    fn wfi() {
+        cortex_m::asm::wfi();
+
+        // see https://cliffle.com/blog/stm32-wfi-bug/
+        #[cfg(context = "stm32")]
+        cortex_m::asm::isb();
     }
 }
 
@@ -183,23 +193,20 @@ unsafe extern "C" fn PendSV() {
 // TODO: make arch independent, or move to arch
 #[no_mangle]
 unsafe fn sched() -> u128 {
+    THREADS.with_mut(|mut threads| {
+        let Some(thread) = threads.current() else {
+            return;
+        };
+        if thread.state == crate::ThreadState::Running {
+            let prio = thread.prio;
+            let pid = thread.pid;
+            threads.runqueue.add(pid, prio);
+        }
+    });
     loop {
-        if let Some(res) = critical_section::with(|cs| {
-            let threads = unsafe { &mut *THREADS.as_ptr(cs) };
-            let next_pid = match threads.runqueue.get_next() {
-                Some(pid) => pid,
-                None => {
-                    cortex_m::asm::wfi();
+        if let Some(res) = THREADS.with_mut(|mut threads| {
 
-                    // see https://cliffle.com/blog/stm32-wfi-bug/
-                    #[cfg(context = "stm32")]
-                    cortex_m::asm::isb();
-
-                    // this fence seems necessary, see #310.
-                    core::sync::atomic::fence(core::sync::atomic::Ordering::Acquire);
-                    return None;
-                }
-            };
+            let next_pid = threads.runqueue.pop_next()?;
 
             // `current_high_regs` will be null if there is no current thread.
             // This is only the case once, when the very first thread starts running.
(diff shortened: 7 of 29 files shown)


src/riot-rs-rt/src/cortexm.rs
```diff
@@ -198,11 +198,4 @@ pub fn init() {
             .vtor
             .write(&__RESET_VECTOR as *const _ as u32 - 4)
     };
-
-    // make sure PendSV has a low priority
-    unsafe {
-        use cortex_m::peripheral::scb::SystemHandler;
-        let mut p = cortex_m::Peripherals::take().unwrap();
-        p.SCB.set_priority(SystemHandler::PendSV, 0xFF);
-    }
```
elenaf9, line 207: Moved to `riot_rs_threads::Cpu::start_threading`.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -98,34 +98,136 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
...
+/// priority runqueue after circling through the queue once.
+pub struct RunQueueIter<'a, const N_QUEUES: usize, const N_THREADS: usize> {
+    queues: &'a clist::CList<N_QUEUES, N_THREADS>,
+    // Predecessor in the circular runqueue list.
+    prev: u8,
+    // Head of the currently iterated runqueue.
+    rq_head: Option<u8>,
+    // Bitcache with the remaining runqueues that have to be iterated.
+    bitcache: usize,
+}
```
elenaf9, line 207: I slightly dislike that I had to add this rather complex iterator here.
It's needed for `core-affinity` because we need to "get the next thread from the runqueue that fulfills condition x". The best way to do that that I came up with is with `Iterator` + `filter`. But I am happy to hear alternative ideas.

src/riot-rs-threads/src/arch/cortex_m.rs
```diff
@@ -162,24 +198,37 @@ unsafe extern "C" fn PendSV() {
...
+        let threads = unsafe { &mut *THREADS.as_ptr(cs) };
+        let Some(thread) = threads.current() else {
+            return;
+        };
+        if thread.state == crate::ThreadState::Running {
+            let prio = thread.prio;
+            let pid = thread.pid;
+            threads.runqueue.add(pid, prio);
+        }
+    });
```
elenaf9, line 214: Needed because the current thread is removed from the runqueue on multicore,

src/riot-rs-threads/src/lib.rs
```diff
@@ -226,14 +369,20 @@ pub fn thread_create<T: Arguable + Send>(
...
 }
 
 /// Low-level function to create a thread without argument
-pub fn thread_create_noarg(func: fn(), stack: &'static mut [u8], prio: u8) -> ThreadId {
-    unsafe { thread_create_raw(func as usize, 0, stack, prio) }
+pub fn thread_create_noarg(
+    func: fn(),
+    stack: &'static mut [u8],
+    prio: u8,
+    core_affinity: Option<CoreAffinity>,
```
elenaf9, line 383: Should `CoreAffinity` really awlays be part of the public API? Or instead have a `_with_affinity` version of each of these 3 thread-creation function?

Cargo.toml
```diff
@@ -109,7 +112,7 @@ embassy-executor = { git = "https://github.com/kaspar030/embassy", branch = "for
 embassy-hal-internal = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-240605" }
 embassy-nrf = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-240605" }
 embassy-net = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-240605" }
-embassy-rp = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-240605" }
+embassy-rp = { git = "https://github.com/elenaf9/embassy", branch = "for-riot-rs-multicore-240821" }
```
elenaf9, line 115: The fork (which itself is a fork of `kaspar:for-riot-rs-240605`) disables the `SIO_IRQ_PROC1` interrupt  handler (for FIFO messages between the cores) in `embassy_rp`.
It would conflict with our own handler. They use it to start/resume core 1, we need it to trigger the scheduler on the other core.

src/riot-rs-threads/src/smp/rp2040.rs
```diff
@@ -0,0 +1,76 @@
+use crate::arch::{Arch, Cpu};
```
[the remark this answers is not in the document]
elenaf9: We need to import it because we call below `Cpu::start_threading()` (and `start_threading` is part of the `Arch` trait).

src/riot-rs-threads/src/smp/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn schedule_on_core(id: CoreId);
+}
+
+cfg_if::cfg_if! {
+    if #[cfg(context = "rp2040")] {
+        mod rp2040;
+        pub use rp2040::Chip;
+    }
+    else {
+        use crate::{Arch, Cpu};
```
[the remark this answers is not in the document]
elenaf9: We need to import it because we call below `Cpu::schedule()` (and `schedule` is part of the Arch trait).

src/riot-rs-threads/src/smp/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+/// Trigger the scheduler on core `id`.
+pub fn schedule_on_core(id: CoreId) {
+    Chip::schedule_on_core(id)
+}
+
+/// Affinity mask that defines on what cores a thread can be scheduled.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
elenaf9: Hmm, I'm not sure if `PartialOrd` and `Ord` make sense for a bitmask, at least not in this case here.

examples/threading-multicore/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+publish = false
+
+[dependencies]
+riot-rs = { path = "../../src/riot-rs", default-features = false, features = [
+  "threading",
+  "multicore",
+  "core-affinity",
+] }
+riot-rs-boards = { path = "../../src/riot-rs-boards" }
+riot-rs-rt = { path = "../../src/riot-rs-rt", features = ["_panic-handler"] }
```
elenaf9, line 15: Should `_panic-handler` be a feature on `riot-rs` so we don't have to import `riot-rs-rt` here?

elenaf9:
>  I am planning to also add support for esp32s3 in a follow-up PR.

I was able to also add multicore support for the dual-core esp32s3, based on #399!

It's in a separate branch for now, until #399 is merged. 
- branch [`elenaf9:multicore/v2_esp32s3`](https://github.com/elenaf9/RIOT-rs/tree/multicore/v2_esp32s3)
- diff: [f38b142...e5d64c8](https://github.com/future-proof-iot/RIOT-rs/compare/f38b142...e5d64c8)

I still need to do some more testing; the benchmarks that I am using for rp2040 only support `cortex_m` right now.

examples/threading-multicore/Cargo.toml
elenaf9: I've added `single-core` to the default features of `riot-rs`, so for `multicore`we have to disable it. The alternative would be to not make `single-core` a default, but then it'd have to enabled manually every time.

elenaf9: Solved now with laze instead: 78171d80b228bd4322d7f6ede5c85c1180ccf475.

elenaf9:
I have now rebased my branch and cleaned up my commit history quite a lot.
 (I have saved the old history with all WIP intermediate steps in a separate branch [elenaf9:multicore/v2_save](https://github.com/elenaf9/RIOT-rs/commits/WIP/multicore/v2/)).

Additionally, with #399 merged, the last commit now also adds SMP support for the dual-core xtensa ESP32S3 :tada:

src/riot-rs-runqueue/src/runqueue.rs
elenaf9: And unfortunately it also fails the hax check because`Iterator` isn't supported. I guess i could work around it to just manually implement a filter function for the runqueue...

elenaf9:
Rebased again to include the recently merged dynamic priority feature & adapt it for multicore.

src/riot-rs-rt/Cargo.toml
```diff
@@ -44,8 +44,14 @@ debug-console = ["riot-rs-debug/debug-console"]
...
-critical-section-single-core = ["cortex-m?/critical-section-single-core"]
-multicore = ["embassy-rp?/critical-section-impl"]
+critical-section-single-core = [
+  "cortex-m?/critical-section-single-core",
+  "portable-atomic/unsafe-assume-single-core",
+]
+multicore = [
+  "embassy-rp?/critical-section-impl",
+  "portable-atomic/critical-section",
+]
```
elenaf9, line 54: cc @ROMemories  as the author of #363. 

Any reason why the  `critical-section` feature instead of `unsafe-assume-single-core` should be picked if the RP2040 is used as single-core?
 If I understand the [docs](https://docs.rs/portable-atomic/latest/portable_atomic/#optional-features-unsafe-assume-single-core) correctly, it has better performance and is safe as long as it's single-core and runs in privileged mode (which RIOT-rs always does).

src/riot-rs-rt/Cargo.toml
```diff
@@ -15,19 +15,27 @@ riot-rs-debug.workspace = true
 riot-rs-threads = { path = "../riot-rs-threads", optional = true }
 riot-rs-utils = { workspace = true }
 rtt-target = { version = "0.5.0", optional = true }
+cortex-m = { workspace = true, optional = true }
+embassy-rp = { workspace = true, optional = true }
```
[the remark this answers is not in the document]
elenaf9: You're right.I though it was necessary because of the below feature selection, but also works without.

src/riot-rs-rt/Cargo.toml
```diff
@@ -15,19 +15,27 @@ riot-rs-debug.workspace = true
 riot-rs-threads = { path = "../riot-rs-threads", optional = true }
 riot-rs-utils = { workspace = true }
 rtt-target = { version = "0.5.0", optional = true }
+cortex-m = { workspace = true, optional = true }
```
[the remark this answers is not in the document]
elenaf9: Remove now anyway, see https://github.com/future-proof-iot/RIOT-rs/pull/397#discussion_r1796534547.

src/riot-rs-rt/Cargo.toml
elenaf9: Hmm, so turns out this is unfortunately not possible, because `unsafe_assume-single-core` doesn't work on platforms that do support CAS. So we can't select it here through the feature. and we obviously can't enable it just for the `rp2040 context`, because then it fails on multicore.
Any idea how to solve this?
I dropped the commit for now.

Cargo.toml
```diff
@@ -126,7 +128,7 @@ embassy-executor = { git = "https://github.com/kaspar030/embassy", branch = "for
 embassy-hal-internal = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-2024-09-19" }
 embassy-nrf = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-2024-09-19" }
 embassy-net = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-2024-09-19" }
-embassy-rp = { git = "https://github.com/kaspar030/embassy", branch = "for-riot-rs-2024-09-19" }
+embassy-rp = { git = "https://github.com/elenaf9/embassy", branch = "for-riot-rs-2024-09-19_multicore" }
```
kaspar030, line 131: So this contains a patch disabling some interrupts.  why is that necessary? can we make that a feature in `embassy-rp` and PR it to our fork at kaspar030/embassy?

src/riot-rs-runqueue/src/runqueue.rs
kaspar030: Moved to #460.

deny.toml
```diff
@@ -56,4 +56,4 @@ allow-git = [
 
 [sources.allow-org]
 # github.com organizations to allow git sources for
-github = ["embassy-rs", "esp-rs", "kaspar030"]
+github = ["embassy-rs", "esp-rs", "kaspar030", "elenaf9"]
```
kaspar030, line 59: reminder: drop this once embassy-rp commit is in `kaspar030/embassy`

src/riot-rs-threads/src/lib.rs
kaspar030: fine as is!

examples/threading-multicore/README.md
```diff
@@ -0,0 +1,14 @@
+# threading multicore
+
+## About
+
+This application demonstrates basic threading on multicore.
```
kaspar030, line 5: I think this is missing some explanations what to expect. Looking at the output and code, I have no idea what to expect or what this demonstrates. :shrug:

kaspar030: btw, I get this output:

```
[INFO] - Hello from mid prio thread ThreadId(1) on core CoreId(0)
[INFO] - Hello from mid prio thread ThreadId(2) on core CoreId(0)
[INFO] - ThreadId(1) running again on core CoreId(0)
[INFO] - Hello from low prio thread ThreadId(0) on core CoreId(0)
[INFO] - Hello from high prio thread ThreadId(3) on core CoreId(1)
[INFO] - Hello from high prio thread ThreadId(4) on core CoreId(1)
[INFO] - ThreadId(2) running again (forever) on core CoreId(0)
```

shouldn't the high prio threads run first?

Cargo.toml
kaspar030: now I spotted the [explanation](https://github.com/future-proof-iot/RIOT-rs/pull/397#discussion_r1745532593).

examples/threading-multicore/README.md
elenaf9: The two high priority threads are pinned to core 1 with affinity masks. The reason why they run so late is because:
1. on start_threading, the scheduler on both cores is triggered
1. the faster one executes the `riot-rs-embassy` init thread, because that's the one with highest priority
1. if that's the scheduler on Core 1, the two high prio threads must wait until its execution finished

To prevent such a scenario in a real-world application, we could allow the user to set the core affinity for the embassy thread has well, so that it can be pinned to another core.

Of course ideally, in the above scenario the riot-rs-embassy thread should be preempted and continued on Core 0. But preventing these kinds of priority inversions is quite complex, and out of scope for this PR IMO.

Concerning this concrete example: agreed that it needs better documentation, and I think it should also be simplified.
The current complexity was mostly for testing purposes.

src/riot-rs-rt/Cargo.toml
```diff
@@ -17,25 +17,32 @@ riot-rs-utils = { workspace = true }
...
 [features]
-#default = ["threading"]
 threading = ["dep:riot-rs-threads"]
 
 debug-console = ["riot-rs-debug/debug-console"]
 executor-single-thread = []
 silent-panic = []
 _panic-handler = []
+critical-section-single-core = ["cortex-m/critical-section-single-core"]
+multicore = ["embassy-rp/critical-section-impl"]
```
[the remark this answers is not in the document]
elenaf9: `critical-section-single-core` is enabled through `riot-rs/single-core`, and that gives a compilation error if multicore is also enabled:

https://github.com/future-proof-iot/RIOT-rs/blob/9dcfc00804cac29728bfa5c33ceeaf80a30cc80b/src/riot-rs/src/lib.rs#L53-L57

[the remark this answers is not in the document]
elenaf9: Would you be okay if I just move the above check into `riot-rs-rt` (given that the conflict is in that crate) or do you think we need it in `riot-rs` as well?

src/riot-rs-threads/src/lib.rs
```diff
@@ -227,40 +299,165 @@ impl Threads {
...
+        }
+
+        // On multicore with core-affinities, get next thread with matching affinity.
+        #[cfg(all(feature = "multicore", feature = "core-affinity"))]
+        {
+            let next = self
+                .runqueue
+                .get_next_filter(|&t| self.is_affine_to_curr_core(t))?;
+            // Delete thread from runqueue to match the `pop_next`.
+            self.runqueue.del(next);
```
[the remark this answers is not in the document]
elenaf9: Not sure I understand what you mean. Where is it iterating twice?

kaspar030 (changes requested):
Initial review round by me, looks already very nice!

src/riot-rs-rt/Cargo.toml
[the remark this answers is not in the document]
kaspar030: so, better create a `riot-rs-rt/single-core` feature so the riot-rs becomes a pass-through.

src/riot-rs-threads/src/lib.rs
```diff
@@ -276,6 +473,28 @@ impl Threads {
...
+
+        // Stacks for the idle threads.
+        // Creating them inside the below for-loop is not possible because it would result in
+        // duplicate identifiers for the created `static`.
+        static STACKS: [ConstStaticCell<[u8; IDLE_THREAD_STACK_SIZE]>; CORES_NUMOF] =
+            [const { ConstStaticCell::new([0u8; IDLE_THREAD_STACK_SIZE]) }; CORES_NUMOF];
+
+        // Create one idle thread for each core with lowest priority.
+        for stack in &STACKS {
+            thread_create_noarg(idle_thread, stack.take(), 0, None);
```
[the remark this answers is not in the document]
elenaf9: Core affinities introduce some overhead (and are thus also feature-gated), and there is not need for them here.
As long as we have as many idle threads as cores, it doesn't matter which core one executes which thread. 

Or is there a case that I am missing, where not having the idle threads pinned to a core would lead to a problem?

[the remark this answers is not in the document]
kaspar030: I don't see where this might cause a problem! Was just wondering, avoiding overhead makes sense of course.

src/riot-rs-threads/src/lib.rs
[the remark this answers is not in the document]
kaspar030: runqueue is a linked list, so `get_next_filter()` has to iterate the list in order to find `next`. `del(next)` now has to iterate again to unlink `next`. If I understand correctly.
Textbook deletion would save one of the iterations by finding `next` and unlinking directly. 
(if at all, this is a minor optimisation. If my understanding is correct, maybe add "// TODO: this would benefit from `del_one_with_filter(...)`" or similar.)

src/riot-rs-rt/Cargo.toml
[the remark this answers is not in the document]
kaspar030: Unfortunately Cargo doesn't support "enable this feature if this other is not enabled".
So I guess we need to use `laze` to select either `riot-rs/single-core` or `riot-rs/multi-core`.

examples/threading-multicore/README.md
elenaf9: Simplified the example now to just start two threads, one with an core-affinity, that both print the core that they are running on: edeaddb.

src/riot-rs-threads/src/smp/esp32s3.rs
```diff
@@ -0,0 +1,88 @@
...
+    const CORES: u32 = 2;
+    const IDLE_THREAD_STACK_SIZE: usize = 2048;
+
+    fn core_id() -> CoreId {
+        esp_hal::get_core().into()
+    }
+
+    fn startup_other_cores() {
+        // TODO: How much stack do we really need here?
+        static STACK: ConstStaticCell<Stack<4096>> = ConstStaticCell::new(Stack::new());
```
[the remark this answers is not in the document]
elenaf9: It's the main stack for the second core, from what I can tell it is also used for ISRs. Or at least I didn't find anything about separate interrupt stacks or ISR-specific SP registers in any Xtensa Lx/ ESP32S3 docs.

src/riot-rs-threads/src/smp/rp2040.rs
```diff
@@ -0,0 +1,98 @@
...
+impl Multicore for Chip {
+    const CORES: u32 = 2;
+
+    fn core_id() -> CoreId {
+        CoreId(SIO.cpuid().read() as u8)
+    }
+
+    fn startup_other_cores() {
+        // TODO: How much stack do we really need here?
+        static STACK: ConstStaticCell<Stack<4096>> = ConstStaticCell::new(Stack::new());
```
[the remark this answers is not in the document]
elenaf9: Out-of-band discussion concerning stack size: 
- If the `interrupt-executor` is used, it runs only on core0 and all embassy interrupts are handled there. Thus only little stack is needed for the second core in that case. (@kaspar030 how much do you think is reasonable?)
- In case of the thread executor the needed stack very much depends on where the embassy init thread runs, which other interrupts are configured on the second core, and how much stack their handler requires. There is a general plan to in the future add logic for picking reasonable stack sizes, but that's out of scope for this PR and also independent of multicore. For now, let's pick the same `ISR_STACKSIZE` as on core 0? 

And maybe add `ISR_STACKSIZE_CORE1` env so it can be overridden by a user?

src/riot-rs-threads/src/lib.rs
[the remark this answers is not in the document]
elenaf9: The kind of iterations we're doing is a bit different between the two, but agreed that this can be optimized.
Will add the TODO.

Cargo.toml
elenaf9: see https://github.com/future-proof-iot/RIOT-rs/pull/397#discussion_r1808728314

deny.toml
elenaf9: dropped.

src/riot-rs-threads/src/smp/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+impl CoreId {
+    /// Creates a new [`CoreId`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if `value` >= [`Chip::CORES`].
+    pub fn new(value: u8) -> Self {
+        if value >= Chip::CORES as u8 {
+            panic!(
+                "Invalid CoreId {}: only {} cores available.",
```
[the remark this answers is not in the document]
elenaf9: Did `... only core ids 0..{N} available` because ranges in rust exclude the `end` value.

src/riot-rs-threads/src/smp/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    const CORES: u32;
+    /// Stack size for the idle threads.
+    const IDLE_THREAD_STACK_SIZE: usize = 256;
+
+    /// Returns the ID of the current core.
+    fn core_id() -> CoreId;
+
+    /// Starts other available cores so that the scheduler
+    /// is triggered there.
+    fn startup_other_cores();
```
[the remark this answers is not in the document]
elenaf9: Included.

kaspar030 (changes requested):
This is getting into shape!
I did run most tests and benchmarks over a couple of boards. The general impression is that in the single-core case, performance gets better and code gets smaller. Which is rare for substantial feature additions *after splitting out* the optimizations. :sunglasses: 

Unfortunately this PR breaks threading on the esp32-s3 for both single and multi-core. That needs fixing.

src/riot-rs-threads/src/lib.rs
```diff
@@ -49,17 +52,41 @@ pub mod macro_reexports {
 pub use riot_rs_runqueue::{RunqueueId, ThreadId};
 pub use thread_flags as flags;
 
+#[cfg(feature = "core-affinity")]
+pub use smp::CoreAffinity;
+#[cfg(feature = "multi-core")]
+pub use smp::CoreId;
```
kaspar030, line 58: Can we have a `CoreId` also on single-core? and `riot_rs_threads::core_id()`? (always returning 0).
(can do as follow up)

kaspar030:
> You mean if it's rebased onto main, right? I believe that's still due to the most recent esp-hal update and [esp-rs/esp-hal#2386](https://github.com/esp-rs/esp-hal/issues/2386).

Ah yes. I was confused b/c the fix for riscv was merged and I thought that would fix the issue also on xtensa. I guess there's a reason why we use INTR0 on xtensa and INTR1 on riscv?

src/riot-rs-threads/src/smp/rp2040.rs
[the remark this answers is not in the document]
elenaf9: Added env-based `ISR_STACKSIZE_CORE1` constant, that fallbacks to  `ISR_STACKSIZE`.

src/riot-rs-threads/src/smp/esp32s3.rs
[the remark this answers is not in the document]
elenaf9: See https://github.com/future-proof-iot/RIOT-rs/pull/397#discussion_r1815101168

src/riot-rs-threads/src/lib.rs
elenaf9: Done

kaspar030 (approved):
LGTM. Good job!

kaspar030:
Gratulations @elenaf9!

PR #392, opened 2024-08-29
refactor(embassy)!: extract `arch` modules as crates

ROMemories:
# Description

This PR extracts the `arch` modules (the manufacturer-specific modules previously found in `riot-rs-embassy`) as separate, manufacturer-specific crates. A major benefit of this is allowing to generate user-visible documentation for these crates, which was not possible before because (a) only one `arch` module could be enabled at a time and (b) it was not possible to enable any of them for documentation, because of them being manufacturer-specific.

Extracting these modules as separate crates has required to:

- Introduce a new crate (whose temporary name is `riot-rs-shared-types`) to break cyclic dependencies between crates/types: this new crate is depended upon by every manufacturer-specific crate.
- Revisit some types in the new `riot-rs-shared-types` crate to accommodate manufacturer-specific variants in enums, by introducing simple generics on them.
- Work around the orphan rules in some internal cases.
- Introduce the relevant Cargo features for individual manufacturer-specific crates. Only features that are actually useful for the crate are introduced. This introduces a slight increase in duplicated features in the tree, but has the benefit of documenting what features are actually provided by each manufacturer-specific crate.

## Issues/PRs references

Depends on #391
Depends on #394

## Testing

The following have been successfully tested in hardware (on dfaca71):

- `example/gpio` on nRF52840-DK, nRF5340-DK, ESP32-C6, RP2040, ST-NUCLEO-WB55RG
- `example/embassy-http-server`
  - on ESP32-C6, RPi Pico W with Wi-Fi
  - on nRF52840-DK, ST-NUCLEO-WB55RG on Ethernet over USB

## TODO

- [x] Clean up the circumventions around the orphan rule limitations

## Future work

- [ ] Generate documentation for these new manufacturer-specific crates

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](../book/src/coding-conventions.md).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #399, opened 2024-09-06
feat(xtensa): add support for esp32s3 with xtensa arch

### LINKED ISSUE #359: Add Xtensa support
### Description

.

### Motivation

It's the next big family to be supported.

### Requirements

in particular ESP32-S3-DevKitC-1 ;)

### Open questions

_No response_

Yes

elenaf9:
# Description

- Add support for `xtensa` arch in `riot-rs-threads`
- Add support for `esp32s3` chip in `riot-rs-embassy` (cc @ROMemories given that these are minimal changes, I don't think there will be many merge conflicts with #392. I'll rebase once that PR is in)
- Add `espressif-esp32-s3-wroom-1` board
- Adapt the `laze-project.yml` based on prio work from @kaspar030.

## Testing

Rust on xtensa requires the installation of the esp toolchain and setting up the environment correctly. See the [Rust on ESP Book](https://docs.esp-rs.org/book/installation/riscv-and-xtensa.html). It can then be run as usual with

```sh
laze build -C examples/threading -b espressif-esp32-s3-wroom-1 run
``` 

## Issues/PRs references

Fixes #359.

## Open TODOs

There are still linker issues that have to be fixed. Right now I simply worked around them by commenting out the conflicting parts, and somehow it worked. See self-review.

## Change checklist

- [ ] I have followed the [Coding Conventions](../book/src/coding-conventions.md).


laze-project.yml
```diff
@@ -12,8 +12,8 @@ contexts:
       - ?defmt
     env:
       RUSTFLAGS:
-        - "-Clink-arg=--nmagic"
-        - "-Clink-arg=${LINK_ARG_PREFIX}--no-eh-frame-hdr"
+        #        - "-Clink-arg=--nmagic"
+        #        - "-Clink-arg=${LINK_ARG_PREFIX}--no-eh-frame-hdr"
```
elenaf9, line 16: @kaspar030 I copied this from your WIP branch for xtensa support.
I am not familiar enough with linking in general to really understand if/ when these flags are needed.

laze-project.yml
```diff
@@ -278,8 +278,8 @@ contexts:
         - --cfg context=\"esp\"
         # linkall first
         - -Clink-arg=-Tlinkall.x
-        - -Clink-arg=-Tlinkme-esp-fixup.x
-        - -Clink-arg=-Tlinkme.x
+        # - -Clink-arg=-Tlinkme-esp-fixup.x
+        # - -Clink-arg=-Tlinkme.x
```
elenaf9, line 282: Commented out because it somehow didn't find the `.rodata` section that is referred to in `linkme.x`. It works right now like this, but I'll try to fix this properly.

src/riot-rs-threads/src/arch/xtensa.rs
```diff
@@ -0,0 +1,163 @@
...
+        thread.sp = stack_pos;
+        thread.data.A1 = stack_pos as u32;
+        thread.data.A6 = arg as u32;
+        // Usually A0 holds the return address.
+        // However, xtensa features so-called Windowed registers, which allow
+        // to shift the used registers when calling procedure.
+        // The xtensa-lx-rt does this when calling the exception handler using
+        // call4, which shifts the window by 4.
+        // See `xtensa_lx_rt::exception::asm::__default_naked_exception`.
+        // (At least that's what I assume is happening)
```
elenaf9, line 35:
```suggestion

```

laze-project.yml
elenaf9: I have now moved `-Clink-arg=--nmagic` to the `cortex-m` module because [according to the rustc book](https://doc.rust-lang.org/beta/rustc/platform-support/arm-none-eabi.html?search=eh-frame-hdr) it's part of the usual **Arm** bare-metal linker setup.
I looked into when/ why the `--no-eh-frame-hdr` flag is needed, and seems like it's only helpful for reducing code size, but I didn't find any docs that actually recommend using it. Might be just a premature optimization. Can anyone give some input/ opinion on whether we have to keep it?

laze-project.yml
elenaf9: Out-of band discussion with @kaspar030: we might not need the linkme script for esp32, so it's okay to remove them (correct me if I misunderstood something @kaspar030). Should we then also remove the esp-fixup files?

laze-project.yml
kaspar030: Let's keep it for now and investigate another time. Moving to "cortex-m" is fine here.

laze-project.yml
kaspar030: > Should we then also remove the esp-fixup files?

yes, if linkme works without!

elenaf9: Tested for esp32c3, esp32c6 and esp32s3 the examples `threading-channel` and `hello-world-async`, works without :+1:

kaspar030:
@elenaf9 do you mind rebasing this?

src/riot-rs-threads/src/arch/xtensa.rs
```diff
@@ -0,0 +1,141 @@
...
+            threads.current_thread = Some(next_pid);
+            *trap_frame = threads.threads[usize::from(next_pid)].data;
+            true
+        }) {
+            break;
+        }
+        // The esp-hal implementation of critical-section doesn't disable all interrupts.
+        // Thus we should release our hold on `THREADS` before we `waiti`, to prevent
+        // that another interrupt handler will try to borrow it while we still have it borrowed.
+        unsafe { core::arch::asm!("waiti 0") };
```
elenaf9, line 139: We could also disabled all interrupts when accessing `THREADS` on xtensa. `esp-hal` originally disabled all interrupts in their critical section, but changed it to only disable interrupts with prio <=5 so that the debugging at interrupt level 6 still works (see https://github.com/esp-rs/esp-hal/pull/360). 
So if we'd change it, debugging wouldn't work inside critical sections anymore I think.
On the other hand, it can happen right now that an `esp_hal::embassy` interrupt interrupts a running thread that is currently accessing `THREADS`. If that interrupt is then also accessing `THREADS` (e.g. because it's a `_pender`), it panics.
But the closures for accessing `THREADS` are very short, so I don't know.

Either way the `waiti` here still would need to be outside of the closure, because IIUC it only returns if an interrupt **at the currently enabled level** occurs. If all interrupts are disabled, it will never return.

tests/benchmarks/bench_sched_yield/laze.yml
```diff
@@ -2,4 +2,5 @@ apps:
   - name: bench_sched_yield
     selects:
       - sw/benchmark
+      - executor-thread
```
kaspar030, line 5: I guess this is unrelated?

src/riot-rs-threads/src/arch/xtensa.rs
kaspar030: if that `prio <= 5` refers to the `Priority` type, this might be fine, see: https://github.com/kaspar030/esp-hal/blob/23d22e32784e8c9b418a4fb4e316397192e6e32b/esp-hal/src/interrupt/xtensa.rs#L355-L356. Also, `Priority::max() == Priority3`. So embassy (or any other Rust) code will probably not be able to access `THREADS`

kaspar030: then again, we might prefer stalling at debug time over possible panic at non-debug time ...

kaspar030 (commented):
Please rebase (again), and squash. and probably, do a `cargo update`, I added a commit to our embasssy fork to use the old waker code, which the current xtensa compiler needs.

kaspar030:
I've added installing the xtensa toolchain for esp32s3 to CI, also building for the `espressif-esp32-s3-wroom-1` board.

IMO this is good to go, apart from the CI (unrelated?) errors and the `bench_sched_yield` change.

src/riot-rs-threads/src/arch/xtensa.rs
elenaf9: Okay so seems like interrupt level up until level 7 is still possible, or at least configured for xtensa-lx-rt in https://github.com/esp-rs/esp-hal/blob/3bb49b049a2d95dd2f6fc1b6710941f5938c1db7/xtensa-lx-rt/config/xtensa_esp32s3.toml#L190-L191.
But looking through that list only the debug interrupt has an interrupt level > 5, and the NMI interrupt / INT14, but that one can't be masked anyway :sweat_smile:  And `CpuInterrupt::Interrupt14NmiPriority7` is among the interrupts that are mapped to `None` in your above code.
So I don't think it'll be an issue for us here.

tests/benchmarks/bench_sched_yield/laze.yml
elenaf9: It was needed so that this benchmark compiles for esp, because otherwise the default `executor-single-thread` was selected, that conflicts with the `threading` feature.

But I guess the correct variant is to instead select `sw/threading`, which will down the line result into `executor-thread` being selected for esp.
```suggestion
      - sw/threading
```

PR #425, opened 2024-09-18
chore: make clippy happy

ROMemories:
# Description

Some of these come from new lints brought by updating the toolchain, others are lingering warning fixes.

## Change checklist

### DIFF
--- a/src/riot-rs-embassy-common/src/gpio.rs
+++ b/src/riot-rs-embassy-common/src/gpio.rs
@@ -20,9 +20,10 @@ impl From<Level> for bool {
 
 impl From<bool> for Level {
     fn from(boolean: bool) -> Self {
-        match boolean {
-            false => Level::Low,
-            true => Level::High,
+        if boolean {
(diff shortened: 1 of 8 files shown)


src/riot-rs-embassy-common/src/gpio.rs
```diff
@@ -20,9 +20,10 @@ impl From<Level> for bool {
 
 impl From<bool> for Level {
     fn from(boolean: bool) -> Self {
-        match boolean {
-            false => Level::Low,
-            true => Level::High,
+        if boolean {
```
kaspar030, line 23: This turns around the result, is that intended?

kaspar030:
lgtm!

src/riot-rs-embassy-common/src/gpio.rs
kaspar030: I wonder if we could encode one `From` direction with the other, so there's only one source of truth?

PR #426, opened 2024-09-18
test(esp): set up host testing of `ariel-os-esp`

ROMemories:
# Description

## Change checklist

### DIFF
--- a/src/ariel-os-esp/Cargo.toml
+++ b/src/ariel-os-esp/Cargo.toml
@@ -149,5 +149,7 @@ wifi-esp = [
 ## Enables the interrupt executor.
 executor-interrupt = ["embassy-executor/executor-interrupt"]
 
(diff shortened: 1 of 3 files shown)


kaspar030:
@ROMemories where are we with this?

stappersg:
Early 2026: Where are we with this?

ROMemories:
> Early 2026: Where are we with this?

I've updated the PR to take into account the changes brought by #650, but I believe the original issue remains: [because the selection of the `riscv` or `xtensa-lx` crate depends on the current architecture](https://github.com/esp-rs/esp-hal/blob/327071c262afaa0682ffd9ac55e0a747966e5214/esp-hal/Cargo.toml#L116-L122), this doesn't work as-is for the host tests. However we don't currently have host tests in the `ariel-os-*` HAL crates, and now that we have more visibility on what goes in them, it seems unlikely to me that we have any in the future.

PR #434, opened 2024-09-19
refactor(runqueue-clist): use `peek_head` as much as possible

ROMemories:
# Description

This is a very small refactor of the runqueue's `CList` which re-uses internal functions as much as possible while preserving performance.
It's an attempt to simplify its implementation to ease future refactors possibly required for formally verifying it.

## Testing

I obtained the exact same results for the `bench_sched_yield` benchmark before and after this PR.

## Change checklist

### DIFF
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -160,48 +160,47 @@ mod clist {
 
         pub fn push(&mut self, n: u8, rq: u8) {
             assert!(n < Self::sentinel());
-            if self.next_idxs[n as usize] == Self::sentinel() {
-                if self.tail[rq as usize] == Self::sentinel() {
-                    // rq is empty, link both tail and n.next to n
-                    self.tail[rq as usize] = n;
-                    self.next_idxs[n as usize] = n;
-                } else {
-                    // rq has an entry already, so
-                    // 1. n.next = old_tail.next ("first" in list)
-                    self.next_idxs[n as usize] = self.next_idxs[self.tail[rq as usize] as usize];
-                    // 2. old_tail.next = n
-                    self.next_idxs[self.tail[rq as usize] as usize] = n;
-                    // 3. tail = n
-                    self.tail[rq as usize] = n;
-                }
+            if self.next_idxs[n as usize] != Self::sentinel() {
+                return;
+            }
+
+            if let Some(head) = self.peek_head(rq) {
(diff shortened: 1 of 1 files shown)


src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -161,56 +161,53 @@ mod clist {
...
-                    // rq's tail bites itself, so there's only one entry.
-                    // so, clear tail.
-                    self.tail[rq as usize] = Self::sentinel();
-                    // rq is now empty
-                } else {
-                    // rq has multiple entries,
-                    // so set tail.next to head.next (second in list)
-                    self.next_idxs[self.tail[rq as usize] as usize] = self.next_idxs[head as usize];
-                }
+            let head = self.peek_head(rq)?;
```
elenaf9, line 181: We *could* manually first get tail and then here, so that we can re-use tail further below. But might be a premature optimization; we'd need a new benchmark do actually test the performance impact of it.

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -161,56 +161,53 @@ mod clist {
         pub fn push(&mut self, n: u8, rq: u8) {
             assert!(n < Self::sentinel());
             if self.next_idxs[n as usize] == Self::sentinel() {
-                if self.tail[rq as usize] == Self::sentinel() {
-                    // rq is empty, link both tail and n.next to n
-                    self.tail[rq as usize] = n;
-                    self.next_idxs[n as usize] = n;
-                } else {
+                if let Some(head) = self.peek_head(rq) {
```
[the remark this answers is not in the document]
elenaf9: Yes :+1:

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -161,56 +161,53 @@ mod clist {
         pub fn push(&mut self, n: u8, rq: u8) {
             assert!(n < Self::sentinel());
             if self.next_idxs[n as usize] == Self::sentinel() {
```
elenaf9, line 163: Nit while we're at it:
Question of taste, but I like to separate the "happy path" from the "error/ unexpected path" by using early returns.
This also reduces the indentation and imo makes the code more readable. Here, this would mean doing

```rs
if self.next_idxs[n as usize] != Self::sentinel() {
    return;
}

elenaf9 (approved):

PR #440, opened 2024-09-25
feat: add an SPI main abstraction

ROMemories:
# Description

This PR provides an abstraction layer for SPI support, in main mode.
It focuses on providing non-DMA-enabled drivers, which still provide an async interface, with timeouts—e.g., to avoid indefinitely blocking when the target is not connected.
The drivers implement [`embedded_hal_async::spi::SpiBus`](https://docs.rs/embedded-hal-async/latest/embedded_hal_async/spi/trait.SpiBus.html), and can thus easily be used by any device driver relying on that trait.

It provides ways to either select frequencies which all MCUs are expected to support (up to 8 MHz), or to request the highest available frequency within a range, resolved at compile-time using `highest_freq_in`.

## Issues/PRs references

Depends on #463

## Open Questions

- Should `Frequency` be renamed to `Bitrate`?
- Should we attempt to make the test more generic, so it doesn't rely on a specific sensor?
- The SPI frequency, selected either using `highest_freq_in` or directly using the arch-independent enum, is only precisely respected on a best-effort basis by Embassy: as the SPI frequency is sometimes configured based on an input frequency divided by a prescaler, which only has a limited number of possible values, the exact frequency actually used can be different from the one selected by the user. I suppose it would be useful to document this. As the documentation of the manufacturer-specific crates is not currently rendered (where I think this should be mentioned), I'm opting for documenting this as part of a follow-up PR.

## Limitations

DMA-enabled drivers are out-of-scope of this PR as the number of DMA channels is limited on some architectures and non-DMA-enabled drivers are therefore useful when no channels are left. Moreover, setting up DMA incurs some overhead, and is thus not relevant for small operations as it is the case with sensors, which was the initial motivation for adding support for SPI.

## Future work

- [ ] Provide DMA-enabled drivers in a `dma` module inside `spi::main`.

## Change checklist

### DIFF
--- a/book/src/support_matrix.html
+++ b/book/src/support_matrix.html
@@ -4,7 +4,7 @@
     <tr>
       <th>Chip</th>
       <th>Testing Board</th>
-      <th colspan="8">Functionality</th>
+      <th colspan="9">Functionality</th>
     </tr>
     <tr>
       <th></th>
@@ -12,6 +12,7 @@
       <th>GPIO</th>
       <th>Debug Output</th>
       <th>I2C Controller Mode</th>
+      <th>SPI Main Mode</th>
       <th>Logging</th>
       <th>User USB</th>
       <th>Wi-Fi</th>
@@ -23,6 +24,7 @@
     <tr>
       <td>ESP32-C6</td>
       <td>ESP32-C6-DevKitC-1</td>
+      <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="supported">✅</td>
@@ -40,6 +42,7 @@
       <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="supported">✅</td>
+      <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="not available on this piece of hardware">–</td>
       <td class="support-cell" title="supported">✅</td>
       <td class="support-cell" title="supported">✅</td>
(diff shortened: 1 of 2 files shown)


kaspar030:
@ROMemories do you mind rebasing this? I'm starting to take a closer look.

kaspar030:
> * Should `Frequency` be renamed to `Bitrate`

IMO, no. If you feel it needs discussion, let's do that at next monthly. (we're not waiting for this here).

kaspar030:
> * Should we attempt to make the test more generic, so it doesn't rely on a specific sensor?

I'm doing simple loopback testing. I can push that test here?

kaspar030:
> > I'm doing simple loopback testing. I can push that test here?
> 
> Sure, please do!

Done.

I've changed the used SPI pins to the ones from the arduino connector where applicable, maybe it makes sense to use that pinout also for `spi-main`?

I've tested the nfr52840dk, nrf5340, nucleo-h755, nucleo-wb55 an nucleo-f401re using `spi-loopback` successfully.

src/riot-rs-rp/src/spi/main/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+pub enum Frequency {
+    F(Kilohertz),
+}
+
+riot_rs_embassy_common::impl_spi_from_frequency!();
+riot_rs_embassy_common::impl_spi_frequency_const_functions!(MAX_FREQUENCY);
+
+impl Frequency {
+    #[expect(non_snake_case, reason = "consistency with fugit")]
+    fn to_Hz(&self) -> u32 {
```
kaspar030, line 45:
```suggestion
    fn to_Hz(self) -> u32 {
```

kaspar030:
Then, lets get this in! LGTM. Please squash.

kaspar030 (approved):

PR #443, opened 2024-09-26
fix(coapcore): Update liboscore dependency

chrysn:
# Description

This pulls in workarounds / fixes for clang-18, which would otherwise fail to define `uint8_t` & co in its stdint.h, causing bindgen failures.

Thanks @kaspar030 for noticing the issue and helping me reproduce it.

## Issues/PRs references

See-Also: https://gitlab.com/oscore/liboscore/-/issues/61

Depends on: https://github.com/future-proof-iot/RIOT-rs/pull/491

## Open Questions

<del>Let's hope everyone can use clang-18 already: as the fix for using clang-18 requires a recent clang (didn't test -17, but -16 doesn't cut it), this may require updates here and there.</del>

## Change checklist

 <del> (Our documentation says to install clang on Ubuntu, but the latest Ubuntu LTS already has clang 18 by default, so that doesn't need an update.)</del>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,17 +124,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -202,24 +191,22 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.63.0"
+version = "0.70.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "36d860121800b2a9a94f9b5604b332d5cffb234ce17609ea479d723dbc9d3885"
+checksum = "f49d8fed880d473ea71efb9bf597651e77201bdd4893efe54c9e5d65ae04ce6f"
 dependencies = [
- "bitflags 1.3.2",
+ "bitflags 2.6.0",
  "cexpr",
  "clang-sys",
- "lazy_static",
- "lazycell",
+ "itertools",
  "log 0.4.22",
- "peeking_take_while",
+ "prettyplease",
  "proc-macro2",
  "quote",
  "regex",
  "rustc-hash",
  "shlex",
- "syn 1.0.109",
- "which",
+ "syn 2.0.85",
 ]
 
 [[package]]
@@ -333,7 +320,6 @@ version = "0.24.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b922faaf31122819ec80c4047cc684c6979a087366c069611e33649bf98e18d"
 dependencies = [
- "clap",
  "heck 0.4.1",
  "indexmap 1.9.3",
  "log 0.4.22",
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -2440,31 +2393,12 @@ dependencies = [
...
-]
-
 [[package]]
 name = "httparse"
 version = "1.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fcc0b4a115bf80b728eb8ea024ad5bd707b615bfed49e0665b6e0f86fd082d9"
 
-[[package]]
-name = "i2c"
```
chrysn, line 2459: It's slightly odd that this is being removed, but those changes are a tad hard to track, so I'd just trust Cargo on this.

kaspar030: I've seen this before when switching branches lately. Whateger thisbis, it is unrelated.

chrysn:
Could the tests that now fail be the ones that previously passed (that have clang <18)? The error style is identical to what I get when I do *not* install `libclang1-18`.

Where is the precise container our builds are run in defined? In `.github/workflows/main.yml` I only see ubuntu-latest, but that's a GitHub internal thing and not a docker spec.

kaspar030:
> Where is the precise container our builds are run in defined? In `.github/workflows/main.yml` I only see ubuntu-latest, but that's a GitHub internal thing and not a docker spec.

IIUC our builds are run in a VM, which runs the `ubuntu-latest` image...

kaspar030:
> Where is the precise container our builds are run in defined? In `.github/workflows/main.yml` I only see ubuntu-latest, but that's a GitHub internal thing and not a docker spec.

[this](https://github.com/actions/runner-images) has more info. I think the page says that `ubuntu-latest` installs the latest three major versions of tools like GCC or Clang.

kaspar030:
Is it possible that when multiple versions of clang (or libclang) are installed, the selection process is indeterministic?
Hm, on CI it is usually stm32's that fail ...

kaspar030:
Would you revert https://github.com/future-proof-iot/RIOT-rs/pull/447/commits/d351ee888ae97ddf5928d92778f0ec5f8243e3dc here (as soon as that landed in main)?

kaspar030:
Github misunderstood my intentions with `Potential fix: #443` here.

src/lib/coapcore/Cargo.toml
```diff
@@ -26,9 +26,9 @@ coap-message-utils = "0.3.3"
 coap-numbers = "0.2.3"
```
kaspar030, line 26: (again as a conversation so we don't forget) Would you revert https://github.com/future-proof-iot/RIOT-rs/commit/d351ee888ae97ddf5928d92778f0ec5f8243e3dc here (as soon as that landed in main)?

chrysn:
The latest update should work both on clang 18 and on earlier versions; this makes this branch now the one to investigate into further (as opposed to #469 which worked on the assumption that the older version works for pre-18 and the new for 18-and-later; its failure to pass CI shows that there was *something else* broken with the older version too, and the updated version at least provides a modernized base to look into the issue).

kaspar030:
You can temporarily limit the build in main.yml if it speeds up testing!

chrysn:
This is now ready for review again, after CI setbacks.

This is now:
* trivial stuff:
  * a plain update-fix (upstream libOSCORE changed, even less than originally anticipated in that it does not depend on clang-18, but nonetheless it is updated)
  * removing the blocker against the CoAP example in CI
* Explicitly selecting a clang to use by setting CLANG_PATH in CI. I think doing this at CI time is suitable because a) most systems just have a recent clang installed, and b) for those who don't we don't have sufficient data to decide which clang is the right one.
* Adds a bunch of version information to the CI output. I found this useful, and it just takes 1s, but if you dislike I can just remove it.

chrysn:
Downconverted to a draft again because enabling all the previously disabled tests showed that the Xtensa setup with its custom (and, for those boards, needed) LIBCLANG_PATH breaks everything once more: It installs a libclang without a corresponding clang.

Sigh.

chrysn:
This is now rebased onto #492 and #491.

It may also work without #492, but then the clang version needs adjusting, and generally, 2024.04 is the current LTS version of Ubuntu, let's not make our lives harder needlessly.

The PR still consists of the 4 parts described [earlier](https://github.com/future-proof-iot/RIOT-rs/pull/443#issuecomment-2439585200), but the rebasing makes all of that also work in the presence of the Xtensa toolchain. Let's hope that before this all becomes troublesome again (eg. if something actually starts needing a proper LIBCLANG_PATH), enough of the Xtensa stuff is upstreamed and can be used through the regular channels.

It'll take some time for the CI to confirm, but it looks promising.

chrysn:
With https://github.com/future-proof-iot/RIOT-rs/pull/491 as the now working base, I think that's it – ready for review, ideally as #492, #491 and then this one.

kaspar030:
LGTM

PR #448, opened 2024-09-27
feat(docs): Add section on CoAP

chrysn:
# Description

As per today's meeting, I've pulled ahead writing CoAP documentation to give all involved a better overview of how the security protocols interact.

Unlike what we otherwise do with the book, this is ahead of the implementation (covering some of the [roadmap](https://github.com/future-proof-iot/RIOT-rs/issues/226) but also some things not in there in the current iteration). As this matures, the annotation of "we don't have this yet" will move from page-wide into sections.

## Open questions

Should we aim for this to be merged with the disclaimer in, or do we rather keep it as a PR until most of it is implemented? I'm leaning toward the former, especially because the book is also a good reference for developers.

## Change checklist

### DIFF
--- a/book/src/tooling/coap.md
+++ b/book/src/tooling/coap.md
@@ -18,14 +18,8 @@ As there is more complexity in a CoAP client than a CoAP server,
 the most minimal use is to just have a CoAP server on the embedded system.
 Having both components in use is also a common choice.
 
-As part of our mission for strong security,
-we use encrypted CoAP traffic by default.
-Of the multiple means to secure CoAP,
-we have picked the OSCORE/EDHOC security mechanisms as our default,
-with the option to use ACE for authorization management.
-These were chosen because they scale down well to the smallest devices.
-Alternatives (DTLS, TLS, IPsec or link-layer encryption)
-are currently not implemented.
+Also, as part of our mission for strong security,
+we use encrypted CoAP traffic by default as explained below.
 
 [CoAP]: https://coap.space/
 [over UDP]: https://datatracker.ietf.org/doc/html/rfc7252
@@ -43,43 +37,14 @@ and a path `/fwup` for interacting with firmware updates.
 The CoAP implementation can put additional resources at well-known locations,
 eg. `/.well-known/core` for discovery or `/.well-known/edhoc` for establishing secure connections.
 
-The handler does not need to concern itself with authorization:
-The CoAP stack is configured with an access policy,
-which is evaluated before the handler even receives the request.
-The handler still needs to concern itself with security to some extent
-(eg. file format parsers should still treat incoming data as possibly malformed),
+The handler only needs to concern itself with basic security aspects
+(eg. file format parsers should treat incoming data as possibly malformed),
 but is not the main enforcement point.
 
-The policy is configured depending on the enabled security mechanisms.
-Examples of described policy entries are:
-
-* This is a fixed public key, and requests authenticated with that key are allowed to GET and PUT to `/limit`.
-* The device has a shared secret from its authorization server, with which the authorization server secures the tokens it issues to clients. Clients may perform any action as long as they securely present a token that allows it. For example, a token may allow GET on `/limit` and PUT on `/led/0`".
-* Any (even unauthenticated) device may GET `/hello/`.
 
 [provided as `examples/coap`]: https://github.com/future-proof-iot/RIOT-rs/tree/main/examples/coap
 [its `run()` function]: https://github.com/future-proof-iot/RIOT-rs/blob/2b76e560394884d3c8f7eaae51beefd59a316d7b/examples/coap/src/main.rs#L70
 
-### Interacting with RIOT-rs CoAP server from the host
-
-A convenient policy (which is the default of RIOT-rs's examples)
-is to grant the user who flashes the device all access on it.
-When that policy is enabled in the build system,
-an unencrypted key is created in the developer's [state home directory]<!-- precise location TBD -->,
-from where it can be picked up by tools such as [aiocoap-client].
-
-Furthermore,
-when a CoAP server is provisioned through the RIOT-rs build system,
-public keys and their device associations are stored
-in the developer's state home directory.
-
-Together, these files act in a similar way as the classic UNIX files `~/.netrc`, `~/.ssh/id_rsa{,.pub}`.
-They can also double as templates for an application akin to `ssh-copy-id`
-in that they enable a server policy like
-"any device previously flashed on this machine may GET all resources".
-
-[aiocoap-client]: https://aiocoap.readthedocs.io/en/latest/tools.html
-[state home directory]: https://specifications.freedesktop.org/basedir-spec/latest/
 
 ## Usage: Client side
 
@@ -103,12 +68,6 @@ A program that triggers a CoAP request provides[^whatsinarequest] some component
   This is optional if there is a global policy,
   or if there is an implied security mechanism for the origin.
 
-  Examples of policies that can be available are
-  "expect the server to present some concrete public key, use this secret key once the server is verified",
-  "use a token for this audience and scope obtained from that authentication server",
-  "expect the server to present a chain of certificates for its hostname down to a set of root certificates" (which is the default for web browsers),
-  "establish an encrypted connection and trust the peer's key on first use",
-  down to "do not use any encryption".
 
 [^whatsinarequest]: These components required for a request are not documented as such in the CoAP RFCs,
     but it is the author's opinion that they are a factual requirement:
@@ -116,12 +75,25 @@ A program that triggers a CoAP request provides[^whatsinarequest] some component
     but the decisions are still made.
     At the time of writing, there [is an open issue](https://github.com/core-wg/corrclar/issues/41) to clarify this in the specifications.
 
-## Details on security mechanisms
+## Security
 
-RIOT-rs uses three pieces of security components:
+The CoAP stack is configured with an access policy,
+which is evaluated before the handler receives the request.
+
+Furthermore, the CoAP communication is secured using mechanisms for
+symmetric encryption, key exchange and authentication. 
+
+At this stage, RIOT-rs uses three pieces of security components:
 OSCORE (for symmetric encryption), EDHOC (for key exchange) and ACE (for authentication).
 
-All have in common that they sit on top of CoAP:
+Alternatives are possible (for instance DTLS, TLS, IPsec or link-layer encryption)
+but are currently not implemented / not yet supported in RIOT-rs.
+
+#### Rationale for OSCORE/EDHOC/ACE
+
+OSCORE/EDHOC/ACE were chosen first because they scale down
+well to the smallest devices, and because they
+all have in common that they sit naturally on top of CoAP:
 Their communication consists of CoAP requests and responses.
 Thus, they work homogeneously across all CoAP transports,
 and provide end-to-end security across untrusted proxies.
@@ -132,8 +104,47 @@ is to allow unauthenticated access everywhere.
 For example, this may make sense on a link layer with tight access control.
 The components also have internal dependencies:
 EDHOC can only practically be used in connection with OSCORE;
-ACE comes with profiles with individual dependencies
-(eg. using the ACE-EDHOC profile requires EDHOC).
+ACE depends on either depending on the profiles used.
+
+### Access policy
+
+The policy is configured depending on the enabled security mechanisms.
+Examples of described policy entries are:
+
+* This is a fixed public key, and requests authenticated with that key are allowed to GET and PUT to `/limit`.
+* The device has a shared secret from its authorization server, with which the authorization server secures the tokens it issues to clients. Clients may perform any action as long as they securely present a token that allows it. For example, a token may allow GET on `/limit` and PUT on `/led/0`".
+* Any (even unauthenticated) device may GET `/hello/`.
+
+#### Interacting with RIOT-rs CoAP server from the host
+
+A convenient policy (which is the default of RIOT-rs's examples)
+is to grant the user who flashes the device all access on it.
+When that policy is enabled in the build system,
+an unencrypted key is created in the developer's [state home directory]<!-- precise location TBD -->,
+from where it can be picked up by tools such as [aiocoap-client].
+
+Furthermore,
+when a CoAP server is provisioned through the RIOT-rs build system,
+public keys and their device associations are stored
+in the developer's state home directory.
+
+Together, these files act in a similar way as the classic UNIX files `~/.netrc`, `~/.ssh/id_rsa{,.pub}`.
+They can also double as templates for an application akin to `ssh-copy-id`
+in that they enable a server policy like
+"any device previously flashed on this machine may GET all resources".
+
+[aiocoap-client]: https://aiocoap.readthedocs.io/en/latest/tools.html
+[state home directory]: https://specifications.freedesktop.org/basedir-spec/latest/
+
+#### Interacting with the RIOT-rs CoAP client
+
+Examples of policies that can be available are
+  "expect the server to present some concrete public key, use this secret key once the server is verified",
+  "use a token for this audience and scope obtained from that authentication server",
+  "expect the server to present a chain of certificates for its hostname down to a set of root certificates" (which is the default for web browsers),
+  "establish an encrypted connection and trust the peer's key on first use",
+  down to "do not use any encryption".
+
 
 ### Symmetric encryption: OSCORE
 
--- a/book/src/tooling/coap.md
+++ b/book/src/tooling/coap.md
@@ -1,0 +1,257 @@
+# CoAP
+
+> **Warning**: This documentation is currently ahead of the implementation roadmap.
+
+[CoAP] is an **application layer protocol similar to HTTP** in its use
+(eg. you could POST data onto a resource on a server that is identified by a URL)
+but geared towards IoT devices in its format and its security mechanisms.
+
+CoAP provides a versatile set of transports
+(with IETF Proposed Standards for running [over UDP] including multicast, [over TCP and WebSockets],
+other standards for running [over SMS and NB-IoT], and more in development).
+It relies on proxies to span across transports and accommodate the characteristics of particular networks,
+and offers features exceeding the classical REST set such as [observation].
+
+**RIOT-rs supports** the use of CoAP
+for implementing clients, servers or both in a single device.
+As there is more complexity in a CoAP client than a CoAP server,
+the most minimal use is to just have a CoAP server on the embedded system.
+Having both components in use is also a common choice.
+
+As part of our mission for strong security,
+we use encrypted CoAP traffic by default.
+Of the multiple means to secure CoAP,
+we have picked the OSCORE/EDHOC security mechanisms as our default,
+with the option to use ACE for authorization management.
+These were chosen because they scale down well to the smallest devices.
+Alternatives (DTLS, TLS, IPsec or link-layer encryption)
+are currently not implemented.
+
+[CoAP]: https://coap.space/
+[over UDP]: https://datatracker.ietf.org/doc/html/rfc7252
+[over TCP and WebSockets]: https://datatracker.ietf.org/doc/html/rfc8323
+[over SMS and NB-IoT]: https://www.omaspecworks.org/wp-content/uploads/2018/10/Whitepaper-11.1.18.pdf
+[observation]: https://datatracker.ietf.org/doc/html/rfc7641
+
+## Usage: Server side
+
+An example of a CoAP server is [provided as `examples/coap`], see [its `run()` function] for the practical steps.
+
+A CoAP server is created by assembling several **resource handlers on dedicated paths**:
+There might be a path `/s/0` representing a particular sensor,
+and a path `/fwup` for interacting with firmware updates.
+The CoAP implementation can put additional resources at well-known locations,
+eg. `/.well-known/core` for discovery or `/.well-known/edhoc` for establishing secure connections.
+
+The handler does not need to concern itself with authorization:
+The CoAP stack is configured with an access policy,
+which is evaluated before the handler even receives the request.
+The handler still needs to concern itself with security to some extent
+(eg. file format parsers should still treat incoming data as possibly malformed),
+but is not the main enforcement point.
+
+The policy is configured depending on the enabled security mechanisms.
+Examples of described policy entries are:
+
+* This is a fixed public key, and requests authenticated with that key are allowed to GET and PUT to `/limit`.
+* The device has a shared secret from its authorization server, with which the authorization server secures the tokens it issues to clients. Clients may perform any action as long as they securely present a token that allows it. For example, a token may allow GET on `/limit` and PUT on `/led/0`".
+* Any (even unauthenticated) device may GET `/hello/`.
+
+[provided as `examples/coap`]: https://github.com/future-proof-iot/RIOT-rs/tree/main/examples/coap
+[its `run()` function]: https://github.com/future-proof-iot/RIOT-rs/blob/2b76e560394884d3c8f7eaae51beefd59a316d7b/examples/coap/src/main.rs#L70
+
+### Interacting with RIOT-rs CoAP server from the host
+
+A convenient policy (which is the default of RIOT-rs's examples)
+is to grant the user who flashes the device all access on it.
+When that policy is enabled in the build system,
+an unencrypted key is created in the developer's [state home directory]<!-- precise location TBD -->,
+from where it can be picked up by tools such as [aiocoap-client].
+
+Furthermore,
+when a CoAP server is provisioned through the RIOT-rs build system,
+public keys and their device associations are stored
+in the developer's state home directory.
+
+Together, these files act in a similar way as the classic UNIX files `~/.netrc`, `~/.ssh/id_rsa{,.pub}`.
+They can also double as templates for an application akin to `ssh-copy-id`
+in that they enable a server policy like
+"any device previously flashed on this machine may GET all resources".
+
+[aiocoap-client]: https://aiocoap.readthedocs.io/en/latest/tools.html
+[state home directory]: https://specifications.freedesktop.org/basedir-spec/latest/
+
+## Usage: Client side
+
+The example [provided as `examples/coap`] also contains client steps.
+
+A program that triggers a CoAP request provides[^whatsinarequest] some components to the CoAP stack before phrasing the actual request:
+
+* A **URL describing the resource**, eg. `coap://coap.summit.riot-os.org/agenda` or `coap+tcp://[2001:db8::1]/.well-known/core`.
+
+  Note that while the address is printed here as a text URL
+  (and may even be entered as such in code),
+  it is kept in memory and sent as binary components.
+
+* Optionally, directions regarding how to reach or find the host.
+
+  This is not needed when there are IP literals involved, or DNS is available to resolve names.
+  Directions may be available if the application discovered a usable proxy,
+  or when it is desired to use opt-in discovery mechanisms such as multicast.
+
+* A policy reference on how to authenticate the server, and which identity to present.
+  This is optional if there is a global policy,
+  or if there is an implied security mechanism for the origin.
+
+  Examples of policies that can be available are
+  "expect the server to present some concrete public key, use this secret key once the server is verified",
+  "use a token for this audience and scope obtained from that authentication server",
+  "expect the server to present a chain of certificates for its hostname down to a set of root certificates" (which is the default for web browsers),
+  "establish an encrypted connection and trust the peer's key on first use",
+  down to "do not use any encryption".
+
+[^whatsinarequest]: These components required for a request are not documented as such in the CoAP RFCs,
+    but it is the author's opinion that they are a factual requirement:
+    Implementations may implicitly make decisions on those,
+    but the decisions are still made.
+    At the time of writing, there [is an open issue](https://github.com/core-wg/corrclar/issues/41) to clarify this in the specifications.
+
+## Details on security mechanisms
+
+RIOT-rs uses three pieces of security components:
+OSCORE (for symmetric encryption), EDHOC (for key exchange) and ACE (for authentication).
+
+All have in common that they sit on top of CoAP:
+Their communication consists of CoAP requests and responses.
+Thus, they work homogeneously across all CoAP transports,
+and provide end-to-end security across untrusted proxies.
+
+These components are optional, but enabled by default --
+when disabled, the only sensible policy that is left <!-- "deny everything" is not sensible, could just not include CoAP then -->
+is to allow unauthenticated access everywhere.
+For example, this may make sense on a link layer with tight access control.
+The components also have internal dependencies:
+EDHOC can only practically be used in connection with OSCORE;
+ACE comes with profiles with individual dependencies
+(eg. using the ACE-EDHOC profile requires EDHOC).
+
+### Symmetric encryption: OSCORE
+
+OSCORE ([RFC8613]) provides symmetric encryption for CoAP requests:
+It allows clients to phrase their CoAP request,
+encrypts the parts that a proxy does not need to be aware of[^parts]
+and sends on the ciphertext in a CoAP request.
+The server can decrypt the request,
+process it like any other request (but with the access policy evaluated according to the key's properties),
+and its response gets encrypted likewise.
+
+Working with symmetric keys requires a lot of care and effort managing keys:
+assigning the same key twice can have catastrophic consequences,
+and even recovering from an unplanned reboot is by far not trivial.
+
+RIOT-rs does not offer direct access to OSCORE for those reasons,
+and uses OSCORE's companion mechanisms to set up keys.
+
+Policies are not described in terms of OSCORE keys.
+
+[RFC8613]: https://datatracker.ietf.org/doc/html/rfc8613
+[^parts]: Most of the message is encrypted.
+  Noteworthy unencrypted parts are the hostname the request is sent to,
+  the parts that link a response to a request,
+  and housekeeping details such as whether a request is for an observation and thus needs to be kept alive longer.
+
+### Key establishment: EDHOC
+
+EDHOC ([RFC9528]) is a key establishment protocol
+that uses asymmetric keys to obtain mutual authentication and forward secrecy.
+In particular, two CoAP devices can run EDHOC over CoAP to obtain key material for OSCORE,
+which can then be used for fast communication.
+
+Unless ACE or certificate chains are used,
+the main use of EDHOC in RIOT-rs is with raw public keys:
+Devices (including the host machine) generate a private key,
+make the corresponding public key known,
+and then send the public key (or its short ID) along with EDHOC messages to be identified by that public key.
+This is similar to how SSH keys are commonly used.
+
+Policies described in terms of EDHOC keys include the public key of the peer,
+which private key to use,
+and whether our public key needs to be sent as a full public key or can be sent by key ID.
+
+[RFC9528]: https://datatracker.ietf.org/doc/html/rfc9528
+
+### Authorization: ACE
+
+The ACE framework ([RFC9200]) describes how a trusted service (the Authorization Server, "AS")
+can facilitate secure connections between devices that are not explicitly configured to be used together.
+It frequently issues CWTs (CBOR Web Tokens) that are to JWTs (JSON Web Tokens) as CoAP is to HTTP.
+
+The general theme of it is that
+a client that wants to access some resource on a server under the AS's control
+first asks the AS for an access token,
+and then presents that token to the server (called Resource Server / RS in that context).
+
+Clients know the AS by having means to establish some secure connection with it,
+eg. through EDHOC.
+Resource Servers know the AS by verifying the tokens they receive from the AS,
+eg. by having a shared secret with the AS or knowing its signing key.
+
+Details of the process are specified in ACE profiles;
+apart from those listed here, popular profiles include the DTLS profile and the profiles for group communication.
+
+[RFC9200]: https://datatracker.ietf.org/doc/html/rfc9200
+
+#### ACE-OSCORE profile
+
+With the ACE-OSCORE profile ([RFC9203]),
+the AS provides a random OSCORE key in the token (which is encrypted for the RS),
+and sends the token along with the same OSCORE key through its secure connection to the client.
+Before the client can send OSCORE requests,
+it POSTs the token to the server over an unprotected connection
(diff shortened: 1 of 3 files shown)


book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
+# CoAP
+
+> **Warning**: This documentation is currently ahead of the implementation roadmap.
+
+[CoAP] is an **application layer protocol similar to HTTP** in its use
+(you can POST data onto a resource on a server that is identified by a URI)
```
chrysn, line 6:
```suggestion
(you can POST data onto a resource on a server that is identified by a URL)
```

Thanks @jkarinkl

book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
...
+
+A program that triggers a CoAP request provides some components to the CoAP stack before phrasing the actual request:
+
+* A **URI describing the resource**, eg. `coap://coap.summit.riot-os.org/agenda` or `coap+tcp://[2001:db8::1]/.well-known/core`.
+
+  Note that while the address is printed here as a text URI
+  (and may even be entered as such in code),
+  it is kept in memory and sent as binary components.
+
+* Optionally, directions how to reach or find the host.
```
chrysn, line 95: From @kaspar030: Is this fact or opinion?

Note to self: elaborate.

chrysn: While URI is kind of self-evident to readers making the connection from HTTP, the others are not.

I *claim* that 2 and 3 are factual requirements, which the relevant standards just failed to spell out. Libraries may make it non-obvious because they make the choice for you -- for example, a browser has a "proxy" configuration field and a "do you want to use DoH / which one" field that give you little but existing choice on the directions, and hardcode a "we accept X.509 certificates as long as the root CA is in the CA/BF list or the system certificates or manually added, and the name on that matches your origin". That's by far not the only model, and for example EMail does something different with DANE (nobody checks the certificate chain there, but there might be a TLSA record in DNSSec that tells which key is authorized to receive mails for that place).

It'd probably be sensible at some point to write this up, and I'm linking in some discussion in CoRE's corr-clar document (technically Carsten's document, but adoption call is underway). However, that'll take time, and if the direction makes sense to you, I'll add a footnote around "provides some components" that explains the situation.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
...
+**RIOT-rs supports** the use of CoAP
+for implementing clients, servers or both in a single device.
+As there is more complexity in a CoAP client than a CoAP server,
+the most minimal use is to just have a CoAP server on the embedded system.
+Having both components in use is also a common choice.
+
+As part of our mission for strong security,
+we use encrypted CoAP traffic by default.
+Of the multiple means to secure CoAP,
+we have picked the OSCORE/EDHOC security mechanisms as our default,
```
elenaf9, line 24: Why did we choose OSCORE/EDHOC?

book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
...
+eg. `/.well-known/core` for discovery or `/.well-known/edhoc` for establishing secure connections.
+
+The handler does not need to concern itself with authorization:
+The CoAP stack is configured with an access policy,
+which is evaluated before the handler even receives the request.
+The handler still needs to concern itself with security to some extent
+(eg. file format parsers should still treat incoming data as possibly malformed),
+but is not the main enforcement point.
+
+Policy is configured depending on the enabled security mechanisms.
```
elenaf9, line 52:
```suggestion
The Policy is configured depending on the enabled security mechanisms.
```

book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
...
+Examples of described policy entries are:
+
+* This is an fixed public key, and requests authenticated with that key are allowed to GET and PUT to `/limit`.
```
elenaf9, line 55:
```suggestion
* This is a fixed public key, and requests authenticated with that key are allowed to GET and PUT to `/limit`.
```

book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
...
+* The device has a shared secret from its authorization server, and any requests coming along a token from that server may do whatever the token says they may do (eg. "GET `/limit` and PUT `/led/0`").
```
elenaf9, line 56: This sentence is a bit confusing to me, especially the second half.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,242 @@
...
+  (and may even be entered as such in code),
+  it is kept in memory and sent as binary components.
+
+* Optionally, directions how to reach or find the host.
+
+  This is not needed when there are IP literals involved, or DNS is available to resolve names.
+  Directions may be available if the application discovered a usable proxy,
+  or when it is desired to use opt-in discovery mechanisms such as multicast.
+
+* A policy reference to on how to authenticate the server, and which identity to present.
```
elenaf9, line 101:
```suggestion
* A policy reference on how to authenticate the server, and which identity to present.
```

elenaf9 (commented):
Great chapter!

book/src/tooling/coap.md
chrysn: There is now a footnote saying that in 024f99324bde9ac42810d64ea93dca07ec0da18f

book/src/tooling/coap.md
chrysn: Done in 0dd605b8d3f6eb5df5360ba5ef658b21ca9d3c7f:

```patch
+These were chosen because they scale down well to the smallest devices.
```

book/src/tooling/coap.md
chrysn: Good point, rephrased in 01998806decfd9b7e218e58751619d446e14dd91:

> The device has a shared secret from its authorization server, with which the authorization server secures the tokens it issues to clients. Clients may perform any action as long as they securely present a token that allows it. For example, a token may allow GET on `/limit` and PUT on `/led/0`".

chrysn:
Thanks, good points, all addressed. I've left the non-trivial ones unresolved (except the one I opened myself).

While most fixups are precise (thanks to `git absorb`), I think this can be squashed into a single commit when accepted.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,249 @@
...
+
+[^whatsinarequest]: These components required for a request are not documented in such in the CoAP RFCs,
+    but it is the author's opinion that they are a factual requirement:
+    Implemenations may implicitly make decisions on those,
+    but the decisions are still made.
+    At the time of writing, there [is an open issue](https://github.com/core-wg/corrclar/issues/41) to clarify this in the specifications.
+
+## Details on security mechanisms
+
+RIOT OS uses three pieces of security components:
```
jkarinkl, line 122: RIOT-rs instead of OS?

chrysn: Thanks, indeed, fixed.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,249 @@
...
+* The device has a shared secret from its authorization server, with which the authorization server secures the tokens it issues to clients. Clients may perform any action as long as they securely present a token that allows it. For example, a token may allow GET on `/limit` and PUT on `/led/0`".
+* Any (even unauthenticated) device GET `/hello/`.
+
+[provided as "examples/coap"]: https://github.com/future-proof-iot/RIOT-rs/tree/main/examples/coap
+[its `run()` function]: https://github.com/future-proof-iot/RIOT-rs/blob/2b76e560394884d3c8f7eaae51beefd59a316d7b/examples/coap/src/main.rs#L70
+
+### Interacting with RIOT-rs CoAP server from the host
+
+A convenient policy (which is the default of RIOT-rs's examples)
+is to allow all access to the programmer.
```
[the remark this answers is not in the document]
chrysn: While I accepted this at first, I've now rephrased it to "is to grant the user who flashes the device all access on it."

book/src/tooling/coap.md
```diff
@@ -0,0 +1,249 @@
...
+Their communication consists of CoAP requests and responses.
+Thus, they work homogeneously across all CoAP transports,
+and provide end-to-end security across untrusted proxies.
+
+### OSCORE
+
+OSCORE ([RFC8613]) provides symmetric encryption for CoAP requests:
+It allows clients to phrase their CoAP request,
+encrypts the parts that a proxy does not need to be aware of[^parts]
+and sends the ciphertext on in a CoAP request.
```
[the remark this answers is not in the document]
chrysn: This is as intended: It is a component that takes some piece of data and sends it on to the next place of processing.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,249 @@
...
+Unless ACE or certificate chains are used,
+the main use of EDHOC in RIOT-rs is with raw public keys:
+Devices (including the host machine) generate a private key,
+make the corresponding public key known,
+and then send the public key (or its short ID) along with EDHOC messages to be identified by that public key.
+This is similar to how SSH keys are commonly used.
+
+Policies described in terms of EDHOC keys include the public key of the peer,
+which own private key to use,
+and whether the own key needs to be sent as a full public key or can be sent by key ID.
```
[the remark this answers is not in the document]
chrysn: Clarified as "own public key": in EDHOC a user has the choice to send its full credential (that includes the public key, or in many our cases is just the public key) or just reference the credential (eg. by a key ID, a certificate URI or a certificate fingerprint).

book/src/tooling/coap.md
```diff
@@ -0,0 +1,249 @@
...
+  This is not needed when there are IP literals involved, or DNS is available to resolve names.
+  Directions may be available if the application discovered a usable proxy,
+  or when it is desired to use opt-in discovery mechanisms such as multicast.
+
+* A policy reference on how to authenticate the server, and which identity to present.
+  This is optional if there is a global policy,
+  or if there is an implied security mechanism for the origin.
+
+  Examples of policies that can be available are
+  "expect the server to present some concrete public key, present this own key"
```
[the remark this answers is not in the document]
chrysn: That is to some extent reaching ahead (how identities are presented depends on the details on security mechanisms); rephrased a bit to simplify.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,249 @@
...
+
+[CoAP]: https://coap.space/
+[over UDP]: https://datatracker.ietf.org/doc/html/rfc7252
+[over TCP and WebSockets]: https://datatracker.ietf.org/doc/html/rfc8323
+[over SMS and NB-IoT]: https://www.omaspecworks.org/wp-content/uploads/2018/10/Whitepaper-11.1.18.pdf
+[observation]: https://datatracker.ietf.org/doc/html/rfc7641
+
+## Usage: Server side
+
+An example of a CoAP server is [provided as "examples/coap"], see [its `run()` function] for the practical steps.
```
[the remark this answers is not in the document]
chrysn: I think that's a bit a matter of taste / style guide, and unless we have a consistent guide we follow, I'd go with whether while reading I'd switch into the "this is an identifier" voice, which for path names that are speakable I wouldn't do.

book/src/tooling/coap.md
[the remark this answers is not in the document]
chrysn: I think the rules of by-the-book English would be to "send the object on to", but what you propose is understood and more legible, so let's do it this way. If English doesn't allow it, maybe it should, so let's let language live.

kaspar030 (changes requested):
I really want the OSCORE/EDHOC/ACE stack optional.

*if* we manage to fully tame the added complexity, this can be changed down the line, but initially *just to make that complexity visible*, it has to be optional.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,248 @@
...
+and offers features exceeding the classical REST set such as [observation].
+
+**RIOT-rs supports** the use of CoAP
+for implementing clients, servers or both in a single device.
+As there is more complexity in a CoAP client than a CoAP server,
+the most minimal use is to just have a CoAP server on the embedded system.
+Having both components in use is also a common choice.
+
+As part of our mission for strong security,
+we use encrypted CoAP traffic by default.
```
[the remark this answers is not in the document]
chrysn: It is optional in that "accessible by unencrypted requests" is a policy that can be assigned. That would even build if none of the encrypting modules are configured. Right now I don't have them optional, but I'm all game for making them optional, as long as there is a conscious decision by the user to have that policy. The modules then don't even need to be on-by-default: coapcore would be a trivially thin layer (quite possibly even not resulting in any code) if `Policy::Unsecured` is the only policy available in the enum, making that zero-size information.

ad 4, this is WIP, and when done, all setup should be outside the code.

ad 6, coapcore is open to supporting others (there's just nobody doing that right now).

book/src/tooling/coap.md
```diff
@@ -0,0 +1,248 @@
+# CoAP
+
+> **Warning**: This documentation is currently ahead of the implementation roadmap.
+
+[CoAP] is an **application layer protocol similar to HTTP** in its use
+(you can POST data onto a resource on a server that is identified by a URL)
```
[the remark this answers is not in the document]
chrysn: Would be "for example, " a sufficient addition? I don't want to go into exhaustively listing what it can do, but just to give an example in how they are alike.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,248 @@
...
+[over UDP]: https://datatracker.ietf.org/doc/html/rfc7252
+[over TCP and WebSockets]: https://datatracker.ietf.org/doc/html/rfc8323
+[over SMS and NB-IoT]: https://www.omaspecworks.org/wp-content/uploads/2018/10/Whitepaper-11.1.18.pdf
+[observation]: https://datatracker.ietf.org/doc/html/rfc7641
+
+## Usage: Server side
+
+An example of a CoAP server is [provided as `examples/coap`], see [its `run()` function] for the practical steps.
+
+A CoAP server is created by assembling several **resource handlers on dedicated paths**:
```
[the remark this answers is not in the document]
chrysn: Generally yes, and I'm happy to update once that is final, but the goal of this PR is to enable the RIOT-rs team to align on an understanding how those components interact, while the APIs are not finished. Once examples/coap is in its final shape, we can have a round of "what works better in examples/, what works better in the book, and can we inline from the former into the latter", but AIU we want to have this chapter in to guide the rest of CoAP that follows.

book/src/tooling/coap.md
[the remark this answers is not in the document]
chrysn: Turned into an "eg. you could POST".

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+Having both components in use is also a common choice.
+
+As part of our mission for strong security,
+we use encrypted CoAP traffic by default.
+Of the multiple means to secure CoAP,
+we have picked the OSCORE/EDHOC security mechanisms as our default,
+with the option to use ACE for authorization management.
+These were chosen because they scale down well to the smallest devices.
+Alternatives (DTLS, TLS, IPsec or link-layer encryption)
+are currently not implemented.
```
emmanuelsearch, line 28: I'd be in favor of having a first part of the doc on aspects concerning CoAP except security aspects, and a second part on aspects concerning only CoAP security

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+Thus, they work homogeneously across all CoAP transports,
+and provide end-to-end security across untrusted proxies.
+
+These components are optional, but enabled by default --
+when disabled, the only sensible policy that is left <!-- "deny everything" is not sensible, could just not include CoAP then -->
+is to allow unauthenticated access everywhere.
+For example, this may make sense on a link layer with tight access control.
+The components also have internal dependencies:
+EDHOC can only practically be used in connection with OSCORE;
+ACE depends on either depending on the profiles used.
```
emmanuelsearch, line 135: This sentence is somewhat unclear. Could you rephrase?

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+
+Policies are not described in terms of OSCORE keys.
+
+[RFC8613]: https://datatracker.ietf.org/doc/html/rfc8613
+[^parts]: Most of the message is encrypted.
+  Noteworthy unencrypted parts are the hostname the request is sent to,
+  the parts that link a response to a request,
+  and housekeeping details such as whether a request is for an observation and thus needs to be kept alive longer.
+
+### EDHOC
```
emmanuelsearch, line 162: would rename this section with "key exchange" or something like that

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+and then send the public key (or its short ID) along with EDHOC messages to be identified by that public key.
+This is similar to how SSH keys are commonly used.
+
+Policies described in terms of EDHOC keys include the public key of the peer,
+which private key to use,
+and whether our public key needs to be sent as a full public key or can be sent by key ID.
+
+[RFC9528]: https://datatracker.ietf.org/doc/html/rfc9528
+
+### ACE
```
emmanuelsearch, line 182: would rename to "Authorization" or something equivalent

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+It differs from the ACE-OSCORE profile
+in that the AS does not provide symmetric key material
+but only points out the respective peer's public keys.
+
+The token is not sent in plain,
+but as part of the EDHOC exchange.
+
+[under development in the ACE working group]: https://datatracker.ietf.org/doc/draft-ietf-ace-edhoc-oscore-profile/
+
+#### ACE during development
```
emmanuelsearch, line 233: you mean "under development"?

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+While full operation of ACE requires having an AS as part of the network,
+CoAP servers running on RIOT-rs can be used in the ACE framework without a live server.
+
+Similar to how an EDHOC key is created on demand on the host,
+an AS's list of Resource Servers is maintained by default.
+Tools at the host can then use the locally stored key to create tokens that grant fine-grained permissions on the RIOT-rs device.
+
+With the [New ACE Workflow developed in ACE],
+such tokens can also be provisioned into Resource Servers on behalf of clients that are being provisioned.
+Thus,
```
emmanuelsearch, line 244: I get the impression there are 3 parts for this section of the book, which would benefit from being more clearly distinguished/disentangled:
1. CoAP
2. Securing CoAP
3. "What's in the pipe" & further reading

What do you think about that?

book/src/tooling/coap.md
chrysn: That's just the link expansion. After rendering, this reads as

> The ACE-EDHOC profile is [under development in the ACE working group].
>
> [under development in the ACE working group]: https://datatracker.ietf.org/doc/draft-ietf-ace-edhoc-oscore-profile/

(which I think is the right semantic span to place the link under, following the spirit of [Don't Click Here](https://www.w3.org/QA/Tips/noClickHere) and making the span a sensible title for the target)

book/src/tooling/coap.md
chrysn: Would it help to split up the paragraphs between "For example" and "The components"? Then, which of those is the unclear one?

book/src/tooling/coap.md
```diff
@@ -0,0 +1,256 @@
...
+
+These components are optional, but enabled by default --
+when disabled, the only sensible policy that is left <!-- "deny everything" is not sensible, could just not include CoAP then -->
+is to allow unauthenticated access everywhere.
+For example, this may make sense on a link layer with tight access control.
+The components also have internal dependencies:
+EDHOC can only practically be used in connection with OSCORE;
+ACE depends on either depending on the profiles used.
+
+### OSCORE
```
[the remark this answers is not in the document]
chrysn: (Replying here instead of individually on the 3 affected sections)

Adding what it is about is good, but shouldn't replace the technologies that are described; changed to "Symmetric encryption: OSCORE" etc. in 4658fee81356d19a44b9b60caa9dc6c6d5f2eb16

book/src/tooling/coap.md
chrysn: Item 3 is rolled in because I can't do everything at the same time and you preferred to have overview docs available while the technical part is being worked on. We can split this up later when we know precisely what is in, but until this project phase is done, splitting out 3 would be a microedit nightmare.

As for 1 and 2, they are intentionally close as part of the security theme. They need some words to explain, and more words than 1 on its own due to the tacit assumption that the reader has worked with REST services such as HTTP before, but both are there: Both the intro and the "Usage:" parts go from the fundamentals to how security works with them. If we split a this from "server basics, server security, client basics, client security" into "server basics, client basics, server security, client security", we'd lose a reader interested in the server side at the switch from server to client. (Also, there is little text there so far in terms of basics anyway -- can be changed later, but right now the goal of this is to have an overview that works both as a book part and to convey the plan to the team).

book/src/tooling/coap.md
chrysn: That point is repeated later at https://github.com/future-proof-iot/RIOT-rs/pull/448#discussion_r1804785550, closing this in favor of the other one.

book/src/tooling/coap.md
emmanuelsearch: basically line 135 is unclear to me. "ACE depends on either depending on..." 
How about spelling it out: what does "either" stand for here? 
i.e. ACE depends on what, explicitly?

chrysn: Rephrased in 5d8e30bd6ae681b73181052dd6abcb1d5fe5913d:

> ACE comes with profiles with individual dependencies (eg. using the ACE-EDHOC profile requires EDHOC).

book/src/tooling/coap.md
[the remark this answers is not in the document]
emmanuelsearch: OK

book/src/tooling/coap.md
emmanuelsearch: My comment was about the title of the section... should it not be "ACE under development" ?

book/src/tooling/coap.md
emmanuelsearch: not sure what you mean with micro-edit nightmare, but what's wrong with the rough reshuffling below?

emmanuelsearch: # CoAP

> **Warning**: This documentation is currently ahead of the implementation roadmap.

[CoAP] is an **application layer protocol similar to HTTP** in its use
(eg. you could POST data onto a resource on a server that is identified by a URL)
but geared towards IoT devices in its format and its security mechanisms.

CoAP provides a versatile set of transports
(with IETF Proposed Standards for running [over UDP] including multicast, [over TCP and WebSockets],
other standards for running [over SMS and NB-IoT], and more in development).
It relies on proxies to span across transports and accommodate the characteristics of particular networks,
and offers features exceeding the classical REST set such as [observation].

**RIOT-rs supports** the use of CoAP
for implementing clients, servers or both in a single device.
As there is more complexity in a CoAP client than a CoAP server,
the most minimal use is to just have a CoAP server on the embedded system.
Having both components in use is also a common choice.

Also, as part of our mission for strong security,
we use encrypted CoAP traffic by default as explained below.

[CoAP]: https://coap.space/
[over UDP]: https://datatracker.ietf.org/doc/html/rfc7252
[over TCP and WebSockets]: https://datatracker.ietf.org/doc/html/rfc8323
[over SMS and NB-IoT]: https://www.omaspecworks.org/wp-content/uploads/2018/10/Whitepaper-11.1.18.pdf
[observation]: https://datatracker.ietf.org/doc/html/rfc7641

## Usage: Server side

An example of a CoAP server is [provided as `examples/coap`], see [its `run()` function] for the practical steps.

A CoAP server is created by assembling several **resource handlers on dedicated paths**:
There might be a path `/s/0` representing a particular sensor,
and a path `/fwup` for interacting with firmware updates.
The CoAP implementation can put additional resources at well-known locations,
eg. `/.well-known/core` for discovery or `/.well-known/edhoc` for establishing secure connections.

The handler only needs to concern itself with basic security aspects
(eg. file format parsers should treat incoming data as possibly malformed),
but is not the main enforcement point.


[provided as `examples/coap`]: https://github.com/future-proof-iot/RIOT-rs/tree/main/examples/coap
[its `run()` function]: https://github.com/future-proof-iot/RIOT-rs/blob/2b76e560394884d3c8f7eaae51beefd59a316d7b/examples/coap/src/main.rs#L70


## Usage: Client side

The example [provided as `examples/coap`] also contains client steps.

A program that triggers a CoAP request provides[^whatsinarequest] some components to the CoAP stack before phrasing the actual request:

* A **URL describing the resource**, eg. `coap://coap.summit.riot-os.org/agenda` or `coap+tcp://[2001:db8::1]/.well-known/core`.

  Note that while the address is printed here as a text URL
  (and may even be entered as such in code),
  it is kept in memory and sent as binary components.

* Optionally, directions regarding how to reach or find the host.

  This is not needed when there are IP literals involved, or DNS is available to resolve names.
  Directions may be available if the application discovered a usable proxy,
  or when it is desired to use opt-in discovery mechanisms such as multicast.

* A policy reference on how to authenticate the server, and which identity to present.
  This is optional if there is a global policy,
  or if there is an implied security mechanism for the origin.


[^whatsinarequest]: These components required for a request are not documented as such in the CoAP RFCs,
    but it is the author's opinion that they are a factual requirement:
    Implementations may implicitly make decisions on those,
    but the decisions are still made.
    At the time of writing, there [is an open issue](https://github.com/core-wg/corrclar/issues/41) to clarify this in the specifications.

## Security

The CoAP stack is configured with an access policy,
which is evaluated before the handler receives the request.

Furthermore, the CoAP communication is secured using mechanisms for
symmetric encryption, key exchange and authentication. 

At this stage, RIOT-rs uses three pieces of security components:
OSCORE (for symmetric encryption), EDHOC (for key exchange) and ACE (for authentication).

Alternatives are possible (for instance DTLS, TLS, IPsec or link-layer encryption)
but are currently not implemented / not yet supported in RIOT-rs.

#### Rationale for OSCORE/EDHOC/ACE

OSCORE/EDHOC/ACE were chosen first because they scale down
well to the smallest devices, and because they
all have in common that they sit naturally on top of CoAP:
Their communication consists of CoAP requests and responses.
Thus, they work homogeneously across all CoAP transports,
and provide end-to-end security across untrusted proxies.

These components are optional, but enabled by default --
when disabled, the only sensible policy that is left 
is to allow unauthenticated access everywhere.
For example, this may make sense on a link layer with tight access control.
The components also have internal dependencies:
EDHOC can only practically be used in connection with OSCORE;
ACE depends on either depending on the profiles used.

### Access policy

The policy is configured depending on the enabled security mechanisms.
Examples of described policy entries are:

* This is a fixed public key, and requests authenticated with that key are allowed to GET and PUT to `/limit`.
* The device has a shared secret from its authorization server, with which the authorization server secures the tokens it issues to clients. Clients may perform any action as long as they securely present a token that allows it. For example, a token may allow GET on `/limit` and PUT on `/led/0`".
* Any (even unauthenticated) device may GET `/hello/`.

#### Interacting with RIOT-rs CoAP server from the host

A convenient policy (which is the default of RIOT-rs's examples)
is to grant the user who flashes the device all access on it.
When that policy is enabled in the build system,
an unencrypted key is created in the developer's [state home directory],
from where it can be picked up by tools such as [aiocoap-client].

Furthermore,
when a CoAP server is provisioned through the RIOT-rs build system,
public keys and their device associations are stored
in the developer's state home directory.

Together, these files act in a similar way as the classic UNIX files `~/.netrc`, `~/.ssh/id_rsa{,.pub}`.
They can also double as templates for an application akin to `ssh-copy-id`
in that they enable a server policy like
"any device previously flashed on this machine may GET all resources".

[aiocoap-client]: https://aiocoap.readthedocs.io/en/latest/tools.html
[state home directory]: https://specifications.freedesktop.org/basedir-spec/latest/

#### Interacting with the RIOT-rs CoAP client

Examples of policies that can be available are
  "expect the server to present some concrete public key, use this secret key once the server is verified",
  "use a token for this audience and scope obtained from that authentication server",
  "expect the server to present a chain of certificates for its hostname down to a set of root certificates" (which is the default for web browsers),
  "establish an encrypted connection and trust the peer's key on first use",
  down to "do not use any encryption".


### Symmetric encryption: OSCORE

OSCORE ([RFC8613]) provides symmetric encryption for CoAP requests:
It allows clients to phrase their CoAP request,
encrypts the parts that a proxy does not need to be aware of[^parts]
and sends on the ciphertext in a CoAP request.
The server can decrypt the request,
process it like any other request (but with the access policy evaluated according to the key's properties),
and its response gets encrypted likewise.

Working with symmetric keys requires a lot of care and effort managing keys:
assigning the same key twice can have catastrophic consequences,
and even recovering from an unplanned reboot is by far not trivial.

RIOT-rs does not offer direct access to OSCORE for those reasons,
and uses OSCORE's companion mechanisms to set up keys.

Policies are not described in terms of OSCORE keys.

[RFC8613]: https://datatracker.ietf.org/doc/html/rfc8613
[^parts]: Most of the message is encrypted.
  Noteworthy unencrypted parts are the hostname the request is sent to,
  the parts that link a response to a request,
  and housekeeping details such as whether a request is for an observation and thus needs to be kept alive longer.

### Key establishment: EDHOC

EDHOC ([RFC9528]) is a key establishment protocol
that uses asymmetric keys to obtain mutual authentication and forward secrecy.
In particular, two CoAP devices can run EDHOC over CoAP to obtain key material for OSCORE,
which can then be used for fast communication.

Unless ACE or certificate chains are used,
the main use of EDHOC in RIOT-rs is with raw public keys:
Devices (including the host machine) generate a private key,
make the corresponding public key known,
and then send the public key (or its short ID) along with EDHOC messages to be identified by that public key.
This is similar to how SSH keys are commonly used.

Policies described in terms of EDHOC keys include the public key of the peer,
which private key to use,
and whether our public key needs to be sent as a full public key or can be sent by key ID.

[RFC9528]: https://datatracker.ietf.org/doc/html/rfc9528

### Authorization: ACE

The ACE framework ([RFC9200]) describes how a trusted service (the Authorization Server, "AS")
can facilitate secure connections between devices that are not explicitly configured to be used together.
It frequently issues CWTs (CBOR Web Tokens) that are to JWTs (JSON Web Tokens) as CoAP is to HTTP.

The general theme of it is that
a client that wants to access some resource on a server under the AS's control
first asks the AS for an access token,
and then presents that token to the server (called Resource Server / RS in that context).

Clients know the AS by having means to establish some secure connection with it,
eg. through EDHOC.
Resource Servers know the AS by verifying the tokens they receive from the AS,
eg. by having a shared secret with the AS or knowing its signing key.

Details of the process are specified in ACE profiles;
apart from those listed here, popular profiles include the DTLS profile and the profiles for group communication.

[RFC9200]: https://datatracker.ietf.org/doc/html/rfc9200

#### ACE-OSCORE profile

With the ACE-OSCORE profile ([RFC9203]),
the AS provides a random OSCORE key in the token (which is encrypted for the RS),
and sends the token along with the same OSCORE key through its secure connection to the client.
Before the client can send OSCORE requests,
it POSTs the token to the server over an unprotected connection
(the token itself is encrypted),
along with a random number and some housekeeping
that go into the establishment of an OSCORE context.

The [documentation of the CoAP/ACE-OAuth PoC project] describes the whole setup in more detail.

[documentation of the CoAP/ACE-OAuth PoC project]: https://gitlab.com/oscore/coap-ace-poc-overview

[RFC9203]: https://datatracker.ietf.org/doc/html/rfc9203

#### ACE-EDHOC profile

The ACE-EDHOC profile is [under development in the ACE working group].

It differs from the ACE-OSCORE profile
in that the AS does not provide symmetric key material
but only points out the respective peer's public keys.

The token is not sent in plain,
but as part of the EDHOC exchange.

[under development in the ACE working group]: https://datatracker.ietf.org/doc/draft-ietf-ace-edhoc-oscore-profile/

#### ACE during development

While full operation of ACE requires having an AS as part of the network,
CoAP servers running on RIOT-rs can be used in the ACE framework without a live server.

Similar to how an EDHOC key is created on demand on the host,
an AS's list of Resource Servers is maintained by default.
Tools at the host can then use the locally stored key to create tokens that grant fine-grained permissions on the RIOT-rs device.

With the [New ACE Workflow developed in ACE],
such tokens can also be provisioned into Resource Servers on behalf of clients that are being provisioned.
Thus,
the offline AS can enable deployed RIOT-rs based CoAP servers
to accept requests from newly created RIOT-rs based CoAP clients
without the need for the CoAP client to create a network connection to the host.
(Instead, the host needs to find the Resource Server over the network).

.. note: Some more exploration of this workflow will be necessary
  as to how the client can trigger the AS to re-install (or renew) its token
  in case the Resource Server retired the token before its expiration.
  For RIOT-rs internal use,
  AS-provisioned tokens might just be retained longer.

[New ACE Workflow developed in ACE]: https://www.ietf.org/archive/id/draft-ietf-ace-workflow-and-params-00.html#name-new-ace-workflow

emmanuelsearch: (sorry for the extra-long recursive post ;))

book/src/tooling/coap.md
chrysn: Ah, that's what you mean:

No, this refers to the first two paragraphs of that -- you *can* have a full-blown ACE Authorization Server, but here we describe how ACE is used in a minimal local-machine mode.

book/src/tooling/coap.md
chrysn: The microedit nightmare referred to what there'd be if we split between what is implemented now and what is not. (Adding this section is now already taking a month when it was planned to be a one-day thing between status updates -- I'd like to only touch this once for every large change in functionality).

I've applied your version as a commit and added a few commits to alter where I thought things would become misleading or inconsistently structured.

book/src/tooling/coap.md
```diff
@@ -0,0 +1,265 @@
...
+
+* Optionally, directions regarding how to reach or find the host.
+
+  This is not needed when there are IP literals involved, or DNS is available to resolve names.
+  Directions may be available if the application discovered a usable proxy,
+  or when it is desired to use opt-in discovery mechanisms such as multicast.
+
+* A policy reference on how to authenticate the server, and which identity to present.
+  This is optional if there is a global policy,
+  or if there is an implied security mechanism for the origin.
```
kaspar030, line 68: "origin" here means URL?

book/src/tooling/coap.md
```diff
@@ -0,0 +1,265 @@
...
+
+
+[^whatsinarequest]: These components required for a request are not documented as such in the CoAP RFCs,
```
kaspar030, line 71:
```suggestion
[^whatsinarequest]: The components required for a request are not documented as such in the CoAP RFCs,
```

book/src/tooling/coap.md
```diff
@@ -0,0 +1,265 @@
...
+  down to "do not use any encryption".
+
+### Available security mechanisms
+
+These components are optional, but enabled by default --
+when disabled, the only sensible policy that is left <!-- "deny everything" is not sensible, could just not include CoAP then -->
+is to allow unauthenticated access everywhere.
+For example, this may make sense on a link layer with tight access control.
+The components also have internal dependencies:
+EDHOC can only practically be used in connection with OSCORE;
```
kaspar030, line 143:
```suggestion
EDHOC can only practically be used in combination with OSCORE;
```

book/src/tooling/coap.md
```diff
@@ -0,0 +1,248 @@
...
+
+CoAP provides a versatile set of transports
+(with IETF Proposed Standards for running [over UDP] including multicast, [over TCP and WebSockets],
+other standards for running [over SMS and NB-IoT], and more in development).
+It relies on proxies to span across transports and accommodate the characteristics of particular networks,
+and offers features exceeding the classical REST set such as [observation].
+
+**RIOT-rs supports** the use of CoAP
+for implementing clients, servers or both in a single device.
+As there is more complexity in a CoAP client than a CoAP server,
```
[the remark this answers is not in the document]
kaspar030: Can we just drop that sentence? IMO it doesn't make much sense - applications need to use either client or server. One would never use a server instead of a client because it is more minimal.

book/src/tooling/coap.md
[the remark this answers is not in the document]
kaspar030: ok

book/src/tooling/coap.md
```diff
@@ -0,0 +1,265 @@
...
+[^whatsinarequest]: These components required for a request are not documented as such in the CoAP RFCs,
+    but it is the author's opinion that they are a factual requirement:
+    Implementations may implicitly make decisions on those,
+    but the decisions are still made.
+    At the time of writing, there [is an open issue](https://github.com/core-wg/corrclar/issues/41) to clarify this in the specifications.
+
+## Security
+
+The CoAP stack is configured with server and client policies.
+The security mechansims used depend on those selected in the policies.
```
kaspar030, line 80:
```suggestion
The security mechanisms used depend on those selected in the policies.
```

(typo)

kaspar030: This might warrant an example. Does this means that when the policy defines "allow access to anyone having this or that token", the cwt token infra is used? what does used mean?

kaspar030:
I much prefer this restructured version!

book/src/tooling/coap.md
chrysn: Generally it means that, but how precisely that is done is something I'll only see during modularization. May be dead code elimination, may be that the policy can be described well enough to be visible to laze, may be that a warning is shown when there are unused mechanisms because the policy doesn't have an entry, may be that it'll need rephrasing for "Security mechanisms need to be enabled in order to be available in the policies".

book/src/tooling/coap.md
chrysn: It means something [more precise](https://datatracker.ietf.org/doc/html/rfc6454), but explaining the differences would be excessive, so I just put URL instead.

chrysn:
I've done an editing pass fixing some remaining typos, grammer errors or inconsistencies (finally found a better subsection title, s/ACE during development/Using ACE from the host during development/ is now consistent with the earlier heading "Interacting with a RIOT-rs CoAP server from the host"), and applied fixes that got lost over non-git contributions.

@kaspar030, if you're fine with this and can press the button, I'll squash.

kaspar030 (approved):

kaspar030:
> @kaspar030, if you're fine with this and can press the button, I'll squash.

let's go!

kaspar030:
Feel free to "gq" the paragraphs to a consistent look ...

chrysn:
> Feel free to "gq" the paragraphs to a consistent look ...

I expect to review later edits on those, so I'd prefer not to; https://sembr.org/ says it better than I can.

Issue #446, opened 2024-09-27
Provisioning roadmap


chrysn:
### Description

Around the flashing step we should have an option to customize what is on the device.

Whether that is part of the built source code (eg. through environment variables), linked in, flashed to a dedicated memory address, flashed through some code placed in RAM and executed there or done by the starting application while it communicates with the debugger is an implementation aspect that will need to be figured out, but may not be the same across all platforms.

### Motivation

To make the usability aspects of CoAP (road map on https://github.com/future-proof-iot/RIOT-rs/issues/226) fly, we need some better provisioning.

### Requirements

Concrete requirements are that

1. Devices get unique keys (or key properties such as usage counters / sender IDs).[^1]
2. When a device gets a key, its previous keys are revoked.
3. Optionally, devices when re-flashed get the same keys again.

Item 1 is just security essentials[^1], item 2 is for practical reasons: If we want to express "All my devices may access this resource" without checking signatures, the public keys of all those devices need to be in some list, ideally distributed to other devices – which is practical for the 20-or-so distinct boards a typical developer ever flashes, but not for the 20k different boards those appear as if every flashing creates a new participant without revoking the old key. Item 3 is convenient if you don't want to limit "All my devices" to "All my devices so far", but update a device without revoking its permissions. (Using this is a trade-off, and I think that by that time one would rather use signed identities, but 1+2 enable it to the point where it is cheap enough).

[^1]: Protocols nowadays sometimes have mitigation in place, so eg. in EDHOC (unlike in similar protocols like earlier versions of Tox) when someone else has your key they can't impersonate anyone else to the owner that key, but that is supposed to be an extra layer of security and not relied on.

### Open questions

* Will probe-rs or other debuggers help us?

  Not immediately: Unlike in debuggers specialized to concrete series (eg. those on EFM32 devices where I've just used that for years and took it for granted), there is no concept of a device identity that a flashing or debugging process can be aimed at – some chips have registers that could be read through a debug interface, some chips have an area in their in-SoC QSPI flash for that, some have nothing but could have an identifying chip on board (eg. the MAC of the soldered-on Ethernet module).

* How does such customization fit in the laze process? Will the `run` target commonly used now, instead of being flash-and-attach, become flash-and-provision-and-attach? When (which I expect can be a step on the way) binaries are customized, how does the pre-flash provisioning preparation feed data back into the build process?

### Are you planning to do it yourself in a pull request?

Yes

chrysn:
[...]

PR #453, opened 2024-10-04
Provisioning sketch

chrysn:
# Description

Building on #444, this is a first demo of how the device ID could be used to do something with it outside the system.

Aligned with the latest plans for https://github.com/future-proof-iot/RIOT-rs/issues/446, this collects some data at startup time (once storage is available: after having generated a private/public key pair), and reports them. For applications with defmt, the only way I found to pick up data from the output was to use `cargo embed`, which can tee the reassembled defmt texts to a TCP socket. (I generally think that cargo-embed is something we should look into: It also allows neat things such a having stdout / defmt (even simultaneously for those so inclined) while also having GDB through probe-rs).

## Issues/PRs references

* Right now, this needs an Embed.toml in the application directory. With https://github.com/probe-rs/probe-rs/issues/2845, that could go away and be generated.

## Open Questions

* Is #446 on the right track?

## Change checklist

- [ ] This is a proper PR and not a sketch.


chrysn:
The latest commit uses the --config-file option to probe-rs -- that has been merged, but will require installation from git.

Pausing activity here until device-id (#444) is done; maybe by that time, a `cargo install probe-rs` will already be the easy way to use it.

chrysn:
Now that probe-rs 0.28 is out, this doesn't need to depend on cargo-embed any more; I'll need to update my branch accordingly.

PR #456, opened 2024-10-09
feat(tests/benchmarks): add `bench_sched_flags`

elenaf9:
# Description

Add new benchmark to test signaling between threads using `thread-flags`.
Compared to the existing benchmark `bench_sched_yield`, it involves IPC between the threads and one thread changing another thread's state.

## Issues/PRs references

Extracted from #397.

## Open Question

Do we want this benchmark?

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://future-proof-iot.github.io/RIOT-rs/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #459, opened 2024-10-09
refactor(threads): avoid unnecessary scheduler invocations

elenaf9:
# Description

This PR implements some changes and checks to avoid unnecessary scheduler invocations:
- the scheduler is now triggered in `set_state` as a central point
- if a thread changes into `ThreadState::Running`, only trigger the scheduler if the thread has a higher priority than the currently running one (**new**)
- always trigger the scheduler when a thread changes from `ThreadState::Running` into another state because this only ever happens if a currently running thread gets blocked
- only trigger the scheduler in `yield_now` if the runqueue contains any other threads

## Issues/PRs references

Extracted from #397.

## Benchmarks

(ticks **per iteration**)
`bench_sched_flags` is PR'd in #456

nRF52840dk: 

benchmark | main | this PR
| - | - | -
`bench_sched_yield` with two threads |  560 | 580
`bench_sched_yield` with one thread | 235 | 98
`bench_sched_flags` | 1848 |  1461

RPI-Pico

benchmark | main | this PR
| - | - | -
`bench_sched_yield` with two threads |  631 | 647
`bench_sched_yield` with one thread | 243 | 78
`bench_sched_flags` | 1739 |  1495

The additional checks come a some cost when the scheduler invocation is really needed, as visible in the `bench_sched_yield` benchmark. 
But I think this is tolerable, because the actual context switching logic (i.e. `sched`) is not affected, and in many cases we save an unnecessary scheduler call that would be much more expensive.

## Open Questions

See self review.

## Change checklist

### DIFF
--- a/src/riot-rs-threads/src/lib.rs
+++ b/src/riot-rs-threads/src/lib.rs
@@ -118,20 +118,16 @@ impl Threads {
         arg: usize,
         stack: &'static mut [u8],
         prio: RunqueueId,
-    ) -> Option<&mut Thread> {
-        if let Some((thread, pid)) = self.get_unused() {
-            Cpu::setup_stack(thread, stack, func, arg);
-            thread.prio = prio;
-            thread.pid = pid;
-            thread.state = ThreadState::Paused;
-
-            Some(thread)
-        } else {
-            None
-        }
+    ) -> Option<ThreadId> {
+        let (thread, pid) = self.get_unused()?;
+        Cpu::setup_stack(thread, stack, func, arg);
+        thread.prio = prio;
+        thread.pid = pid;
+        thread.state = ThreadState::Paused;
+        Some(pid)
     }
 
-    /// Returns access to any thread data.
+    /// Returns immutable access to any thread data.
     ///
     /// # Panics
     ///
@@ -172,22 +168,29 @@ impl Threads {
         }
     }
 
-    /// Sets the state of a thread.
+    /// Sets the state of a thread and triggers the scheduler if needed.
     ///
-    /// This function handles adding/ removing the thread to the Runqueue depending
+    /// This function also handles adding/ removing the thread to the Runqueue depending
     /// on its previous or new state.
     ///
     /// # Panics
     ///
     /// Panics if `pid` is >= [`THREADS_NUMOF`].
     fn set_state(&mut self, pid: ThreadId, state: ThreadState) -> ThreadState {
-        let thread = &mut self.threads[usize::from(pid)];
-        let old_state = thread.state;
-        thread.state = state;
-        if old_state != ThreadState::Running && state == ThreadState::Running {
-            self.runqueue.add(thread.pid, thread.prio);
-        } else if old_state == ThreadState::Running && state != ThreadState::Running {
-            self.runqueue.pop_head(thread.pid, thread.prio);
+        let thread = self.get_unchecked_mut(pid);
+        let old_state = core::mem::replace(&mut thread.state, state);
+        let prio = thread.prio;
+
+        if state == ThreadState::Running {
+            self.runqueue.add(pid, prio);
+            self.schedule_if_higher_prio(pid, prio);
+        } else if old_state == ThreadState::Running {
+            // A running thread is only set to a non-running state
+            // if it itself initiated it.
+            debug_assert_eq!(Some(pid), self.current_pid());
+
+            self.runqueue.pop_head(pid, prio);
+            schedule();
         }
 
         old_state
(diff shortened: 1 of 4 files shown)


src/riot-rs-threads/src/lib.rs
```diff
@@ -172,22 +168,29 @@ impl Threads {
...
+        if state == ThreadState::Running {
+            self.runqueue.add(pid, prio);
+            self.schedule_if_higher_prio(pid, prio);
+        } else if old_state == ThreadState::Running {
+            // A running thread is only set into a non-running state
+            // if it itself initiated it.
+            debug_assert_eq!(Some(pid), self.current_pid());
+
+            self.runqueue.pop_head(pid, prio);
+            schedule();
```
elenaf9, line 193: I removed the checks that the respective other state is `!= ThreadState::Running`, because we never have that case in our code.

Interestingly, if I don't remove them `bench_sched_yield` only takes 574 instead of 580 ticks for the nRF52840, which is very weird given that the benchmark doesn't even use this method apart from the initial thread creation. Any idea why that could be @kaspar030?
(`bench_sched_flags` performs as expected better without the checks).

src/riot-rs-threads/src/lib.rs
```diff
@@ -118,20 +118,16 @@ impl Threads {
...
-        } else {
-            None
-        }
+    ) -> Option<ThreadId> {
+        let (thread, pid) = self.get_unused()?;
+        Cpu::setup_stack(thread, stack, func, arg);
+        thread.prio = prio;
+        thread.pid = pid;
+        thread.state = ThreadState::Paused;
+        Some(pid)
```
elenaf9, line 127: Drive-by improvement because I noticed that the caller doesn't need the `&mut Thread`.

src/riot-rs-threads/src/lib.rs
```diff
@@ -202,37 +205,62 @@ impl Threads {
...
+
         if self.runqueue.peek_head(old_prio) == Some(thread_id) {
             self.runqueue.pop_head(thread_id, old_prio);
         } else {
             self.runqueue.del(thread_id);
         }
         self.runqueue.add(thread_id, prio);
-        true
+
+        // Check if the thread currently running and trigger the scheduler if
```
elenaf9, line 237:
```suggestion
        // Check if the thread is currently running and trigger the scheduler if
```

src/riot-rs-threads/src/lib.rs
```diff
@@ -202,37 +205,62 @@ impl Threads {
...
         let thread = self.get_unchecked_mut(thread_id);
         let old_prio = thread.prio;
         if old_prio == prio {
-            return false;
+            return;
         }
         thread.prio = prio;
         if thread.state != ThreadState::Running {
-            return false;
+            // No runqueue changes or scheduler invocations needed.
```
elenaf9, line 226:
```suggestion
            // No runqueue changes or scheduler invocation needed.
```

src/riot-rs-threads/src/lib.rs
```diff
@@ -172,22 +168,29 @@ impl Threads {
...
-            self.runqueue.pop_head(thread.pid, thread.prio);
+        let thread = self.get_unchecked_mut(pid);
+        let old_state = core::mem::replace(&mut thread.state, state);
+        let prio = thread.prio;
+
+        if state == ThreadState::Running {
+            self.runqueue.add(pid, prio);
+            self.schedule_if_higher_prio(pid, prio);
+        } else if old_state == ThreadState::Running {
+            // A running thread is only set into a non-running state
```
elenaf9, line 188:
```suggestion
            // A running thread is only set to a non-running state
```

src/riot-rs-threads/src/lib.rs
kaspar030: > I removed the checks that the respective other state is `!= ThreadState::Running`, because we never have that case in our code.

Hm, while that may well be, are we ok if we made a mistake here? E.g., in the first case, `runqueue.add()` calls `runqueue.queues.push()`, which with a quick glance probably won't like if a thread is added a second time.

elenaf9: `CList::push` prevents that thread is added twice, by early-returning if `self.next_idxs[n as usize] != Self::sentinel()`.

And in the second `else if` branch the check `state != ThreadState::Running` then is redundant because that's already catched by the first branch.

src/riot-rs-threads/src/lib.rs
```diff
@@ -202,37 +205,62 @@ impl Threads {
...
+        }
+
+        // Thread isn't running.
+        // Only schedule if the thread has a higher priority than the running one.
+        if prio > old_prio {
+            self.schedule_if_higher_prio(thread_id, prio);
+        }
+    }
+
+    /// Triggers the scheduler if the thread has higher priority than the current running thread.
```
kaspar030, line 253:
```suggestion
    /// Triggers the scheduler if the thread has higher priority than the currently running thread.
```

src/riot-rs-threads/src/lib.rs
kaspar030: no, fine, just verifying. :slightly_smiling_face:

src/riot-rs-threads/src/lib.rs
elenaf9:
```suggestion
    /// Triggers the scheduler if the thread has a higher priority than the currently running thread.
```

elenaf9:
Rebased so that recently merged `bench_sched_flags` can be run on this branch.

kaspar030:
LGTM!

PR #460, opened 2024-10-09
feat(runqueue): add RunQueue iterator

elenaf9:
# Description

Add an iterator over the threads in `RunQueue`. The iterator can start from any thread and will then iterate over all succeeding threads in the same queue, and then all lower index queues one by one.

The motivation for this iterator is that then `Iterator::filter` + `Iterator::next` can be used to pick a thread from the runqueue that matches a certain condition. 
In the multicore PR #397 that's needed so that we can get the next thread from the runqueue that has the correct core-affinity.

## Issues/PRs references

Extracted from #397.

## Open Questions

Is there any simpler way to implement this?

## Change checklist

### DIFF
--- a/src/riot-rs-runqueue/src/runqueue.rs
+++ b/src/riot-rs-runqueue/src/runqueue.rs
@@ -121,8 +121,8 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
         if rq_ffs == 0 {
             return None;
         }
-        let rq = RunqueueId::new(rq_ffs as u8 - 1);
-        self.queues.peek_head(rq.0).map(ThreadId::new)
+        let rq = rq_ffs as u8 - 1;
+        self.queues.peek_head(rq).map(ThreadId::new)
     }
 
     /// Advances runqueue number `rq`.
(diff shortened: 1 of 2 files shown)


elenaf9:
Hax doesn't know the `Iterator` trait.

I briefly tried before to just implement a filter function instead of using `Iterator::filter` + `::next`. But it performed slightly worse in the benchmarks than the Iterator, and imo using an Iterator for this is more Rust idiomatic .
 Also, it would require a generic `FnMut` predicate function, and those aren't supported yet either.

cc @franziskuskiefer is it possible/ much effort to add support for one of the two (either the `Iterator` trait, or `FnMut`/ `Fn` generics) to hax?

src/riot-rs-runqueue/src/runqueue.rs
```diff
@@ -108,21 +108,36 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
...
         }
-        let rq = RunqueueId::new(rq_ffs as u8 - 1);
-        self.queues.peek_head(rq.0).map(ThreadId::new)
+        let rq = rq_ffs as u8 - 1;
+        self.queues
+            .peek_head(rq)
+            .map(|id| (ThreadId::new(id), RunqueueId::new(rq)))
+    }
+
+    /// Returns the next thread from the runqueue that fulfils the predicate.
```
kaspar030, line 131:
```suggestion
    /// Returns the next thread from the runqueue that fulfills the predicate.
```

kaspar030:
please squash!

kaspar030 (approved):

PR #469, opened 2024-10-16
fix(examples/coap): Enable again

chrysn:
# Description

This reverts d351ee88 / [PR447].

[PR447]: https://github.com/future-proof-iot/RIOT-rs/pull/447

## Issues/PRs references

There have been issues with libOSCORE when building with clang 18, due to which the example has been disabled in #447; #443 proposes to update to a libOSCORE that works with (but also requires) clang 18.

The single commit in this is also part of #443 now.

## Change checklist

This is a draft PR because I need to see whether CI maybe went back on the clang 18 update (#443 only works if everything is clang 18). Only one of those makes sense.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,7 +83,7 @@ jobs:
           git config --global user.name "CI"
           cargo binstall --no-confirm --no-symlinks --force --no-discover-github-token laze
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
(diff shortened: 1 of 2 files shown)


kaspar030:
I guess as soon as this passes CI it can be merged.

chrysn:
Now that https://github.com/future-proof-iot/RIOT-rs/pull/443 is not potentially problematic through a std=c23 dependency any more: Closing this, working on the rest there.

Issue #201, opened 2024-03-19
RFC for a sensor abstraction layer


ROMemories:
# RFC for a sensor abstraction layer

## Definitions

For the purpose of this RFC, we introduce the following definitions:

- A *sensor device* is a device measuring one or multiple physical quantities and reporting them as one or more digital values.
- Sensor devices measuring the same physical quantity are said to be part of the same *sensor class*.
- A *reading* is the result returned by a sensor device of measuring a physical quantity.
- A *sensor driver* refers to a sensor device exposed by the sensor abstraction layer.
- *Consumers* refer to consumers of the abstraction.
- *Users* refer to users of the hardware product.

## Goals

- Provide an abstraction over sensors, exposing readings in a way that consumers do not need to know about implementation details of specific sensor devices within a sensor class.
- Allow sensor drivers to expose metadata relevant for formatting measurements in a physically correct and consistent fashion.
- Allow to list connected sensor devices, with names understandable by users.
- Allow sensor drivers to notify consumers of relevant changes, in an energy-efficient manner.

### Sensor driver management

- A sensor driver MUST be discoverable for listing by a central entity.
- It MUST be possible to disable and re-enable sensor drivers—e.g., in case of failure and for energy efficiency.
- A sensor driver MUST provide the sensor device's part number, when known.
- A sensor driver MUST provide its version number.
- A sensor driver SHOULD provide a user-friendly display name.

### Measurements

- A sensor driver MUST provide a polling interface and/or a notification-based interface.
- Within the notification-based interface, consumers SHOULD be able to request sensor drivers to provide notifications when:
    - a new reading value is available, with a configurable debouncing frequency,
    - the reading value has crossed a configured threshold,
    - the first-order time derivative of the reading value has crossed a configured threshold.
- A sensor driver MUST be able to report multiple values at once, as data tuples, with a common prescaler. The cardinality of such data tuples and the prescaler are fixed per sensor driver.
- A sensor driver MUST be able to report the unit of measurement of its readings. The unit of measurement is fixed per sensor driver.
- A sensor driver SHOULD be able to report its measurement accuracy/error, for individual readings.
- A sensor driver SHOULD be able to report its measurement precision.

## Open questions

- Should sensor drivers be able to tie readings to a timestamp of the time of measurement?

## References

[RIOT-c's SAUL](https://doc.riot-os.org/group__drivers__saul.html)
[BTHome](https://bthome.io/format/)
[...]

PR #474, opened 2024-10-18
feat: introduce a sensor abstraction

### LINKED ISSUE #334: "portable" sensor reading

ROMemories:
# Description

This is the first step in providing a sensor abstraction/API whose goal is to provide consistent access to sensor drivers and their readings, abstracting over implementation details of specific sensor devices. It is a (partial) implementation of RFC #201 (handling of thresholds and sensor interrupts will come later and is expected to be relatively orthogonal).

The main change is the introduction of a `ariel-os-sensors` crate, which is re-exported in `ariel-os` as the `sensors` module, which is feature-gated behind the `sensors` Cargo feature. Until this API and the first sensor drivers are landed, this feature is considered experimental/unstable and breaking changes will therefore not be reflected on the version number of the `ariel-os` crate.

## Issues/PRs references

Partial implementation of RFC #201.
Closes #334.

## Open Questions

- Should we derive `serde::Serialize` on the relevant items?
- Are floats completely out of the question, even as helpers? Integers should stay at the core of the API, but maybe we could have additional (feature-gated?) helpers on `Value`+`ReadingAxis`+`Accuracy` which would return floats taking into account everything required to make sense of the reading (in particular the scaling value)?
- The `Reading` trait was initially introduced when it was thought it would be possible to return different concrete types from `Sensor::wait_for_reading()` (it now always return a `Values`); should the trait be removed and its methods implemented on `Values` itself, slightly simplifying the API (it is not *impossible* that Rust would allow returning different concrete types from `Sensor::wait_for_reading()` in the far future)?
- Open for bikeshedding:
  - We should agree on names for what is currently called:
    - Sensor category: could be renamed to sensor class to match SAUL naming, but "class" is a reserved keyword in many languages.
    - Scaling (value): could be "scaling", "prescaler", something else?
  - Some of the documentation is split between the top module, the `Value` and the `Accuracy` types, in an attempt to keep it close to its associated implementation. Should we instead group (some of) it in the top-level module?
  - The set of items we re-export at root level could use some refinement (or err on the side of caution and not re-export much for now).

## Future work

The following steps would be:

- Adding support for external interrupts triggered by sensor devices: *physical event* interrupts and *data available* interrupts are to be treated differently even though they use the same pins.
- Adding the first sensor drivers provided by Ariel OS (the current API does also allow to provide and use third-party sensor drivers developed for this sensor API): drivers for push buttons, some temperature sensors and 3-axis accelerometers exist in a preliminary state but are not part of this PR.
- Introducing codegen for automating the instantiating and configuration of sensor drivers based on static, structured configuration files.

## How to review this PR

I think this PR is better reviewed by generating the rustdoc docs locally and starting from there, before glancing at the implementation.

## TODO

- [x] The `Category`, `Label`, and `MeasurementUnit` enums should be expanded with more variants, even though these can (and will) be expanded later.

## Change checklist

### DIFF
--- a/src/ariel-os-macros/Cargo.toml
+++ b/src/ariel-os-macros/Cargo.toml
@@ -30,4 +30,13 @@ trybuild = "1.0.89"
 proc-macro = true
 
 [features]
+# These features could be codegened
+max-sample-min-count-2 = []
+max-sample-min-count-3 = []
+max-sample-min-count-4 = []
+max-sample-min-count-6 = []
+max-sample-min-count-7 = []
+max-sample-min-count-9 = []
+max-sample-min-count-12 = []
+
 _test = []
--- a/src/ariel-os-sensors/Cargo.toml
+++ b/src/ariel-os-sensors/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "ariel-os-sensors"
+version = "0.1.0"
+license.workspace = true
+edition.workspace = true
+rust-version = "1.85"
+repository.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os-macros = { workspace = true }
+defmt = { workspace = true, optional = true }
+embassy-sync = { workspace = true }
+linkme = { workspace = true }
+pin-project = { workspace = true }
+
+[features]
+defmt = ["dep:defmt"]
+
+# These features could be codegened
+max-sample-min-count-2 = ["ariel-os-macros/max-sample-min-count-2"]
+max-sample-min-count-3 = ["ariel-os-macros/max-sample-min-count-3"]
+max-sample-min-count-4 = ["ariel-os-macros/max-sample-min-count-4"]
+max-sample-min-count-6 = ["ariel-os-macros/max-sample-min-count-6"]
+max-sample-min-count-7 = ["ariel-os-macros/max-sample-min-count-7"]
+max-sample-min-count-9 = ["ariel-os-macros/max-sample-min-count-9"]
+max-sample-min-count-12 = ["ariel-os-macros/max-sample-min-count-12"]
--- a/src/ariel-os-sensors/src/sensor.rs
+++ b/src/ariel-os-sensors/src/sensor.rs
@@ -1,0 +1,326 @@
+//! Provides a [`Sensor`] trait abstracting over implementation details of a sensor driver.
+use core::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, channel::ReceiveFuture};
+
+use crate::{Category, Label, MeasurementUnit};
+
+pub use crate::{
+    Reading,
+    sample::{Accuracy, Sample},
+};
+
+ariel_os_macros::define_count_adjusted_sensor_enums!();
+
+/// This trait must be implemented by sensor drivers.
+///
+/// See [the module level documentation](crate) for more.
+pub trait Sensor: Send + Sync {
+    /// Triggers a measurement.
+    /// Clears the previous reading.
+    ///
+    /// To obtain readings from every sensor drivers this method can be called in a loop over all
+    /// sensors returned by [`Registry::sensors()`](crate::registry::Registry::sensors), before
+    /// obtaining the readings with [`Self::wait_for_reading()`] in a second loop, so that the
+    /// measurements happen concurrently.
+    ///
+    /// # For implementors
+    ///
+    /// This method should return quickly.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TriggerMeasurementError::NonEnabled`] if the sensor driver is not enabled.
+    fn trigger_measurement(&self) -> Result<(), TriggerMeasurementError>;
+
+    /// Waits for the reading and returns it asynchronously.
+    /// Depending on the sensor device and the sensor driver, this may use a sensor interrupt or
+    /// data polling.
+    /// Interpretation of the reading requires data from [`Sensor::reading_channels()`] as well.
+    /// See [the module level documentation](crate) for more.
+    ///
+    /// # Note
+    ///
+    /// It is necessary to trigger a measurement by calling [`Sensor::trigger_measurement()`]
+    /// beforehand, even if the sensor device carries out periodic measurements on its own.
+    ///
+    /// # Errors
+    ///
+    /// - Quickly returns [`ReadingError::NonEnabled`] if the sensor driver is not enabled.
+    /// - Quickly returns [`ReadingError::NotMeasuring`] if no measurement has been triggered
+    ///   beforehand using [`Sensor::trigger_measurement()`].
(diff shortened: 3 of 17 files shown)


bergzand:
Some initial comments based on the docs

Bikeshedding:
- I would prefer `Sample` instead of `Value`. 
- To prevent confusion I would call the different quantities returned by a sensor different `channels` and not `axes`, so `Sensor::reading_channels()`. Mostly to remove ambiguity when dealing with accelero + gyro sensors.
- I think category is good enough, but I agree that there might be a better name.


Does it make sense to include the sensor range in a same way the accuracy is provided? I think this can be valuable with gas sensors and accelerometers, to know when the measurement is out of (or close to) the max range.

 
> Are floats completely out of the question, even as helpers? Integers should stay at the core of the API, but maybe we could have additional (feature-gated?) helpers on `Value`+`ReadingAxis`+`Accuracy` which would return floats taking into account everything required to make sense of the reading (in particular the scaling value)?

I think a set of (gated) helpers to convert to floats is useful, even if it just to make printing measurements easier.

On the API side, the only question I have so far is whether it is easy enough to zip a set of `Value`s and their `ReadingAxis` metadata. But there's probably some Rust thingy for that ([zip](https://doc.rust-lang.org/core/iter/struct.Zip.html) probably :)

ROMemories:
> * I would prefer `Sample` instead of `Value`.
> 
> * To prevent confusion I would call the different quantities returned by a sensor different `channels` and not `axes`, so `Sensor::reading_channels()`. Mostly to remove ambiguity when dealing with accelero + gyro sensors.

Renamed `Value` to `Sample` and axes to channels.

src/ariel-os-sensors/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "ariel-os-sensors"
+version = "0.1.0"
+license.workspace = true
+edition.workspace = true
+rust-version = "1.85"
```
ROMemories, line 6: Both the crate version number and its MSRV are decoupled from the other crates, as it is expected that every sensor driver crate depend on `ariel-os-sensors`.

src/ariel-os-macros/Cargo.toml
```diff
@@ -30,4 +30,13 @@ trybuild = "1.0.89"
 proc-macro = true
 
 [features]
+# These features could be codegened
+max-sample-min-count-2 = []
```
kaspar030, line 34: why no 5,8,10,11?

ROMemories: We can just add them later if need be, they didn't seem like a priority.

nponsard:
I've been trying to understand how GPS would work with the sensor abstraction.
If we want to give access to GPS data using this, some new units are needed:
- Meters per second (m/s) for the velocity
- Meters (m) for the altitude
- Degrees (deg) for latitude, longitude and heading (converting to radians may be confusing)

A `Gps` category would need to be added.

GPS also communicates date and time, should it be exposed via sensor readings ? If yes, do we represent time like its sent in NMEA packets (day, month, year, hour, minute, seconds) or as seconds since an Epoch ?

chrysn:
> Meters per second (m/s) for the velocity
>  Meters (m) for the altitude

Those should definitely be in -- actually I think that the full list of (primary and recommended) SenML units from https://www.iana.org/assignments/senml/senml.xhtml should be there.

> Degrees (deg) for latitude, longitude and heading (converting to radians may be confusing)

SenML actually uses different units for latitude and longitude, under the same rationale as one would treat J and Nm differently (in a sense they are both technically orthogonal). For heading, "degree (angle)" would make sense.

For time, I think that scalar time (number of seconds) is the right API interface. Note that the annotations afforded by SenML units are not quite sufficient there (it'd just be "s", possibly with a nonzero decimal exponent) but needs a time scale annotation too. There is now [RFC9581](https://www.rfc-editor.org/rfc/rfc9581.html) that gives the right terminology -- it does *not* provide an entry for GPS but states:

> Note that there should be no timescales "GPS" or "NTP" [[RFC5905](https://www.rfc-editor.org/rfc/rfc9581.html#RFC5905)] -- instead, the time should be converted to TAI or UTC using a single addition or subtraction.)
>     
>        t    = t    - 2208988800
>         utc    ntp
>     
>        t    = t    + 315964819
>         tai    gps
>
> [Figure 2](https://www.rfc-editor.org/rfc/rfc9581.html#figure-2): [Converting Common Offset Timescales](https://www.rfc-editor.org/rfc/rfc9581.html#name-converting-common-offset-ti)

which is a rather simple addition to do (compared to processing ASCII NMEA records, but also compared to converting SIRF's week-and-centiseconds).

src/riot-rs-sensors/src/value.rs
```diff
@@ -0,0 +1,130 @@
...
+/// # Accuracy
+///
+/// The accuracy can be obtained with [`Self::accuracy()`].
+// NOTE(derive): we do not implement `Eq` or `PartialOrd` on purpose: `Eq` would prevent us from
+// possibly adding floats in the future and `PartialOrd` does not make sense because interpreting
+// the value requires the `ReadingAxis` associated with this `Value`.
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Value {
+    value: i32,
```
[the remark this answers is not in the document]
kaspar030: Doesn't make much sense IMO, we'd just waste a byte of precision.

src/riot-rs-sensors/src/measurement_unit.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Please open an issue to discuss it.
+// Built upon https://doc.riot-os.org/phydat_8h_source.html
+// and https://bthome.io/format/#sensor-data
+// and https://www.iana.org/assignments/senml/senml.xhtml
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum MeasurementUnit {
+    /// [Acceleration *g*](https://en.wikipedia.org/wiki/G-force#Unit_and_measurement).
+    AccelG,
```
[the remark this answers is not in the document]
kaspar030: all accel sensors I've worked with returned `g`s, let's rather provide conversion  methods ourselves (later).

src/ariel-os-sensors/src/sensor.rs
```diff
@@ -0,0 +1,326 @@
...
+    fn part_number(&self) -> Option<&'static str>;
+
+    /// Returns the sensor driver version number.
+    #[must_use]
+    fn version(&self) -> u8;
+}
+
+/// Future returned by [`Sensor::wait_for_reading()`].
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+#[pin_project::pin_project(project = ReadingWaiterProj)]
```
kaspar030, line 107: I admit I don't understand why this needs pin projection. Maybe add an explanation?

ROMemories: I'll add an explanation as a follow-up.

ROMemories:
Rebased on main and fixed the lockfile, which had duplicated packages.

PR #481, opened 2024-10-22
fix(threads): xtensa: enable saving/restoring float regs

kaspar030:
# Description

... seems like something we want. `esp-wifi` also enables this.

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2339,15 +2339,15 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 name = "hello-world"
 version = "0.1.0"
 dependencies = [
+ "embassy-executor",
  "riot-rs",
  "riot-rs-boards",
 ]
 
 [[package]]
-name = "hello-world-async"
+name = "hello-world-threading"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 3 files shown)


elenaf9:
Does this impact context switching performance? Because now on each context switch the 15 float registers need to be saved/ restored as well.

kaspar030:
> Does this impact context switching performance? Because now on each context switch the 15 float registers need to be saved/ restored as well.

It probably does. But if we don't save them, they're potentially garbled after a context switch ...

elenaf9:
> But if we don't save them, they're potentially garbled after a context switch

If the feature isn't enabled, the registers aren't "there" from a user perspective, and as soon as they are in the Trapframe we save them because we always switch the whole trapframe.
I was just wondering if we should feature-gate it in case floats aren't needed. And if esp-wifi still enables them than that's not a problem.

kaspar030:
> So IMO users that need floating point operations can just enable it and it will work with our implementation (also thanks to [0e84775](https://github.com/ariel-os/ariel-os/commit/0e847753b0324274046371dfd7d69b8a98c99365)). No need for us to enable it?

Agreed.

kaspar030:
For anyone watching: #665 now enables this regardless of `esp-wifi` being used or not.

PR #485, opened 2024-10-24
fix(examples/coap): Various fixes

chrysn:
# Description

These are simple fixes taken out of https://github.com/future-proof-iot/RIOT-rs/pull/348 – that PR is going back a few steps because I took a wrong turn, and this now I'm consolidating things into manageable changes.

This does introduce one FIXME, but the need for getting that piece of information was already there and just hidden by the broken code.

## Change checklist

### DIFF
--- a/examples/coap/Cargo.toml
+++ b/examples/coap/Cargo.toml
@@ -45,5 +45,5 @@ scroll-ring = "0.1.1"
 [features]
 default = ["proto-ipv4"] # shame
 # actually embedded-nal features, we have to match them here while developing udp_nal in here
-proto-ipv4 = []
-proto-ipv6 = []
+proto-ipv4 = ["embassy-net/proto-ipv4"]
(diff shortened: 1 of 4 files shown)


kaspar030:
Depends on #469.

chrysn:
I'm on that in parallel, but that's running through tests time and again. I did do a full local build test of the CoAP example, and exceeding what CI did also with and without IPv6 enabled.

Code that is not in CI has a tendency to rot, let's fix this in parallel. Even with no CI running, this PR makes the code better, and reduces the warnings I'm sifting through while looking for hints at what is keeping #469 back.

kaspar030:
> I'm on that in parallel, but that's running through tests time and again. I did do a full local build test of the CoAP example, and exceeding what CI did also with and without IPv6 enabled.

ok!

kaspar030:
LGTM.

PR #489, opened 2024-10-25
chore(deps): update deps, add `embassy-executor-macros` patch

kaspar030:
# Description

Fixes "RIOT-rs doesn't build for me anymore after running cargo update.
The reason is that it updates embassy-executor to v0.6.1, and that causes our patch to not be applied anymore because the version in the patch is v0.6.0."

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ portable-atomic = { version = "1.8.0", default-features = false, features = [
 ] }
 
 embassy-embedded-hal = { version = "0.2.0", default-features = false }
-embassy-executor = { version = "0.6", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -59,7 +59,7 @@ portable-atomic = { version = "1.8.0", default-features = false, features = [
 ] }
 
 embassy-embedded-hal = { version = "0.2.0", default-features = false }
-embassy-executor = { version = "0.6", default-features = false }
+embassy-executor = { version = "0.6.0", default-features = false }
```
kaspar030, line 62: I know this doesn't pin it. :)

elenaf9 (approved):

PR #491, opened 2024-10-26
fix(CI): Keep Xtensa clang environment variables off the general system

chrysn:
# Description

The Xtensa toolchain sets a general `LIBCLANG_PATH`, which is part of what causes the clang troubles for libOSCORE (to be fixed in https://github.com/future-proof-iot/RIOT-rs/pull/443).

This approach <del>unsets it from the GitHub action environment</del>, and </del>will later set it specifically for the ESP builds on demand</del>.

[edit: added] Now we can't *unset* it from the GitHub actions because that just doesn't work, but we prepare a sourceable .profile style file that is then sourced by all relevant run scripts.

## Open Questions

* [x] Will it build?\
  [edit, added:] … and it didn't even need that environment variable anywhere else 🤦

* [x] Were alternatives considered?
  Yes: The alternative is to fork the Xtensa script and make it optional to export the variable.
  Or fork GitHub runners and fix https://github.com/actions/runner/issues/1126 so that we can just unset what the action sets.
  But those would all be way more effort than maintaining this workaround.

## PR dependencies

Depended on https://github.com/future-proof-iot/RIOT-rs/pull/492

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,6 +64,19 @@ jobs:
         with:
           buildtargets: esp32s3
 
+      - name: Keep general build environment clean of Xtensa peculiarities
+        # The Xtensa setup unconditionally sets LIBCLANG_PATH in GITHUB_ENV --
+        # we can't undo that. (It also sets it in ~/exports, but that is just
+        # what it uses internally to get that data around into the action
+        # script)
+        #
+        # As the shell apparently does not source any files, and until
+        # https://github.com/actions/runner/issues/1126 is fixed by GitHub, the
+        # best remedy is to stow the workaround and source it wherever we can.
+        run: |
+          set -x
+          echo 'unset LIBCLANG_PATH' >> ~/fix-environment-after-xtensa-setup
+
       - name: rust cache
         if: steps.result-cache.outputs.cache-hit != 'true'
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


chrysn:
Turned out it did build, so the LIBCLANG_PATH (which is now stored away in _XTENSA_LIBCLANG_PATH in case anything we do needs to pull it back in) was exported completely needlessly and just broke stuff.

I've rebased this PR onto #492 – originally they were distinct approaches, but with how that one turned out to just update the system, and how this one turned out to just work, both contribute to having a consistent story towards resolving the CoAP build issues (which eventually #443 should fix).

chrysn:
Set back to draft when I found that as tracked in https://github.com/actions/runner/issues/1126, GitHub action environment variables can simply not be unset in general.

chrysn:
This is now ready for review again -- it finally does something, and conveniently, the (back then unfounded) observation that the environment variable is not really needed held.

kaspar030 (commented):
Damn ugly but pragmatic. Maybe make it generic, call it 'unset-variables.sh' or use an even more generic name for "sourced always first" script that we can just do "ecgi TWEAK >> name" into?

kaspar030:
Lgtm

chrysn:
If we don't go with a squash, #443 would work as-is, and I don't think it hurts to keep this one, but if you prefer squashing, I can do that too.

PR #496, opened 2024-10-29
docs(network)!: document items from the `network` module

ROMemories:
# Description

BREAKING CHANGE: stop re-exporting `NetworkStack` from `riot_rs`'s root.

## Change checklist

### DIFF
--- a/src/riot-rs-embassy/src/network.rs
+++ b/src/riot-rs-embassy/src/network.rs
@@ -1,4 +1,10 @@
-//! To provide a custom network configuration, use the `riot_rs::config` attribute macro.
+//! Provides network access.
+//!
+//! The network link to use is selected through Cargo features.
+//! Additionally, the [`riot_rs::config`](riot_rs_macros::config) attribute macro allows to provide
+//! custom network configuration.
+
+#![deny(missing_docs)]
 
 use embassy_net::{Runner, Stack};
 use embassy_sync::once_lock::OnceLock;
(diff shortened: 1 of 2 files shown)


src/riot-rs-embassy/src/network.rs
```diff
@@ -1,17 +1,30 @@
-//! To provide a custom network configuration, use the `riot_rs::config` attribute macro.
+//! Provides network access.
+//!
+//! The network link to use is selected through Cargo features.
```
chrysn, line 3: Should this also refer to the laze selection? Can a user of this crate expect enabling a backend through features to work?

src/riot-rs-embassy/src/network.rs
```diff
@@ -1,17 +1,30 @@
...
+//! custom network configuration.
+
+#![deny(missing_docs)]
 
 use embassy_net::{Runner, Stack};
 use embassy_sync::once_lock::OnceLock;
 
 use crate::{sendcell::SendCell, NetworkDevice};
 
+/// [MTU](https://en.wikipedia.org/wiki/Maximum_transmission_unit) supported by Ethernet, in bytes.
```
chrysn, line 14: Is this accurate? https://en.wikipedia.org/wiki/Maximum_transmission_unit#Applicability says it's 1500 or 1518, depending on whether you count overhead.

src/riot-rs-embassy/src/network.rs
chrysn: I was more thinking about the underlying laze mechanisms than the concrete features -- but as we don't have text on those in the book, I don't think there's something concrete to reference so far.

PR #497, opened 2024-10-30
chore(CI): Line break long command line invocations

chrysn:
# Description

This should help keep merge conflicts resulting from different PRs adding modules at a minimum.

## Issues/PRs references

Builds on https://github.com/future-proof-iot/RIOT-rs/pull/493 (which is already being merged)

## Open Questions

* [ ] Will it build? (There's a long way from the YAML to a `sh` parser, let's see whether my guesses about which escaping applies where are correct)

## Change checklist

- [x] I have performed a self-review of my own code. (`git show --color-words='[^,\\|"]'`  looks very boring)
- [x] I have made corresponding changes to the documentation. (N/A)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -168,14 +168,14 @@ jobs:
                 -p riot-rs-embassy \
                 -p riot-rs-embassy-common \
                 -p riot-rs-identity \
+                -p riot-rs-macros \
                 -p riot-rs-runqueue \
                 -p riot-rs-threads \
-                -p riot-rs-macros \
 
             cargo test \
+                -p coapcore \
                 -p rbi \
                 -p ringbuffer \
-                -p coapcore \
 
       # We need to set `RUSTDOCFLAGS` as well in the following jobs, because it
       # is used for doc tests.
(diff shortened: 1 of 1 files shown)


chrysn:
Works well in here; now it's just a matter of `V}:!sort`.

Most are sorted anyway. There's no sharp criterion of what is sortable. I'll go visually, and trust that if there is something small enough that it didn't show up on my glasses-off view, the next author who does something in there will do it.

chrysn:
With the critical CI steps passed, I think this is now ready for review. Fixups were kept in place because they're easier to review; not sure yet how I'd squash this (I'd like to at least keep the sorting visible).

Note that as for "is this correct", it'll help that if any backslashes were missing (not even @ROMemories can be *that* good at reviewing to spot single missed ones?!), what comes after is generally not a valid command, and thus causes an error even if the original command that now has fewer argument went OK.

chrysn:
Added it for the cargo test jobs, and squashed into wrapping and sorting.

kaspar030:
If we have more multiline than regular commands, maybe it is more readable to use `>`  instead of `|` for the [yaml multiline mode](https://yaml-multiline.info/)?
Multiple commands can still be chained with `&&` or by inserting two empty lines.

kaspar030:
(Ok I'm late, I'll try next week)

PR #506, opened 2024-11-06
deps: bump embassy fork to 2024-11-06

kaspar030:
# Description

This bumps embassy-net & smoltcp to ip address types from `core`.
Needed bumping of `coap-embedded-nal`.

Also makes `embassy-nrf` a new, chiptool generated consolidated `nrf-pac`, which also required some changes.

## Change checklist

- [ ] I have cleaned up my commit history and squashed fixup commits.
- [ ] I have followed the [Coding Conventions](https://future-proof-iot.github.io/RIOT-rs/dev/docs/book/coding-conventions.html).
- [ ] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

PR #523, opened 2024-11-13
feat(nrf): use NFC pins as normal GPIOs

ROMemories:
# Description

On nRF MCUs, some pins are by default reserved for NFC. This PR allows to use them as normal GPIOs as we don't currently support NFC. This was already done for the nRF52833 as part of #463.

This fixes the following probe-rs warning:

```
WARN  You have requested to use P0.09 and P0.10 pins for NFC, by not enabling the Cargo feature `nfc-pins-as-gpio`.
However, UICR is already programmed to some other setting, and can't be changed without erasing it.
To fix this, erase UICR manually, for example using `probe-rs erase` or `nrfjprog --eraseuicr`.
```

## Issues/PRs references

Depends on #506.

## Change checklist

### DIFF
--- a/src/riot-rs-nrf/Cargo.toml
+++ b/src/riot-rs-nrf/Cargo.toml
@@ -30,17 +30,23 @@ riot-rs-embassy-common = { workspace = true }
 riot-rs-random = { workspace = true, optional = true }
 
 [target.'cfg(context = "nrf52832")'.dependencies]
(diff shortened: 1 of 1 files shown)


kaspar030 (approved):

kaspar030:
This'll conflict with #506. :/

PR #533, opened 2024-11-15
fix(startup): move out peripherals required for USB drivers earlier

ROMemories:
# Description

Previously, it would have been possible for user tasks to take peripherals out of `OptionalPeripherals` before the required peripherals had been extracted for USB drivers.
This now takes the peripherals required for USB before starting the user tasks, making this impossible for tasks.

## Hardware testing

I've successfully tested the `udp-echo` example on the nRF52840-DK, the RPi Pico and the ST NUCLEO WB55RG, and the `usb-serial` on the ST NUCLEO H755ZI-Q (because it doesn't currently support Ethernet over USB). These two ST NUCLEO boards use different USB modules of ours.

## Issues/PRs references

Extracted from #337. **This PR is only concerned with USB peripherals.** I'll update #337 later to do Wi-Fi peripherals after this one.

## Change checklist

### DIFF
--- a/src/riot-rs-arch/src/dummy/usb.rs
+++ b/src/riot-rs-arch/src/dummy/usb.rs
@@ -33,7 +33,15 @@ impl<'a> Driver<'a> for UsbDriver {
     }
 }
 
-pub fn driver(_peripherals: &mut crate::OptionalPeripherals) -> UsbDriver {
+pub struct Peripherals {}
+
+impl Peripherals {
+    pub fn new(_peripherals: &mut crate::OptionalPeripherals) -> Self {
+        unimplemented!();
+    }
+}
+
+pub fn driver(_peripherals: Peripherals) -> UsbDriver {
     unimplemented!();
 }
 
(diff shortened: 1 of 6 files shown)


src/riot-rs-arch/src/dummy/usb.rs
```diff
@@ -33,7 +33,15 @@ impl<'a> Driver<'a> for UsbDriver {
     }
 }
 
-pub fn driver(_peripherals: &mut crate::OptionalPeripherals) -> UsbDriver {
+pub struct Peripherals {}
```
kaspar030, line 36: why no `define_peripherals!()`?

kaspar030: OK I see. that's unfortunate but the way to go I guess.

kaspar030: Can we move those to `riot-rs-arch`?

kaspar030 (approved):

Issue #410, opened 2024-09-11
tracking: cleanup examples


kaspar030:
### Description

Initial targeted list of examples in this [pad](https://pad.riot-os.org/aPbbp77KTEKZb--l6GwdUA?both).

For each example,

- [ ] minimize
- [ ] check name
  - [ ] no embassy
  - [ ] no async, is default
  - [ ] "-threading"
- [ ] polish
- [ ] use `ariel_os::reexports` when relevant
- [ ] use the `tcp` and `udp` Cargo features introduced by #525, instead of depending on `embassy-net`
- [ ] Be consistent in the usage of laze modules vs ariel_os features.
- [ ] has entry in `examples/README.md`

Please copy this list into cleanup PRs!

#### Missing examples

- [x] CoAP Minimal server -- static hello world -> coap-server TODO
- [x] CoAP Minimal client -- get pre-configured resource -> coap-client TODO
- [ ] i2c demonstrator
- [ ] spi demonstrator

### Tracking

From [pad](https://pad.riot-os.org/aPbbp77KTEKZb--l6GwdUA?both): 

- [ ] benchmark
- [x] blinky
- [ ] Secure LED blinking over CoAP + OSCORE -> coap-blinky TODO
- [ ] coap-client
- [ ] coap-server
- [x] core-sizes (move to tests/, turn into regression test?) removed by #563
- [x] device-metadata #608 
- [x] gpio #611 
- [x] ~hello-world-async~ -> hello-world (maybe becoming blinky-async?( K: no, why?)) #478
- [x] ~hello-world~ -> hello-world-threading #478 
- [x] http-client #627
- [x] ~embassy-http-server~ -> http-server #570
- [x] log #547 
- [ ] multicore-affinities
- [x] random #562
- [x] storage #676
- [x] ~embassy-net-tcp~ -> tcp-echo #522
- [x] ~embassy~ -> thread-async-interop #505 
- [x] threading #605 
- [x] threading-channel #607
- [x] threading-event #606
- [ ] threading-multicore
- [x] ~embassy-net-udp~ -> udp-echo #521
- [x] usb-keyboard #538
- [x] usb-serial #573

elenaf9:
- Be consistent in the usage of laze modules vs `riot_rs` features.

emmanuelsearch:
Consider adding examples on
- basics: eg timers, sensor abstraction etc. inspired by RIOT-course https://riot-os.github.io/riot-course/slides/03-riot-basics/#11
[...]

PR #562, opened 2024-11-22
docs(random-example): clean up the `random` example

ROMemories:
# Description

As part of https://github.com/ariel-os/ariel-os/issues/410, this PR:
- Avoids using threading as it is not necessary for this example.
- Prints 10 random values instead of just one, to demonstrate the randomness.
- Adds the example to the list in `examples/README.md`.

## Change checklist

### DIFF
--- a/examples/random/src/main.rs
+++ b/examples/random/src/main.rs
@@ -1,15 +1,17 @@
 #![no_main]
 #![no_std]
+#![feature(impl_trait_in_assoc_type)]
 #![feature(used_with_arg)]
 
 use ariel_os::debug::log::*;
+use rand::Rng as _;
 
-#[ariel_os::thread(autostart)]
-fn main() {
-    use rand::Rng as _;
+#[ariel_os::task(autostart)]
+async fn main() {
     let mut rng = ariel_os::random::fast_rng();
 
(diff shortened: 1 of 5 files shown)


examples/random/src/main.rs
```diff
@@ -2,14 +2,15 @@
...
 #![feature(used_with_arg)]
 
-use ariel_os::debug::log::*;
+use ariel_os::{asynch::Spawner, debug::log::*};
+use rand::Rng as _;
 
-#[ariel_os::thread(autostart)]
-fn main() {
-    use rand::Rng as _;
+#[ariel_os::spawner(autostart)]
```
kaspar030, line 8: I think we should use a regular task here.

kaspar030: I'm thinking educational, examples should be as uniform and consistent as possible. If the random example differs from hello-world only in the random use, perfect. if it suddenly uses the spawner mechanism, it is "too different".
I'm OK with the extra feature, as that is boilerplate we'll explain away once, centrally. :)

chrysn: Looks to me like that's done.

chrysn (approved):

PR #568, opened 2024-11-25
fix(coap): Update dependencies

chrysn:
# Description

This updates libOSCORE (just to a released version from a practically identical git snapshot) and Lakers (major update, including a compatibility fix).

## Caveats

This breaks compatibility with aiocoap when credentials are sent by value, as is the case in some examples. This is part of the fallout of a previous Lakers bug [303], and due to be fixed in an upcoming aiocoap update, which I expect to be available before this PR is merged.

## Change checklist


chrysn:
aiocoap 0.4.12 is released that fixes the bug in earlier lakers, and has all the dependencies right. In parallel, this PR has been updated to use the latest aiocoap and related tools.

If the examples don't work for you, consider `rm -rf ~/.cache/pipx` -- might just not have updated yet. Conversely, any fresh use of the examples in main will now fail.

I'm still facing some libOSCORE build failure (yea, again :'-( ), but other while tests run on that, I think the rest is good to review.

Some notes from self-review:
* I'm glad we're getting the checked-out COSE key out of the example; aiocoap can now come up with a random one on its own.
* Some small bug fixes as bycatch are hopefully OK for that kind of PR (the [fix(coap): don't crash on crafted OSCORE request](https://github.com/ariel-os/ariel-os/pull/568/commits/082ef83f1219b87f7f0a5c9656601b05dae6b756) commit).
* Updating fauxhoc feels a bit weird, because that file is a bit of a relic of before aiocoap supported EDHOC. I'm doing the minimal fixes here; to get rid of that file completely I'll want to go through it with a fine comb to see if we're losing anything in terms of coverable scenarios, so that'd be another PR.

chrysn:
Fully ready now, just missing review: Using the released crate caused trouble, and while that's something that needs to be fixed, it can't be fixed in this PR.

kaspar030 (approved):

chrysn:
Thanks, squashed and applied `git filter-branch -f --tree-filter 'cargo check; git add Cargo.lock' f6ebe415e10fa748ae43ffd8bc5f827245ba2425..` to keep the Cargo.lock file in sync with the changes in the individual commits.

PR #583, opened 2024-11-28
[test PR please ignore] maybe oscore fails because it only works from git

chrysn:
This tests possible solutions to the build issues of https://github.com/ariel-os/ariel-os/pull/568, without disturbing the review process there.

See https://github.com/chrysn-pull-requests/RIOT-rs/pull/2 for an attempt to do that without making noise in here.


chrysn:
Closing as what this PR should have shown has been shown: that it's using the published crate rather than the git reference that broke the builds.

PR #588, opened 2024-12-02
feat(coapcore): refactor authentication sources, add ACE-OSCORE profile

chrysn:
This contains changes to coapcore that enable running in an ACE context, eg. using the ACE OSCORE profile.

This started small (eg.generalizing crypto factory from `fn()` to `F` where `F: Fn()`: This is not needed in Ariel's concrete use case, but helpful in applications with less global state), but eventually now contains a large refactoring, which was due anyway to get rid of key material that was originally either hard-coded, or passed in without regard for future flexibility (like, starting a CoAP server does not *always* entail having a single EDHOC identity, could be 0 or 2).

## General status

This <del>has most of </del>i<del>t</del>s<del> migrations</del> done<del>, but is missing a few reapings of the changed responsibilities' benefits yet, and also examples</del>.

[edit, added:] There's always more that can be done, but this is a relatively self-contained step.

## Change checklist

  [edit, clarifying why this is ticked]: Item documentation should be fine, example documentation does not need to change, and CoAP book chapter is updated out-of-band once more changes are in.


chrysn:
This is now in a state that I'd call feature-complete for this PR -- it moves all parts that are supposed to be configured out of the coapcore crate (to ariel-os-coap, where now it is waiting for a follow-up around provisioning).

I'm starting a self-review run -- things I expect to change at this stage are documentation enhancements, maybe moving around or renaming types/traits, altering what is sealed, adding some testing code.

It is rebased onto main to resolve conflicts from recently addressed lints.

chrysn:
From live review during weekly break:
* randcore might have a workspace crate.
* yes there are hardcoded credentials still; they are moving toward the application.

chrysn:
Pushed the merge of main in there. Turned out that even on current main, rand_core is *not* a workspace crate, so that should be fine.

chrysn:
This should now have all Clippy and other linter complaints addressed. Many fixes are squashed in to their origins; some are kept as "fix(coap): address clippy issues" or similar because threading them in would require going through different commits (like when something was introduced in one commit, extended in another, and then needed renaming).

PR #596, opened 2024-12-05
feat(build): pass `[patch.crates-io]` via extra cargo toml

### LINKED ISSUE #524: make crates.io patches available to out-of-tree builds
Currently we specify the patches in the riot-rs workspace Cargo.toml, but there they won't get picked up by out of tree builds.

Those patch lines cannot IIUC supplied via environment (`CARGO_PATCH_CRATES_IO_FOO_PATH=...` doesn't work).

We can supply those patches via Cargo command line, but then they're not carried over to sub-cargo calls in e.g., `build.rs`, which affects everything that is using info from cargo_metadata from within build.rs.

We can supply those oatches in an extra .toml that we pass via `--config <path>`, with similar issues.

IMO a `CARGO_CONFIG_INCLUDE` or similar would be ideal, but that's been denied so far: https://github.com/rust-lang/cargo/pull/7894, https://github.com/rust-lang/cargo/issues/6728.
I've been sifting through most of the [cargo configuration issues / ideas](https://github.com/rust-lang/cargo/issues/9769), I don't find a way even considering unstable options other than:

1. asking users to copy&paste our patches into the application's Cargo.toml
1. asking users to copy&paste our patches into the application's .cargo/config.toml
3. doing the latter with laze (potentially overwriting what the user already had there

This seems clumsy to me. Also, if all patches are copied, the user get warnings like "warning: the esp-wifi patch was not used", when the application is only built for other architectures.

Ideas?

kaspar030:
# Description

This moves the crates.io patches into another file that is then be parsed to Cargo via `--config`. This allows re-using the patches in out-of-tree builds.
For local tooling, the repo's `.cargo/config.toml` is modified to include the file via the unfortunately still unstable [config-include](https://github.com/rust-lang/cargo/issues/7723).

## Issues/PRs references

Fixes #524.

## Open Questions

- [x] How to handle CI `cargo check/clippy/...`?
- [x] How to handle local `cargo check/clippy/...`?
- [x] How to handle tooling like rust analyzer?

## Change checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,8 @@
-# This is needed for cbindgen to work in no_std cross compiles.
+include = "../patch.crates-io.toml"
+
 [unstable]
+# This is needed for cbindgen to work in no_std cross compiles.
 features = ['all']
+
+# This is needed so the "include" statement above works.
+config-include = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,36 +142,6 @@ opt-level = "s"
 [profile.release.package.esp-wifi]
 opt-level = 3
 
-[patch.crates-io]
-# ariel-os embassy fork
-embassy-executor = { git = "https://github.com/ariel-os/embassy", branch = "for-ariel-os-2024-11-28" }
-embassy-executor-macros = { git = "https://github.com/ariel-os/embassy", branch = "for-ariel-os-2024-11-28" }
-embassy-hal-internal = { git = "https://github.com/ariel-os/embassy", branch = "for-ariel-os-2024-11-28" }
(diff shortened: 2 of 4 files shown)


kaspar030:
I'm wondering if we should put all global build settings in such a file.

At least this could make sense:

```
[profile.release.package.esp-wifi]                                                                                                                           
opt-level = 3
```

Also `debug = true` when defmt is enabled (or, just always).
And `opt-level = "s"`.

Alternatively we could set those via cli. these settings are not really needed for the local test builds etc., and don't look too ugly when passed as arguments to Cargo.

kaspar030:
> I'm wondering if we should put all global build settings in such a file.

... but not in this PR :)

PR #601, opened 2024-12-09
refactor(delegate)!: enforce more safety requirements

ROMemories:
# Description

- Make sure that `with()` is only called once, panic otherwise. This costs one extra byte to store the required flag.
- Mark the `lend()` method as unsafe.
- Update and slightly polish the docs as a result of these changes.

## Change checklist

### DIFF
--- a/src/ariel-os-embassy/src/delegate.rs
+++ b/src/ariel-os-embassy/src/delegate.rs
@@ -1,11 +1,12 @@
-//! Delegate or lend an object to another task
+//! Delegate or lend an object to another task.
 
 use embassy_executor::Spawner;
 use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, signal::Signal};
+use portable_atomic::{AtomicBool, Ordering};
 
 use crate::sendcell::SendCell;
 
-/// [`Delegate`] or lend an object to another task.
+/// [`Delegate`]s or lends an object to another task.
 ///
 /// This struct can be used to lend a `&mut T` to another task on the same executor.
 /// The other task can then call a closure on it. After that, the `&mut T` is returned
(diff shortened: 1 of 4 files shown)


src/ariel-os-embassy/src/delegate.rs
```diff
@@ -30,43 +34,54 @@ use crate::sendcell::SendCell;
...
+    ///
+    /// This blocks until [`Delegate::lend()`] is called on the instance.
     ///
-    /// This blocks until another task called [`lend(something)`](Delegate::lend).
+    /// # Panics
+    ///
+    /// Panics if called multiple times on the same [`Delegate`] instance.
     pub async fn with<U>(&self, func: impl FnOnce(&mut T) -> U) -> U {
+        // Enforce that the value be only populated once, panic otherwise.
+        assert!(!self.was_exercised.load(Ordering::Acquire));
```
kaspar030, line 82: I think we got a race here: a) assserts, b) asserts, both succeed -> bad.

Not sure if it can be triggered with only a single executor, but IMO we should still just do `assert!(!self.was_exercised.swap(true, Ordering::Acquire));` here. (not sure about `Acquire`).

src/ariel-os-embassy/src/delegate.rs
```diff
@@ -30,43 +34,53 @@ use crate::sendcell::SendCell;
...
 /// ```
-///
-/// TODO: this is a PoC implementation.
-/// - takes 24b for each delegate (on arm), which seems too much.
-/// - doesn't protect at all against calling [`lend()`](Delegate::lend) or
-///   [`with()`](Delegate::with) multiple times
-///   each, breaking safety assumptions. So while the API seems OK, the implementation
-///   needs work.
+// TODO: this is a PoC implementation.
+// - Takes 25 B for each delegate (on arm), which seems too much.
```
kaspar030, line 38: `25` means `28` usually (4 byte aligned)

src/ariel-os-embassy/src/delegate.rs
```diff
@@ -30,43 +34,53 @@ use crate::sendcell::SendCell;
...
 
+impl<T> !Send for Delegate<T> {}
+
 impl<T> Delegate<T> {
     /// Creates a new [`Delegate`].
     pub const fn new() -> Self {
         Self {
             send: Signal::new(),
             reply: Signal::new(),
+            was_exercised: AtomicBool::new(false),
```
kaspar030, line 54: maybe, if we turn this into a tristate enum (or `AtomicU8` with an enum wrapper to have it atomic), we can make the lend also safe?
(can do as follow up)

kaspar030 (approved):

PR #616, opened 2024-12-11
docs(debug): clean up the documentation

ROMemories:
# Description

This completes the documentation of the `debug` module.

## Issues/PRs references

Depends on #615.
Depends on #629.

## Open Questions

I'm not sure about the correct terminology here; I'm currently using the following:

- "debug facilities": general term for the module
- "debug logging": `defmt` macros and helpers
- "debug output" and "debug output session": semihosting's `println!` and `exit()` ("session" sounds too much like a GDB session to me)

## Change checklist

### DIFF
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -1,5 +1,9 @@
+//! Provides debug interface facilities.
+
 #![cfg_attr(not(test), no_std)]
 #![cfg_attr(test, no_main)]
+#![deny(missing_docs)]
+#![deny(clippy::pedantic)]
 
 #[cfg(all(feature = "rtt-target", feature = "esp-println"))]
 compile_error!(
@@ -41,6 +45,11 @@ impl ExitCode {
     }
 }
 
+/// Terminates the debug output session.
+///
+/// # Note
+///
+/// This may or may not stop the MCU.
 pub fn exit(code: ExitCode) {
     #[cfg(feature = "semihosting")]
     semihosting::process::exit(code.to_semihosting_code());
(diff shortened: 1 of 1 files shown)


src/ariel-os-debug/src/lib.rs
```diff
@@ -14,17 +18,36 @@ compile_error!(
...
+    #[allow(dead_code, reason = "not always used due to conditional compilation")]
+    fn to_semihosting_code(self) -> i32 {
+        match self {
+            Self::Success => 0,
+            Self::Failure => 1,
+        }
+    }
+}
 
+/// Terminates the debug output session.
```
chrysn, line 45: Should it say "and leaves the MCU running" (in which case some impls may be wrong) or should it say "and may or may not reboot"?

I'd prefer the former unless there is a documented (here?) reason to do the latter.

chrysn: Pinging @bergzand to verify that the storage example on his hardware does still reboot. (Nonetheless, it may make sense to document that this does *not* terminate execution of the current function, thread or task)

bergzand: I agree that this should clarify that it terminates the debug session and detaches from the mcu, but does not stop/halt the mcu.

bergzand (approved):

src/ariel-os-debug/src/lib.rs
bergzand: With the [added note](https://github.com/ariel-os/ariel-os/pull/616#discussion_r1930733031), I'm fine with this.

PR #626, opened 2024-12-12
chore(attr-macros): allow internal, macro-generated non-snake fn names

ROMemories:
# Description

This seems to made necessary by our recent nightly version bump in #476.

The `threading`, `hello-world`, and `http-server` examples can be used to test the effect of this (with the `#[thread]`, `#[task]`, and `#[spawner]` attribute macros respectively): before, the `threading` raised warnings OOTB when inspected by rust-analyzer, and the three macros raised duplicate warnings when the function names were not snake case (e.g., containing two consecutive underscores `__`).

## Change checklist

### DIFF
--- a/src/ariel-os-macros/src/spawner.rs
+++ b/src/ariel-os-macros/src/spawner.rs
@@ -84,6 +84,7 @@ pub fn spawner(args: TokenStream, item: TokenStream) -> TokenStream {
     };
 
     let expanded = quote! {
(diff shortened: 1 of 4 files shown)


kaspar030:
We can't `expect` here?

kaspar030:
> We can't `expect` here?

No :(

kaspar030 (approved):

PR #627, opened 2024-12-13
docs: add a basic HTTP client example

ROMemories:
# Description

As part of #410, this introduces an HTTP client example.
As explained in the README, an endpoint URL must be provided as an environment variable, and the example will make a GET request to that endpoint every few seconds and print the response body as a string.

## Issues/PRs references

Depends on #633.
Depends on #699.

## Change checklist

### DIFF
--- a/examples/http-client/Cargo.toml
+++ b/examples/http-client/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "http-client"
+version = "0.1.0"
+license.workspace = true
+edition.workspace = true
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "csprng",
+  "dns",
+  "mdns",
+  "override-network-config",
+  "tcp",
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+heapless = { workspace = true }
+rand = { workspace = true }
+reqwless = { version = "0.13.0", default-features = true, features = [
+  "embedded-tls",
+] }
--- a/examples/http-client/src/main.rs
+++ b/examples/http-client/src/main.rs
@@ -1,0 +1,111 @@
+#![no_main]
+#![no_std]
+#![feature(impl_trait_in_assoc_type)]
+#![feature(type_alias_impl_trait)]
+#![feature(used_with_arg)]
+
+use ariel_os::{
+    config,
+    debug::{exit, log::*, ExitCode},
+    net,
+    reexports::embassy_net,
+};
+use embassy_net::{
+    dns::DnsSocket,
+    tcp::client::{TcpClient, TcpClientState},
+};
+use reqwless::{
+    client::{HttpClient, TlsConfig, TlsVerify},
+    request::Method,
+};
+
+// RFC8449: TLS 1.3 encrypted records are limited to 16 KiB + 256 bytes.
+const MAX_ENCRYPTED_TLS_13_RECORD_SIZE: usize = 16640;
+// Required by `embedded_tls::TlsConnection::new()`.
+const TLS_READ_BUFFER_SIZE: usize = MAX_ENCRYPTED_TLS_13_RECORD_SIZE;
+// Can be smaller than the read buffer (could be adjusted: trade-off between memory usage and not
+// splitting large writes into multiple records).
+const TLS_WRITE_BUFFER_SIZE: usize = 4096;
+
+const TCP_BUFFER_SIZE: usize = 1024;
+const HTTP_BUFFER_SIZE: usize = 1024;
+
+const MAX_CONCURRENT_CONNECTIONS: usize = 2;
+
+const ENDPOINT_URL: &str = config::str_from_env_or!(
+    "ENDPOINT_URL",
+    "https://crab.ariel-os.org",
+    "endpoint to send the GET request to",
+);
+
+#[ariel_os::config(network)]
+const NETWORK_CONFIG: embassy_net::Config = {
+    use embassy_net::Ipv4Address;
+
+    embassy_net::Config::ipv4_static(embassy_net::StaticConfigV4 {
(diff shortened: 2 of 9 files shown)


examples/http-client/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+// splitting large writes into multiple records).
+const TLS_WRITE_BUFFER_SIZE: usize = 4096;
+
+const TCP_BUFFER_SIZE: usize = 1024;
+const HTTP_BUFFER_SIZE: usize = 1024;
+
+const MAX_CONCURRENT_CONNECTIONS: usize = 2;
+
+// Endpoint to send the GET request to.
+const ENDPOINT_URL: &str = env!("ENDPOINT_URL");
```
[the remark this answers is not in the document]
kaspar030: yes, add some example.html with a nice **text** message into this folder, then use its full github blob url as default.

examples/http-client/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+    let tls_verify = TlsVerify::None;
+    let tls_config = TlsConfig::new(tls_seed, &mut tls_rx_buffer, &mut tls_tx_buffer, tls_verify);
+
+    let mut client = HttpClient::new_with_tls(&tcp_client, &dns_client, tls_config);
+
+    loop {
+        if let Err(err) = send_http_get_request(&mut client, ENDPOINT_URL).await {
+            error!(
+                "Error while sending an HTTP request: {:?}",
+                defmt::Debug2Format(&err)
```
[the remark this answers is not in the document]
kaspar030: isn't that b/c the `ariel_os::debug::log::*` [imports defmt](https://github.com/ariel-os/ariel-os/blob/6141b3bdea95e63ba7c5f369a842c85abf153e8b/src/ariel-os-debug/src/lib.rs#L112)?

examples/http-client/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+) -> Result<(), reqwless::Error> {
+    let mut http_rx_buf = [0; HTTP_BUFFER_SIZE];
+
+    let mut handle = client.request(Method::GET, url).await?;
+    let response = handle.send(&mut http_rx_buf).await?;
+
+    info!("Response status: {}", response.status.0);
+
+    if let Some(ref content_type) = response.content_type {
+        info!("Response Content-Type: {}", content_type.as_str());
```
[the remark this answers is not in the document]
kaspar030: interesting. :)

examples/http-client/src/main.rs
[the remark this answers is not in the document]
kaspar030: esp-hal is using [`http://www.mobild-j.de`](https://github.com/esp-rs/esp-hal/blob/3a03dd88c774b68533f7c539c7480d19f05fc9ec/examples/src/bin/wifi_embassy_dhcp.rs#L137C54-L137C69), with some html that looks ok when dumped to a terminal. IMO that's nice.

examples/http-client/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+        dns_servers: heapless::Vec::new(),
+        gateway: Some(Ipv4Address::new(10, 42, 0, 1)),
+    })
+};
+
+#[ariel_os::task(autostart)]
+async fn main() {
+    let stack = network::network_stack().await.unwrap();
+
+    let tcp_client_state =
```
kaspar030, line 53: ... and these lines seem to be identical for a tls http client, with *maybe* configurable buffer sizes. I guess for now we're OK but we'll need to abstract this.

kaspar030: would you move it into a function already? and moce that to the bottom of the file?

examples/http-client/src/main.rs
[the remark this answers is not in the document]
kaspar030: typo, it is `mobile-j.de`, but you're correct

examples/http-client/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+// splitting large writes into multiple records).
+const TLS_WRITE_BUFFER_SIZE: usize = 4096;
+
+const TCP_BUFFER_SIZE: usize = 1024;
+const HTTP_BUFFER_SIZE: usize = 1024;
+
+const MAX_CONCURRENT_CONNECTIONS: usize = 2;
+
+// Endpoint to send the GET request to.
+const ENDPOINT_URL: Option<&str> = option_env!("ENDPOINT_URL");
```
kaspar030, line 35: can we re-use our config macros here?

examples/http-client/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "csprng",
+  "dns",
+  "mdns",
+  "override-network-config",
```
kaspar030, line 16:
```suggestion
```

base on #715 already?

examples/http-client/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+const NETWORK_CONFIG: embassy_net::Config = {
+    use embassy_net::Ipv4Address;
+
+    embassy_net::Config::ipv4_static(embassy_net::StaticConfigV4 {
+        address: embassy_net::Ipv4Cidr::new(Ipv4Address::new(10, 42, 0, 61), 24),
+        dns_servers: heapless::Vec::new(),
+        gateway: Some(Ipv4Address::new(10, 42, 0, 1)),
+    })
+};
+
```
kaspar030, line 51:
```suggestion
```

(with #715)

examples/http-client/Cargo.toml
kaspar030: ... hm, this one is g2g otherwise, so I'll update this in #715

examples/http-client/src/main.rs
kaspar030: postponed

kaspar030 (approved):

PR #628, opened 2024-12-13
feat(random): re-export some `rand` traits

kaspar030:
# Description

To re-export traits or not. That's the question.

While adding dependencies is easy, not adding them is a nop, so easiest. A software engineer has not achieved perfection if they have implemented all features, but if they have removed all boilerplate.

only somewhat related: our `CryptoRng` name-clashes with the `rand_core::CryptoRng` trait.

## Change checklist


kaspar030:
No consensus on this, closing.

PR #630, opened 2024-12-16
chore(embassy): ensure we call the embassy `raw::Executor::initialize()`

kaspar030:
# Description

[`raw::Executor` now has an `unsafe` `initialize()` function](https://github.com/embassy-rs/embassy/blob/main/embassy-executor/CHANGELOG.md) that needs to be called. Do it.

## Change checklist


chrysn:
I don't think this needs to be unsafe: The `&'static mut self -> !` already means that there is only one call ever regularly, I think that it being `'static` rather than `'a` means that we even don't have to worry about the panic unwind case. If we can't convince ourselves that we're safe against unwinds, let's rather drop a abort-on-drop bomb in there and make the outer function safe.

PR #646, opened 2024-12-20
doc(book): point out how multiple log levels are done

chrysn:
# Description

This extends the book documentation around `-D LOG`.

Reason we should do this explicitly is that we do comma-join LOG into DEFMT_LOG, so it's not immediately obvious to those who know defmt how this is done.

## Change checklist

### DIFF
--- a/book/src/tooling/defmt.md
+++ b/book/src/tooling/defmt.md
@@ -5,12 +5,13 @@ Ariel OS supports [defmt] on all platforms. It is enabled by default.
 See the [defmt documentation] for general info on `defmt`.
 
 In Ariel OS, the log level defaults to `info`. It can be configured using the
-laze variable `LOG`.
+laze variable `LOG`. As usual with defmt, different levels can be given per
+crate or per module.
 
 Example:
 
 ```shell
-# laze build -C examples/log --builders nrf52840dk -DLOG=warn run
+# laze build -C examples/log --builders nrf52840dk -DLOG=warn,ariel_os_rt=trace run
 ```
 
 Then within Rust code, import `ariel_os::debug::log` items, then use `defmt` log


book/src/tooling/defmt.md
```diff
@@ -5,12 +5,13 @@ Ariel OS supports [defmt] on all platforms. It is enabled by default.
...
 In Ariel OS, the log level defaults to `info`. It can be configured using the
-laze variable `LOG`.
+laze variable `LOG`. The variable can be given multiple times for crate- or
+module-level granularity.
 
 Example:
 
 ```shell
-# laze build -C examples/log --builders nrf52840dk -DLOG=warn run
+# laze build -C examples/log --builders nrf52840dk -DLOG=warn -DLOG=ariel_os_rt=trace run
```
kaspar030, line 14:
```suggestion
# laze build -C examples/log --builders nrf52840dk -DLOG=warn -DLOG=ariel_os_rt=trace run
# laze build -C examples/log --builders nrf52840dk -DLOG=warn,ariel_os_rt=trace run
```

Comma-seperated is also possible. ... did I say, tested it, and it doesn't work. That's a laze bug, the `--define/-D` had `,` as value separator like builders/apps.

kaspar030: https://github.com/kaspar030/laze/pull/590

kaspar030:
I'd prefer the soon-to-be possible comma separated variant mentioned as default, as that matches how rust logs are usually defined.

kaspar030:
Wow, *this* might be the reason why sccache didn't recognize changes to the DEFMT_LOG variable - laze just didn't pass them on ...

chrysn:
Updated the PR to reflect what works once the latest laze is around. I've even removed the two-`-D` form -- no good reason to confuse users.

Once the single form works, this whole PR is really needed so much any more, though; it was a good idea because it did *not* work the way people would expect, but now that it will…

kaspar030:
> Once the single form works, this whole PR is really needed so much any more, though; it was a good idea because it did _not_ work the way people would expect, but now that it will…

IMO docs improve with this anyways! And we found & fixed a nasty bug. :) Thanks a lot!

kaspar030:
`6759b90: error Disallowed type `doc` used, please use one of ["build", "chore", "ci", "docs", "fix", "feat", "perf", "refactor", "style", "test"]`

linter wants the commitprefix to be `docs` (plural)

chrysn:
Thanks, should go green now.

PR #634, opened 2024-12-17
feat: debug logs without defmt (wip)

kaspar030:
# Description

This adds a `log` crate backend to `ariel_os::debug::log`.

Putting it as draft as https://github.com/ariel-os/ariel-os/pull/629 needs to be resolved first.

## Issues/PRs references

Depends on #629

## Open Questions

- [ ] when to select dependency, always but not use in defmt case
- [ ] how to allow both "debug logs" (over rtt/... for debugging) and "application logs" (runtime logging for applications)

## Change checklist

- [ ] I have cleaned up my commit history and squashed fixup commits.
- [ ] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [ ] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

PR #647, opened 2024-12-21
feat: native board

kaspar030:
# Description

This PR provides an initial implementation of an infini-core scheduler, and a native board.

Try `laze -Cexamples/hello-world build -b native run`.

The scheduler implementation basically just removes the runqueue, and priorities are ignored. Each thread gets an `std::thread::Thread`.

The most tricky thing, (apart from enabling std for just some parts of the system), was to keep the isr based stopping of threads, where a call to `schedule()` pends an ISR, which triggers right at the end of a critical section. If that ISR de-schedules a thread into a non-Running state, that thread needs to stop. All our IPC logic depends on this behavior.
As std mutexes can only be unlocked from the thread that locked them, I used [`atomic-wait`](https://github.com/m-ou-se/atomic-wait) futexes, and provide a wrapper `critical-section` implementation that waits for a futex on `release()` if a thread is not supposed to continue. Runqueue operations are replaced a simple atomic stating whether a thread is "Running" or not, and the futex waits based on that.

[`atomic-wait`](https://github.com/m-ou-se/atomic-wait) hasn't seen much activity lately, but was written by Mara Bos herself, and seems to work well so far.

## Issues/PRs references

Depends on #1131.

~~Depends on #634 for output.~~

## Open Questions

- [ ] no peripherals so far, GPIO using tests/examples compile-fail
- [ ] networking
- [x] embedded-test

## Change checklist

### DIFF
--- a/ariel-os-cargo.toml
+++ b/ariel-os-cargo.toml
@@ -25,6 +25,7 @@ embassy-nrf = { git = "https://github.com/ariel-os/embassy", branch = "embassy-n
 embassy-net = { git = "https://github.com/ariel-os/embassy", branch = "embassy-net-v0.6.0+ariel-os" }
 embassy-rp = { git = "https://github.com/ariel-os/embassy", branch = "embassy-rp-v0.4.0+ariel-os+trng-panic-fix" }
 embassy-stm32 = { git = "https://github.com/ariel-os/embassy", branch = "embassy-stm32-v0.2.0+ariel-os" }
+embassy-time = { git = "https://github.com/ariel-os/embassy", branch = "embassy-time-v0.4.0+ariel-os" }
 
 # ariel-os esp-hal fork
 esp-alloc = { git = "https://github.com/ariel-os/esp-hal", branch = "v0.23.1+ariel-os-threads" }
@@ -36,3 +37,5 @@ xtensa-lx-rt = { git = "https://github.com/ariel-os/esp-hal", branch = "v0.23.1+
 
 # patched to use portable-atomics <https://github.com/seanmonstar/try-lock/pull/11>
 try-lock = { git = "https://github.com/seanmonstar/try-lock", rev = "45c39685b56a4dba1b71bdbbbe5f731c3c77dc50" }
+
+embedded-test = { git = "https://github.com/ariel-os/embedded-test", branch = "v0.6.2+ariel-os" }
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -994,6 +994,8 @@ modules:
     selects:
       - ?debug-console
       - ?rtt-target
+    provides_unique:
+      - embedded-test-runner
 
     env:
       global:
@@ -1464,6 +1466,17 @@ modules:
     selects:
       - doc-only
 
+  - name: infini-core
+    help: use platform as if it had one core per thread
+    provides_unique:
+      - critical-section
+    context:
+      - native
+    env:
+      global:
+        FEATURES:
+          - ariel-os/infini-core
+
   - name: rtt-target
     help: use rtt-target in ariel-os-debug
     provides_unique:
@@ -1526,21 +1539,54 @@ modules:
   - name: embedded-test
     context: ariel-os
     selects:
+      - embedded-test-extra
       # There is an issue with multi-core on esp.
       - ?esp-single-core
       - sw/threading
-      - probe-rs
+      - embedded-test-runner
     tasks:
       test:
         workdir: ${appdir}
         cmd:
           - ${CARGO_ENV} ${CARGO} ${CARGO_TOOLCHAIN} ${CARGO_ARGS} test ${FEATURES} $@
         build: false
-
     env:
       global:
         RUSTFLAGS:
           - -Clink-arg=-Tembedded-test.x
+
+  - name: embedded-test-extra
+    help: Context-specific settings for embedded-test
+    context:
+      - ariel-os
+    env:
+      global:
+        FEATURES:
+          - embedded-test/semihosting
+
+  - name: embedded-test-extra
+    help: native specific settings for embedded-test
+    context:
+      # This overrides ariel-os::embedded-test-extra
+      - native
+    env:
+      global:
+        FEATURES:
+          - embedded-test/std
+
+  - name: embedded-test-runner
+    help: provided if there is an embedded-test runner
+    selects:
+      - doc-only
+
+  - name: embedded-test-runner-std
+    context: native
+    provides_unique:
+      - embedded-test-runner
+    env:
+      global:
+        # `cargo install embedded-test-std-runner`
+        CARGO_RUNNER: 'embedded-test-std-runner'
 
   # This helper module works around laze not supporting sth like "if context == foo then depend on bar".
   - name: esp-single-core
@@ -1821,6 +1867,22 @@ builders:
       CARGO_ENV:
         # This ISR is unused on a naked board. Configured here for testing.
         - CONFIG_SWI=USART2_LPUART2
+
+  - name: native
+    parent: ariel-os
+    selects:
+      - infini-core
+      - ?debug-console
+    provides:
+      - sw/benchmark
+    disables:
+      - semihosting
+      - defmt
+      - probe-rs
+    env:
+      RUSTC_TARGET: x86_64-unknown-linux-gnu
+      CARGO_TARGET_PREFIX: CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU
+      CARGO_RUNNER: '"sh -c"'
 
 apps:
   # define a dummy host application so the host tasks work
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -1,6 +1,6 @@
 //! Provides debug interface facilities.
 
-#![cfg_attr(not(test), no_std)]
+#![cfg_attr(not(any(test, feature = "std")), no_std)]
 #![cfg_attr(test, no_main)]
 #![deny(missing_docs)]
 
@@ -42,6 +42,9 @@ impl ExitCode {
 pub fn exit(code: ExitCode) {
     #[cfg(feature = "semihosting")]
     semihosting::process::exit(code.to_semihosting_code());
+
+    #[cfg(feature = "std")]
+    std::process::exit(code.to_semihosting_code());
 
     #[allow(unreachable_code, reason = "stop nagging")]
     let _ = code;
@@ -175,6 +178,17 @@ pub mod backend {
     }
 }
 
+#[cfg(all(feature = "debug-console", feature = "std"))]
+mod backend {
+    pub use std::println;
+
+    #[doc(hidden)]
+    pub fn init() {
+        #[cfg(feature = "log")]
+        crate::logger::init();
+    }
+}
+
 #[cfg(not(feature = "debug-console"))]
 mod backend {
     #[doc(hidden)]
@@ -252,7 +266,7 @@ mod logger {
             });
         }
 
-        log::trace!("debug logging enabled");
+        log::debug!("debug logging enabled at level {}", MAX_LEVEL);
     }
 
     struct DebugLogger;
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -32,6 +32,13 @@ impl ExitCode {
             Self::Failure => 1,
         }
     }
+    #[allow(dead_code, reason = "not always used due to conditional compilation")]
+    fn to_std_code(self) -> i32 {
+        match self {
+            Self::Success => 0,
+            Self::Failure => 1,
+        }
+    }
 }
 
 /// Terminates the debug output session.
@@ -44,7 +51,7 @@ pub fn exit(code: ExitCode) {
     semihosting::process::exit(code.to_semihosting_code());
 
     #[cfg(feature = "std")]
-    std::process::exit(code.to_semihosting_code());
+    std::process::exit(code.to_std_code());
 
     #[allow(unreachable_code, reason = "stop nagging")]
     let _ = code;
--- a/src/ariel-os-hal/src/dummy/mod.rs
+++ b/src/ariel-os-hal/src/dummy/mod.rs
@@ -13,6 +13,7 @@
     reason = "this dummy module mimics manufacturer-specific crates"
 )]
 
+#[cfg(not(context = "native"))]
 mod executor;
 
 #[doc(hidden)]
@@ -48,13 +49,18 @@ pub mod storage;
 #[cfg(feature = "usb")]
 pub mod usb;
 
+#[cfg(not(context = "native"))]
 pub use executor::{Executor, Spawner};
 pub use peripheral::{OptionalPeripherals, Peripheral};
 
 #[doc(hidden)]
 #[must_use]
 pub fn init() -> OptionalPeripherals {
-    unimplemented!();
+    if cfg!(context = "native") {
+        OptionalPeripherals {}
+    } else {
+        unimplemented!()
+    }
 }
 
 #[doc(hidden)]
--- a/src/ariel-os-hal/src/dummy/mod.rs
+++ b/src/ariel-os-hal/src/dummy/mod.rs
@@ -13,7 +13,6 @@
     reason = "this dummy module mimics manufacturer-specific crates"
 )]
 
-#[cfg(not(context = "native"))]
 mod executor;
 
 #[doc(hidden)]
@@ -49,18 +48,13 @@ pub mod storage;
 #[cfg(feature = "usb")]
 pub mod usb;
 
-#[cfg(not(context = "native"))]
 pub use executor::{Executor, Spawner};
 pub use peripheral::{OptionalPeripherals, Peripheral};
 
 #[doc(hidden)]
 #[must_use]
 pub fn init() -> OptionalPeripherals {
-    if cfg!(context = "native") {
-        OptionalPeripherals {}
-    } else {
-        unimplemented!()
-    }
+    unimplemented!()
 }
 
 #[doc(hidden)]
--- a/src/ariel-os-hal/src/lib.rs
+++ b/src/ariel-os-hal/src/lib.rs
@@ -31,10 +31,18 @@ pub use define_peripherals::*;
 
 cfg_if::cfg_if! {
     if #[cfg(context = "native")] {
-        // TODO: fill this with life:
-        // pub use ariel_os_native::*;
-        mod dummy;
-        pub use dummy::{gpio, peripheral, identity, SWI, init, OptionalPeripherals};
+        // TODO: fill me with life
+        //pub use ariel_os_native::*;
+        #[path="dummy/gpio.rs"]
+        pub mod gpio;
+        #[path="dummy/identity.rs"]
+        pub mod identity;
+        #[path="dummy/peripheral.rs"]
+        pub mod peripheral;
+        pub use peripheral::{OptionalPeripherals};
+        pub fn init() -> OptionalPeripherals {
+            OptionalPeripherals {}
+        }
     } else if #[cfg(context = "nrf")] {
         pub use ariel_os_nrf::*;
     } else if #[cfg(context = "rp")] {
--- a/src/ariel-os-hal/src/lib.rs
+++ b/src/ariel-os-hal/src/lib.rs
@@ -30,7 +30,12 @@ pub mod define_peripherals;
 pub use define_peripherals::*;
 
 cfg_if::cfg_if! {
-    if #[cfg(context = "nrf")] {
+    if #[cfg(context = "native")] {
+        // TODO: fill this with life:
+        // pub use ariel_os_native::*;
+        mod dummy;
+        pub use dummy::{gpio, peripheral, identity, SWI, init, OptionalPeripherals};
+    } else if #[cfg(context = "nrf")] {
         pub use ariel_os_nrf::*;
     } else if #[cfg(context = "rp")] {
         pub use ariel_os_rp::*;
--- a/src/ariel-os-hal/src/lib.rs
+++ b/src/ariel-os-hal/src/lib.rs
@@ -31,18 +31,9 @@ pub use define_peripherals::*;
 
 cfg_if::cfg_if! {
     if #[cfg(context = "native")] {
-        // TODO: fill me with life
-        //pub use ariel_os_native::*;
-        #[path="dummy/gpio.rs"]
-        pub mod gpio;
-        #[path="dummy/identity.rs"]
-        pub mod identity;
-        #[path="dummy/peripheral.rs"]
-        pub mod peripheral;
-        pub use peripheral::{OptionalPeripherals};
-        pub fn init() -> OptionalPeripherals {
-            OptionalPeripherals {}
-        }
+        mod dummy;
+        pub use ariel_os_native::*;
+        pub use dummy::{gpio, identity, peripheral};
     } else if #[cfg(context = "nrf")] {
         pub use ariel_os_nrf::*;
     } else if #[cfg(context = "rp")] {
--- a/src/ariel-os-native/Cargo.toml
+++ b/src/ariel-os-native/Cargo.toml
@@ -1,0 +1,47 @@
+[package]
+name = "ariel-os-native"
+version = "0.2.0"
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+cfg-if = { workspace = true }
+defmt = { workspace = true, optional = true }
+embassy-embedded-hal = { workspace = true, optional = true }
+embassy-executor = { workspace = true, default-features = false }
+embassy-time = { workspace = true, default-features = false, features = [
+  "std",
+] }
+embedded-hal-async = { workspace = true }
+ariel-os-debug = { workspace = true, features = ["std"] }
+ariel-os-embassy-common = { workspace = true }
+ariel-os-random = { workspace = true, optional = true }
+rand = { workspace = true, default-features = false, optional = true, features = [
+  "getrandom",
+] }
+
+[features]
+## Enables GPIO interrupt support.
+external-interrupts = ["ariel-os-embassy-common/external-interrupts"]
+
+## Enables seeding the random number generator from hardware.
+hwrng = ["dep:ariel-os-random"]
+
+## Enables I2C support.
+i2c = ["ariel-os-embassy-common/i2c"]
+
+## Enables SPI support.
+spi = ["ariel-os-embassy-common/spi"]
+
+## Enables storage support.
+storage = ["dep:embassy-embedded-hal"]
+
+## Enables USB support.
+usb = []
+
+## Enables defmt support.
+defmt = ["dep:defmt"]
--- a/src/ariel-os-native/src/lib.rs
+++ b/src/ariel-os-native/src/lib.rs
@@ -2,3 +2,11 @@
 
 #![no_std]
 #![cfg_attr(nightly, feature(doc_auto_cfg))]
+
+pub struct OptionalPeripherals {}
+
+pub fn init() -> OptionalPeripherals {
+    OptionalPeripherals {}
+}
+
+pub struct SWI {}
--- a/src/ariel-os-native/src/lib.rs
+++ b/src/ariel-os-native/src/lib.rs
@@ -1,0 +1,4 @@
+//! Items specific to the "native" implementation
+
+#![no_std]
+#![cfg_attr(nightly, feature(doc_auto_cfg))]
--- a/src/ariel-os-power/src/lib.rs
+++ b/src/ariel-os-power/src/lib.rs
@@ -1,7 +1,7 @@
 //! Provides power management functionality.
 
 #![deny(missing_docs)]
-#![no_std]
+#![cfg_attr(not(context = "native"), no_std)]
 
 /// Reboots the MCU.
 ///
@@ -16,6 +16,8 @@ pub fn reboot() -> ! {
             esp_hal::reset::software_reset();
             #[expect(clippy::empty_loop, reason = "software_reset() should eventually never return")]
             loop {}
+        } else if #[cfg(context = "native")] {
+            std::process::exit(0)
         } else if #[cfg(context = "ariel-os")] {
             compile_error!("reboot is not yet implemented for this platform")
         } else {
--- a/src/ariel-os-threads/src/arch/native.rs
+++ b/src/ariel-os-threads/src/arch/native.rs
@@ -1,0 +1,114 @@
+use std::cell::Cell;
+
+use std::sync::atomic::{AtomicU32, Ordering};
+
+use crate::{Arch, SCHEDULER, ThreadId, ThreadState, thread::Thread};
+
+pub struct Cpu;
+
+mod critical_section;
+
+static THREAD_RUNNABLE: [AtomicU32; crate::THREAD_COUNT] =
+    [const { AtomicU32::new(0) }; crate::THREAD_COUNT];
+
+#[derive(Debug)]
+pub struct ThreadData {
+    thread: Option<std::thread::Thread>,
+}
+
+impl ThreadData {
+    pub const fn new() -> Self {
+        Self { thread: None }
+    }
+}
+
+impl ThreadData {
+    thread_local! {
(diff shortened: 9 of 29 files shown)


chrysn:
Awesome – looking forward to trying this out after the holidays!

Until we have a better place to spool ideas on peripherals, here's something I'd like to deposit (likely as a note-to-self): It would be highly convenient to have a BLE back-end that on the host side emulates a BLE adapter within radio reach of the device. That would speed up development a lot, both on the device side and for the host applications that interact with the device. (May even work in a WASM port of native that can run in the browser – with both the host webapp and a native board running in the same browser tab).

chrysn:
This will need an entry in the `install-toolchain` target, but apart from that: The log example and hello-world-example work. Sure, peripherals will be needed, but that'll be added one by one.

Is this in a state where a code-level review can start?

(There's the aspect of documentation too, and we [gathered](https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$FFoig9i5mYt1yVQqoGl7oQbzhPgiN5kc0SqeWOJLDdI?via=matrix.org&via=utwente.io&via=club1.fr) a bit of "we should document its limitations", but I think this can go into the main tree and mature before we even have to do much in terms of having it as a documented feature.)

kaspar030:
> Is this in a state where a code-level review can start?

I was kinda holding off until I get embedded-test to work, which upstream now has [initial support](https://github.com/probe-rs/embedded-test/pull/58). I'm almost there. :/

src/ariel-os-hal/src/dummy/mod.rs
```diff
@@ -13,6 +13,7 @@
     reason = "this dummy module mimics manufacturer-specific crates"
 )]
 
+#[cfg(not(context = "native"))]
```
kaspar030: Just three special cases in the dummy module allows using it for native w/o warnings.
`ariel-os-native` is basically empty now. I kept it so we have the scaffolding to add impls there.

src/ariel-os-native/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+  "std",
+] }
+embedded-hal-async = { workspace = true }
+paste = { workspace = true }
+portable-atomic = { workspace = true }
+ariel-os-debug = { workspace = true, features = ["std"] }
+ariel-os-embassy-common = { workspace = true }
+ariel-os-random = { workspace = true, optional = true }
+
+[features]
```
kaspar030, line 27: These don't have any implementation behind, and are filtered by laze before they end up here. I kept them as scaffolding.

src/ariel-os-threads/src/sync/mutex.rs
```diff
@@ -85,7 +85,7 @@ impl<T> Mutex<T> {
     /// # Panics
     ///
     /// Panics if called outside of a thread context.
-    pub fn lock(&self) -> MutexGuard<T> {
+    pub fn lock(&self) -> MutexGuard<'_, T> {
```
kaspar030: These were suggested by clippy:
```
error: lifetime flowing from input to output with different syntax can be confusing                                                                            
  --> src/ariel-os-threads/src/sync/mutex.rs:88:17                                                                                                             
   |                                                                                                                                                           
88 |     pub fn lock(&self) -> MutexGuard<T> {                                                                                                                 
   |                 ^^^^^     ------------- the lifetime gets resolved as `'_`                                                                                
   |                 |                                                                                                                                         
   |                 this lifetime flows to the output                                                                                                         
   |                                                                                                                                                           
   = note: `-D mismatched-lifetime-syntaxes` implied by `-D warnings`                                                                                          
   = help: to override `-D warnings` add `#[allow(mismatched_lifetime_syntaxes)]`                                                                              
help: one option is to remove the lifetime for references and use the anonymous lifetime for paths                                                             
   |                                                                                                                                                           
88 |     pub fn lock(&self) -> MutexGuard<'_, T> {                                                                                                             
   |                                      +++

File: src/ariel-os-hal/src/dummy/mod.rs (hunk 1 above)
Comment (kaspar030) [reply src/ariel-os-hal/src/dummy/mod.rs]: Fixed in [`293b488` (#647)](https://github.com/ariel-os/ariel-os/pull/647/commits/293b48837d9554baa110331ab30d28dc9eb8758e).

File: src/ariel-os-threads/src/arch/native.rs [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+        pub static ID: Cell<Option<ThreadId>> = Cell::new(None);
+    }
+}
+
+impl Arch for Cpu {
+    type ThreadData = ThreadData;
+    const DEFAULT_THREAD_DATA: Self::ThreadData = ThreadData::new();
+
+    fn setup_stack(thread: &mut Thread, _stack: &mut [u8], func: usize, arg: usize) {
+        let func: fn(arg: usize) = unsafe { core::mem::transmute(func) };
```
ROMemories: That transmute is possibly unsound, the docs say you need to [cast to a raw pointer before transmuting](https://doc.rust-lang.org/stable/core/primitive.fn.html#casting-to-and-from-integers). The `setup_stack()` method would also really need to be marked `unsafe` (with a proper `# Safety` doc-comment). But really even casting function pointers to a single `usize` is still [ABI dependent](https://doc.rust-lang.org/stable/core/primitive.fn.html#abi) and I would *really* prefer we first finally refactor the `thread::create_raw()` function to take a proper function pointer instead of [confusing addresses with pointers](https://doc.rust-lang.org/stable/core/ptr/index.html#provenance).

src/ariel-os-threads/src/arch/native/critical_section.rs
ROMemories, line 1: Seems to be mostly copied from [`rust-embedded/critical-section/src/std.rs`](https://github.com/rust-embedded/critical-section/blob/cebd3d76cc5237d77abd0e03c090577fad1c689f/src/std.rs)

src/ariel-os-threads/src/arch/native.rs
kaspar030: agreed. "works for now". Let's tackle that here: https://github.com/ariel-os/ariel-os/pull/1131

src/ariel-os-threads/src/sync/mutex.rs
ROMemories: Should be addressed by #1133.

src/ariel-os-native/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+] }
+embedded-hal-async = { workspace = true }
+paste = { workspace = true }
+portable-atomic = { workspace = true }
+ariel-os-debug = { workspace = true, features = ["std"] }
+ariel-os-embassy-common = { workspace = true }
+ariel-os-random = { workspace = true, optional = true }
+rand = { workspace = true, default-features = false, optional = true, features = [
+  "getrandom",
+] }
```
ROMemories: Seems like half these dependencies are unused, could you do some cleanup?

src/ariel-os-native/src/hwrng.rs
```diff
@@ -0,0 +1,11 @@
+/// Constructs the hardware random number generator (RNG) for the nRF family.
+///
```
ROMemories: Copy-paste leftover:

```suggestion
/// Constructs the hardware random number generator (RNG).
```

src/ariel-os-native/src/hwrng.rs
```diff
@@ -0,0 +1,11 @@
...
+pub fn construct_rng(_peripherals: &mut crate::OptionalPeripherals) {
+    cfg_if::cfg_if! {
+        if #[cfg(any(context = "native"))] {
```
ROMemories:
```suggestion
        if #[cfg(context = "native")] {
```

src/ariel-os-runqueue/src/runqueue.rs
```diff
@@ -12,6 +12,12 @@ const USIZE_BITS: usize = mem::size_of::<usize>() * 8;
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct RunqueueId(u8);
 
+impl core::fmt::Display for RunqueueId {
+    fn fmt(&self, formatter: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
+        core::fmt::Debug::fmt(self, formatter)
+    }
+}
```
ROMemories: Do we really need these in this PR?

ROMemories (commented):
Just a partial review for now, no need to start acting on my comments. Will do another pass later.

src/ariel-os-threads/src/arch/native.rs
kaspar030: This is now depending on #1131.

src/ariel-os-debug/src/lib.rs
```diff
@@ -43,6 +43,9 @@ pub fn exit(code: ExitCode) {
     #[cfg(feature = "semihosting")]
     semihosting::process::exit(code.to_semihosting_code());
 
+    #[cfg(feature = "std")]
+    std::process::exit(code.to_semihosting_code());
```
ROMemories: The exit code is platform-specific, the semihosting code can't just be reused like that.

src/ariel-os-debug/src/lib.rs
```diff
@@ -195,7 +209,7 @@ pub use backend::*;
 #[doc(hidden)]
 #[cfg(feature = "log")]
 mod logger {
```
ROMemories: Seems like changes to this module are bugfixes that could be extracted into a separate PR.

laze-project.yml
```diff
@@ -1822,6 +1870,23 @@ builders:
...
+      - has_hwrng
+      - sw/benchmark
+    disables:
+      - semihosting
+      - defmt
+      - probe-rs
+    env:
+      RUSTC_TARGET: x86_64-unknown-linux-gnu
+      CARGO_TARGET_PREFIX: CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU
+      CARGO_RUNNER: '"sh -c"'
```
ROMemories: I think most of this should be a laze context (like MCUs) and we should have a separate builder (maybe also named `native`). That would also us to simulate whole boards when we have OoT board definitions.

src/ariel-os-threads/src/arch/native/critical_section.rs
ROMemories: Could you split this into two files: the `std` coming from upstream, and the Ariel OS one, so that it's easier to diff against the upstream one in the future?

src/ariel-os-threads/src/arch/native/critical_section.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::sync::{Mutex, MutexGuard};
+
+static GLOBAL_MUTEX: Mutex<()> = Mutex::new(());
+
+// This is initialized if a thread has acquired the CS, uninitialized otherwise.
+static mut GLOBAL_GUARD: MaybeUninit<MutexGuard<'static, ()>> = MaybeUninit::uninit();
+
+std::thread_local!(static IS_LOCKED: Cell<bool> = Cell::new(false));
+
+struct StdCriticalSection;
```
ROMemories: Seems to me we could use the upstream one if (a) it was public and (b) didn't [set itself as the one critical section](https://github.com/rust-embedded/critical-section/blob/cebd3d76cc5237d77abd0e03c090577fad1c689f/src/std.rs#L13)?

src/ariel-os-threads/src/arch/native/critical_section.rs
```diff
@@ -0,0 +1,87 @@
...
+
+struct ArielCriticalSection;
+unsafe impl critical_section::Impl for ArielCriticalSection {
+    unsafe fn acquire() -> bool {
+        unsafe { StdCriticalSection::acquire() }
+    }
+
+    unsafe fn release(nested_cs: bool) {
+        if !nested_cs {
+            let thread_id = crate::ThreadData::ID.get();
```
ROMemories, line 52: Seems like the ordering of this statement is important, could you add a comment explaining the relationship with the next statement?

laze-project.yml
```diff
@@ -1526,22 +1539,57 @@ modules:
...
+    context:
+      # This overrides ariel-os::embedded-test-extra
+      - native
+    env:
+      global:
+        FEATURES:
+          - embedded-test/std
+        RUSTFLAGS:
+          - -Clink-arg=-Wl,-znostart-stop-gc
+          - -Clink-arg=-Wl,-static
```
ROMemories: Can you explain what these do/why they're needed?

laze-project.yml
```diff
@@ -1526,22 +1539,57 @@ modules:
...
+    selects:
+      - doc-only
+
+  - name: embedded-test-runner-std
+    context: native
+    provides_unique:
+      - embedded-test-runner
+    env:
+      global:
+        CARGO_RUNNER: 'embedded-test-std-runner'
```
ROMemories, line 1589: I've found [this](https://github.com/kaspar030/embedded-test-std-runner), but not sure I understand how that's articulated.

src/ariel-os-hal/src/lib.rs
```diff
@@ -30,7 +30,18 @@ pub mod define_peripherals;
...
+        pub use ariel_os_native::*;
+        #[path="dummy/gpio.rs"]
+        pub mod gpio;
+        #[path="dummy/identity.rs"]
+        pub mod identity;
+        #[path="dummy/peripheral.rs"]
+        pub mod peripheral;
+        pub fn init() -> OptionalPeripherals {
+            OptionalPeripherals {}
+        }
```
ROMemories: Could you extract the commit that splits these dummy modules into a separate PR?

ROMemories: It seems to me that these could then be more simply rewritten with something along the lines of:

```rust
mod dummy;

pub use dummy::{gpio, identity, peripheral};
```

laze-project.yml
kaspar030: they actually aren't, I dropped them.

laze-project.yml
kaspar030: interesting and good idea. builders share the namespace with contexts, so we'd need a different name.
Maybe tackle later? and, it is possible to have another board with native as parent.

src/ariel-os-debug/src/lib.rs
kaspar030: copy&pasted to `to_std_code()`

src/ariel-os-native/Cargo.toml
kaspar030: I dropped paste and portable-atomic, I don't think native will ever need them. I kept the ones I think it will use (defmt, embassy-e-h, embedded-hal-async,ariel-os-embassy-common, ariel-os-debug)

src/ariel-os-runqueue/src/runqueue.rs
kaspar030: no, dropped and using `{:?}` now in the only user in this PR.

src/ariel-os-threads/src/arch/native/critical_section.rs
kaspar030: Yeah. They might accept a PR.

src/ariel-os-threads/src/arch/native/critical_section.rs
kaspar030: I've added some module level docs, and another comment here.

laze-project.yml
kaspar030: hm. how do we articulate this? that test runner needs to be installed (via `cargo install embedded-test-std-runner`). I've added a comment to the laze file.

src/ariel-os-hal/src/lib.rs
kaspar030: That would require re-adding the `cfg(context = "native")` that I had removed in the first place

ROMemories: > That would require re-adding the `cfg(context = "native")` that I had removed in the first place

Not sure I get what you mean. Using my suggestion and adding
```rust
pub struct SWI;
```

to `ariel-os-native/src/lib.rs` seems to be building successfully.

kaspar030: yes, done.

src/ariel-os-debug/src/lib.rs
kaspar030: split out the log comparison fix: https://github.com/ariel-os/ariel-os/pull/1152

kaspar030: let's piggy-back the other two changes.

ROMemories:
Seems like the HWRNG implementation is orthogonal to the rest, could you also move it to a follow-up, so this PR is even smaller and easier for me to review?

kaspar030:
> Seems like the HWRNG implementation is orthogonal to the rest, could you also move it to a follow-up, so this PR is even smaller and easier for me to review?

Removed the hwrng impl from this PR, will open another once this is in!

ariel-os-cargo.toml
```diff
@@ -25,6 +25,7 @@ embassy-nrf = { git = "https://github.com/ariel-os/embassy", branch = "embassy-n
 embassy-net = { git = "https://github.com/ariel-os/embassy", branch = "embassy-net-v0.6.0+ariel-os" }
 embassy-rp = { git = "https://github.com/ariel-os/embassy", branch = "embassy-rp-v0.4.0+ariel-os+trng-panic-fix" }
 embassy-stm32 = { git = "https://github.com/ariel-os/embassy", branch = "embassy-stm32-v0.2.0+ariel-os" }
+embassy-time = { git = "https://github.com/ariel-os/embassy", branch = "embassy-time-v0.4.0+ariel-os" }
```
ROMemories, line 28: For the record: currently all this does is drop the critical section impl auto-enabled by the `std` feature, so that it does not conflict with the implementation provided in this PR.

laze-project.yml
```diff
@@ -1822,6 +1868,22 @@ builders:
...
+    provides:
+      - sw/benchmark
+    disables:
+      - semihosting
+      - defmt
+      - probe-rs
+    env:
+      RUSTC_TARGET: x86_64-unknown-linux-gnu
+      CARGO_TARGET_PREFIX: CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU
+      CARGO_RUNNER: '"sh -c"'
```
ROMemories, line 1885: Why do we need the double quotes?

laze-project.yml
ROMemories: That'll definitely need to go into the docs when we have some (later on).

src/ariel-os-debug/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! Provides debug interface facilities.
 
-#![cfg_attr(not(test), no_std)]
+#![cfg_attr(not(any(test, context = "native")), no_std)]
```
ROMemories: Shouldn't that use the newly introduced Cargo feature instead?

```suggestion
#![cfg_attr(not(any(test, feature = "std")), no_std)]
```

laze-project.yml
```diff
@@ -1526,22 +1539,55 @@ modules:
...
         cmd:
           - ${CARGO_ENV} ${CARGO} ${CARGO_TOOLCHAIN} ${CARGO_ARGS} test ${FEATURES} $@
         build: false
-
     env:
       global:
         RUSTFLAGS:
           - -Clink-arg=-Tembedded-test.x
 
+  - name: embedded-test-extra
```
ROMemories, line 1558: Not a huge fan of the name of these modules, but considering they're internal we can just renamed them later.

src/ariel-os-hal/src/lib.rs
```diff
@@ -30,7 +30,11 @@ pub mod define_peripherals;
 pub use define_peripherals::*;
 
 cfg_if::cfg_if! {
-    if #[cfg(context = "nrf")] {
+    if #[cfg(context = "native")] {
+        pub use ariel_os_native::*;
+        pub mod dummy;
+        pub use dummy::{gpio, identity, peripheral};
```
ROMemories: No need to make the module `pub`, and I would really prefer we kept module *declarations* before `use`s:

```suggestion
        mod dummy;

        pub use ariel_os_native::*;
        pub use dummy::{gpio, identity, peripheral};
```

src/ariel-os-native/src/lib.rs
```diff
@@ -0,0 +1,11 @@
...
+
+#![no_std]
+#![cfg_attr(nightly, feature(doc_auto_cfg))]
+
+pub struct OptionalPeripherals {}
+pub struct SWI {}
+
+pub fn init() -> OptionalPeripherals {
+    OptionalPeripherals {}
+}
```
ROMemories: I would prefer we kept the same order as in the dummy module, that makes comparing them easier:

```suggestion

pub fn init() -> OptionalPeripherals {
    OptionalPeripherals {}
}

pub struct SWI {}
```

src/ariel-os-power/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 //! Provides power management functionality.
 
 #![deny(missing_docs)]
-#![no_std]
+#![cfg_attr(not(context = "native"), no_std)]
```
ROMemories, line 4: I think this crate should get an `std` Cargo feature and use it here.

src/ariel-os-threads/src/arch/native.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl ThreadData {
+    pub const fn new() -> Self {
+        Self { thread: None }
+    }
+}
+
+impl ThreadData {
+    thread_local! {
+        pub static ID: Cell<Option<ThreadId>> = Cell::new(None);
```
ROMemories: Seems like `thread_local!` supports the following:

```suggestion
        pub static ID: Cell<Option<ThreadId>> = const { Cell::new(None) };
```

src/ariel-os-threads/src/arch/native.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    fn schedule() {}
+
+    fn wfi() {
+        unimplemented!()
+    }
+
+    fn set_running(thread_id: ThreadId) {
+        let n = usize::from(thread_id);
+        if THREAD_BLOCKS[n].swap(1, Ordering::Acquire) == 0 {
```
ROMemories: Seems like both `0` and `1` are magic values of `THREAD_BLOCKS`; they would benefit from having constants defined for them (especially as these specific integers may well not be magic in other contexts).

src/ariel-os-threads/src/arch/native.rs
```diff
@@ -0,0 +1,111 @@
...
+                    //
+                    // Whether this results in a function pointer that is safe to call:
+                    //
+                    // If `arg.is_some()`, `func` has been transmuted from `fn(T: Arguable + Send)`,
+                    // here we transmute it again into a function pointer with one `usize` argument.
+                    //
+                    // According to
+                    // https://doc.rust-lang.org/stable/std/primitive.fn.html#abi-compatibility,
+                    // this "function call behaves as if every argument was transmuted from the
+                    // type in the function pointer to the type at the function declaration".
```
ROMemories, line 58: > **If the declared signature and the signature of the function pointer are ABI-compatible**, then the function call behaves as if every argument was [transmuted](https://doc.rust-lang.org/stable/std/mem/fn.transmute.html) from the type in the function pointer to the type at the function declaration, and the return value is [transmuted](https://doc.rust-lang.org/stable/std/mem/fn.transmute.html) from the type in the declaration to the type in the pointer.

Also:

> For two signatures to be considered ABI-compatible, they must use a compatible ABI string, must take the same number of arguments […]

`fn()` and `fn(T)` are of course not ABI-compatible

src/ariel-os-threads/src/arch/native.rs
```diff
@@ -0,0 +1,111 @@
...
+    const DEFAULT_THREAD_DATA: Self::ThreadData = ThreadData::new();
+
+    fn setup_stack(thread: &mut Thread, _stack: &mut [u8], func: fn(), arg: Option<usize>) {
+        let thread_id = thread.tid;
+
+        let handle = std::thread::spawn(move || {
+            ThreadData::ID.with(|x| x.set(Some(thread_id)));
+            atomic_wait::wait(&THREAD_BLOCKS[usize::from(thread_id)], 0);
+
+            let res = std::panic::catch_unwind(|| {
```
ROMemories, line 45: What's the purpose of this `catch_unwind` here?

src/ariel-os-threads/src/arch/native.rs
```diff
@@ -0,0 +1,111 @@
...
+            });
+        });
+
+        thread.data.thread = Some(handle.thread().clone());
+    }
+
+    fn start_threading() {
+        loop {
+            SCHEDULER.with(|scheduler| {
+                for (n, thread) in scheduler.threads.iter().enumerate() {
```
ROMemories, line 85: optional: I think you could `zip()` `THREAD_BLOCKS.iter()` to avoid indexing.

src/ariel-os-threads/src/lib.rs
```diff
@@ -201,10 +218,11 @@ impl Scheduler {
         _core_affinity: Option<CoreAffinity>,
     ) -> Option<ThreadId> {
         let (thread, tid) = self.get_unused()?;
-        Cpu::setup_stack(thread, stack, func, arg);
         thread.prio = prio;
         thread.tid = tid;
         thread.state = ThreadState::Parked;
+        Cpu::setup_stack(thread, stack, func, arg);
```
ROMemories, line 227: Is this a bug fix or is the difference only visible for native?

src/ariel-os-threads/src/arch/native/critical_section.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+}
+
+critical_section::set_impl!(ArielCriticalSection);
+
+struct ArielCriticalSection;
+unsafe impl critical_section::Impl for ArielCriticalSection {
+    unsafe fn acquire() -> bool {
+        unsafe { StdCriticalSection::acquire() }
```
ROMemories, line 44: This `unsafe` block should have a `SAFETY` comment.

src/ariel-os-threads/src/arch/native/critical_section.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    unsafe fn release(nested_cs: bool) {
+        if !nested_cs {
+            let thread_id = crate::ThreadData::ID.get();
+            unsafe { StdCriticalSection::release(false) }
```
ROMemories, line 54: This `unsafe` block should have a `SAFETY` comment.

src/ariel-os-threads/src/arch/native/critical_section.rs
```diff
@@ -0,0 +1,110 @@
+//! critical_section implementation for std & infinicore
+//!
+//! In Ariel OS, most IPC code expects a context switch that is triggered within a
+//! critical section to be deferred until the critical section ends, and then run
+//! in an ISR.
+//! This means, a thread might get scheduled away right after a critical section,
```
ROMemories:
```suggestion
//! This means a thread might get scheduled away right after a critical section,
```

src/ariel-os-threads/src/arch/native/critical_section.rs
```diff
@@ -0,0 +1,110 @@
...
+//! maybe because it is not runnable anymore (but waiting for a lock, message or
+//! flag).
+//!
+//! On native with each Ariel thread running in its own host kernel thread, we
+//! need to emulate this behavior. So we wrap the regular `StdCriticalSection`.
+//! We keep an `AtomicLock` per thread. Instead of the regular runqueue for keeping
+//! track of runnable threads, the scheduler locks/unlocks those `AtomicLock`s
+//! to specify whether a thread is runnable, and the wrapped critical section, on `release()`
```
ROMemories: I'm having trouble parsing this line.

src/ariel-os-native/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "ariel-os-native"
+version = "0.1.0"
```
ROMemories: Seems like it would be easier if we kept it in sync with the other crates:

```suggestion
version = "0.2.0"
```

src/ariel-os-power/src/lib.rs
```diff
@@ -16,6 +16,8 @@ pub fn reboot() -> ! {
             esp_hal::reset::software_reset();
             #[expect(clippy::empty_loop, reason = "software_reset() should eventually never return")]
             loop {}
+        } else if #[cfg(context = "native")] {
+            std::process::exit(0)
```
ROMemories, line 20: This doesn't actually reboot; maybe rephrase the documentation in a way that allows that behavior, without actually referring to native (as I don't we want to document it just yet?) and without making that a breaking change?

ariel-os-cargo.toml
kaspar030: yes, and that patch has been merged upstream

src/ariel-os-threads/src/arch/native.rs
ROMemories: After some OoB discussions and having another read at [the docs](https://doc.rust-lang.org/stable/std/primitive.fn.html#abi-compatibility), it seems that the ABI-compatibility requirements for function pointers and function signatures are actually different, and that `fn()` and `fn(T)` (as function pointers) are actually ABI-compatible.

laze-project.yml
kaspar030: because we need this to extrapolate to `CARGO_RUNNER="sh -c"` (with quotes). Otherwise the command line becomes `CARGO_RUNNER=sh -c cargo ...`.
using single and double quotes so the double quotes don't get eaten by yaml.

src/ariel-os-power/src/lib.rs
kaspar030: I'm somewhat reluctant here. The file uses contexts to distinguish between targets later on.

src/ariel-os-threads/src/arch/native.rs
kaspar030: I'll add a comment.

`catch_unwind` catches a panicking thread and allows us to handle that. Without that, threads would panic but Ariel would happily continue (with that thread in some weird state). So to match other platforms' behavoir, abort on any thread panic.

src/ariel-os-threads/src/lib.rs
kaspar030: native's `setup_stack()` needs the `tid` field populated. I'll add a comment.

src/ariel-os-threads/src/arch/native/critical_section.rs
kaspar030: reworded

src/ariel-os-debug/src/lib.rs
kaspar030: both merged in #1152

src/ariel-os-power/src/lib.rs
kaspar030: Alternatively, print sth like "native: reboot called, exiting..." ?
IMO there is no concept of "exiting" on MCUs.
And I checked, there seems to be no portable way of re-execing a process, so no way of mimicking a reboot by the binary itself.
Another option would be to exit with a specific exit code and re-start the binary from the outside, maybe using a runner that is smarter than "sh -c".

src/ariel-os-threads/src/arch/native.rs
kaspar030: done

src/ariel-os-threads/src/arch/native.rs
kaspar030: I have renamed the array to `THREAD_RUNNABLE`. I'd rather not do defines like `const THREAD_IS_RUNNABLE: usize = 1` and `const THREAD_IS_NOT_RUNNABLE: usize = 0`, IMO that is just cruft and mental overhead.

laze-project.yml
ROMemories: So the double quotes are not for shell expansion? Usually I would prefer to stick to single quotes then (so in this case that would be using the double the double quotes for YAML, and single quotes for the shell); but fine to keep them here.

src/ariel-os-power/src/lib.rs
ROMemories: It was a bit surprising at first not getting the same debug output, but I get what you say and we can reconsider later.

ROMemories (approved):
Successfully re-did the testing of my previous review, and the `tests/threading-lock` test is now disabled.

There is quite a few dead code warnings when compiling the examples, but I would rather we address them in a follow-up.

PR #654, opened 2025-01-08
refactor(threading)!: use "TID" instead of "PID" for threads

ROMemories:
# Description

PIDs are for processes, which of cours we don't have, so this PR replaces occurrences of "pid" with "tid" (thread ID).
The only remaining occurrences are in test files where "PID" comes from an external library, and from a configuration file which specifies an external symbol.

## Change checklist


kaspar030 (approved):
lgtm

PR #657, opened 2025-01-09
refactor(debug-logging)!: rename debug logging laze modules

ROMemories:
# Description

Only `debug-logging-facade` is intended to be used by users.

## Change checklist


kaspar030:
Hm, the point of the "-default" module is that we have control over the selection of the implementer of "debug-logging-facade", which would otherwise be definition order dependent.

PR #661, opened 2025-01-10
refactor(threading)!: rename `*_NUMOF` to `*_COUNT`

ROMemories:
# Description

This uses a convention that's more commonly used and matches English syntax more closely.
There is no remaining occurrences for `numof` when grepped for case insensitively.

## Change checklist


kaspar030:
If we touch this, can we express that this is the maximum and not the current count?

e.g., `THREAD_COUNT_MAX` or similar?

PR #664, opened 2025-01-10
fix(thread): seal `Arguable` so users cannot impl it

ROMemories:
# Description

Otherwise users would be able to pass any type they want to functions that create threads (e.g., `thread_create`), which we cannot guarantee to be sound.

## Change checklist

### DIFF
--- a/src/ariel-os-threads/src/lib.rs
+++ b/src/ariel-os-threads/src/lib.rs
@@ -522,11 +522,16 @@ pub unsafe fn start_threading() {
     Cpu::start_threading();
 }
 
+trait Sealed {}
+
 /// Trait for types that fit into a single register.
-pub trait Arguable {
+#[expect(private_bounds, reason = "sealed trait")]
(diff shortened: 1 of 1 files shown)


kaspar030:
Offline discussion: let's not seal this but make the trait unsafe.

chrysn:
I agree that going unsafe is the right call here. Not sure the trait needs to be unsafe, I think it suffices that `into_arg` is unsafe, with text like this:

```
/// # Safety
///
/// This function (and therefore trait) must only be implemented if it is sound
/// to pass the output of this function to an `fn(T)` in the first argument position
/// on the current architecture's ABI.
```

That being said, I'm curious how the impl for `&'static T` is sound, if we neither request that T is sized (I think it'd just break if not?) nor that T is sync. (That's technically another issue, but given that this PR will describe how to satisfy the usage criteria for the trait, I think it fits).

chrysn:
Quite possibly so; implied bounds always confuse me.

PR #676, opened 2025-01-15
 docs(storage): Clean up the storage example

bergzand:
# Description

The storage example pr for #410

This:
  - cleans up and adds extra info prints to the example.
  - Adds the example to the index
  - Extends the readme with extra info

## Issues/PRs references

Part of #410

## Change checklist

### DIFF
--- a/examples/storage/README.md
+++ b/examples/storage/README.md
@@ -2,7 +2,17 @@
 
 ## About
 
-This application demonstrates use of the ariel-os-storage API.
+This application demonstrates use of the
+[ariel-os-storage](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/storage/index.html)
+API.
+
+The example writes a number of key-value pairs to storage and retrieves them.
+Different types of values are written to storage to demonstrate the capabilities.
+
+Values are also retrieved with different types as they were stored with,
+to show how values are decoded when different types are used
+between `insert` and `get`.
+While doing so won’t cause unsafety, it might return garbage data, or panic.
 
 Note: The application is not stateless, as it writes to flash.
 
@@ -13,3 +23,45 @@ In this directory, run
     laze build -b nrf52840dk run
 
 The application will store and read back a value to the flash.
+
+When the maximum number of cycles has been reached
+and the application exits early to avoid wearing out the flash,
+the `flash-erase-all` laze task can be used to completely erase
+the flash memory:
+
+    laze build -b nrf52840dk flash-erase-all
+
+## Expected Output
+
+When run, this example shows the following output:
+
+    INFO  storage: using flash range 0xe000..0x10000
+    INFO  storage: initializing
+    INFO  Start storage example
+    INFO  no counter value in storage. Is this the first time running this example?
+    INFO  
+    INFO  Old 'another_counter' value at 0
+    INFO  
+    INFO  Don't try this in your code!
+    INFO  Storing "string_key": "string_value" into storage
+    INFO  got heapless string value: "string_value"
+    INFO  Attempting to retrieve string value as ArrayString: [73, 74, 72, 69, 6e, 67, 5f, 76, 61, 6c, 75, 65]
+    INFO  
+    INFO  Storing cfg object MyConfig { val_one: "some value", val_two: 99 } as struct
+    INFO  got cfg object: MyConfig { val_one: "some value", val_two: 99 }
+    INFO  Attempting to retrieve cfg as ArrayVec: [73, 6f, 6d, 65, 20, 76, 61, 6c, 75, 65]
+    INFO  Attempting to retrieve cfg as array: [0a, 73, 6f, 6d, 65, 20, 76, 61, 6c, 75]
+    INFO  
+    INFO  Storing raw bytes [00, 01, 02, 03, 04]
+    INFO  got bytes as array: [00, 01, 02, 03, 04]
+    INFO  Attempting to retrieve bytes as heapless vec arr: []
+    INFO  
+    INFO  Exit storage example
+
+When the initial counter reached the maximum number of executions, the following
+output is produced:
+
+    INFO  storage: using flash range 0xe000..0x10000
+    INFO  Start storage example
+    INFO  got counter value 11 from storage
+    INFO  counter value > 10, aborting test to save flash cycles
--- a/examples/storage/src/main.rs
+++ b/examples/storage/src/main.rs
@@ -13,103 +13,125 @@ use ariel_os::debug::{
 use ariel_os::storage;
 use serde::{Deserialize, Serialize};
 
-// Some object. For storing it, derive the serde Serialize / Deserialize traits.
+/// Example object.
+///
+/// The serde Serialize / Deserialize traits are required for storage
 #[derive(Serialize, Deserialize, Debug, defmt::Format)]
 struct MyConfig {
-    val_uno: heapless::String<64>,
-    val_dos: u64,
+    val_one: heapless::String<64>,
+    val_two: u64,
 }
 
 #[ariel_os::task(autostart)]
 async fn main() {
-    info!("Hello from storage test!");
+    info!("Start storage example");
 
     // Storing a primitive type (e.g., u32)
     let value: Option<u32> = storage::get("counter").await.unwrap();
     let value = if let Some(value) = value {
-        info!("got counter value {}", value);
+        info!("got counter value {} from storage", value);
         value
     } else {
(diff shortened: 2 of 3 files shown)


examples/storage/README.md
```diff
@@ -2,7 +2,14 @@
...
 
-This application demonstrates use of the ariel-os-storage API.
+This application demonstrates use of the
+[ariel-os-storage](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/storage/index.html)
+API.
+
+The example writes a number of key-value pairs to storage and retrieves them.
+Different types of values are written to storage to demonstrate the capablities.
+Values are also retrieved with different types as they were stored with, to show
+how values are decoded when different types are used between `insert` and `get`.
```
chrysn, line 12: Should this contain a note to Not Do This At Home?

examples/storage/README.md
```diff
@@ -13,3 +20,42 @@ In this folder, run
...
+    INFO  Storing cfg object MyConfig { val_one: "some value", val_two: 99 } as struct
+    INFO  got cfg object: MyConfig { val_one: "some value", val_two: 99 }
+    INFO  Attempting to retrieve cfg as ArrayVec: [73, 6f, 6d, 65, 20, 76, 61, 6c, 75, 65]
+    INFO  Attempting to retrieve cfg as array: [a, 73, 6f, 6d, 65, 20, 76, 61, 6c, 75]
+    INFO  
+    INFO  Storing raw bytes [0, 1, 2, 3, 4]
+    INFO  got bytes as array: [0, 1, 2, 3, 4]
+    INFO  Attempting to retrieve bytes as heapless vec arr: []
+    INFO  
+    INFO  Exit storage example
```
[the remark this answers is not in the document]
bergzand: My main goal here is to set expectations for the user of these examples. Help the user to determine whether the output is what is expected. For example, I asked @kaspar030 about the heapless vec arr output because it looked incorrect to me, I didn't realize that was intentional.

examples/storage/src/main.rs
```diff
@@ -13,91 +13,100 @@ use ariel_os::debug::{
...
     storage::insert("counter", value + 1).await.unwrap();
 
+    // By getting the Storage mutex directly, changing e.g., a counter,
+    // can be done atomically w.r.t. concurrent access from the same firmware:
     {
-        // By getting the Storage mutex directly, changing e.g., a counter,
-        // can be done atomically w.r.t. concurrent access from the same firmware:
         let mut s = storage::lock().await;
-        let value: Option<u32> = s.get("counter").await.unwrap();
+        let value: Option<u32> = s.get("atomic_counter").await.unwrap();
```
chrysn, line 50: The comment explains what is meant with atomicity; yet maybe calling it atomic counter is maybe not ideal, as there are other aspects to atomicity.

Or: if the counter is atomic in the sense that after insert completes, we are guaranteed not to read the old value any more at the next reboot, maybe we could comment on that and then it's atomic for all senses I could come up with.

examples/storage/src/main.rs
```diff
@@ -13,91 +13,100 @@ use ariel_os::debug::{
...
     // Getting an object
     // Type used for `get()` needs to match what was used for `insert()`.
     let cfg: Option<MyConfig> = storage::get("my_config").await.unwrap();
     if let Some(cfg) = cfg {
-        info!("got cfg {:?}", cfg);
+        info!("got cfg object: {:?}", cfg);
     }
 
     // Getting a value as raw bytes probably does not return what you want due
     // to the way postcard works
```
chrysn, line 94: When and how *does* it make sense to use that?

bergzand: cc @kaspar030

examples/storage/README.md
```diff
@@ -13,3 +20,42 @@ In this folder, run
...
+    INFO  Old atomic counter value at 0
+    INFO  
+    INFO  Storing "string_key": "string_value" into storage
+    INFO  got heapless string value: "string_value"
+    INFO  Attempting to retrieve string value as ArrayString: [73, 74, 72, 69, 6e, 67, 5f, 76, 61, 6c, 75, 65]
+    INFO  
+    INFO  Storing cfg object MyConfig { val_one: "some value", val_two: 99 } as struct
+    INFO  got cfg object: MyConfig { val_one: "some value", val_two: 99 }
+    INFO  Attempting to retrieve cfg as ArrayVec: [73, 6f, 6d, 65, 20, 76, 61, 6c, 75, 65]
+    INFO  Attempting to retrieve cfg as array: [a, 73, 6f, 6d, 65, 20, 76, 61, 6c, 75]
```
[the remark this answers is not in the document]
bergzand: Sounds good. Do you also have a preference whether they should be prefixed with '0x'?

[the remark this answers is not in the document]
kaspar030: +1 for no 0x

examples/storage/src/main.rs
kaspar030: hm. I used that as a test, and kept it in to showcase what can go wrong.
Maybe the whole section should be prefixed with `info!("See what happens, Larry:");`, and explain why and how the deserializing goes wrong the way it does. Maybe, keep only one of the error cases. Obviously, code comments don't cut it.

examples/storage/src/main.rs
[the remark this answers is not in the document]
kaspar030: (update below also)

examples/storage/README.md
bergzand: Reiterated the message from the storage module docs

examples/storage/src/main.rs
```diff
@@ -13,91 +13,100 @@ use ariel_os::debug::{
 use ariel_os::storage;
 use serde::{Deserialize, Serialize};
 
-// Some object. For storing it, derive the serde Serialize / Deserialize traits.
+// Example object.
+// The serde Serialize / Deserialize traits are required for storage
```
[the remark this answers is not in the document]
bergzand: Done!

examples/storage/README.md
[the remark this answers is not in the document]
bergzand: Added the leading zero in the fmt.

examples/storage/src/main.rs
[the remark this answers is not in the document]
bergzand: Renamed!

examples/storage/src/main.rs
```diff
@@ -13,103 +13,115 @@ use ariel_os::debug::{
...
         let value = value.unwrap_or_default();
-        s.insert("counter", value + 1).await.unwrap();
+        s.insert("another_counter", value + 1).await.unwrap();
+        info!("Old 'another_counter' value at {}", value);
     }
+    info!("");
 
     // Storing a string value
     // For insertion, a literal can be used.
+    info!("Storing \"string_key\": \"string_value\" into storage");
```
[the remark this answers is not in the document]
kaspar030: I actually prefer the escaped version. too much C in my life probably.

examples/storage/src/main.rs
kaspar030: maybe this just needs an `info!("Don't try this at home:");` ?

kaspar030:
where are we here? I'm fine as is, maybe add another message before the not-recommended part, as suggested above.

examples/storage/src/main.rs
chrysn: "Don't try this at home" catches the spirit well.

chrysn: Maybe also "Note that this may break at any time, in particular when we start sorting that hash"

bergzand: Added a `info!` note

chrysn: That's added way up when there is the first mismatch -- is everything below "Don't try this in your code!" playing fast and lose?

chrysn: Nonetheless, it's an improvement, and it *does* point out the necessary things. Let's go and improve gradually (or skip it because the don't-try-at-home parts break once postcard hashes are in).

chrysn (approved):

examples/storage/README.md
```diff
@@ -13,3 +23,45 @@ In this folder, run
     laze build -b nrf52840dk run
```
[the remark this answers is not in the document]
bergzand: With the extra fixup in here to revert 494dfc1a9562a631aa85136bd5dd08cc61d1a2d3 for this example, the CI indeed fails: https://github.com/ariel-os/ariel-os/actions/runs/13049320147/job/36405848854#step:5:6

Issue #679, opened 2025-01-16
Coding convention: Log levels


chrysn:
### Description

There's a lot of things that parts of Ariel may or may not want to log, and different opinions about what level to log on.

Romain found the [Swift guidelines](https://www.swift.org/documentation/server/guides/libraries/log-levels.html) helpful, which encourates to error! sparringly, use debug! a lot, and no info! in libraries. I'd personally use more info! to help users get unstuck without rebuilding.

### Motivation

coapcore produces errors over the network that follow the spec but are not too helpful in finding out *why* something did not work.

### Current Implementation

None.

Some discussion around https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$FVGlZBQnk5ROOHcNLc4QxrDhPSbzCCepFIv5iSjPswo?via=matrix.org&via=utwente.io&via=club1.fr

coapcore [states](https://ariel-os.github.io/ariel-os/dev/docs/api/coapcore/index.html) that it may log secrets at exotic levels, which we should do away with ASAP.

### Are you planning to do it yourself in a pull request?

No

chrysn:
I'm still not quite convinced that errors should only be setup errors, and not caused by clients.

One aspect I'd like to bring up is visual usability: Suppose we limit our library components to trace and debug (unless something is really an internal error), and during development, a client sends "wrong" credentials.

The developer sees that something negative comes back, sees nothing in the logs, and ramps up debug levels. But as soon as debug is on, there's a lot of output, from which nothing stands out; they'll have to read line by line until they find what is wrong.

By comparison, if the final point where things fail (essentially when the error response is rendered on the network, and the cause is printed) is sent at the "error" level, that stands out in red, and is easy to find.

I think we should rather silence modules (maybe "just show critical") than turn all error details gray.

PR #691, opened 2025-01-17
fix(ci): remove dependencies-action

chrysn:
# Description

The dependencies action performs natural language processing on PR texts, but does it so badly that we have to word around what the NLP may think, necessitating at least one of pseudocensorship, doublethink or just being silent about things that would be useful to say. Neither is productive.

Let's reintroduce this when GitHub has proper means of declaring issue and PR dependencies; seems they're working on it in beta: https://github.com/orgs/community/discussions/4928.

## Issues/PRs references

In #690, I had to write

> ## Issues/PRs references
>
> This may need updating when https://github.com/ariel-os/ariel-os/issues/679 concludes.
>
> Comments already indicate how to improve it once https://github.com/knurling-rs/defmt/pull/916 lands (but that is currently bloc_k__ed by https://github.com/knurling-rs/defmt/pull/922)

because had I used the term "blocked by", it would have thought #922 is what it is blocked by.

## Open Questions

* Are there softer means to do this?


kaspar030:
Sorry, I depend on this a lot. Can't merge without providing an alteenative. Please reopen once we got one!

PR #693, opened 2025-01-18
feat(coap): add ACE EDHOC profile

chrysn:
# Description

The ACE EDHOC profile is a useful addition to coapcore as it allows having EDHOC (and thus forward secrecy) with ACE tokens.

<del>This first iteration of it uses signed CWTs instead of symmetrically encrypted CWTs – these may not be the logical first step building things up (as they are mainly useful thinking of group audiences), but it is what I can test first.</del>

[edit:] It supports both signed and encrypted CWTs. (Things Just Worked once the other case was added 🥰).

## Issues/PRs references

This builds on #690.

* This d_epends on https://github.com/twittner/minicbor/pull/9 -- consequently it adds a patch.crates-io. Let's see how responsive upstream is; the alternative is a rather awkward manual parsing of the map.
* Practical use d_epends on https://github.com/openwsn-berkeley/lakers/pull/331 (the old `quadruple_sizes` leads to kilobytes of stack usage); they might release based on that soon (either in the 0.7 series, or as a 0.8 with some breaking changes they already accumulated on main, but for which I already have the optional adjustment at hand).

[edit:] CI checks misinterpret what this means (see https://github.com/ariel-os/ariel-os/pull/691 for remedy); we can (and this branch does) work from branches, but it would be nice to get those merged at some point.

## Open Questions

* <del>As part of the patch series I'll need to overhaul the ScopeGenerator a bit, because when accepting a signed token, one has to check the audience (which is pretty much a no-op on symmetrically encrypted tokens, as their key can only be used by the AS for a single RS).</del>
  * ScopeGenerator was removed; the decrypt/verify functions now do a bit more than I'd ideally want to have in a single function, but document why that is, and a better version of ScopeGenerator (then probably: "opaque data carried between this and that method") can be reintroduced when it is better understood in the general case.

## Change checklist

### DIFF
--- a/src/lib/coapcore/Cargo.toml
+++ b/src/lib/coapcore/Cargo.toml
@@ -29,8 +29,14 @@ lakers-crypto-rustcrypto = "0.7.2"
 liboscore = "0.2.2"
 liboscore-msgbackend = "0.2.2"
 
+# actually we depend on <https://github.com/twittner/minicbor/pull/9> so it
+# should be
+#     minicbor = { git = "https://github.com/chrysn-pull-requests/minicbor", branch = "negativ-indices", features = ["derive"] }
+# and it could be patched locally because it is a non-pub dependency, but as
+# minicbor-adapters needs to use the same types, this still has to go into
+# patch.crates-io to also catch minicbor-adapters.
 minicbor = { version = "0.25.1", features = ["derive"] }
-minicbor-adapters = "0.0.1"
+minicbor-adapters = "0.0.3"
 heapless = "0.8.0"
 defmt-or-log = { version = "0.2.1", default-features = false }
 defmt = { workspace = true, optional = true }
@@ -43,6 +49,8 @@ document-features = "0.2.10"
 # dependencies.
 ccm = { version = "0.5.0", default-features = false }
 aes = { version = "0.8.4", default-features = false }
+
+p256 = { version = "0.13.2", features = ["ecdsa"], default-features = false }
 
 [features]
 #! # Cargo features
--- a/src/lib/coapcore/src/ace.rs
+++ b/src/lib/coapcore/src/ace.rs
@@ -367,6 +367,7 @@ pub fn process_acecbor_authz_info<Scope>(
         return Err(CredentialErrorDetail::UnsupportedAlgorithm.into());
     }
 
+    // FIXME: Duplicated with process_edhoc_token
     #[derive(minicbor::Encode)]
     struct Encrypt0<'a> {
         #[n(0)]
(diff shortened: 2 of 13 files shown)


kaspar030:
Nice!

chrysn:
I'm sending the commits through CI piecemeal as I did too many fixups to be sure all is fine:
* 8aafae44646925ec84b230f9047c0362004d9bde passed everything relevant (the spell check lint complains about a branch name)
* d072e2c4b3759af7bd7ac9169a3d3a1356b1540c errs out earlier, but all errors are just about unused fields (and they get used next); the build steps do pass.

Anything atop of that is failing so far, being investigated.

src/lib/coapcore/Cargo.toml
```diff
@@ -29,8 +29,15 @@ lakers-crypto-rustcrypto = "0.7.2"
...
 
+# actually we depend on <https://github.com/twittner/minicbor/pull/9> so it
+# should be
+#     minicbor = { git = "https://github.com/chrysn-pull-requests/minicbor", branch = "negativ-indices", features = ["derive"] }
+# and it could be patched locally because it is a non-pub dependency, but as
+# minicbor-adapters needs to use the same types, this still has to go into
+# patch.crates-io to also catch minicbor-adapters.
 minicbor = { version = "0.25.1", features = ["derive"] }
-minicbor-adapters = "0.0.1"
+
```
chrysn, line 39:
```suggestion
```

src/lib/coapcore/src/ace.rs
```diff
@@ -67,14 +69,45 @@ pub struct HeaderMap<'a> {
 
 impl HeaderMap<'_> {
     /// Merge two header maps, using the latter's value in case of conflict.
-    fn updated_with(self, other: Self) -> Self {
+    fn updated_with(&self, other: Self) -> Self {
```
chrysn, line 72: It's just an module-internal function; its users changed, and now this is more convenient.

src/lib/coapcore/Cargo.toml
```diff
@@ -44,6 +51,8 @@ document-features = "0.2.10"
 ccm = { version = "0.5.0", default-features = false }
 aes = { version = "0.8.4", default-features = false }
 
+p256 = { version = "0.13.2", features = ["ecdsa"], default-features = false }
```
chrysn, line 54: This should only get pulled in with code when there is a token assigned; I should verify that.

src/lib/coapcore/src/ace.rs
```diff
@@ -89,39 +122,101 @@ struct CoseEncrypt0<'a> {
...
+    ) -> Result<(HeaderMap<'_>, impl AsRef<[u8]>, &'t mut [u8]), CredentialError> {
+        trace!("Preparing decryption of {:?}", self);
+
+        // Could have the extra exception for empty byte strings expressing the empty map, but we don't
+        // encounter this here
+        let protected: HeaderMap = minicbor::decode(self.protected)?;
+        trace!("Protected decoded as header map: {:?}", protected);
+        let headers = self.unprotected.updated_with(protected);
+
+        // FIXME: Duplicated with process_acecbor_authz_info
```
chrysn, line 144:
```suggestion
```

src/lib/coapcore/src/ace.rs
```diff
@@ -89,39 +122,101 @@ struct CoseEncrypt0<'a> {
...
+            protected: &'a [u8],
+            #[cbor(b(2), with = "minicbor::bytes")]
+            external_aad: &'a [u8],
+        }
+        let aad = Encrypt0 {
+            context: "Encrypt0",
+            protected: self.protected,
+            external_aad: &[],
+        };
+        // FIXME: This is overkill by far; usually we get away with like 16 byte.
```
chrysn, line 159: I should address that now already; it's just another MAX parameter.

src/lib/coapcore/src/ace.rs
```diff
@@ -397,3 +447,95 @@ pub fn process_acecbor_authz_info<Scope>(
...
+            sign1.payload,
+            crate::PrivateMethod,
+        )?;
+
+        (scope, claims)
+    } else {
+        return Err(CredentialErrorDetail::UnsupportedExtension.into());
+    };
+
+    // FIXME: Check iat/exp against raytime
```
chrysn, line 512: Addressing this is out of scope for now: We don't have trusted wall time on our system yet at all.

src/lib/coapcore/Cargo.toml
chrysn: The compiler currently does not see that -- it might (because there is only one thing that writes anything other than None into `ConfigBuilder.as_key_neg7`, and if that doesn't happen, that branch is never taken), but it seems it doesn't do that right now.

I have a pending item to make more crypto stuff optional; if it's OK with you, I'll do that there rather than starting to half-do it here.

kaspar030 (approved):

chrysn:
Thanks. Rebased to resolve the conflicts, and Cargo.lock was regenerated per-commit as changes there are generally not straightforward to merge.

PR #694, opened 2025-01-19
refactor!(net): `ariel_os_embassy::network` -> `ariel_os_embassy::net`

kaspar030:
# Description

This renames the `ariel_os_embassy::network` module to just `net`.

## Change checklist


chrysn (approved):
Straightforward enough; conveniently this is one of those "if it builds, all is fine" changes.

Should be good to merge once the last checkbox is in (I don't know if there is any documentation that mentions the module by name but is not inline docs – could just as well be ticking the box if there is nothing mentioning it).

If this is our pub API, it may we worth considering whether ariel_os_**embassy**::net is the right path there, but IIRC there's a separate issue about that.

kaspar030:
> If this is our pub API, it may we worth considering whether ariel_os_**embassy**::net is the right path there, but IIRC there's a separate issue about that.

IIUC this was already exported as `ariel_os::net` :/

kaspar030:
> Should be good to merge once the last checkbox is in (I don't know if there is any documentation that mentions the module by name but is not inline docs – could just as well be ticking the box if there is nothing mentioning it).

I grepped for more `network::`, and for `network` in `book/`, no hits. So I checked the box. :)

PR #665, opened 2025-01-10
chore(deps): bump embassy

kaspar030:
# Description

Bumbs embassy & esp-hal.

~~Doesn't build yet, esp-hal needs to update.~~ included here.

This PR also changes the embassy fork to per-crate branches, so it'll hopefully be the last "bump embassy" PR.

Todo:

- [x] wait for https://github.com/esp-rs/esp-hal/pull/2701

## Issues/PRs references

~~Depends on 555.~~ (included now)

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

PR #709, opened 2025-01-22
fix(examples): re-enable `examples/thread-async-interop` for esp

kaspar030:
# Description

## Issues/PRs references

Depends on #665.

## Change checklist


kaspar030:
> What was the reason they were disabled in the first place?

I think this is from before the thread executor. not sure. the original commit message for those lines says `examples/embassy: limit to cortex-m (interrupt executor) for now`.

PR #726, opened 2025-01-24
chore(manifests): remove version keys from tests and examples

ROMemories:
# Description

The `version` key is not a required key anymore and these crates are not intended to be published.

[Documentation of the `version` key](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field)

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,6 +57,9 @@ jobs:
       - name: Check for leftover occurrences of the `authors` key in manifests
         run: set +e; grep -r '^authors\s*=' --line-number src examples tests; test $? -eq 1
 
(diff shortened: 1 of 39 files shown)


chrysn (approved):
LGTM.

Does the absence of a version imply publish=false? If so, could clean that up in the same step.

chrysn:
I just tried:
```
mkdir versionless-crates-can-not-be-published
cd versionless-crates-can-not-be-published
cargo init
# remove version from Cargo.toml
cargo publish --dry-run
```

and it gave me:

```
error: `versionless-crates-can-not-be-published` cannot be published.
`package.publish` must be set to `true` or a non-empty list in Cargo.toml to publish.
```

PR #731, opened 2025-01-27
coap: prepare for DHCP

chrysn:
# Description

There were two places where we had IP addresses hardcoded:

* in the ariel-os-coap server (due to https://github.com/smoltcp-rs/smoltcp/issues/1037 for which there is now a workaround)
* in the credentials files (but wildcards work there as well)

## Issues/PRs references

Blocks: https://github.com/ariel-os/ariel-os/pull/715 (subtly: that PR would go through, but the examples described in the README would not work any more)

## Change checklist

  Note that no changes to the docs are required at this time; #715 will update READMEs en masse.

### DIFF
--- a/src/ariel-os-coap/src/lib.rs
+++ b/src/ariel-os-coap/src/lib.rs
@@ -79,9 +79,7 @@ pub async fn coap_run(handler: impl coap_handler::Handler + coap_handler::Report
 
     info!("Starting up CoAP server");
 
-    // Can't that even bind to the Any address??
-    // let local_any = "0.0.0.0:5683".parse().unwrap(); // shame
-    let local_any = "10.42.0.61:5683".parse().unwrap(); // shame
+    let local_any = "0.0.0.0:5683".parse().unwrap(); // shame
     let mut unconnected = udp_nal::UnconnectedUdp::bind_multiple(socket, local_any)
         .await
         .unwrap();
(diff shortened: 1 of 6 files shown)


src/ariel-os-coap/src/lib.rs
```diff
@@ -79,9 +79,7 @@ pub async fn coap_run(handler: impl coap_handler::Handler + coap_handler::Report
 
     info!("Starting up CoAP server");
 
-    // Can't that even bind to the Any address??
-    // let local_any = "0.0.0.0:5683".parse().unwrap(); // shame
-    let local_any = "10.42.0.61:5683".parse().unwrap(); // shame
+    let local_any = "0.0.0.0:5683".parse().unwrap(); // shame
```
kaspar030, line 82:
```suggestion
    let local_any = "0.0.0.0:5683".parse().unwrap();
```

kaspar030 (approved):
LGTM. mind removing the `// shame` ?

src/ariel-os-coap/src/lib.rs
chrysn: These are my reminders for every place where we have IPv4 addresses around. Happy to adjust them to something else, but we should have some reminders to grep for them once we can get rid of them.

In this concrete case, I *think* that due to how smoltcp works, we can actually switch already. Adding and testing; in the meantime: any suggestions for a better name for where there are still any around? (The CoAP client examples are probably the only ones left)

chrysn:
All addressed and re-tested.

There's an unrelated bug in the security config that keeps the server /hello from being accessible to authenticated users, but that's part of the upcoming fixes anyway.

PR #737, opened 2025-01-27
docs(ariel-os): hide from users laze-managed Cargo features

ROMemories:
# Description

This hides from user documentation the set of Cargo features that must be enabled through laze modules instead of Cargo features directly, so that users don't run into hard-to-debug issues.

## Issues/PRs references

Depends on #761.
Depends on #797.
Depends on #800.

## Change checklist


kaspar030:
Hm, down the line, we want laze to infer its modules from selected features.

kaspar030:
Hm, and the discoverability ... I wish we'd have the laze auto inference a bit closer

bergzand (approved):
Ack, lgtm

PR #743, opened 2025-01-28
ci(rustdoc): use `--no-deps` to clean up the crate list

ROMemories:
# Description

This should also make doc generation slightly faster.

## Change checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -60,6 +60,7 @@ jobs:
       - name: Build rustdoc docs
         run: |
           cargo doc \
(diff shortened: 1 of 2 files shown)


.github/workflows/build-deploy-docs.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       - name: Build rustdoc docs
         run: |
           cargo doc \
+              --no-deps \
               -p ariel-os \
```
chrysn, line 64: Are those of ariel-os indeed the only docs we want to show?

chrysn (approved):

PR #746, opened 2025-01-28
docs(rand): use `rand_core` instead of `rand` when possible

ROMemories:
# Description

Following the [policy change/clarification of `rand`](https://github.com/rust-random/rand/blob/master/CHANGELOG.md#security-and-unsafe) regarding whether it is a "crypto library", this makes sure to use `rand_core` instead of `rand` to reduce the moving pieces where possible.

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2975,7 +2975,7 @@ dependencies = [
  "ariel-os",
  "ariel-os-boards",
  "heapless 0.8.0",
- "rand",
+ "rand_core",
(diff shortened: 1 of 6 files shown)


chrysn (approved):
I'm not sure there is as much urgency to it (*we* don't give any legally-binding guarantees on anything either), but going with the core version of rand rather than the everything-and-std-if-you-enable it rand crate is certainly a good thing.

PR #752, opened 2025-01-29
feat(coapcore): evaluate tokens' lifetime

chrysn:
# Description

Issued ACE tokens often have limited lifetime; this adds the APIs to coapcore to evaluate that aspect.
As it required touching some signatures and pub'nesses, the PR also contains changes to reduce and harmonize the public surface before coapcore's API can stabilized for a 0.1 release.

## Issues/PRs references

Currently we don't have "wall time" in Ariel OS (but even if we had, we'd need it trusted).

For this first iteration, Ariel OS just uses the zero-sized "we don't know the time" configuration. Once we start using configuration storage more actively, we can progress from "we don't know anything" to "this can't be valid because we've seen a way newer token" on to [full raytime](https://datatracker.ietf.org/doc/draft-amsuess-t2trg-raytime/).

## Change checklist

### DIFF
--- a/src/ariel-os-coap/src/lib.rs
+++ b/src/ariel-os-coap/src/lib.rs
@@ -111,6 +111,7 @@ pub async fn coap_run(handler: impl coap_handler::Handler + coap_handler::Report
             .with_known_edhoc_credential(admin_key, admin_scope),
         || lakers_crypto_rustcrypto::Crypto::new(ariel_os_random::crypto_rng()),
         ariel_os_random::crypto_rng(),
+        coapcore::time::TimeUnknown,
     );
 
     info!("Server is ready.");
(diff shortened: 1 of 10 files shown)


chrysn:
This is ready for review now. I think it is best reviewed by looking at the commit messages, and then looking at the whole diff: Introducing the time was better done in an isolated step, but made the API worse, which is cleaned up later, but I couldn't easily do that cleanup first and then the extension, so there's a bit of back-and-forth (like, introducing a 4th parameter to a function, just to go back to a 3-version where the 3rd is now something that expresses the meaning of two-or-more things).

kaspar030 (approved):
Only changing coap files, nothing glaringly wrong -> trusting your testing, LGTM.

PR #753, opened 2025-01-29
fix(threading): mark the `Arguable` trait as unsafe

ROMemories:
# Description

Make the `Arguable` trait `unsafe` as incorrectly implementing its method would lead to wild UB.

## Change checklist

### DIFF
--- a/src/ariel-os-threads/src/lib.rs
+++ b/src/ariel-os-threads/src/lib.rs
@@ -527,28 +527,33 @@ pub unsafe fn start_threading() {
     Cpu::start_threading();
 }
 
-/// Trait for types that fit into a single register.
-pub trait Arguable {
-    #[doc(hidden)]
+/// Trait for types that can be used as argument for threads.
+///
+/// # Safety
+///
+/// This trait must only be implemented on types whose binary representation fits into a single
+/// general-purpose register on *all supported architectures*.
+pub unsafe trait Arguable {
+    /// Returns the ABI representation.
     fn into_arg(self) -> usize;
 }
 
-impl Arguable for usize {
+// SAFETY: this is the identity.
+unsafe impl Arguable for usize {
     fn into_arg(self) -> usize {
         self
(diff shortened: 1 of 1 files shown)


src/ariel-os-threads/src/arch/cortex_m.rs
```diff
@@ -5,6 +5,9 @@ use cortex_m::peripheral::{scb::SystemHandler, SCB};
 #[cfg(not(any(armv6m, armv7m, armv8m)))]
 compile_error!("no supported ARM variant selected");
 
+// Represents a machine word.
+pub type Uword = u32;
```
kaspar030, line 9: This is unneeded extra indirection

"The usize type is an unsigned integer type with the same number of bits as the platform’s pointer type. It can represent every memory address in the process."

A pointer fits into a usize. A usize fits into a register. Let's not overcomplicate things.

Maybe, rather change the trait docs from "Trait for types that fit into a single register." into "Trait for types that can be used as argument for threads".
Then we can hack usize-as-arg for avr8 once we want to support that.

kaspar030: > The fact that this is true on our currently supported architectures is mostly happenstance, this doesn't derive at all from the definition of `usize`.

"Trait for types that fit into a single register." is maybe off as well, `Arguable` represents types that we can nudge into thread functions as argument, by implementing the trait for the type.

*everything else is an implementation detail*. Marking the trait as unsafe should be enough to reflect this.

Adding some `UWord` newtype will not save us from a) testing all `Arguable` implementations on all platforms for ABI compatibility, and b) from changing this again, when (if) if want to support AVR with 8bit registers. So adding that type here adds to the cognitive complexity.

I'd be happy if we can improve the `Arguable` docs, and mark that unsafe, in this PR, and leave out the `UWord`.

kaspar030 (approved):

Issue #674, opened 2025-01-14
coap: Model dependencies in laze


chrysn:
### Description

From https://github.com/ariel-os/ariel-os/pull/585#discussion_r1913148767:

coap depends on random, and also on network, and selects udp … all things that the current examples and tests do manually.

### Motivation

Not just the code, also the setup should be just the minimal lines.

### Current Implementation

In examples / tests: Manual activation from Cargo.toml, and selection laze.yml.

### Are you planning to do it yourself in a pull request?

Yes

PR #755, opened 2025-01-29
feat(coap): model in laze and autostart

chrysn:
# Description

This is a first step for modelling ariel-os's `coap` feature in laze. Reasons:

* Examples currently need to manually select random and network (yuck)
* We'll want to do more later (eg. run an CoAP "main loop" automatically if client and not server is selected, making the coap-client easier)

## Open Questions

* Will I go beyond what is in the first few commits in this PR, or in a separate one? (Next steps are auto-starting the server and setting server policy and even identity through this.)

  * A: Yes to auto-starting the server (because that corresponds to the initial comments around the laze entries), no to more (because some of that may need discussion).

    **Remaining things to do**: model server policy (even if it's just a "pick the demo stuff"), provide public function that people should build their request handler on (even if it's just never_found() right now).

## Issue relations

Closes: https://github.com/ariel-os/ariel-os/issues/674

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -760,6 +760,33 @@ modules:
         FEATURES:
           - ariel-os/hwrng
 
+  - name: coap
+    help: Basic support for the CoAP protocol.
+
+      On its own, this does nothing but load code; this should be auto-selected
+      by coap-client and/or coap-server as per the needs of other modules and
+      the application.
+    depends:
+      - random
+      - network
+
+  - name: coap-server
+    help: Support for applications to set up CoAP server handlers.
+
+      When an application selects this, it needs to run `ariel_os::coap_run()`
+      in a task; otherwise, other components (eg. system components that also
+      run on the CoAP server, or the CoAP client that depends on the server
+      loop to run) get stalled.
+    # Modules that want to inject a CoAP server will *not* depend on this, and
+    # will have their own entry point when they exist.
+    depends:
+      - coap
+
+  - name: coap-client
+    help: Support for CoAP client functionality.
+    depends:
+      - coap
+
   - name: random
     help: A system-wide RNG is available (through the ariel_os::random module).
 
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -785,6 +785,10 @@ modules:
     # will have their own entry point when they exist.
     depends:
       - coap
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap-server
 
   - name: coap-client
     help: Support for CoAP client functionality.
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -769,6 +769,10 @@ modules:
     depends:
       - random
       - network
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap
 
   - name: coap-server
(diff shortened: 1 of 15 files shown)


laze-project.yml
```diff
@@ -760,6 +760,41 @@ modules:
...
+    depends:
+      - random
+      - network
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap
+
+  - name: coap-server
+    help: Support for applications to set up own CoAP server handlers.
```
kaspar030, line 778:
```suggestion
    help: Support for applications to set up CoAP server handlers.
```

src/ariel-os/Cargo.toml
```diff
@@ -59,6 +59,13 @@ dns = ["ariel-os-embassy/dns"]
 mdns = ["ariel-os-embassy/mdns"]
 ## Enables support for [CoAP](https://ariel-os.github.io/ariel-os/dev/docs/book/tooling/coap.html).
 coap = ["dep:ariel-os-coap", "random"]
+## Support for applications to set up own CoAP server handlers.
```
kaspar030, line 62:
```suggestion
## Support for applications to set up CoAP server handlers.
```

src/ariel-os-coap/Cargo.toml
```diff
@@ -43,9 +44,19 @@ embedded-io-async = "0.6.1"
 workspace = true
 
 [features]
+# Enables the coap_run function through which an own server is passed in. The
```
kaspar030, line 47:
```suggestion
# Enables the coap_run function through which a server is passed in. The
```

kaspar030:
Looks good to me.

chrysn:
Some [background](https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$pjQcmQMXJ_NGV_s6ITvogaiikPvOqhwBRBx5W8OuMso?via=matrix.org&via=utwente.io&via=club1.fr) for documentation on why the switch from OnceLock to Waker happened:

> Now that's funny: apparently, https://github.com/ariel-os/ariel-os/pull/755/commits/f31c0e997f9b33a3b691c48c27afb4731fc02e0c [the one where the CoAP server gets a startup delay] breaks the network to the point where the USB network interface doesn't even show up on the host.
>
> The startup of CoAP may have depended all the time on the sequence between the coap_run and the coap_client obtaining, and if that was changed by autostarting...
>
> it is blocking in the main task
>
> [the client example] does behave like a `core::future::poll_fn(|cx| { cx.waker().wake_by_ref(); core::task::Poll::Pending }).await`
>
> … which is not completely unexpected, given I modelled the CoAP stack global after the network stack global, which is built on a OnceCellOnceLock, which is described to "busy loop" before it is initialized.

There's the one caveat that the "Too many CoAP clients are waiting for the network to come up." error will come when there is more than one client waiting during startup.

That's an enhancement over "the clients can't wait at all and race the system into a deadlock if they do", so we go with it, but I'd like to explore better options that are not bounded.

chrysn:
Ran all tests again.

Thanks for the review, merging.

PR #760, opened 2025-01-30
docs(getting-started): reduce the set of build dependencies

ROMemories:
# Description

This asks new users to install fewer build dependencies.

The set of required build dependencies can be verified using the following `Containerfile`:

```containerfile
# Build with:
# podman build -f Containerfile
# If the image builds successfully it means that the build dependencies
# specified are sufficient.

# Use a very minimal image to make as few assumptions as possible about users'
# environment.
# NOTE: the testing branch is used for now as rustup is not yet on stable.
FROM docker.io/debian:testing-slim

ENV PATH="$PATH:/root/bin"

RUN apt-get update \
  && apt-get install -y \
    git \
    rustup \
    ninja-build \
    pkg-config \
    libudev-dev \
    clang \
    gcc-arm-none-eabi \
  && rm -rf /var/lib/apt/lists/*

RUN rustup install stable
# Requires a C compiler at build time and ninja-build at runtime
RUN cargo install --root=/root laze
# Requires libudev at build time
RUN cargo install --root=/root --locked probe-rs-tools

RUN git clone https://github.com/ariel-os/ariel-os.git \
  && cd ariel-os \
  && laze build install-toolchain

# CoAP requires gcc-arm-none-eabi at build time
RUN cd ariel-os \
  && laze -C examples/coap-server build -b nrf52840dk
```

(This takes about 6 minutes to build on my machine, YMMV.)

This `Containerfile` could be used in CI later or for quite a few other things, but the goal of this PR is really just to reduce the number of dependencies users need to install to start.

## Change checklist

### DIFF
--- a/book/src/getting_started.md
+++ b/book/src/getting_started.md
@@ -11,9 +11,7 @@ It explains how to compile and run the `hello-word` example to verify your setup
    On Ubuntu, the following is sufficient:
 
 ```sh
-apt install build-essential curl git python3 pkg-config \
-  libssl-dev llvm-dev cmake libclang-dev gcc-arm-none-eabi \
-  clang unzip lld ninja-build
+apt install git rustup ninja-build pkg-config libudev-dev clang gcc-arm-none-eabi
 ```
 
 1. Install the Rust installer [rustup](https://rustup.rs/) using the website's instructions or through your distribution package manager.


book/src/getting_started.md
```diff
@@ -11,9 +11,7 @@ It explains how to compile and run the `hello-word` example to verify your setup
    On Ubuntu, the following is sufficient:
 
 ```sh
-apt install build-essential curl git python3 pkg-config \
-  libssl-dev llvm-dev cmake libclang-dev gcc-arm-none-eabi \
-  clang unzip lld ninja-build
+apt install git rustup ninja-build pkg-config libudev-dev clang gcc-arm-none-eabi
```
bergzand, line 14: according to [their instructions](https://probe.rs/docs/getting-started/installation/#debian-based-linux) probe-rs also requires cmake and pkg-config.

bergzand: The alternative is to us the binstall instructions from probe-rs, that drops libudev-dev from this list, but adds `cargo-binstall` to the deps. No strong opinion either way from me.

bergzand: The alternative is to us the binstall instructions from probe-rs, that drops libudev-dev from this list, but adds `cargo-binstall` to the deps. No strong opinion either way from me.

bergzand: The alternative is to us the binstall instructions from probe-rs, that drops libudev-dev from this list, but adds `cargo-binstall` to the deps. No strong opinion either way from me.

bergzand (approved):

PR #768, opened 2025-02-03
ci(commited): restrict the subject length to a more conventional length

ROMemories:
# Description

This is also the length after which GitHub starts eliding the subject line.

## Change checklist

### DIFF
--- a/committed.toml
+++ b/committed.toml
@@ -1,6 +1,6 @@
 style = "conventional"
 ignore_author_re = "(dependabot|renovate)"
-subject_length = 80
+subject_length = 72
 line_length = 80
 subject_capitalized = false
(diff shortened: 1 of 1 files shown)


chrysn:
This is overly restrictive, especially because the conventional format already dictates parts of the text (`refactor(config): ` is already 18); I already have to needlessly shorten messages to fit in the 80.

The goal is to have readable and understandable commit messages; generally short is good, but 72 is too short.

bergzand:
I'm with @chrysn on this. I sometimes already have a hard time to craft a subject within the limits because some name in the commit subject eats up halve the space.

Shortening it even more will just decrease the quality of the subject in content, I expect.

PR #772, opened 2025-02-04
feat(stm32): add partial support for persistent storage

ROMemories:
# Description

This adds partial support for flash-backed persistent storage to the currently supported STM32 MCUs, which only support blocking drivers. This makes `ariel_os::storage::remove()` unavailable on that MCU family, as flash drivers do not implement `MultiwriteNorFlash`. We may want to revisit how we handle this limitation.

This only supports homogeneous flash organization (therefore precluding support for STM32F401RE); support for heterogeneous organization can be addressed in the future.

## How to review this PR

- The `storage` example works as expected on STM32H755ZI, STM32WB55RG.

## Change checklist

### DIFF
--- a/src/ariel-os-storage/build.rs
+++ b/src/ariel-os-storage/build.rs
@@ -1,24 +1,47 @@
 use std::{env, path::PathBuf};
 
-fn main() {
-    // TODO: These should be configurable. Like this, it works for MCUs with
-    // a flash page size <= 4KiB.
-    const FLASH_PAGE_SIZE: u32 = 0x1000;
-    const STORAGE_SIZE_TOTAL: u32 = 0x2000;
+const KIBIBYTES: u32 = 1024;
 
-    // need at least two flash pages
-    // TODO: uncomment once this is not always true
-    //assert!(STORAGE_SIZE_TOTAL / FLASH_PAGE_SIZE >= 2);
+fn main() {
+    // NOTE(hal): values of `flash_page_size` from the datasheets, confirmed by HAL's constants.
+    // Important: only homogeneous flash organizations are currently supported.
+    // Trying to restrict the storage size to the subset of homogeneous flash would not work as it
+    // could be pushed out of it by a large enough binary.
+    let (storage_size_total, flash_page_size) =
+        if is_in_current_contexts(&["nrf52", "nrf5340", "rp2040", "stm32wb55rgvx"]) {
+            (8 * KIBIBYTES, 4 * KIBIBYTES)
+        } else if is_in_current_contexts(&["stm32h755zitx"]) {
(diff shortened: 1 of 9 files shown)


kaspar030:
Works fine on the wb55, fails on h755, similar to how it did on the f401:

```
    Finished in 5.02s
 WARN probe_rs::probe::stlink: send_jtag_command 242 failed: SwdApWait
 WARN probe_rs::probe::stlink: got SwdDpWait/SwdApWait, retrying.
 WARN probe_rs::probe::stlink: send_jtag_command 242 failed: SwdApFault
 WARN probe_rs::probe::stlink: send_jtag_command 242 failed: SwdApFault
 WARN probe_rs::session: Could not clear all hardware breakpoints: An ARM specific error occurred.

Caused by:
    0: The debug probe encountered an error.
    1: An error which is specific to the debug probe in use occurred.
    2: Command failed with status SwdApFault.
 WARN probe_rs::probe::stlink: send_jtag_command 242 failed: SwdApFault
 WARN probe_rs::session: Failed to deconfigure device during shutdown: Arm(Probe(ProbeSpecific(BoxedProbeError(CommandFailed(SwdApFault)))))
Error: Error communicating with the probe.

Caused by:
    0: An ARM specific error occurred.
    1: The debug probe encountered an error.
    2: An error which is specific to the debug probe in use occurred.
    3: Command failed with status SwdApFault.
```

src/ariel-os-storage/src/lib.rs
```diff
@@ -131,6 +136,8 @@ where
 /// All items in flash have to be read and deserialized to find the items with the key.
 /// This is unlikely to be cached well.
 /// </div>
+// STM32 flash drivers do not implement `MultiwriteNorFlash`.
+#[cfg(not(context = "stm32"))]
```
kaspar030, line 140: I'd prefer this to be like `cfg(has-MultiwriteNorFlash)`, let's revisit when the next comes around.

src/ariel-os-storage/build.rs
```diff
@@ -1,24 +1,47 @@
...
 
     std::fs::write(out.join("storage.x"), &storage_template).unwrap();
 
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_CONTEXT");
     println!("cargo:rerun-if-changed=storage.ld.in");
     println!("cargo:rustc-link-search={}", out.display());
 }
+
+/// Returns whether any of the current `cfg` contexts is one of the given contexts.
+fn is_in_current_contexts(contexts: &[&str]) -> bool {
```
kaspar030, line 40: interesting. I built sth similar using rustflags and reading `CARGO_ENCODED_RUSTFLAGS`. Let's move sth to an ariel-or-build helper down the line.

kaspar030 (approved):

PR #786, opened 2025-02-10
feat(rt): more flexible stacksize configuration

kaspar030:
# Description

This is my try at making the executor stack size configuration a bit more flexible, and more to the point.
E.g., our examples increase the `CONFIG_ISR_STACKSIZE`, but probably mean to increase the executor stack size.

Marking as draft, IMO we need some way of actually measuring the stack size to work on this properly.

## Change checklist

### DIFF
--- a/examples/coap-client/laze.yml
+++ b/examples/coap-client/laze.yml
@@ -2,7 +2,7 @@ apps:
   - name: example-coap-client
     env:
       global:
-        CARGO_ENV:
-          - CONFIG_ISR_STACKSIZE=32768
+        executor_stacksize_required:
+          - "32768"
     selects:
       - coap-client
--- a/src/ariel-os-rt/src/lib.rs
+++ b/src/ariel-os-rt/src/lib.rs
@@ -41,11 +41,27 @@ cfg_if::cfg_if! {
 
 #[cfg(any(context = "cortex-m", context = "riscv"))]
 mod isr_stack {
-    const ISR_STACKSIZE: usize = ariel_os_utils::usize_from_env_or!(
-        "CONFIG_ISR_STACKSIZE",
-        8192,
-        "ISR stack size (in bytes)"
-    );
+    pub(crate) const ISR_STACKSIZE: usize = {
+        const CONFIG_ISR_STACKSIZE: usize = ariel_os_utils::usize_from_env_or!(
+            "CONFIG_ISR_STACKSIZE",
+            2048,
+            "ISR stack size (in bytes)"
+        );
+
+        #[cfg(feature = "executor-interrupt")]
+        {
+            const CONFIG_EXECUTOR_STACKSIZE: usize = ariel_os_utils::usize_from_env_or!(
+                "CONFIG_EXECUTOR_STACKSIZE",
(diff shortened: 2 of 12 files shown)


kaspar030:
> Maybe only tangentially related, but is it possible in an application to configure different executor stack sizes for different architectures?

Good point. It should be.

chrysn:
From the team meeting: with ISR executor, the allocated size needs to be executor stack + ISR stack.

chrysn:
This just bit me again when running the CoAP example on stm32wb55 -- generally runs fine (RNG, writing key to flash at first startup, everything), but needs `s/CONFIG_ISR_STACKSIZE/CONFIG_EXECUTOR_THREAD_STACKSIZE/` in the example's laze.yml -- I'd rather not write that hack in the tutorial.

src/ariel-os-rt/src/lib.rs
```diff
@@ -39,8 +39,27 @@ cfg_if::cfg_if! {
...
+
+    #[cfg(feature = "executor-interrupt")]
+    {
+        const CONFIG_EXECUTOR_STACKSIZE: usize = ariel_os_utils::usize_from_env_or!(
+            "CONFIG_EXECUTOR_STACKSIZE",
+            8192,
+            "System executor stack size (in bytes)"
+        );
+
+        core::cmp::max(CONFIG_ISR_STACKSIZE, CONFIG_EXECUTOR_STACKSIZE)
```
chrysn, line 57:
```suggestion
        CONFIG_ISR_STACKSIZE + CONFIG_EXECUTOR_STACKSIZE
```

chrysn (commented):
This already looks quite good.

I think we should go ahead with this even if we don't have a means of measuring stack sizes: Right now, our sizes are informed guesses, and they'll still be after, but at least after this they will be guesses on a more realistic model.

kaspar030:
I tested all examples on nrf52830dk, and the http-client on rpi-pico-w (this one with both executor-interrupt and executor-thread).

src/ariel-os-rt/src/lib.rs
kaspar030: done

kaspar030:
@chrysn I think this is g2g, take a look?

examples/coap-client/laze.yml
```diff
@@ -2,7 +2,7 @@ apps:
   - name: example-coap-client
     env:
       global:
-        CARGO_ENV:
-          - CONFIG_ISR_STACKSIZE=32768
+        executor_stacksize_required:
+          - "32768"
```
kaspar030, line 6: Having to quote here is unfortunate. IIUC this is due to an architecture quirk in serde: https://github.com/dtolnay/serde-yaml/issues/388

chrysn (approved):
LGTM, trusting your testing.

Thanks for implementing this!

examples/coap-client/laze.yml
chrysn: I'd have been surprised if the number worked in the first place, considering that environment variables are generally stringly typed.

PR #788, opened 2025-02-10
docs(README): add logo

emmanuelsearch:
# Description

Adds a logo.


emmanuelsearch:
The tricky part is to get the "I" to be central, on the axis from the upmost to the downmost corners of the cube (subject to the word appearing to be overall central to the cube ;)

chrysn:
I've logically re-done the logo based on your version, replacing the PNG (which happened at some point during the editing process) and re-doing the line widths and chamfers, which suffered heavily from original construction on a 70° grid, manual edits to get the top lines in order, and the manual reslanting to the current upright version. All points are snapped to a grid, the cube is logically centered on the page, and the cut-outs are made into actual cut-outs.

I wouldn't consider the color shade normative (it's just RGB), but we can tune that over time and as we interact with print.

The version I'm uploading here has a history in my git repo, but I think we can copy that over and treat the code repo as the authoritative version.

logo

kaspar030 (changes requested):
Not convinced.

chrysn:
One of the things one could do with the logo: 
preview

That is not to say I'm happy with all aspects (like the text-to-image size ratio, or the orange -- especially with yellow LEDs, some blue would go well), but it's the best we have around, I think.

emmanuelsearch:
> I've logically re-done the logo based on your version

Thanks a bunch for this @chrysn . One detail I detected on this version: the lines are white, thus part of the transparency layer. To avoid this, I had originally made the lines with a super-duper-light grey.

chrysn:
That looked just like a bug to me; it'd add an extra color that is not used otherwise, and invisible half the time.

emmanuelsearch:
@chrysn I see what you mean, but then with svg/png the logo changes depending on the background => not sure you'd like green lines (in case of  a green background ;) What I'm suggesting is: let's make sure we control the colors of the lines (right?)

kaspar030:
> @chrysn I see what you mean, but then with svg/png the logo changes depending on the background => not sure you'd like green lines (in case of a green background ;) What I'm suggesting is: let's make sure we control the colors of the lines (right?)

Usually you'd just do multiple versions

emmanuelsearch:
> Usually you'd just do multiple versions

Of course, but often you simply don't have control over the background (eg OS/browser dark _vs_ light mode). So having a fits-all version is good to have...

chrysn:
I think we have two options here that are practical:
* The lines are supposed to be transparent. (That's the current version)
* The lines are supposed to be white, and the surroundings are supposed to be white.
  This can be implemented either by a "we only use this on a white background" rule (impractical, but I've seen this in CIs), or by giving the logo a white outline of similar width than the inbound lines. (I can produce such a version)

There is the third option of "use on any background but the lines are white", but I don't think that this is a good idea, because it alters the perceived shape (and the perceived set of colors that are part of the logo) depending on the surroundings.

emmanuelsearch:
Ariel-OS-Logo-Cube4-cropped

emmanuelsearch:
offline discussions brought out the fact that some disliked the font. There's another try I had in my drawers -- thought I'd share, just in case it was _miraculously_  more convincing

chrysn:
I think we'd need much bigger text in relation to the logo space, it was already on the brink with the previous one.

At any rate, if that's the level of changes we're still doing, we might be better off going back to the pad than in the issue discussion (which works well for discussing a single proposal but not so well for branching around in a brainstorm).

emmanuelsearch:
No consensus. Closing this.

PR #796, opened 2025-02-12
fix(debug-console): enable the debug logging facade only with a console

ROMemories:
# Description

The debug logging facade should only be enabled when a debug console is there to print on.
Previously the following would fail to link:
```sh
laze -C examples/log/ build -d debug-console -b nrf52840dk size
```
and would require to explicitly disable `debug-logging-facade` as well:
```sh
laze -C examples/log/ build -d debug-console -d debug-logging-facade -b nrf52840dk size
```

With this PR `debug-logging-facade` is only enabled when `debug-console` is enabled instead of always, so the first command works as expected.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -22,7 +22,6 @@ contexts:
     selects:
       - executor-default
       - ?critical-section
-      - ?debug-logging-facade-default
       - ?lto
       - ?semihosting
     env:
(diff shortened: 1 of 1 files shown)


kaspar030 (approved):

kaspar030:
Hm, I don't have it clear, yet. If there is _no_ facade, will all the `info` still work? Is there a noop backend? Or does this work b/c all boards depend on a flash module (`probe-rs`, `esp-flash`) that pull in `debug-module`?

PR #800, opened 2025-02-12
docs(randomness): document the RNGs

ROMemories:
# Description

This adds a book page documenting our RNGs and how to obtain and use them.
The immediate goal is to document the laze modules at play so that their Cargo features can be hidden as part of #737.

## Change checklist

### DIFF
--- a/book/src/randomness.md
+++ b/book/src/randomness.md
@@ -1,0 +1,23 @@
+# Randomness and Entropy Sources
+
+Ariel OS provides RNGs which fulfill needs for both fast and cryptographically secure sources of randomness.
+
+## Provided RNGs
+
+The `random` [laze module][laze-modules-book] needs to be enabled to be able to obtain the provided RNGs.
+Two different RNG interfaces are provided, which both implement `rand_core` traits:
+
+- A fast RNG interface, not suitable for cryptography use, which can be obtained with [`random::fast_rng()`][fast-rng-fn-rustdoc].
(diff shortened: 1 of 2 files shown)


book/src/randomness.md
```diff
@@ -0,0 +1,21 @@
+# Randomness and Entropy Sources
+
+Ariel OS provides RNGs which fulfill needs for both fast and cryptographically secure sources of randomness.
+
+## Provided RNGs
+
+The `random` [laze module][laze-modules-book] needs to be enabled to be able to obtain the provided RNGs.
+Two different RNGs are provided, which both implement `rand_core` traits:
```
chrysn, line 8: There is not one global "fast RNG" and a global "crypto RNG"; there is a single (hidden) RNG, and we haven't committed yet to whether the user gets access to any of them or copies. (Currently, the user gets a copy of the fast RNG, but it's made non-Send so we can make that thread-local if we prefer).

Reading this feels a bit like both are global; any idea how we could retain clarity while also retaining flexibility? Should we be explicit about why and how we've kept options open here?

chrysn: The internal one may be slow, but you get a fast one which you can then use for multiple quick operations.

There is no big difference in observable behavior (right now the APIs even ensure that we could make the fast RNG thread-local, maybe even global, and ZST); it's just about not implying anything about the implementation.

chrysn (approved):
Thanks, that looks good!

PR #803, opened 2025-02-12
feat(coap): laze based configuration

chrysn:
# Description

Right now, all our examples use the demo keys, and we can't provision own keys, and can't run completely nosec.

This introduces laze (really cargo) features in which ariel-os-coap decides how to set up security, with the examples' config remaining "pick the demo keys" (but this time, explicitly).

With some coapcore internal changes (well they change its API, but not in a way ariel-os-coap notices), the nosec mode builds with considerably less flash usage.

Next steps from here are introducing more building blocks (like "flash my moral equivalent of ~/.ssh/id_ecdh.pub as the root key instead of a key file we ship" or "there is provisioning like in #453 so take those keys"), and those can extend both the -demokeys and can become part of a -default setup, but I'd take those gradually from here.

## Issues

Depends on #805 (so please ignore that first commit)

## Open Questions

* <del>I've seen a stack overflow again; I don't think this PR increases it, but nonetheless running some more tests.</del>

## Reviewing hints

* "consolidate demo setup" is best viewed with --ignore-all-space
* "feat(coapcore): do not link OSCORE and EDHOC unless usable ": chatting with tanriol at #rust:matrix.org has led me on a path for doing this better (uninhabited types to make some internal state machines vanish completely in the nosec mode), but that's a next step.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -807,6 +807,9 @@ modules:
     # will have their own entry point when they exist.
     depends:
       - coap
+      # This does *not* depend on (or select) coap-server-config yet, because
+      # doing so would cause a config to be auto-selected, and there's none
+      # that can currently be recommended without explicit configuration.
     env:
       global:
         FEATURES:
@@ -819,6 +822,7 @@ modules:
       accept opportunistic EDHOC.
     depends:
       - coap
+    provides_unique: [coap-server-config]
     env:
       global:
         FEATURES:
@@ -830,6 +834,7 @@ modules:
       This configuration may change over time, in lockstep with the examples' setup.
     depends:
       - coap
+    provides_unique: [coap-server-config]
     env:
       global:
         FEATURES:
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -812,6 +812,29 @@ modules:
         FEATURES:
           - ariel-os/coap-server
 
+  - name: coap-server-config-unprotected
+    help: Configure the CoAP server to accept any request without authorization checks.
+
+      The device will have no cryptographic identity for CoAP, and not even
+      accept opportunistic EDHOC.
+    depends:
+      - coap
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap-server-config-unprotected
+
+  - name: coap-server-config-demokeys
+    help: Configure the CoAP server as expected by simple examples that expect a fixed server key.
+
+      This configuration may change over time, in lockstep with the examples' setup.
+    depends:
+      - coap
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap-server-config-demokeys
+
   - name: coap-client
     help: Support for CoAP client functionality.
     depends:
--- a/src/ariel-os-coap/Cargo.toml
+++ b/src/ariel-os-coap/Cargo.toml
@@ -32,6 +32,8 @@ ariel-os-random = { workspace = true, features = ["csprng"] }
 ariel-os-macros = { path = "../ariel-os-macros" }
 static_cell = "2.1.0"
 
+cfg-if = { workspace = true }
+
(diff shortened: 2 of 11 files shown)


src/ariel-os-coap/Cargo.toml
```diff
@@ -54,6 +56,9 @@ workspace = true
 # laze's name for this (where coap-server makes more sense).
 coap-server = []
 
+coap-server-config-unprotected = []
```
chrysn, line 59: Should I refer to the docs of other items, or do we understand "this is a feature documented in ariel-os or laze" to be a default assumption?

src/ariel-os/Cargo.toml
```diff
@@ -64,6 +64,11 @@ coap = ["dep:ariel-os-coap", "random"]
 ## Enables applications to set up CoAP server handlers.
 ## See [`coap::coap_run()`].
 coap-server = ["coap", "ariel-os-coap/coap-server"]
+# Plain forwarded features that are not documented as features but just as laze
+# modules, because while those here work without any extra help from laze, most
+# later ones will likely need some build system help.
```
chrysn, line 69: Examples of that build system help are "laze checks whether a file is present, decides which to select, passes it in through the environment, and creates one if none is present"; even for those here, we could have some build system interaction like a yellow blinking "hey that's not secure" (for both, actually).

laze-project.yml
```diff
@@ -812,6 +812,29 @@ modules:
...
+    help: Configure the CoAP server as expected by simple examples that expect a fixed server key.
+
+      This configuration may change over time, in lockstep with the examples' setup.
+    depends:
+      - coap
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap-server-config-demokeys
+
```
chrysn, line 837: Is there better practice than just having multiple modules? Is there a way to express that a module conflicts with any other of a group?

tests/coap/README.md
```diff
@@ -21,6 +21,12 @@ This application is a work in progress demo of running CoAP with OSCORE/EDHOC se
   up to the maximum number of security contexts that are stored (currently 4).
 * There is also `./fauxhoc.py`, which did EDHOC manually before it was integrated in aiocoap.
 
+### Variation
+
+* CoAP in NoSec mode: Building a smaller binary at the cost of confidentiality and integrity protection.
+    * In `laze.yml`, replace `coap-server-config-demokeys` with `coap-server-config-unprotected`.
```
chrysn, line 27: Is this the right way to do a demo, or can the example do something softer than "selects", and the user can push things over to another one (deselcting the demokeys by virtue of the conflict)?

chrysn: Obsoleted by the `provides_unique` construction: now it can be overridden on the command line, and the README is updated to reflect that.

src/ariel-os-coap/Cargo.toml
```diff
@@ -32,6 +32,8 @@ ariel-os-random = { workspace = true, features = ["csprng"] }
 ariel-os-macros = { path = "../ariel-os-macros" }
 static_cell = "2.1.0"
 
+cfg-if = { workspace = true }
```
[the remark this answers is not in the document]
chrysn: Good point, thanks.

May I squash that in right away?

chrysn:
Rebasing with --keep-base succeeded in making GitHub not see that first commit any more (apparently, a force push makes it reconsider the contained commits).

Test results of selecting both causes an error in laze:

```
laze: not building binary "test-coap" for builder "particle-xenon"

Caused by:
    0: "test-coap" cannot resolve "coap-server-config-unprotected"
    1: "coap-server-config-unprotected" conflicts already provided "coap-server-config" (by "coap-server-config-demokeys")
```

Issue #338, opened 2024-07-02
Pre-screen CoAP requests based on credential's claims


chrysn:
To
* enable flexible policy settings,
* avoid security checks coded into every single handler, and
* avoid the trouble of passing data from the request into the handler that is not in the request (which on its own is a bad reason, but in combination with the others, there's no good reason to change that setup)

we should process [AIF REST-specific](https://www.rfc-editor.org/rfc/rfc9237.html) scope into permissions.

The current implementation is crude: It is not parsed, and hardcoded to a single resource.

Going forward, there are two paths:
1. At build time, we could build a type representing where a resource is in the tree (effectively a big nested enum, which is efficient at runtime), and when a credential is presented, parse the scope into a `[(PathEnum, PermissionSet); N]` structure.
2. When a credential is presented, verify its structure, but store it as `AsRef<[u8]>` and parse it on every request.

<del>Right now I am leaning toward</del><ins>I've implemented</ins> 2 because it does not require changes to the handler types, because we may need the full credential later anyway, and because a size limit on the full credential (typically: token) is easier to communicate with an AS.

[edit: Added text, set title – sorry fort the noise, won't use GitHub's "convert into issue" feature again]

# Steps

* [x] Implement
* [ ] Expose

chrysn:
This is largey closed by https://github.com/ariel-os/ariel-os/pull/588

chrysn:
This is implemented now in [AifValue](https://ariel-os.github.io/ariel-os/dev/docs/api/coapcore/scope/struct.AifValue.html), but not easily accessible, as the ariel_os_coap configuration is not yet steered by the build system.

chrysn:
Good enough for first release: Exposing it happens through laze, so it's not an API change to migrate away from example creds.

chrysn:
This is closed in https://github.com/ariel-os/ariel-os/pull/814: In that PR, the build.rs reads a per-application peers.yml, which is a prelude to <https://ariel-os.github.io/ariel-os/dev/docs/book/tooling/coap.html#outlook-interacting-with-an-ariel-os-coap-server-from-the-host>, but already allows setting arbitrary AIF sequences as per the user's needs.

PR #814, opened 2025-02-13
feat(coap): support stored security configurations

chrysn:
# Description

Shared secret keys bad. Private keys never leaving device good.

This builds up local key generation, paving the way for https://github.com/ariel-os/ariel-os/issues/446.

## Scope

For this step, the scope is to enable the tests/coap case to switch away from the shipped keys, and to enable custom permissions for keys:

* Device keys are auto-generated at first boot and persisted.
* The user can enter their picks their key (README says to reuse the demo key to avoid depending on aiocoap-keygen details) into a per-application permissions file, those get baked into the firmware and used at runtime

In particular, this closes https://github.com/ariel-os/ariel-os/issues/338 by virtue of now having a way for the user to tell which AIF permissions are associate with any key.

### Future steps

Steps from there are somewhat orthogonal:
* Extraction of the generated key can become prettier (https://github.com/knurling-rs/defmt/pull/916; as that is now ACK'd, we could patch it in from a branch)
* More of peers.yml could be inferred from the system (in particular towards [the book description of using keys stored a bit like SSH keys](https://ariel-os.github.io/ariel-os/dev/docs/book/tooling/coap.html#outlook-interacting-with-an-ariel-os-coap-server-from-the-host))
* The same source can also convey information about trusted Authorization Servers.
* The host keys are currently only in flash; we could just as well have some in storage, but we'd need a (CoAP?) interface to actually edit them. (Possibly having some in ROM, some in storage, and a an option for those in storage to override those in ROM, given we don't prepopulate the storage).

Semi-relatedly, seeing how concise `expand_id_cred_x` can be after having implemented it a second time, I'll want to make some simplifications in coapcore's code as well, but that'll be pure refactoring.

## Testing

Run the new "with more than just demo keys" variant of tests/coap.

## Change checklist


chrysn:
One pain point in the current README is https://github.com/knurling-rs/defmt/pull/916 -- let's see if we can get this unstuck.

src/ariel-os-coap/src/stored.rs
```diff
@@ -0,0 +1,124 @@
...
+        None
+    }
+}
+
+impl StoredPolicy {
+    async fn load() -> Self {
+        // Storage format: ([u8], [u8; 32]), where the former is a CCS, and the latter the
+        // corresponding key. We may need to extend the latter to be a COSE_Key when crypto agility
+        // becomes a thing.
+        const OWN_CREDENTIAL_KEY: &str = "coapcore.own-edhoc-credential";
```
chrysn, line 53: What kind of conventions do we want to use for those names?

laze-project.yml
```diff
@@ -856,6 +856,19 @@ modules:
         FEATURES:
           - ariel-os/coap-server
 
+  - name: coap-server-config-storage
+    help: Configure the CoAP server to accept requests depending on build- and runtime configuration
+    depends:
+      - coap
+      - sw/storage
+    provides_unique: [coap-server-config]
```
kaspar030, line 864:
```suggestion
    provides_unique:
      - coap-server-config
```

we'll have a convention for that soon

laze-project.yml
```diff
@@ -856,6 +856,19 @@ modules:
...
+    env:
+      global:
+        FEATURES:
+          - ariel-os/coap-server-config-storage
+        CARGO_ENV:
+          - PEERS_YML=$\\${PWD}/peers.yml
```
kaspar030, line 870:
```suggestion
          - PEERS_YML=${appdir}/peers.yml
```

laze-project.yml
```diff
@@ -856,6 +856,19 @@ modules:
...
```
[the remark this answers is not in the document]
kaspar030: I think have a `PEER_YML` _laze_ variable that defaults to `{appdir}/peers.yml` would be the way to go. (And use that here in `CARGO_ENV` with `- PEER_YML=${PEER_YML}`).

src/ariel-os-coap/src/stored.rs
kaspar030: "be sufficiently individual in your naming scheme that it won't match the next guys"

src/ariel-os-coap/src/stored.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        let mut storage = ariel_os_storage::lock().await;
+        let (credential, key) = match storage
+            .get(OWN_CREDENTIAL_KEY)
+            .await
+            .expect("flash error prevents startup")
+        {
+            Some(credpair) => credpair,
+            None => {
+                use lakers::CryptoTrait;
```
kaspar030, line 63: Style wise I find it off to have the generation, and serialization, of the keypair within a function that is called `load()`, while holding a lock.
Maybe split out some helper functions?

chrysn: Good point, will do. (Held the lock because it is a read-or-set-default thing, but realistically there won't be parallel calls, and there is no harm as long as there is one that is eventually set).

src/ariel-os-coap/src/stored.rs
kaspar030: more seriously, you could prefix the module name?

laze-project.yml
chrysn: This will be obsolete when using a laze variable. (It may apply to the laze variable default, though).

laze-project.yml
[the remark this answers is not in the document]
chrysn: Tested with `-D PEERS_YML=<(echo foo)` failing :-)

chrysn:
I think all points are addressed in the latest fixups, and lints are through.

kaspar030 (approved):
lgtm

PR #824, opened 2025-02-14
feat(coap): remove blockers for ESP32

chrysn:
# Description

This threads features of libOSCORE through the whole build system so that a particular symbol (`abort()`) that libOSCORE needs to provide by itself on pure-Rust builds is disabled on ESP32S3.

## Caveats and testing

This does not yet enable the ESPs to build (that'll need build script and CI magic); nonetheless, these changes' effectiveness can be tested by removing the conflict from the coap-server example's laze file, and running:

* as a baseline that ESP works (this one doesn't need any particular features)

   ```sh
   $ laze build -b particle-xenon
   $ (source ~/export-esp.sh && CONFIG_WIFI_NETWORK=... CONFIG_WIFI_PASSWORD=... CFLAGS="-mlongcalls" CC=xtensa-esp32-elf-gcc laze -C examples/coap-server/ build -b espressif-esp32-devkitc)
   ```

* to show its effectiveness:

   ```sh
   $ (source ~/export-esp.sh && CONFIG_WIFI_NETWORK=... CONFIG_WIFI_PASSWORD=... CFLAGS="-mlongcalls" CC=xtensa-esp32s3-elf-gcc laze -C examples/coap-server/ build -b espressif-esp32-s3-devkitc-1)
   ```

I've only tested this with ESP32 (not S or other letter) and don't have ESP32S2 hardware around.

## Issues/PRs references

* I guess @kaspar030 has some branch already around where the `export-esp.sh` sourcing happens inside the build system, and on top of that we set the CC and CFLAGS. In combination with that, we could enable building on any esp32.

* Builds on #823

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -302,6 +302,7 @@ contexts:
     parent: esp
     selects:
       - xtensa
+      - something-inside-the-esp32s3-blobs-that-contains-a-c-abort
     env:
       CARGO_TOOLCHAIN: +esp
       RUSTFLAGS:
@@ -452,6 +453,9 @@ modules:
       global:
         RUSTFLAGS:
           - --cfg context=\"xtensa\"
+
+  - name: something-inside-the-esp32s3-blobs-that-contains-a-c-abort
+    provides_unique: [c-function-abort]
 
   - name: riscv
     env:
@@ -794,6 +798,8 @@ modules:
     depends:
       - random
       - network
+      - c-function-assert
+      - c-function-abort
     env:
       global:
         FEATURES:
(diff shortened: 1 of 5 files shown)


laze-project.yml
```diff
@@ -302,6 +302,7 @@ contexts:
     parent: esp
     selects:
       - xtensa
+      - something-inside-the-esp32s3-blobs-that-contains-a-c-abort
```
chrysn, line 305: That might be done prettier in laze, and if anyone has any insight as to whether a non-WiFi build of ESP32S2 would also have that in, that might also help.

chrysn: There's https://github.com/kaspar030/laze/issues/638#issue-2854170723 to track that in laze; until I get around to patching that, I think we'll need this workaround (but probably with a better name).

Issue #667, opened 2025-01-13
build: lingering build issues around liboscore


kaspar030:
This came up at the today's weekly:

- Q: kaspar reports some lingering building issues with coapcore (doing full builds, maybe LLVM version issues...)
  - A: Try this environment: `LLVM_CONFIG_PATH=/usr/bin/llvm-config-19 LIBCLANG_PATH=/usr/lib/llvm-19/lib/`
  - Maybe this is something only laze / the build environment can decide.
  - → Kaspar will create issue

chrysn:
Is this https://github.com/ariel-os/ariel-os/issues/418 or more?

kaspar030:
I think it is more:

```
[45/444] CARGO build/bin/nrf52840dk/coap-client/coap-client.elf                                                                          
   Compiling liboscore v0.2.2                                                                                                            
error: failed to run custom build command for `liboscore v0.2.2`                                                                         
                                                                                                                                         
Caused by:                                                                                                                                                                                                                                                                        
  process didn't exit successfully: `/home/kaspar/src/own/rust/ariel-os/examples/coap-client/../../build/bin/nrf52840dk/cargo/release/build/liboscore-8dbd681f8bef6e47/build-script-build` (exit status: 101)                                                                     
  --- stdout                                                                                                                                                                                                                                                                      
  cargo:PLATFORMHEADERS=/home/kaspar/src/own/rust/ariel-os/examples/coap-client/../../build/bin/nrf52840dk/cargo/thumbv7em-none-eabi/release/build/liboscore-207cb1a01f045b8d/out/rust-built-headers                                                                              
  cargo:rerun-if-env-changed=TARGET                                                                                                                                                                                                                                               
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_thumbv7em-none-eabi                                                                                                                                                                                                         
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS_thumbv7em_none_eabi                                                                                                                                                                                                         
  cargo:rerun-if-env-changed=BINDGEN_EXTRA_CLANG_ARGS                                                                                                                                                                                                                             
  cargo:rerun-if-changed=oscore_all_headers.h                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                  
  --- stderr                                                                                                                                                                                                                                                                      
[...]

PR #825, opened 2025-02-15
Esp32 envs

chrysn:
# Description

Users having `~/export-esp.sh` sourced constantly has been a [source of trouble](https://github.com/ariel-os/ariel-os/issues/667#issuecomment-2659116085) -- after all, that bends the user's LLVM path. The logical consequence is to not let the user do that, but do it inside laze -- especially given that a laze build might build multiple platforms.

@kaspar030 might have had something similar in planning, but I want to do a full test of ESP32 CoAP builds, and this is a piece thereof.

## Change checklist


chrysn:
This currently fails because for $REASONS, ~/export-esp.sh is not populated by the esp-rs/xtensa-toolchain action -- instead, they place that file in `$HOME/exports`, so we need to have a tunable for "what do you need to source when doing ESP stuff", and that will need to be passed in CI.

chrysn:
Rebased onto and taking over #827 -- the latest upstream PR used (https://github.com/esp-rs/xtensa-toolchain/issues/38) has a dedicated tunable to not export ~/export-esp.sh. To combat the path confusion around where to actually take the exports from, that upstream action now also exports ESPUP_EXPORT_FILE, which is their location for where the export file is.

chrysn:
From chat, ToDo: use CFLAGS as list rather than value.

chrysn:
Abandoning in favor or #826 which wraps up all changes needed to effectively test this.

PR #826, opened 2025-02-15
chore: source extensa environment on demand, unblocking coap

chrysn:
# Description

This combines #825, #824 and a commit that removes the blockers for the tests/examples and xtensa.

This will need a rebase when both dependencies are in.

## Open Questions

* <del>Will it blend?</del> It even loses the taste of stale workarounds and inaccessible platforms after blending.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #827, opened 2025-02-16
ci: disable ESP32 action's overrides

chrysn:
# Description

This brings what happens in CI closer to what happens on a host set up as per the book's instructions.

A second commit pulls in https://github.com/esp-rs/xtensa-toolchain/issues/35 to not only unset the cargo override but also to avoid messing with LLVM variables.

## Issues/PRs references

#824, #825 and #826 a related, but I'm trying to keep things testable independently.

## Open Questions

* I think there's a follow-up step that even needs to be done in here, where LIBCLANG_PATH is not exported in that step any more, and we use something #825-ish instead.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,6 +75,7 @@ jobs:
         with:
           version: "1.84.0.0"
           buildtargets: esp32s3
+          override: false
 
       - name: Keep general build environment clean of Xtensa peculiarities
         if: steps.result-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


chrysn:
I'm marking this as "ready for review" because it still works in CI, and I'd like to do deeper testing based on it (ie. rebase some other branches onto it.)

chrysn:
Closing this one because it became so good in what it does (keeping the env clear of ESP stuff) that it broke things, and thus only works in combination with https://github.com/ariel-os/ariel-os/pull/825 which is taking over its commits.

PR #833, opened 2025-02-17
feat(storage): allow structured keys

chrysn:
# Description

This picks up an item discussed during recent technical meetings: Storing the keys as CBOR items rather than length-delimited strings, as that enables structured keys.

Commits start from noncontroversial refactoring: a new internal (sealed, because the storage module is private) trait Key is introduced that initially works only for &str – "feat(storage): generalize to Key type" doesn't change serialization but just adds this as an interface for internal tuning.

Going to CBOR comes with a choice of either going through a library (dependency and flash size increase are only relevant if this is the only place it is used -- but then again, currently CBOR is only in CoAP security) or by hand-rolling it (barely more lines of code as long as we know to only expect very limited types in keys, but I wouldn't want to explain all the magical constants).

The last commit is more an outlook at what kinds of keys are possible; it'll change depending on the open questions (the current form is purely for illustration).

## Open Questions

* Do we want to keep a minimal "string-only" variant of storage keys for binary simplicity (at higher maintenance cost)?
* How far do we want to open up the key space? Options range from "explicitly allow &str, (&str, usize), (&str, usize, usize) and (u64, usize, usize)[^1]" to "anything that is minicbor::Encode".
  My gut feeling is that we don't want to do the latter (because that's really a lot, and can be influenced a lot by users), but maybe have some allow-listed types, and simple tuple combinations thereof. This limitation allows us to easily substitute in a minimal CBOR processor when the direction is "go slim" without too big of an ifdef hell.

I'd also very much like to have value-typed keys, where rather than doing `const KEY: str = "key"; let received: usize = get(KEY).await; store(KEY, 4u8 /* oups */);` I can do `const KEY: TypedKey<usize> = "key"; let received = get(KEY).await;` because that key only opens that lock, but I *think* that that is largely an orthogonal discussion that can build on altering the sealed key trait.

[^1]: or whatever YANG SID needs

## Change checklist

## Status

Marked as "ready for review" to get the high-level discussion; implementation details are subject to change depending on the outcome.


kaspar030 (changes requested):
I'm missing a "why".

chrysn:
# Why

The "why" is that not all configurations are scalar. A setting like "allow this other device network access" has those options:
1. What is in the PR: store items with keys `("peer", 0)`, `("peer", 1)` etc (possibly also using binary key IDs)
2. Render ASCII keys at runtime: store items with keys `format!("peer-{}", id)`
3. Store data for all peers in a single big `"peers"` item that gets updated all the time.

If we don't go with 1, my next best choice is 3, which has a few downsides:
* The runtime stack cost is proportional to the maximum number of peers we allow. If we set aside 512B of ROM for peers (roughly 10 peers, depending on complexity of their permissions), those 512B need to be loaded onto the stack before a single peer can be picked out. Granted, right now the data gets loaded into RAM at startup for lack of async processing inside CoAP credential loading, but I regard that as a bug to be fixed.
* Small updates amplify a lot. Once we start storing more than just the credential (with OSCORE that'd be resumption counters, with ACE that'd be AS timestamps), any update would rewrite the whole maximal storage. This is better when going with 2, but even better when going with 1, because there the structure can be extended to only update `("peer", 2, "count")` with its 32-bit data item, rather than the whole `"peer-2"` (or the `"peers"` of 3).

This is the "why I'd prefer to have this" view; in parallel, at some point I hope we'll have CoREconf, and that has keys shaped `(u64, /* any number of */ u8)`; if the any-number is 0 (because it is a global setting), there's the option of having a text string per ID, but for anything with multiple occurrences (eg. interface setups once we have them), the same options 1/2/3 as above come in again.

chrysn:
Status update as this has been sitting unduely long: This is waiting for me to demonstrate that it can also be done without blowing the flash size of a minimal example out of proportion.

PR #836, opened 2025-02-18
feat(storage): Wrap errors in own error type

bergzand:
# Description

This wraps both the sequential-storage and arrayvec errors in our own error type. In turn this removes the condition under which the storage functions panic due to too long keys. Instead a relevant error is returned

## Issues/PRs references

None

## Open Questions

-

## Change checklist


src/ariel-os-storage/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+#![deny(missing_docs)]
+#![deny(clippy::pedantic)]
+use arrayvec::CapacityError;
+use embedded_storage_async::nor_flash::NorFlashError;
+use sequential_storage::Error as StorageError;
+
+/// Storage-related errors
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum Error<S> {
```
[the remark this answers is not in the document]
bergzand: Sounds fine, I can check which errors we need and which ones we should never run into.

If I understand correctly, your suggestion includes removing the `NorFlashError` generic from our error and map those errors to a single error in our error type?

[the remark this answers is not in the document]
bergzand: I've flattened the full set of errors. This means that the current list is a bit long as it covers also the  NorFlash related and sequential-storage related errors. 

I've added a small bit of documentation on what the user can expect. This can be extended later to emphasize that some errors should not occur during normal operation, but might occur during development or initial board configuration.

src/ariel-os-storage/src/lib.rs
```diff
@@ -9,13 +9,14 @@
 // TODO: overhaul errors
 #![expect(clippy::missing_errors_doc)]
```
ROMemories, line 10: So now we can remove this (and move it to the internal modules), and document on each public function which set of errors it can return (e.g., `remove()` is never going to return `Error::ItemTooBig`).

```suggestion
```

src/ariel-os-storage/src/storage.rs
```diff
@@ -124,10 +102,10 @@ impl<F: NorFlash> Storage<F> {
...
 
     /// Resets the flash in the entire flash range of this [`Storage`] instance.
-    pub async fn erase_all(
-        &mut self,
-    ) -> Result<(), sequential_storage::Error<<F as ErrorType>::Error>> {
-        erase_all(&mut self.flash, self.storage_range.clone()).await
+    pub async fn erase_all(&mut self) -> Result<(), Error> {
+        erase_all(&mut self.flash, self.storage_range.clone())
+            .await
+            .map_err(Error::from)
```
ROMemories, line 108: optional: I usually write this this way (`?` already uses `From`); I'm not sure it results in the same binary however. This is really just a suggestion (the other occurrences would need to be updated as well if you went with this):

```suggestion
        Ok(erase_all(&mut self.flash, self.storage_range.clone()).await?)
```

bergzand: I used your suggestion in a previous version of this PR and you` suggested `map_err` [here](https://github.com/ariel-os/ariel-os/pull/836#discussion_r1959936211) :laughing:

ROMemories: Yes, but to map/wrap upstream errors into specific custom errors :slightly_smiling_face:

src/ariel-os-storage/src/error.rs
```diff
@@ -0,0 +1,102 @@
...
+///    Either decrease the key length or increase `MAX_KEY_LENGTH`.
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum Error {
+    /// Flash operation arguments are not properly aligned.
+    FlashNotAligned,
+    /// Flash operation arguments are out of bounds.
+    FlashOutOfBounds,
+    /// Other flash implementation specific error.
+    FlashOther,
```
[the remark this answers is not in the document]
bergzand: I agree this can be moved to a `panic!()`.

> If these happen it's because there is a bug in our storage implementation,

I expect this to also get triggered by any incorrectly configured custom board by users. A `panic!()` should still be fine, but I expect to iterate a bit over the exact panic message :smile:

PR #841, opened 2025-02-18
fix(build): add fallback `CARGO_PRESHELL`

kaspar030:
# Description

#826 exposes a weirdness in laze's task escaping. No idea what it is but here is a workaround.

## Issues/PRs references

#826 

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -8,6 +8,7 @@ contexts:
       CARGO:
         - ${CARGO_WRAPPER}
         - cargo
+      CARGO_PRESHELL: true &&
 
     tasks:
       info-modules:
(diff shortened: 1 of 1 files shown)


chrysn (approved):
I have no idea why @ROMemories's builds failed as they did, because both CI and local testing passed, but this is an improvement, and if there are cases that work with this, all the better.

We can later improve, eg. by moving the `&&` into the CARGO_PRESHELL users (then it is more clear how this is all supposed to be used), and we can still find out where this comes from when this is merged.

chrysn:
The odd thing is that I saw something similar on my machine when I tried going through `env` instead.

What are all your /bin/sh (which I'd guess laze invokes)? Mine is from dash, and so is probably the one on CI.

PR #843, opened 2025-02-19
ci(build): restrict build runs based on PR labels

ROMemories:
# Description

The initial goal of this was to require specific labels on PR that would allow us to specify whether to run the `Build` CI workflow, and if so, for which set of boards.
Now, the goal is simply to make the Build workflow fail quickly if no `ci-build:*` label is attached, and to collect information which would help us actually deploy the initial idea later, with minimal breakage.

This is intended to support the following scenarios when running for PRs:

- No `ci-build:*` labels: fail job
- `ci-build:small`: build with restricted build set
- ~`ci-build:large`: build with full build set (at least for now)~ (removed from this initial PR)
- ~`ci-build:skip`: skip building~ (removed from this initial PR)
- Two or more `ci-build:*` labels: fail job

**Runs for the merge queue and on main must take these labels into account as well.**

## Future work

- We may want to introduce other kinds of build labels, to target specific MCU subfamilies as well (e.g., `ci-build:stm32f4`).
- We could consider some forms of auto-labeling, e.g., attaching `ci-build:skip` if only the book is changed by a PR.

## Issues/PRs references

Supersedes #581.

## Open Questions

- ~This looks up the top commit ID of the PR in an attempt to find "its" PR; however multiple PRs could contain that commit (in case of stacked PRs).~ This should now be addressed, it could now only fail to find the proper PR if multiple PRs were opened for *exactly the same commit id*, which should never happen anyway.
- We may want to introduce more "T-shirt sizes" for build sets later (e.g., `ci-build:medium`, `ci-build:xxlarge`); should the current `ci-build:small` be set in stone now? If we change the definition of labels later, this may look slightly misleading when looking back on PRs later.  

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,8 +40,54 @@ jobs:
         run: |
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL &
 
+      # Need to get two commits so we can get the PR's commit, not just the
+      # merge commit.
       - name: Check out repository code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+
+      # Get the second parent of the merge commit created by the `checkout`
+      # action.
+      - name: "Get the commit ID of the pull request branch"
+        id: get-commit-id
+        shell: bash
+        run: echo "commit_id=$(git rev-parse HEAD^2)" >> $GITHUB_OUTPUT
+
+      # We use this instead of accessing the PR labels through the trigger event
+      # because that would not work for the merge queue.
+      - name: "Get PR build labels through the commit hash of the pull request"
+        id: get-pr-labels
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: |
+            const commitId = "${{ steps.get-commit-id.outputs.commit_id }}";
+            console.log(`Commit ID: ${commitId}`);
+            const { data: { items } } = await github.rest.search.issuesAndPullRequests({
+              q: `${commitId} type:pr`,
+            });
+            console.log(`Number of PRs containing commit ${commitId}: ${items.length}`);
+            // Find a PR where that commit is the tip of the PR's branch.
+            // Would be incorrect if multiple PRs are open for exactly the same commit.
+            for (const pr of items) {
+              const prNumber = pr.number;
+              const { data: commits } = await github.rest.pulls.listCommits({ owner: 'ariel-os', repo: 'ariel-os', pull_number: prNumber });
+              const tipCommitId = commits[commits.length - 1].sha;
+              if (tipCommitId != commitId) {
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -40,8 +40,54 @@ jobs:
...
+            const commitId = "${{ steps.get-commit-id.outputs.commit_id }}";
+            console.log(`Commit ID: ${commitId}`);
+            const { data: { items } } = await github.rest.search.issuesAndPullRequests({
+              q: `${commitId} type:pr`,
+            });
+            console.log(`Number of PRs containing commit ${commitId}: ${items.length}`);
+            // Find a PR where that commit is the tip of the PR's branch.
+            // Would be incorrect if multiple PRs are open for exactly the same commit.
+            for (const pr of items) {
+              const prNumber = pr.number;
```
ROMemories, line 74: @kaspar030 Suggested that it may be possible to obtain the PR number from the merge commit title (e.g., `(#843)`).

ROMemories: This unfortunately does not seem to be possible for the CI run *before the merge*; I tried this by getting the merge commit subject with `git log --pretty=%s -1` in CI, and the subject seems to be `Merge a9d6f4946f58bfcde6b7ebaa578e0ee8d9005eb6 into ca3b8d556319e761d2c00aa14b31550d8799f3e4`.
Falling back to the previous strategy.

ROMemories:
CI failed with no `ci-build:*` label attached as expect; labeling as `ci-build:small` now.

.github/workflows/build.yml
kaspar030: Was that for the PR build (not the merge commit build)?
I just saw this in the merge queue pipeline:

```
commit 5e28e732cba366736a6d0e05e4a430d967deb461 (upstream/gh-readonly-queue/main/pr-1078-2ad8a572c16be75e3878053c3ba1922ca20c5bbc)
Merge: 2ad8a572 edf8386a
Author: ROMemories <152802150+ROMemories@users.noreply.github.com>
Date:   Thu Jun 5 06:51:42 2025 +0000

    refactor(coapcore): avoid using Copy on types that should not be (#1078)
```

ROMemories: > Was that for the PR build (not the merge commit build)?

Yes, we had already checked the merge queue merge commit, but the PR build merge commit seems to have an entirely different commit subject as I was concerned about.

kaspar030: But, for the PR event, we do know the PR number (through the event data)?

Instead of juggling commit hashes, I'd prefer to a) parse current commit merge commit, for merge queue builds, and b) use the event data, for PR builds, to get the PR number.

kaspar030: offline discussion: let's go with the current version.

PR #844, opened 2025-02-20
build: stop exposing the laze builder as the `builder` `cfg` flag

ROMemories:
# Description

This removes usage of the `builder` `cfg` flag and stops exposing it in the laze configuration.
I'm arguing this is not a breaking change as we did document the `context` `cfg` flag but did not document that `builder` `cfg` flag:

https://github.com/ariel-os/ariel-os/blob/d2b093e93c86214d9b8c9c5c1a7c441ecf537595/book/src/build_system.md?plain=1#L55-L56

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -27,7 +27,6 @@ contexts:
       - ?semihosting
     env:
       RUSTFLAGS:
-        - --cfg builder=\"${builder}\"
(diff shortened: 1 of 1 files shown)


bergzand (approved):
Ack, context includes the builder values in my builds.

PR #838, opened 2025-02-18
 feat(board): add support for the ST-NUCLEO-C031C6 board 

ROMemories:
# Description

This adds support for one of the smallest and cheapest 32-bit MCU and its devkit.

Thanks to #871 which allows using the interrupt executor, it's not even needed anymore to disable `debug-console` or reduce the `CONFIG_ISR_STACKSIZE` value.

## Issues/PRs references

Closes #839.
Depends on #871.
Depends on #852.
Depends on #912.

## TODO

- [x] The `bss`/`uninit` sections of the `blinky` example do not fit in RAM because of the stack sizes. (Fixed by #884 and #871)

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

Issue #640, opened 2024-12-19
refactor: Reduce board crates


chrysn:
### Description

Most of the board crates could be replaced with more descriptive components, and there could be fewer of them.

### Motivation

Simplification

### Current Implementation

Many crates.

### Are you planning to do it yourself in a pull request?

No

PR #848, opened 2025-02-21
refactor(laze): invert how capabilities are provided

ROMemories:
# Description

Right now, each capability declares the list of contexts (i.e., MCUs, boards) on which it is provided. However, this means that adding support for new MCUs/boards involves modifying the existing definitions of these capabilities, which wouldn't work for out-of-tree board definitions, which we want to support.
This PR therefore instead makes each MCU/board context list the capabilities it provides, instead of the other way around.

## Issues/PRs references

This is also related to #640.
Depends on #838.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -194,6 +194,8 @@ contexts:
 
   - name: nrf52
     parent: nrf
+    provides:
+      - has_storage_support
     env:
       CARGO_RUNNER:
         - ${SCRIPTS}/debug-openocd.sh
@@ -226,6 +228,8 @@ contexts:
 
   - name: nrf5340
     parent: nrf53
+    provides:
+      - has_storage_support
     selects:
       - cortex-m33f
     env:
@@ -244,6 +248,8 @@ contexts:
       - rp-link-arg
       - cortex-m0-plus
       - ?probe-rs
+    provides:
+      - has_storage_support
     env:
       PROBE_RS_CHIP: RP2040
       CARGO_RUNNER:
@@ -260,6 +266,8 @@ contexts:
     selects:
       - cortex-m33f
       - ?probe-rs
+    provides:
+      - has_storage_support
     env:
       PROBE_RS_CHIP: RP235x
       CARGO_RUNNER:
@@ -389,6 +397,7 @@ contexts:
       # TODO: also has a cortex-m4f
       - cortex-m7f
     provides:
+      - has_storage_support
       - has_swi
     env:
       PROBE_RS_CHIP: STM32H755ZITx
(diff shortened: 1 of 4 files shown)


laze-project.yml
```diff
@@ -192,6 +196,9 @@ contexts:
 
   - name: nrf52
     parent: nrf
+    provides:
+      - has_hwrng
+      - has_storage_support
```
ROMemories, line 201: This additionally enables storage on `nrf52832` and `nrf52833` which are supported (and support on `nrf52833` was documented in the support matrix); tested successfully on the bbc-microbit-v2.

laze-project.yml
```diff
@@ -330,6 +354,9 @@ contexts:
     parent: esp
     selects:
       - xtensa
+    provides:
+      - has_multi_core_support
+      - sw/benchmark
```
ROMemories, line 359: It's seems this doesn't allow the benchmarks to run on ESPs; looking into it.

ROMemories: This seems to be caused by the fact that now the `executor-single-thread` executor gets selected, which conflicts with `sw/threading`, which is required by the benchmarks.

ROMemories: @kaspar030 Swapping `sw/benchmark` and `sw/threading` so that `sw/threading` comes first in benchmarks' laze files worked, but I'm not sure whether this is a workaround or how we should expect it to work.

laze-project.yml
kaspar030: `ai-c3` seems wrong, I think it is `nrf52dk` instead.

laze-project.yml
kaspar030: I think the reason is that the context modules are selected after the application modules, selecting `sw/benchmark` goes through the providers of `sw/benchmark`, causing `context::esp` to be handled early, which fixes the default executor. I'm looking for a more general solution. Also, https://github.com/ariel-os/ariel-os/issues/904 doesn't make this easier, neither does the lack of insight which module got selected why.
In the meantime, IMO we can merge this with the ordering workaround.

laze-project.yml
```diff
@@ -1323,18 +1325,26 @@ builders:
 
   - name: espressif-esp32-s3-devkitc-1
     parent: esp32-s3-wroom-1
+    provides:
+      - hw/usb-device-port
```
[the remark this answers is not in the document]
kaspar030: When I wired it up, usb-serial worked but usb-ethernet did not. I tried again some days ago, usb-serial didn't enumerate anymore. I'll try again now.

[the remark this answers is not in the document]
kaspar030: Anyhow, this is orthogonal to this PR.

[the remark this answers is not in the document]
kaspar030: for the record: I tried again today, usb-serial works fine.

PR #851, opened 2025-02-24
refactor(identity): depend on `ariel-os-hal` directly

ROMemories:
# Description

This makes `ariel-os-identity` depend on `ariel-os-hal` directly instead of depending on `ariel-os-embassy`. This avoids future cyclic dependencies.

## Change checklist


chrysn (approved):
Uses less indirection, fewer workarounds, and is one of those "if it builds it is fine" PRs, no extra testing needed.

Thanks, let's go.

PR #865, opened 2025-02-26
chore(coapcore): prepare 0.1 release

chrysn:
# Description

Since #814, coapcore is usable enough standalone; let's get this out.

Publishing means that the docs should be in order, this PR mainly takes care of that before setting the 0.1.0 version.

Releasing to crates.io is manual right now, I think that's the good way (it doesn't require particular trust in the GitHub infrastructure).

## Issues/PRs references

Blocked by #814.

## Change checklist


chrysn:
It's up to projects to define what they mean by having the version number incremented on git; I think that some auto-release based on it (I generally don't).

It is a (weak because git is not crates.io) promise that this crate version is compatible with the 0.1.0-to-be on crates.io.

PR #878, opened 2025-03-01
fix(rtt): switch channel sequence

chrysn:
# Description

probe-rs refactored its internals, and the heuristics changed, and our defmt channel is not recognized any more by current probe-rs HEAD.

Upstream is very helpful and active on fixing this -- nonetheless, we need a quick workaround. Changing the order of channels should have zero practical impact, but ensures that the current heuristics match.

## Issues/PRs references

Workaround-For: https://github.com/probe-rs/probe-rs/issues/3130

The change can persist even after that is fixed: the code is no better or worse either way.

## Change checklist

### DIFF
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -86,20 +86,20 @@ mod backend {
                 up: {
                     0: {
                         size: 1024,
-                        mode: NoBlockTrim,
-                        name: "Terminal"
-                    }
-                    1: {
-                        size: 1024,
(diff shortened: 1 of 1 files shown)


chrysn:
Putting this on hold by marking it as draft: https://github.com/probe-rs/probe-rs/pull/3140 is around as a proper fix and might be in upstream before we even get eyes on this for review.

chrysn:
Things already work again with probe-rs HEAD; no broken version was released, so no need to do anything here. Thanks to the probe-rs team for handling this in an excellent way.

PR #888, opened 2025-03-05
refactor(laze): make the flash driver selection more generic

ROMemories:
# Description

This introduces a capability (`sw/async-flash-driver`) representing whether to use the async-capable flash driver when available.
Even though this is currently only used in the STM32 HAL, the capability is not prefixed with `stm32` unlike other similar capabilities because:

- This concern is not STM32-specific unlike these similar capabilities (`stm32-usb`/`stm32-usb-synopsys` and `stm32-rng`/`stm32-hash-rng`), and could be used on other HALs if needed.
- It also acts as user configuration, because Embassy provides a blocking flash driver on all STM32 MCUs, so this "capability" actually is optional and only opts into using the (optional) *async* flash driver. Therefore this capability would be a candidate for user documentation, but this is not done in this PR because we don't want to commit to using laze capabilities for this just yet.

## Issues/PRs references

Follow-up to #772.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -331,6 +331,8 @@ contexts:
     env:
       PROBE_RS_CHIP: STM32F401RETx
       PROBE_RS_PROTOCOL: swd
+      RUSTFLAGS:
+        - --cfg capability=\"async-flash-driver\"
       CARGO_ENV:
         # This ISR is unused on a naked board. Configured here for testing.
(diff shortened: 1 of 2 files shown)


kaspar030:
triple eight! 🎆

kaspar030 (approved):

Issue #900, opened 2025-03-11
feat(coap): storage setup for unilateral authentication


chrysn:
The demo setup assigns some authorizations to unauthenticated users; storage should have facilities to do the same.

This helps the tutorial, as it's a stepping stone for telling the story.

PR #902, opened 2025-03-11
feat(coap): express unauthenticated credentials

chrysn:
# Description

Users can't utilize all the coapcore facilities from ariel-os-coap yet; this one addresses giving authorizations to an unauthenticated user.

Note that this doesn't necessarily mean "insecure": Such a client could be given read-only access to a resource, and while any client can choose to go insecure, they can also start EDHOC with a KCCS, verify the server's identity, and it's about just as confidential as HTTPS (nobody other than the client and server knows what is going on precisely).

## Issues/PRs references

Closes: #900

## Testing

* Go to tests/coap, and follow the README step by step.

## Open Questions

* Should we make it easy to pull in "whatever Ariel OS thinks should be allowed" more easily? That's best answered when we actually do have OS provided services.
* Should we add a way to express "do anything but just read-only"? Probably can be done later when we have a good example (possibly on OS provided services).

## Change checklist

- [x] I have made corresponding changes to the documentation. (**Here this means mainly the READMEs**)

### DIFF
--- a/src/ariel-os-coap/build.rs
+++ b/src/ariel-os-coap/build.rs
@@ -6,7 +6,8 @@ use std::fmt::Write;
 /// (The top level is a list thereof).
 #[derive(Deserialize)]
 struct Peer {
-    kccs: String,
+    kccs: Option<String>,
+    from: Option<KnownSource>,
     scope: Scope,
 }
 
@@ -61,6 +62,12 @@ enum KnownScope {
     AllowAll,
 }
 
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+enum KnownSource {
+    Unauthenticated,
+}
+
 fn main() {
     if !build::cargo_feature("coap-server-config-storage") {
         return;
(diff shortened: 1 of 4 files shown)


chrysn:
Rebased onto main to ease testing (now it's easier to use different ports without stack overflows).

chrysn:
There's one caveat still around that is that the published version of aiocoap doesn't use the latest Lakers version yet and thus doesn't understand the credentials. I'll fix that by just releasing there; you could already test if you have a local version of aiocoap main branch.

ROMemories (approved):
Ran the `test/coap` test and used the following successfully:
```
pipx run --spec 'aiocoap[oscore,prettyprint]' aiocoap-client coap://<ip_address>/poem
```
Then modified the `client.diag` as instructed and ran the same command but with `--credentials client.diag` successfully, so I suppose that works?

chrysn:
I've released new versions of aiocoap over the last few days that would enable the remaining cases. Thanks for testing!

Issue #906, opened 2025-03-14
macOS build instructions


Ollrogge:
Hey guys, any specific reasons why the [getting started](https://ariel-os.github.io/ariel-os/dev/docs/book/getting_started.html) is currently Linux only ? 
I got the hello-world & blinky examples running without any problems on my Apple silicon Mac 👍

kaspar030:
No reasons other than noone took the time to write it up. PR welcome! 🙂

Ollrogge:
>
> That's excellent to hear. Would you have the time to write it up into a PR for us? Or maybe just jut the necessary steps down here so that we can write them into the guide?

Sure :) will write it up and open a PR this weekend

Ollrogge:
Closing since #909 got merged :)

PR #909, opened 2025-03-15
docs(book): Add macOS to getting started

Ollrogge:
# Description
This PR adds instructions for getting started with Ariel-OS on macOS. I built the `hello-world` example for all available boards and found that `gcc-arm-none-eabi` and `gcc-riscv64-unknown-elf` were not required, so I removed them from the macOS install command. 

Additionally, I made minor adjustments to the `rustup` installation commands since it was listed twice—once in step 1 and again in step 2. I also believe the recommended installation method is via the official website.

## Issues/PRs references
#906


kaspar030 (approved):
LGTM, thanks!

PR #914, opened 2025-03-18
refactor!: lower stacksize defaults to 1KiB for ISRs, 3KiB for executor

kaspar030:
# Description

This lowers the stacksize defaults, looking at #838.

## Issues/PRs references

Required by #838.

## Open Questions

This needs proper testing at lease on all platforms unfortunately, for all examples and tests that don't already increase the stacksizes. And, test with defmt and log, isr executor and stack executor.

- [ ] nrf52xxx
- [ ] rpi-pico
- [ ] esp32
- [ ] esp32s3
- [ ] esp32c3
- [ ] esp32c6
- [ ] stm32

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -63,9 +63,9 @@ contexts:
         # These defines the defaults. Other modules can append their needs,
         # the maximum value will be chosen.
         # Note: make sure to append to the list, not set single value!
-        - "8192"
+        - "3072"
       isr_stacksize_required:
-        - "2048"
+        - "1024"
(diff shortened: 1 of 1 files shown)


kaspar030:
The `executor-isr` should hard-fail on cortex-m and riscv. Thread stack overflow is not so easy to notice currently :(

ROMemories:
Renamed the PR:
- This is a breaking change, as it affects any out-of-tree application just as well
- The new title uses actual units

emmanuelsearch:
@ROMemories @kaspar030 can others help test this on the hardware they have? If so, what should be tested with which expected output?

PR #915, opened 2025-03-18
feat(embassy-net): always enable IPv6 support

ROMemories:
# Description

Enable support for IPv6 in `embassy-net`. Until now it was only enabled by other features, like `coap`.

## Open Questions

- Opening as draft as we need to consider whether this needs feature-gating.

## Change checklist


laze-project.yml
```diff
@@ -989,33 +989,80 @@ modules:
...
     selects:
-      - ?network-config-dhcp
-      - ?network-config-static
-      - network-config
+      - ?network-config-dhcpv4
+      - ?network-config-static-ipv4
+      - network-config-ipv4
 
   - name: network-config-dhcp
+    help: "deprecated: use `network-config-dhcpv4` instead"
```
ROMemories: Doesn't work for deprecation, when attempting to disable a module

src/ariel-os-embassy/Cargo.toml
```diff
@@ -98,6 +97,12 @@ random = ["dep:ariel-os-random", "dep:rand_core"]
 ## Use a hardware RNG to seed into the ariel-os-random system-wide RNG
 hwrng = ["ariel-os-hal/hwrng"]
 
+## Enables support for IPv4.
+ipv4 = ["embassy-net/proto-ipv4"]
+## Enables support for IPv6.
+ipv6 = ["embassy-net/proto-ipv6"]
```
ROMemories: @chrysn Should these features be enabled by `ariel-os-coap` instead of enabling them on `embassy-net` directly? I'm not sure how [this comment](https://github.com/ariel-os/ariel-os/blob/672b4b5ba9da80c4ca1def09ba8358c7915f9e23/src/ariel-os-coap/Cargo.toml#L14-L23) relates to this.

laze-project.yml
```diff
@@ -1088,8 +1088,9 @@ modules:
 
   - name: network
     selects:
+      - ?ipv4
+      - ip
```
ROMemories: @kaspar030 Why does this seem to also unconditionally select the `ipv6` laze module? (The goal is only to enforce that at least one of `ipv4` or `ipv6` is selected.)

src/ariel-os-embassy/Cargo.toml
chrysn: CoAP itself doesn't specifically need IPv4 or IPv6; it's only the embedded-nal embassy-net glue code that'd go into a mess of ifdefs if things were inconsistent.

So I'd think that unless having more features enabled in embassy-net than Ariel demands is a problem, I think that the CoAP code is fine to still enable only embassy-net/proto-ipv[46], and (through laze) ask that whichever IP version(s) are default are on (practically it needs one).

laze-project.yml
kaspar030: It selects "ip", so any module that provides that, which ipv6 does.

ROMemories: Removed this for now; it doesn't really seem possible to enforce that at least `ipv4` or `ipv6` is enabled, but that's not really an issue either (it just wouldn't compile).

src/ariel-os-embassy/src/net.rs
```diff
@@ -165,35 +165,73 @@ impl embassy_net::driver::RxToken for DummyDriver {
...
+        } else if #[cfg(feature = "dhcpv4")] {
+            config.ipv4 = embassy_net::ConfigV4::Dhcp(embassy_net::DhcpConfig::default());
+        }
+    }
+
+    #[cfg(feature = "network-config-ipv6-static")]
+    {
+        let ipaddr = ariel_os_utils::ipv6_addr_from_env!(
+            "CONFIG_NET_IPV6_STATIC_ADDRESS",
+            "static IPv6 address",
```
bergzand: We have defaults for the IPv4 address, why not also for IPv6?

src/ariel-os-embassy/src/net.rs
bergzand, line 1: Can we add a `compile_error` in this file when both ipv4 and ipv6 are disabled. When both are disabled, the first build error I receive is from Smoltcp:
```
error: You must enable at least one of the following features: proto-ipv4, proto-ipv6, proto-sixlowpan
  --> /home/koen/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/lib.rs:99:1
   |
99 | compile_error!("You must enable at least one of the following features: proto-ipv4, proto-ipv6, proto-sixlowpan");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This hints in the right direction, but an error from our side with the correct laze features will help the unsuspecting user.

src/ariel-os-embassy/src/net.rs
ROMemories: What kind of defaults would make sense?

src/ariel-os-embassy/src/net.rs
ROMemories: I've tried adding the following here:

```rust
#[cfg(not(any(feature = "ipv4", feature = "ipv6")))]
compile_error!("Both IPv4 and IPv6 have been disabled. At least one of them must be enabled.");
```

However that doesn't work because the `smoltcp` (transitive) dependency is compiled *before* this crate. I don't think we can do anything about it, except waiting for laze to support requiring that at least one of `ipv4` and `ipv6` is enabled.

src/ariel-os-embassy/src/net.rs
bergzand: [`2001:db8::1/64`](https://www.rfc-editor.org/rfc/rfc3849) would be fine here

src/ariel-os-embassy/src/net.rs
bergzand: Makes sense. Nothing to do about it for now then.

src/ariel-os-embassy/src/net.rs
```diff
@@ -165,35 +165,73 @@ impl embassy_net::driver::RxToken for DummyDriver {
...
+        let gw_addr = ariel_os_utils::ipv6_addr_from_env!(
+            "CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS",
+            "static IPv6 gateway address",
+        );
+
+        let prefix_len = ariel_os_utils::u8_from_env_or!(
+            "CONFIG_NET_IPV6_STATIC_CIDR_PREFIX_LEN",
+            64,
+            "static IPv6 CIDR prefix length"
+        );
```
bergzand, line 229: Does it make sense to do some build-time sanity checking on <= 128?

ROMemories: Indeed! Done as follow-up in https://github.com/ariel-os/ariel-os/pull/1569 because we can actually do the same for IPv4.

src/ariel-os-embassy/src/net.rs
ROMemories: Do we really want to use a non-routeable address as a default? That means that users could just forget to pass the address they actually want, instead of getting a compile-time error.

ROMemories: Also, don't we always need a gateway address as well? Would we use `2001:db8::2`?

bergzand: > Do we really want to use a non-routeable address as a default? That means that users could just forget to pass the address they actually want, instead of getting a compile-time error.

My intent here is to have similar behaviour between IPv4 and IPv6. IPv4 has a default for the address, so having that also for IPv6 makes sense to me. The documentation prefix is for these kind of purposes. The IPv4 is also not routeable, so I don't see an issue there. 

> Also, don't we always need a gateway address as well? Would we use `2001:db8::2`?

Mirroring the IPv4 side, we could use `2001:db8::61` as address and `2001:db8::1` as gateway

ROMemories: Added these defaults in a new commit, and updated the documentation in #1377.

The following now compiles without providing any environment variables:

```sh
laze -C examples/udp-echo build -s ipv6 -b nrf52840dk
```

bergzand (approved):
LGTM, tested the supplied build variants

src/ariel-os-embassy/src/net.rs
chrysn: Just for the record: I think this default was a bad outcome due to not being routable -- that kind of broken-ness is widespread in IPv4 land, but unexpected in IPv6. (I'd also advocate for not having any IPv4 defaults, but it's IPv4 so I don't particularly care).

Follow-up happening in https://github.com/ariel-os/ariel-os/pull/1377/files#r2565438614 or wherever things go from there.

Issue #724, opened 2025-01-24
feat: ipv6 support


kaspar030:
- [x] support both (static) ipv6 and ipv4
- [x] enable choosing between either ipv4 and/or (static) ipv6 config (at build-time)

McScotti:
Hi,
I am working on this issue , please can explain the second point of your text ? Because I think that when , static configuration has been chosen , i have initialized both static ipv4 configuration and static ipv6 configuration .

McScotti:
hi,
So sorry , I made a mistake in my code . I think it is not possible to initialize both (static) ipv4 and static (ipv6) configurations.
So I just want more details about the second point of the text , because we can already choose to have a static configuration with ipv4 , and do you mean to just do the same with ipv6 that's to say we create a new laze module for that?
.

emmanuelsearch:
@kaspar030 by in the queue you mean based on what @bergzand is upstreaming to smoltcp / embassy-net, basically https://github.com/smoltcp-rs/smoltcp/pull/1039 right?

PR #916, opened 2025-03-18
feat:ipv6 support 

McScotti:
# Description
Enable choosing between either ipv4 or (static) ipv4 and ipv6 config  (at build-time)

## Issues/PRs references

Feat #724 
## Change checklist

### DIFF
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -23,6 +23,8 @@ embassy-hal-internal = { workspace = true }
 embassy-net = { workspace = true, optional = true, features = [
   "dhcpv4",
   "medium-ethernet",
+  "proto-ipv4",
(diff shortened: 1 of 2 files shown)


kaspar030:
Thanks for tackling this! I'm not sure we can come up with a sensible default address though. Maybe only set & use one if it is provided (via env)?

McScotti:
yes it is possible , but if we make it like that , the build will fail if the whole configuration is not provided (via env)
but yes we can

chrysn:
I like that this doesn't even try to pick a default address, and removes the v4 default addresses: there is no sensible default for those things.

chrysn:
As I understand, this was replaced with https://github.com/ariel-os/ariel-os/pull/936.

McScotti:
exact , there was an error in this version i prefered to put a new version

PR #917, opened 2025-03-18
refactor(power)!: simplify the `power::reboot()` implementation

ROMemories:
# Description

This simplifies the `reboot()` implementation:
- Removes the `ariel-os-power` crate and moves the the implementation into `ariel-os-embassy`
- Removes the `power` Cargo feature

It also adds a README file to the example and adds it to the list in `examples/README.md`.

BREAKING CHANGE: this removes the `power` Cargo feature

## Hardware testing

Successfully tested the `power` example on:

- bbc-microbit-v2
- espressif-esp32-c6-devkitc-1
- nrf52840dk
- nrf5340dk
- rpi-pico-w
- st-nucleo-h755zi-q
- st-nucleo-wb55

## Change checklist


kaspar030:
Quite drastic refactor. :slightly_smiling_face: 

Let's keep in mind future additions. There's at least `power_off()` and `boot_reason()`. Next up is some low power mode handling - power modes (idle, retain ram, deep sleep (but can wake up from wakeup sources like rtc alarm or button), off available everywhere, plus possible arch intermediate modes).

1. crate vs. module - I think we should keep the crate so we don't clutter `ariel-os-embassy` more
1. There's an upcoming dependency issue. `ariel-os-embassy` depends on `-threads`. `-threads` will need to interact with the power module. So if we move that into embassy, we'll have to split it out later again.
1. Let's keep `ariel-os-power` (or module) as the API facing module, and defer for implementations to the hals (or rt)? (maybe trait up at some point ...) I'd like to keep the dispatching throughout the codebase to a minimum. 
1. removing the feature is fine, no need to gate `reboot()`. we can add back the feature later if it makes sense.

PR #925, opened 2025-03-20
 refactor(stm32): unify the USB module to ease future changes 

ROMemories:
# Description

This merges the previously split USB modules in `ariel-os-stm32` to make future changes easier. Such future changes include supporting other USB OTG variants. This should also allow the changes from #919 to actually be usable.

# Hardware testing

Tested successfully on st-nucleo-h755zi-q and st-nucleo-wb55.

## Issues/PRs references

Depends on #919.

## Change checklist


src/ariel-os-stm32/src/usb.rs
```diff
@@ -19,13 +36,55 @@ impl Peripherals {
...
     }
 }
 
+#[cfg(not(capability = "hw/stm32-usb-synopsis"))]
 pub fn driver(peripherals: Peripherals) -> UsbDriver {
     Driver::new(peripherals.usb, Irqs, peripherals.dp, peripherals.dm)
 }
+
+#[cfg(capability = "hw/stm32-usb-synopsis")]
+pub fn driver(peripherals: Peripherals) -> UsbDriver {
```
ROMemories, line 55: I didn't attempt to merge these two `driver` functions as I don't think that would be clearer.

src/ariel-os-stm32/src/usb.rs
```diff
@@ -1,16 +1,33 @@
 use embassy_stm32::{bind_interrupts, peripherals, usb, usb::Driver};
 
 bind_interrupts!(struct Irqs {
+    #[cfg(capability = "hw/stm32-usb-synopsis")]
```
ROMemories, line 4: I think we should rename this capability to allow for the other OTG variants, but this is for a follow-up.

kaspar030 (approved):
LGTM, trusting your testing.

Issue #858, opened 2025-02-25
Support out-of-tree board definitions


ROMemories:
Currently, applications can be defined out of tree.

We also want to allow users to define their own boards out of tree.
However, supporting out-of-tree MCU definitions is not a goal.

This is related to #639 as it must be possible for users to specify different flash (and RAM) sizes, when using external flash chips (e.g., on the RP2040).

- Requires #848.
- Relates to #859.
- Relates to #210.
- Relates to #639.

PR #928, opened 2025-03-21
feat(stm32): codegen a crate enabling the right `embassy-stm32` feature

ROMemories:
# Description

This introduces a new crate for STM32, whose job is to enable the right `embassy-stm32` MCU Cargo feature, based on the MCU laze context. This is needed because the laze context contains the package and temperature range suffixes (e.g., `T6`) because this is currently needed for probe-rs, while the `embassy-stm32` Cargo features do not, as these are not functional properties for software (the pin count has its own identifier, separate from the package).

Because there is currently 1400 supported STM32 MCUs, with 2207 package+temperature range variants, this crate is codegened from data extracted from [`embassy-rs/stm32-data-generated`](https://github.com/embassy-rs/stm32-data-generated) (see the comment in the build script).

The codegen logic is currently in a build script which populates that crate's Cargo manifest. Because all dependencies must be known to Cargo when it starts running, the resulting manifest is also committed (so it gets overwritten by the build script with exactly the same content, which isn't a problem). Changes to the build script or the data requires compiling twice: the first time to update the Cargo manifest, the second time so the changes get picked up by Cargo.
We could later move this to a Cargo script, to possibly slightly reduce build time, but this is not done because the DX of Cargo scripts is currently subpar compared to build scripts (e.g., no `cargo fmt`, no Clippy).

I don't necessarily see this approach as the endgame solution, but this seems like a useful block to work towards #858.

## Open Questions

- How to name that crate (currently `ariel-os-smt32-mapping`)?

## Change checklist


src/ariel-os-stm32/Cargo.toml
```diff
@@ -29,18 +30,6 @@ ariel-os-embassy-common = { workspace = true }
...
-
-[target.'cfg(context = "stm32h755zitx")'.dependencies]
-embassy-stm32 = { workspace = true, features = ["stm32h755zi-cm7"] }
-
-[target.'cfg(context = "stm32wb55rgvx")'.dependencies]
-embassy-stm32 = { workspace = true, features = ["stm32wb55rg"] }
-
-[target.'cfg(context = "stm32wba55cgux")'.dependencies]
-embassy-stm32 = { workspace = true, features = ["stm32wba55cg"] }
-
```
ROMemories, line 43: The whole point is to codegen this.

src/ariel-os-stm32-mapping/build.rs
```diff
@@ -0,0 +1,73 @@
...
+            // Remove the trailing quotation mark.
+            mcu_base.pop();
+            format!("{mcu_base}-cm7\"")
+        } else if CM0P_CM4_MCU_PREFIXES
+            .iter()
+            .any(|prefix| skip_leading_quotation_mark.starts_with(prefix))
+        {
+            // Remove the trailing quotation mark.
+            mcu_base.pop();
+            format!("{mcu_base}-cm4\"")
```
ROMemories, line 60: This selects the Cortex-M4 core on the MCUs listed in `CM0P_CM4_MCU_PREFIXES`, which seems to be the main core. This is similar to what we're already doing for the MCU we currently support in `CM4_CM7_MCU_PREFIXES` (`stm32h755zi-cm7`).

src/ariel-os-stm32-mapping/Cargo.toml
ROMemories, line 1: Even though GitHub folds this one, you still want to review this file (at least the `package` section and the delimiter comments).

src/ariel-os-stm32-mapping/stm32_mapping.csv
ROMemories, line 1: This file is generated as explained in the build script.

src/ariel-os-stm32/Cargo.toml
```diff
@@ -9,6 +9,7 @@ repository.workspace = true
 workspace = true
 
 [dependencies]
+ariel-os-stm32-mapping = { path = "../ariel-os-stm32-mapping" }
```
ROMemories, line 12: We could technically codegen this very manifest instead of introducing a new crate, but it seemed cleaner as (a) this avoids a 10k-line manifest which we still need to manually edit and (b) if we wanted to switch to another approach in the future, we could simply remove the entire codegened crate.

kaspar030:
Cool!

I'm not happy about the `build.rs` modifying `Cargo.toml` in-place. Don't have an alternate solution though, a `./generate.rs` cargo script also has downsides.

LGTM.

kaspar030 (approved):

Issue #242, opened 2024-04-10
Roadmap


kaspar030:
This is the Roadmap tracking issue.
Taken mostly unmodified from the initial [workplan](https://pad.riot-os.org/riot-rs-workplan?both#)

## Development tools

- [ ] Build system
  - [ ] evaluate & confirm or drop use of laze
- [ ] Testing Framework

## Core OS functionality

- [ ] Multicore support
  - [ ] Async executor on second core
  - [ ] Scheduler support for "regular" threads on both cores
- [ ] unify "console"
- [ ] Support defmt

## Embassy integration

- [x] Dispatch to correct embassy init()
- [x] Support per-board ~~drivers~~ initialization -> covered through riot-rs-boards & autostart tasks
- [x] Support application portability
  - [ ] 1. "portable" blinky
     This needs simple gpio or led abstraction (enumeration?)
 - [ ] 2. "portable" sensor reading
     This needs a way (in build system, code) to easily add sensor drivers.
 - [x] "portable" simple network application
 - [ ] make 2. / 3. networked
- [x] Get back linkme support

## Hardware Support

- [x] Extended Cortex-M support
  - [x] rpi-pico support
- [ ] RISCV support
  - [ ] Add any risc-v board
  - [ ] riot-rs-rt bringup
  - [ ] riot-rs-threads riscv task switching [PR](https://github.com/future-proof-iot/RIOT-rs/pull/182)
- [ ] "native" support
  
## RIOT-c support

(postponed)

## Basic (Layer<5) networking support

- [ ] Basic (Layer<5) networking support
- [ ] TCP/IP and UDP
- [ ] 802.15.4 / 6lowpan
- [ ] Bluetooth
- [ ] LoRa

## Extended (Layer>4) networking + security 

- [ ] MQTT
- [ ] [Low-power IPv6 IETF stack (CoAP/OSCORE/EDHOC)](./226)
- [ ] 802.15.4 link layer security
- [ ] TLS and DTLS
- [ ] Crypto library 

## Device Livecycle

- [ ] On-boarding
  - [ ] TOFU
  - [ ] Zero-Touch
- [ ] SUIT

## Extend embassy integration

- [ ] low-power
- [ ] Multiple timers
- [ ] ~~Board support trait~~ not using a trait

## Extend hardware support

- [ ] re-think riot-rs-boards
- [ ] nrf5340
- [ ] stm32

## CI

- [ ] set up CI testing on hardware

[...]

PR #930, opened 2025-03-21
docs(roadmap): Convert roadmap issue to file 

bergzand:
# Description

This commit converts https://github.com/ariel-os/ariel-os/issues/242
into a roadmap file. Implementation details have been removed and
descriptions *why* a feature is required and on the roadmap have been
added.

## Issues/PRs references

See #242 for the original

## Open Questions

There are a few open sections that do not explain yet *why* we want a certain feature:
- [Unified Console](https://github.com/ariel-os/ariel-os/compare/main...bergzand:Ariel-os:pr/roadmap.md?expand=1#diff-683343bdf93f55ed3cada86151abb8051282e1936e58d4e0a04beca95dff6e51R21)
- [Zero-touch onboarding](https://github.com/ariel-os/ariel-os/compare/main...bergzand:Ariel-os:pr/roadmap.md?expand=1#diff-683343bdf93f55ed3cada86151abb8051282e1936e58d4e0a04beca95dff6e51R125)

## Change checklist

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,157 @@
+# Ariel OS Roadmap
+
+This document contains the roadmap for Ariel OS.
+It provides a high-level overview of the project's strategy
+and what the project wants to achieve in the next years.
+
+Please take a look at the [contributing guidelines]
+when you want to start contributing to one of these goals.
+
+## Development tools & Documentation
+
+### Extensive Laze documentation
+
+As Laze is a core component of the Ariel OS workflow,
+extensive documentation for both users and developers is required.
+The current documentation of Laze is lacking, with multiple blank pages in the
+manual.
+
+## OS functionality
+
+### Unified Console
+
+### Common Sensor Interface
+
+A consistent access method for sensors provides a convenient and portable way
+for users to access readings.
+This allows for generic access to different sensors and makes it easy to export
+these sensor readings via different protocols.
+
+See also: https://github.com/ariel-os/ariel-os/pull/474
+
+### Over the air updates
+
+OTA updates are required to fix bugs, including vulnerabilities, on devices
+deployed in the field.
+
+### Stack usage insight
+
+Insights and metrics on stack usage in Ariel OS provides the developer with
+valuable information on memory usage and where it can be decreased.
+
+In the future this can also be extended with automated tuning and verification.
+
+## Application portability
+
+- [ ] 2. "portable" sensor reading
+       This needs a way (in build system, code) to easily add sensor drivers.
+- [x] "portable" simple network application
+- [ ] make 1. / 2. networked
+
+## Hardware Support
+
+### Native hosted mode
+
+Running Ariel OS as native process on the developers computer without requiring
+extra hardware provides a useful way to test and debug OS modules.
+
+See also: 
+ - https://github.com/ariel-os/ariel-os/issues/617
+ - https://github.com/ariel-os/ariel-os/pull/647
+
+### Stack overflow detection
+
+Detecting stack overflows in threads protect against vulnerabilities and other
+bugs around stack overflows.
+
+### Sandboxed thread mode
+
+Fully sandboxing a thread via hardware mechanisms allow for isolated threads.
+This improves the resiliency against memory bugs and can protect critical OS
+threads against potentially untrusted threads.
+
+## RIOT-c support
+
+Improving compatibility with RIOT APIs provide cross OS compatibility.
+
+## Basic (Layer<5) networking support
+
+### Bluetooth
+
+Bluetooth provides a low power mechanism via which to interface
+with other computers such as smartphones.
+It allows for exposing multiple standardized functionalities
+such as sensors and human interface devices.
+
+### IEEE 802.15.4 / 6LoWPAN
+
+This provides a standardized multi-hop network mechanism
+geared towards low power and low throughput devices.
+
+### LoRa
+
+LoRa support provides a mechanism through which simple low power sensor devices
+can be deployed in the field and report their measurements to a central
+authority.
+
+## Extended (Layer>4) networking + security 
+
+### MQTT client
+
+MQTT is a popular publish-subscribe protocol for communicating with constrained
+devices. One of the users is [Home-Assistant] where it is used extensively for
+provisioning and sensor/actuator interaction
+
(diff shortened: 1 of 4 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,156 @@
...
+### Extensive Laze documentation
+
+As Laze is a core component of the Ariel OS workflow,
+extensive documentation for both users and developers is required.
+The current documentation of Laze is lacking, with multiple blank pages in the
+manual.
+
+## OS functionality
+
+### Unified Console
```
chrysn, line 21:
```suggestion
### Unified Console
```

I'm not sure where this is coming from. I do think we should eventually be able to have any stdout or log output equally available over the network. If that's the intent, I can provide text, but it may be about something different as well.

chrysn, line 21: I'm not sure where this is coming from. I do think we should eventually be able to have any stdout or log output equally available over the network. If that's the intent, I can provide text, but it may be about something different as well.

bergzand: I think that's the intent, but I'm not familiar enough with the intent here to write a motivation.

chrysn:
```suggestion
### Unified Console

Currently, console (both std/log and defmt) only prints to a fixed output, typically rtt.
Instead, different modules should be capable of registering at build time, and deliver log
output as they suitable, eg. into storage or into a ring buffer for the developer to attach
to over the network.
```

But pinging @kaspar030 on this (because IIUC he put this in the roadmap issue and should thus be the source of truth)

chrysn:
On the topic that came up during today's meeting on "we have too many places": Might make sense to have this as part of the book, it would fit well next to the "hardware support" page.

bergzand:
> On the topic that came up during today's meeting on "we have too many places": Might make sense to have this as part of the book, it would fit well next to the "hardware support" page.

I think that's a great idea and works well here. I went with `ROADMAP.md` as simple initial step here.

ROADMAP.md
kaspar030: My initial aim for this was to have a way to have output for all our devices. To me this has been solved with the log backend in `ariel-os-debug`.
network logs and enduser-facing application logs would be great to have, but each deserve their own entry. So I'd say, lets remove this.

chrysn: Fine with me. Network logging is a quality-of-life improvement that I want to get in, but don't see a strong necessity to have on the road map.

kaspar030:
I'm not happy with this file as is, not sure it can be monged into form.

1. The representation loses the checkboxes, which makes it difficult to visualize multi-step entries with some parts already done.
2. It loses the option to link to PRs (in-flight or merged) and githubs visual states for those.
3. It is way more static, and is thus prone to not get updated.
4. We already have the Delta. We have procedures around that. There is automation around that. There is information in there (assignees, dependencies, ...).
5. Managing this in a PR might even lead to conflicts.

IMO we will need GH based lists / projects for dynamic updating, where a file in the repo is too rigid.

... yadda yadda only put high-level targets there, that could work.

ROADMAP.md
bergzand: Removed it from the document. We can always add something new in there later if we decide on it.

chrysn:
> thus prone to not get updated.

But it affords a way to get updated. Right now we have the issue, where most of us have edited the top post. This has no way for outside contributors to mark something as solved or add it (granted, is not happening yet, but we already gain outside contributors for small items), and I for one am always at odds with editing consensus documents where there's no clear review process.

bergzand:
> The representation loses the checkboxes, which makes it difficult to visualize multi-step entries with some parts already done.

Those can be added back in if they make sense at this level, for the different on-boarding mechanisms for example.

> It loses the option to link to PRs (in-flight or merged) and githubs visual states for those.

Yup, is that a deal breaker for you?

> It is way more static, and is thus prone to not get updated.

A roadmap is on the scale of a year, it is not supposed to get updated every day or week. Maybe once a month on average. So I don't see it as an issue that it is way more static. Also, for an outsider perspective, theres not much difference, a non-maintainer would have to ask nicely somehow to get a checkbox marked.

> We already have the Delta. We have procedures around that. There is automation around that. There is information in there (assignees, dependencies, ...).

This complements the delta.
- The delta is for us to track our work. This roadmap is for users to show them where we want to go. 
- The delta is at PR/issue level, the roadmap is a layer above that stating coarse goals. This is why I stated around Bluetooth *why* we want bluetooth. The PRs will probably be some integration of Trouble given what we've discussed, but for the roadmap that's an implementation detail.

> Managing this in a PR might even lead to conflicts.

I find that *highly* unlikely to happen. It's not an often changed document. This sounds to me like you also don't want our book in the repo because that might lead to conflicts. And if you want to stick with this train of thought, race conditions also happen when editing issues.

> ... yadda yadda only put high-level targets there, that could work.

I don't follow your train of thought here.

ROADMAP.md
```diff
@@ -0,0 +1,154 @@
+# Ariel OS Roadmap
+
+This document contains the roadmap for Ariel OS.
+It provides a high-level overview of the project's strategy
+and what the project wants to achieve in the next years.
```
jkarinkl, line 5: I see the roadmap and the practical goals we want to reach, but do not see the higher level strategy. Would be good to add that too. If we do not have/do that yet, and will not have that soon in text, than maybe better to delete the 'strategy' here.

jkarinkl (commented):
Nice to put this into a separate document! 

I would like to see long term strategy added, the WHY, as a mission and vision (10 yr vision, coming 3 years mission where we want to be at) and the roadmap as the HOW of this. Adding up to the WHAT into the concrete PR's we're working on.

Example of Mautic [mission and vision](https://mautic.org/vision-mission/) and [roadmap](https://mautic.org/roadmap/).

bergzand:
And now in a pad for collaborative nagging: https://pad.riot-os.org/U3ZrlJDHRWuH694EEYZABQ#

emmanuelsearch:
@bergzand should we shelf this PR for now?

kaspar030:
> @bergzand should we shelf this PR for now?

Yes, let's proceed with PR's to the newly merged file.

ROMemories:
Wouldn't links to the roadmap still need to be updated?

Issue #924, opened 2025-03-19
Reduce the `.bss` size of `init_task()`


ROMemories:
We should reduce the impact on the `bss` section of our `init_task()` task from `ariel-os-embassy`, which comes from the size of the task's `Future`, as we statically allocate tasks.

The two following strategies can be useful:

- Reducing the size of data kept over `.await` points (experiment with [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=32d7c6e1900c072c0102888d6e5bafa9) to get an idea of that involves).
- Removing the `.await` points over which the most data is kept (#653 could be useful in this direction), as the size of a `Future` enum is the size of its bigger variant (plus the tag/discriminant). 

We could also consider using `block_on` as we may not need concurrent execution during startup, or maybe handling this async startup code completely differently.

PR #931, opened 2025-03-21
perf(storage): block on storage init to spare RAM

ROMemories:
# Description

This reduces the size of the startup task ~~on non-RP MCUs~~, which is statically allocated and therefore ends up in the `bss` section, while only running at startup.

The following table summaries the change in `bss` size of the `storage` example:

| Board              | Old `bss` size | New `bss` size | Change       |
| --------           | -------------: | -------------: | -----------: |
| nrf52833dk         | 5140           | 4220           | -920 (17 %)  |
| nrf52840dk         | 5140           | 4220           | -920 (17 %)  |
| nrf5340dk          | 5124           | 4204           | -920 (17 %)  |
| rpi-pico-w         | 6012           | 4780           | -1232 (20 %) |
| st-nucleo-h755zi-q | 5752           | 4832           | -920 (16 %)  |
| st-nucleo-wb55     | 5288           | 4368           | -920 (17 %)  |

The `.text` size also decreases by a few hundred bytes.

I didn't observe any performance difference at startup, but I also didn't measure it properly.

## Testing

Successfully tested in hardware on the boards listed in the table.

## Issues/PRs references

Part of #924.

## Change checklist

### DIFF
--- a/src/ariel-os-embassy/src/lib.rs
+++ b/src/ariel-os-embassy/src/lib.rs
@@ -190,8 +190,10 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     // Clock startup and entropy collection may lend themselves to parallelization, provided that
     // doesn't impact runtime RAM or flash use.
 
+    // Block on this Future to reduce the size of this startup task, which is statically
+    // allocated.
     #[cfg(feature = "storage")]
-    ariel_os_storage::init(&mut peripherals).await;
+    embassy_futures::block_on(ariel_os_storage::init(&mut peripherals));
 
     #[cfg(all(feature = "usb", context = "nrf"))]
     hal::usb::init();
--- a/src/ariel-os-storage/src/lib.rs
+++ b/src/ariel-os-storage/src/lib.rs
@@ -87,7 +87,7 @@ pub async fn init(p: &mut OptionalPeripherals) {
     // interferes with flash write operations.
     // https://github.com/knurling-rs/defmt/pull/683
     #[cfg(context = "rp")]
-    embassy_time::Timer::after_millis(10).await;
+    embassy_time::block_for(embassy_time::Duration::from_millis(10));
 
     // Use a marker to ensure that this storage is initialized.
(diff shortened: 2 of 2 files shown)


src/ariel-os-embassy/src/lib.rs
```diff
@@ -191,7 +191,15 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     // doesn't impact runtime RAM or flash use.
 
     #[cfg(feature = "storage")]
-    ariel_os_storage::init(&mut peripherals).await;
+    {
+        // Block on this Future to reduce the size of this startup task, which is statically
+        // allocated.
+        #[cfg(not(context = "rp"))]
+        embassy_futures::block_on(ariel_os_storage::init(&mut peripherals));
+        // The flash driver on RP does not seem to work with `block_on`.
```
ROMemories, line 199: It panics with the following message:
```
Found waker not created by the Embassy executor. `embassy_time::Timer` only works with the Embassy executor.
```

kaspar030: It might be possible to move out the 10ms delay from that init function.

(I thought with the latest embassy-time this should actually work. Would you try this also with `executor-thread` and our `asynch::block_on`?)

ROMemories: Oh indeed, I totally misattributed that! I just replaced that delay with a blocking wait and from what I can judge it seems to be working. I removed this special treatment of RP.

ROMemories: > Would you try this also with `executor-thread` and our `asynch::block_on`?)

I can't make it work with `-s executor-thread`, an async delay in our storage module and our `asynch::blocker::block_on()`.

kaspar030 (approved):
lgtm

src/ariel-os-embassy/src/lib.rs
kaspar030: Ok, thanks for trying!

PR #933, opened 2025-03-21
fix(coap): remove needless use of ReportingHandlerBuilder

chrysn:
# Description

Removes needless `use` items.

The need for those went away when `.with_wkc()` was moved into ariel_os::coap.

## Open Questions

* Why didn't linting catch those? Maybe things are not so strict for examples.

## Change checklist


chrysn:
Self-reviewed; queuing for merge under the "this is so trivial I can't figure out why anyone would do more than look at it and ACK it" policy.

PR #936, opened 2025-03-24
feat(network): (static)ipv6 support

McScotti:
# Description

Enable choosing static ipv4 and ipv6 configuration at  build time
## Issues/PRs references
feat #724

## Change checklist


src/ariel-os-utils/src/env.rs
```diff
@@ -29,8 +29,38 @@ macro_rules! define_env_with_default_macro {
     };
 }
 
+macro_rules! define_env_macro {
```
kaspar030, line 32: can we implement one macro with the other?

McScotti: I think it was a good idea to add this macro , because in a further task we could need another environment variable with no default values.

McScotti:
hello guys, what about this issue ? I have not had a feedback for a long time.

kaspar030:
> hello guys, what about this issue ? I have not had a feedback for a long time.

Sorry for the delay.
So the way this is implemented now, this requires both ipv4 and ipv6 to be configured via environment. It runtime fails if *any* of the six variables is missing. IMO that's suboptimal, for one b/c probably users would want one or the other.

Can we have it this way:

When `network-config-static` is enabled,

1. if a usable ipv4 config is provided via env, use that
2. if a usable ipv6 config is provided via env, use that
3. if both are provided, use both
4. if none is provided, compile-fail.

What do you think? (are there reasonable defaults?)

McScotti:
I think it could be a good idea to create new modules like ```network-static-config-v6``` and ```network-static-config-v4``` so that give the users the possibility to take only one static configuration . what do you think ?

kaspar030:
> I think it could be a good idea to create new modules like `network-static-config-v6` and `network-static-config-v4` so that give the users the possibility to take only one static configuration . what do you think ?

sounds perfect!

ROMemories:
Thank you for your work on this PR, we've in the meantime moved adding IPv6 configuration into the [PR that enables the use of IPv6](https://github.com/ariel-os/ariel-os/pull/915). Closing as that other PR now supersedes this one.

PR #944, opened 2025-03-25
 fix(boards): fix the context `rpi-pico-w` inherits from 

ROMemories:
# Description

The `rpi-pico-w` should inherit from `rp2040` directly instead of inheriting from `rpi-pico` because the capabilities of the `rpi-pico-w` are not a superset of those of the `rpi-pico`. For instance the GPIO25 is not connected to the onboard LED, which made the `blinky` example not work.

## Issues/PRs references

Depends on #848.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1362,7 +1362,7 @@ builders:
       - has_usb_device_port
 
   - name: rpi-pico-w
-    parent: rpi-pico
+    parent: rp2040
     provides:
       - has_usb_device_port
       - has_wifi_cyw43


laze-project.yml
```diff
@@ -1362,7 +1362,7 @@ builders:
       - has_usb_device_port
 
   - name: rpi-pico-w
-    parent: rpi-pico
+    parent: rp2040
     provides:
       - has_usb_device_port
```
kaspar030, line 1367: this is redundant but doesn't hurt

kaspar030: ah no, the opposite. :)

kaspar030 (approved):

PR #947, opened 2025-03-26
ci: free up space

kaspar030:
# Description

This adds a step to our main build plan that removes some unneeded build tooling that comes preinstalled on the github workers. This gives us ~13gb extra space to fill.

The `rm ...` takes over a minute when run in foreground, so it run in the background.
Looking at CI times with a sample size of 1 for this change, this does not seem to increase total build time.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,10 @@ jobs:
         partition: ["1/20", "2/20", "3/20", "4/20", "5/20", "6/20", "7/20", "8/20", "9/20", "10/20", "11/20", "12/20", "13/20", "14/20", "15/20", "16/20", "17/20", "18/20", "19/20", "20/20"]
 
     steps:
+      - name: Free up space
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -36,6 +36,12 @@ jobs:
         partition: ["1/20", "2/20", "3/20", "4/20", "5/20", "6/20", "7/20", "8/20", "9/20", "10/20", "11/20", "12/20", "13/20", "14/20", "15/20", "16/20", "17/20", "18/20", "19/20", "20/20"]
 
     steps:
+      - name: Free up space
+        run: |
+          du sh .
```
chrysn:
```suggestion
          du -sh .
```

ROMemories (approved):
My only concern is that this makes the workflow quite GitHub-specific, but given `-f` is provided to `rm`, this should ignore nonexistent files if that workflow were to run in another image/container, or when the GitHub image evolves.

chrysn:
Let's not kid ourselves here: Everything about .github is GitHub specific, and the only standardization route their actions format goes is that others copy it and keep adjusting as GitHub changes things. This change is the least of our lock-ins.

PR #948, opened 2025-03-26
chore: enable Clippy pedantic at workspace level

ROMemories:
# Description

This fixes the few remaining warnings of Clippy pedantic, and sets Clippy's `pedantic` lint group to `warn` at workspace level, so that it issues warnings when developing locally but fails CI (because we make these warnings errors in CI with `-- --deny warnings`).
The attributes denying `clippy::pedantic` in each crates are removed, as they are now unnecessary.

I checked that every crate from `src/` is properly inheriting lints from the workspace.

## How to review this PR

- This PR is better reviewed commit by commit.
- CI runs Clippy, failing on warnings; it can however be useful to check locally that the pedantic lints are still properly triggered, for instance by commenting out one of the `#[must_use]` attribute:

    https://github.com/ariel-os/ariel-os/blob/6d67089b91f9dcbcc0a8c51416191d4cb2c9d161/src/ariel-os-embassy/src/gpio.rs#L35-L38

## Issues/PRs references

Supersedes #279.
The required refactoring for individual HAL crates has been done in #949, #1031, #1032, and #1033.

## Change checklist

### DIFF
--- a/src/ariel-os-coap/build.rs
+++ b/src/ariel-os-coap/build.rs
@@ -51,8 +51,8 @@ impl Permission {
 impl SinglePermission {
     /// The `Tperm` unsigned integer representation of the REST-specific AIF model described in
     /// RFC9237.
-    fn mask(&self) -> u32 {
-        1 << (*self as u8 - 1)
+    fn mask(self) -> u32 {
+        1 << (self as u8 - 1)
     }
 }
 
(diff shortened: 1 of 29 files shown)


src/ariel-os-coap/build.rs
```diff
@@ -51,8 +51,8 @@ impl Permission {
 impl SinglePermission {
     /// The `Tperm` unsigned integer representation of the REST-specific AIF model described in
     /// RFC9237.
-    fn mask(&self) -> u32 {
-        1 << (*self as u8 - 1)
+    fn mask(self) -> u32 {
```
ROMemories, line 54: This isn't a breaking change because this is only a build script.

src/ariel-os-runqueue/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 #![cfg_attr(not(test), no_std)]
 #![deny(missing_docs)]
+#![expect(clippy::cast_possible_truncation)]
```
ROMemories, line 5: May or may not be addressed later; this module will eventually be proven panic-free anyway.

src/ariel-os-coap/build.rs
chrysn: … and it's not pub :-)

chrysn (approved):
LGTM!

kaspar030 (approved):
LGTM

Issue #298, opened 2024-05-21
tracking: use of unstable features


kaspar030:
We're aiming to build with stable Rust.

Currently, we're using these unstable features:

1. can't do without:
  - [ ] [`used_with_arg`](https://github.com/rust-lang/rust/issues/93798): this is used by linkme for its distributed slices

2. could do without:
- [ ] [`type_alias_impl_trait`](https://github.com/rust-lang/rust/issues/63063) => embassy uses this or pre-allocate futures memory
- [ ] `const_mut_refs`: used by ringbuffer
- [ ] `custom_test_frameworks`: will go when we have migrated to `defmt-test`
- [ ] `doc_auto_cfg`
- [ ] `error_in_core`
- [ ] `fn_traits`
- [ ] `inline_const`
- [ ] [naked_functions](https://github.com/rust-lang/rust/issues/90957): used by `riot-rs-threads`

josephcsible:
>     * [ ]  [`negative_impls`](https://github.com/rust-lang/rust/issues/68318); used for `!Send` on `MutexGuard`

Could you get rid of that by adding something like `_not_send: PhantomData<*const ()>` to `MutexGuard` instead?

kaspar030:
Where are `fn_traits` used?

```shell
❯ git grep fn_traits
ariel-os on  main took 24ms 
❯ 
```

ROMemories:
> Where are `fn_traits` used?

I'm not sure, I searched for it in my fork as well (in case that was for the sensor abstraction), but couldn't find anything. I removed it from the list.

kaspar030:
> This leaves only `used_with_arg`, and that we might manage by virtue of having good control of linker scripts ([#639](https://github.com/ariel-os/ariel-os/issues/639)?)

I'll give solving `used_with_arg` another shot.

kaspar030:
`ariel-os-debug-log` is using `macro_metavar_expr`

ROMemories:
Closing this, we've been able to build on stable since https://github.com/ariel-os/ariel-os/pull/951, and even with stable Cargo since #1811. The only remaining unstable features in the OS are `doc_cfg` (which is used only for documentation) and `asm_experimental_arch` on Xtensa (but the whole Xtensa toolchain is unstable and nightly). Such features can trivially be grepped for using `feature\(`.

The only other occurrence is in the `http-server` *application*, because `picoserve` needs `impl_trait_in_assoc_type`, but it's not part of the OS itself.

PR #950, opened 2025-03-29
feat(rt): drop use of used(linker)

kaspar030:
# Description

It seems adding `KEEP` to the listed sections in `linkme.x` is sufficient to keep the linker from garbage collecting our distributed slices, which would make the `used_with_arg` unstable feature unnecessary.

This one should be carefully tested, linkme errors produces non-working binaries...

## Issues/PRs references

#298 

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,7 +98,7 @@ esp-wifi-sys = { version = "0.7.1", default-features = false }
 esp-storage = { version = "0.4.0" }
 xtensa-lx-rt = { version = "0.18.0", default-features = false }
 
-linkme = { version = "0.3.31", features = ["used_linker"] }
+linkme = { version = "0.3.32" }
 
 ariel-os = { path = "src/ariel-os", default-features = false }
 ariel-os-alloc = { path = "src/ariel-os-alloc", default-features = false }
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -616,6 +616,8 @@ modules:
         RUSTFLAGS:
           - --cfg context=\"riscv\"
           - -Clink-arg=-Tisr_stack.x
+          - -Clink-arg=-Tlinkme-region-alias.x
+          - -Clink-arg=-Tlinkme.x
 
   - name: rp-link-arg
     help: helper module that ensures link-rp.x is added behind cortex-m ld scripts
--- a/src/ariel-os-rt/linkme.x
+++ b/src/ariel-os-rt/linkme.x
@@ -1,12 +1,12 @@
 SECTIONS {
-  linkme_INIT_FUNCS : { *(linkme_INIT_FUNCS) } > FLASH
-  linkm2_INIT_FUNCS : { *(linkm2_INIT_FUNCS) } > FLASH
-  linkme_EMBASSY_TASKS : { *(linkme_EMBASSY_TASKS) } > FLASH
-  linkm2_EMBASSY_TASKS : { *(linkm2_EMBASSY_TASKS) } > FLASH
-  linkme_USB_BUILDER_HOOKS : { *(linkme_USB_BUILDER_HOOKS) } > FLASH
-  linkm2_USB_BUILDER_HOOKS : { *(linkm2_USB_BUILDER_HOOKS) } > FLASH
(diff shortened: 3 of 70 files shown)


Cargo.toml
```diff
@@ -98,7 +98,7 @@ esp-wifi-sys = { version = "0.7.1", default-features = false }
 esp-storage = { version = "0.4.0" }
 xtensa-lx-rt = { version = "0.18.0", default-features = false }
 
-linkme = { version = "0.3.31", features = ["used_linker"] }
+linkme = { version = "0.3.32" }
```
ROMemories, line 101: I think the version bump is unrelated? [The changelog](https://github.com/dtolnay/linkme/releases/tag/0.3.32) only says "Documentation improvements". Anyway I'm fine with keeping it.

src/ariel-os-rt/linkme.x
```diff
@@ -1,12 +1,12 @@
...
-  linkme_THREAD_FNS : { *(linkme_THREAD_FNS) } > FLASH
-  linkm2_THREAD_FNS : { *(linkm2_THREAD_FNS) } > FLASH
+  linkme_INIT_FUNCS : { KEEP(*(linkme_INIT_FUNCS)) } > FLASH
+  linkm2_INIT_FUNCS : { KEEP(*(linkm2_INIT_FUNCS)) } > FLASH
+  linkme_EMBASSY_TASKS : { KEEP(*(linkme_EMBASSY_TASKS)) } > FLASH
+  linkm2_EMBASSY_TASKS : { KEEP(*(linkm2_EMBASSY_TASKS)) } > FLASH
+  linkme_USB_BUILDER_HOOKS : { KEEP(*(linkme_USB_BUILDER_HOOKS)) } > FLASH
+  linkm2_USB_BUILDER_HOOKS : { KEEP(*(linkm2_USB_BUILDER_HOOKS)) } > FLASH
+  linkme_THREAD_FNS : { KEEP(*(linkme_THREAD_FNS)) } > FLASH
+  linkm2_THREAD_FNS : { KEEP(*(linkm2_THREAD_FNS)) } > FLASH
```
ROMemories, line 9: [Documentation of `KEEP()`](https://www.sourceware.org/binutils/docs/ld/Input-Section-Keep.html) :book:

laze-project.yml
```diff
@@ -616,6 +616,8 @@ modules:
         RUSTFLAGS:
           - --cfg context=\"riscv\"
           - -Clink-arg=-Tisr_stack.x
+          - -Clink-arg=-Tlinkme-region-alias.x
+          - -Clink-arg=-Tlinkme.x
```
ROMemories, line 620: I understand that the `linkme-region-alias.x` link script is generated by the build script of `ariel-os-rt`, so that the `DROM`/`ROM` memory regions are aliased to `FLASH`, which is the name expected by the `linkme.x` linker script. Why is this now necessary on RISC-V? Was that `linkme.x` linker script not picked up before?

ROMemories (approved):
Successfully tested:

- the `threading` (`linkme_THREAD_FNS`) and `usb-keyboard` (`linkme_EMBASSY_TASKS` and `linkme_USB_BUILDER_HOOKS`) examples on nrf52840dk
- the `threading` (`linkme_THREAD_FNS`) and `log` (`linkme_EMBASSY_TASKS`) examples on espressif-esp32-c6-devkitc-1 (cannot test the USB books as this only works on the S3 I think)

The `linkme_INIT_FUNCS` section is always used at startup.

laze-project.yml
kaspar030: Actually I before, the `linkme.x` linker script was not needed at all, `used(linker)` ensured that the sections were not dropped.

kaspar030: ... now it is needed, but the region is called differently on riscvs

kaspar030:
> Successfully tested:
> 
>     * the `threading` (`linkme_THREAD_FNS`) and `usb-keyboard` (`linkme_EMBASSY_TASKS` and `linkme_USB_BUILDER_HOOKS`) examples on nrf52840dk

I tested `threading` and `usb-serial` on esp32s3.

PR #951, opened 2025-03-29
feat: enable building on stable Rust

kaspar030:
# Description

This PR adds stable Rust support. It is ~~WIP~~ a first iteration.

- nightly is still default
- CI only builds nightly
- `examples/http-server` is using `type_alias_impl_trait` and thus conflicts `stable`.

## Issues/PRs references

unstable tracking list in #298.

Depends on #950.

## Open Questions

- [x] Do we wait for the embassy-executor(-macros) release? => no, that is already merged in #982 
- [x] how to choose toolchain with espup? currently does `+esp` => xtensa always nightly

## Change checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -61,7 +61,7 @@ jobs:
       # faster generation.
       - name: Build rustdoc docs
         run: |
-          cargo doc \
+          RUSTDOCFLAGS='--cfg nightly' cargo doc \
               --no-deps \
               -p ariel-os \
               -p coapcore \
@@ -90,7 +90,7 @@ jobs:
                   usb-hid,
                   coapcore/_nightly_docs
                   "
-          RUSTDOCFLAGS='-D warnings --cfg context="esp32c6"' cargo doc \
+          RUSTDOCFLAGS='-D warnings --cfg context="esp32c6" --cfg nightly' cargo doc \
               --target=riscv32imac-unknown-none-elf \
               --no-deps \
               --features "
@@ -101,7 +101,7 @@ jobs:
                   spi,
                   " \
               -p ariel-os-esp
-          RUSTDOCFLAGS='-D warnings --cfg context="rp" --cfg context="rp2040"' cargo doc \
+          RUSTDOCFLAGS='-D warnings --cfg context="rp" --cfg context="rp2040" --cfg nightly' cargo doc \
               --no-deps \
               --features "
                   embassy-rp/rp2040,
@@ -110,7 +110,7 @@ jobs:
                   spi,
                   " \
               -p ariel-os-rp
-          RUSTDOCFLAGS='-D warnings --cfg context="nrf52840"' cargo doc \
+          RUSTDOCFLAGS='-D warnings --cfg context="nrf52840" --cfg nightly' cargo doc \
               --no-deps \
               --features "
                   embassy-nrf/nrf52840,
@@ -119,7 +119,7 @@ jobs:
                   spi,
                   " \
               -p ariel-os-nrf
-          RUSTDOCFLAGS='-D warnings --cfg context="stm32wb55rg"' cargo doc \
+          RUSTDOCFLAGS='-D warnings --cfg context="stm32wb55rg" --cfg nightly' cargo doc \
               --no-deps \
               --features "
                   embassy-stm32/stm32wb55rg,
@@ -140,8 +140,8 @@ jobs:
 
       - name: Check for unexpected cmdrun uses
         run: |
-            cd book
-            ./cmdrun-check.sh
+          cd book
+          ./cmdrun-check.sh
 
       - name: Build the book
         run: |
@@ -163,8 +163,8 @@ jobs:
         continue-on-error: true
         id: check
         run: |
-            hash="$(curl https://ariel-os.github.io/ariel-os/_hash | head -n 1)"
-            test "$hash" != "$(cat ./_site/_hash)"
+          hash="$(curl https://ariel-os.github.io/ariel-os/_hash | head -n 1)"
+          test "$hash" != "$(cat ./_site/_hash)"
 
       - name: Upload artifact (PR)
         if: github.event_name == 'pull_request' && steps.check.conclusion == 'success'
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -25,6 +25,8 @@ contexts:
       - ?hw/device-identity
       - ?lto
       - ?semihosting
+      - ?nightly
+      - ?stable
     env:
       RUSTFLAGS:
         - ${RUSTFLAGS_CONTEXTS}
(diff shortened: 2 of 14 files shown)


kaspar030:
Current roadblock: `macro_metavar_expr` in `ariel-os-debug-log`.

src/ariel-os-embassy/src/delegate.rs
```diff
@@ -41,10 +43,9 @@ pub struct Delegate<T> {
     send: Signal<CriticalSectionRawMutex, SameExecutorCell<*mut T>>,
     reply: Signal<CriticalSectionRawMutex, ()>,
     was_exercised: AtomicBool,
+    _not_send: PhantomData<*const ()>,
```
kaspar030, line 46: This actually breaks the usb_builder_hook use case. why did we add the `!Send` bound, can we just drop it instead?

src/ariel-os-threads/src/sync/mutex.rs
```diff
@@ -15,6 +16,7 @@ use crate::{SCHEDULER, thread::ThreadState, threadlist::ThreadList};
 pub struct Mutex<T> {
     state: UnsafeCell<LockState>,
     inner: UnsafeCell<T>,
+    _not_send: PhantomData<*const ()>,
```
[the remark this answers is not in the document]
josephcsible: On its own it would, but we already do `unsafe impl<T: Sync> Sync for MutexGuard<'_, T> {}` which overrides this and makes it `Sync` anyway.

src/ariel-os-embassy/src/delegate.rs
kaspar030: ah yes, thanks! ... I totally misread the compilation error. I've added the explicit `Sync` impl.

kaspar030:
> IIRC, embassy's story for non-TAIT/ATIT require a fixed size to be given for structs.

My understanding is that since https://github.com/embassy-rs/embassy/pull/4020 this is not needed anymore.

laze-project.yml
```diff
@@ -635,6 +636,27 @@ modules:
...
+          - --cfg stable
+
+  - name: nightly
+    help: build with nightly Rust
+    conflicts:
+      - stable
+    env:
+      global:
+        RUSTFLAGS:
+          - --cfg nightly
```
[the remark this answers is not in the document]
kaspar030: I guess so, but for the Ariel use case, global `cfg(stable)` and `cfg(nightly)` is pretty convenient.

src/ariel-os-embassy/src/delegate.rs
```diff
@@ -41,9 +43,10 @@ pub struct Delegate<T> {
     send: Signal<CriticalSectionRawMutex, SameExecutorCell<*mut T>>,
     reply: Signal<CriticalSectionRawMutex, ()>,
     was_exercised: AtomicBool,
+    _not_send: PhantomData<T>,
```
josephcsible, line 46: Is it desirable that this type will now be `Send` if `T` is `Send`? It wasn't before.

ROMemories:
> Current roadblock: `macro_metavar_expr` in `ariel-os-debug-log`.

This one was addressed in #965.

ROMemories:
@kaspar030 The `cargo-test` CI check is failing because of an unstable feature.

src/ariel-os-embassy/src/delegate.rs
kaspar030: no, my bad. thanks for noticing!

src/ariel-os-threads/src/sync/mutex.rs
```diff
@@ -58,6 +60,7 @@ impl<T> Mutex<T> {
         Self {
             state: UnsafeCell::new(LockState::Unlocked),
             inner: UnsafeCell::new(value),
+            _not_send: PhantomData,
```
ROMemories, line 63: The `Mutex` is `Send`, what must not is the `MutexGuard`.

ROMemories (changes requested):
I don't have a strong opinion regarding whether to use Cargo features for `nightly`, I'm fine with a raw `cfg` flag. I'd like to check the doc rendering in CI, but <surge.sh> is currently down.

ROMemories:
From my local testing, I think this additionally requires `RUSTDOCFLAGS='--cfg nightly'` in the CI doc generation.

src/ariel-os-threads/src/sync/mutex.rs
kaspar030: Oh yes. Well spotted!

kaspar030: fixed.

.github/workflows/build-deploy-docs.yml
```diff
@@ -90,7 +90,7 @@ jobs:
                   usb-hid,
                   coapcore/_nightly_docs
                   "
-          RUSTDOCFLAGS='-D warnings --cfg context="esp32c6"' cargo doc \
+          RUSTDOCFLAGS='-D warnings --cfg context="esp32c6" --cfg nightly' cargo doc \
```
ROMemories, line 93: The main cargo doc run above needs the same, the "feature flairs" are not shown in the docs preview.

laze-project.yml
```diff
@@ -644,6 +645,36 @@ modules:
...
+        RUSTFLAGS:
+          - -Clink-arg=-Tkeep-stack-sizes.x
+          - -Z emit-stack-sizes
+
+  - name: stable
+    help: build with stable Rust
+    conflicts:
+      - nightly
+      # currently, esp needs to use `+esp` as toolchain
+      - xtensa
```
ROMemories, line 662: optional: the `stable` module needs not be aware of the `xtensa` module, this could be moved to the `xtensa` module. This would also result in better error messages:

```
    3: "context::esp32s3" cannot resolve "xtensa"
    4: "xtensa" conflicts "stable"
```

instead of

```
    3: "context::esp32s3" cannot resolve "xtensa"
    4: "xtensa" is disabled/conflicted
```

ROMemories (approved):

```
laze -C examples/hello-world build -s stable -b nrf52840dk run
```

```
laze -C src/ariel-os-identity build -s stable -b nrf52840dk test
```

laze-project.yml
kaspar030: have you updated `laze` lately? The error message nowadays is like this:

```
laze: not building binary "hello-world" for builder "espressif-esp32-s3-devkitc-1"

Caused by:
    0: "hello-world" cannot resolve "context::espressif-esp32-s3-devkitc-1"
    1: "context::espressif-esp32-s3-devkitc-1" cannot resolve "context::esp32-s3-wroom-1"
    2: "context::esp32-s3-wroom-1" cannot resolve "context::esp32s3"
    3: "context::esp32s3" cannot resolve "xtensa"
    4: "xtensa" is disabled/conflicted by stable
```

... I'll move it anyways.

kaspar030: done

ROMemories (approved):

PR #955, opened 2025-03-31
feat(boards) added support for the STM32F401CE and F401CD

zekiz774:
# Description

Provides initial support for the Blackpill boards STM32F401CD and STM32F401CD. They're all pretty similar so adding support for other STM32F4xx should be pretty trivial

## TODO
- [x] GPIO support
- [ ] SPI support
- [ ] I2C support
- [x] USB support
- [ ] Edit example code
- [ ] Test integration

## Issues/PRs references

Depends on #961.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -443,6 +443,7 @@ contexts:
       - cortex-m4f
     provides:
       - has_swi
+      - has_usb_device_port
     env:
       PROBE_RS_CHIP: STM32F401CE
       RUSTFLAGS:
--- a/src/ariel-os-stm32/src/lib.rs
+++ b/src/ariel-os-stm32/src/lib.rs
@@ -89,9 +89,9 @@ fn board_config(config: &mut Config) {
         config.rcc.pll_src = PllSource::HSE;
         config.rcc.pll = Some(Pll {
             prediv: PllPreDiv::DIV25,  // 1MHz
-            mul: PllMul::MUL168,       // 168MHz
-            divp: Some(PllPDiv::DIV2), // sysclk 84MHz
-            divq: Some(PllQDiv::DIV3), // USB 48MHz
+            mul: PllMul::MUL336,       // 336MHz
+            divp: Some(PllPDiv::DIV4), // sysclk 84MHz
+            divq: Some(PllQDiv::DIV7), // USB 48MHz
             divr: None,
         });
         config.rcc.ahb_pre = AHBPrescaler::DIV1;
--- a/examples/usb-keyboard/laze.yml
+++ b/examples/usb-keyboard/laze.yml
@@ -3,5 +3,7 @@ apps:
     context:
       - nrf52840dk
       - nrf5340dk
+      - stm32f401cdux
+      - stm32f401ceux
     selects:
       - usb
--- a/examples/usb-keyboard/src/main.rs
+++ b/examples/usb-keyboard/src/main.rs
@@ -106,18 +106,13 @@ mod buttons {
 
     const PULL: Pull = Pull::Up;
 
-    pub const KEY_COUNT: usize = 4;
+    pub const KEY_COUNT: usize = 1;
 
     pub struct Buttons([Input; KEY_COUNT]);
 
     impl Buttons {
         pub fn new(button_peripherals: crate::pins::Buttons) -> Self {
-            Self([
-                Input::new(button_peripherals.btn1, PULL),
-                Input::new(button_peripherals.btn2, PULL),
-                Input::new(button_peripherals.btn3, PULL),
-                Input::new(button_peripherals.btn4, PULL),
-            ])
+            Self([Input::new(button_peripherals.btn1, PULL)])
         }
 
         pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut Input> {
--- a/examples/usb-keyboard/src/pins.rs
+++ b/examples/usb-keyboard/src/pins.rs
@@ -15,3 +15,6 @@ ariel_os::hal::define_peripherals!(Buttons {
     btn3: P0_08,
     btn4: P0_09,
 });
+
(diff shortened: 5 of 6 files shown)


.gitignore
ROMemories, line 1: Please drop changes to this file, this is out of scope of this PR; we might consider gitignoring editor-specific files/directories separately.

examples/blinky/laze.yml
```diff
@@ -14,3 +14,5 @@ apps:
       - st-nucleo-h755zi-q
       - st-nucleo-wb55
       - st-nucleo-wba55
+      - stm32f401ce
+      - stm32f401cd
```
ROMemories, line 18: For STM32, only boards with an LED should be listed, I think you're adding support for Blackpill boards, so that's what should be there instead (ideally sorted by alphabetical order).

examples/blinky/src/pins.rs
```diff
@@ -33,6 +33,9 @@ ariel_os::hal::define_peripherals!(LedPeripherals { led: PA5 });
 #[cfg(context = "st-nucleo-f401re")]
 ariel_os::hal::define_peripherals!(LedPeripherals { led: PA5 });
 
+#[cfg(any(context = "stm32f401ceux", context = "stm32f401cdux"))]
```
ROMemories, line 36: So these should be updated with board laze contexts.

laze-project.yml
```diff
@@ -425,6 +425,14 @@ contexts:
...
         - CONFIG_SWI=USART2
 
+  - name: stm32f401cx
+    parent: stm32
+    selects:
+      - cortex-m4f
+    env:
+      PROBE_RS_CHIP: STM32F401CD
+      PROBE_RS_PROTOCOL: swd
+
```
ROMemories, line 435: Factoring out `stm32f401cd` and `stm32f401ce` this way should indeed work, we don't do that now and this should eventually be codegenerated, so we'd prefer a flatter hierarchy without this intermediate context here.

laze-project.yml
```diff
@@ -425,6 +425,14 @@ contexts:
         # This ISR is unused on a naked board. Configured here for testing.
         - CONFIG_SWI=USART2
 
+  - name: stm32f401cx
+    parent: stm32
+    selects:
+      - cortex-m4f
+    env:
+      PROBE_RS_CHIP: STM32F401CD
+      PROBE_RS_PROTOCOL: swd
```
ROMemories, line 434: Following #957, specifying `PROBE_RS_PROTOCOL` is not needed anymore.

laze-project.yml
```diff
@@ -1413,6 +1421,18 @@ builders:
   - name: st-nucleo-f401re
     parent: stm32f401retx
 
+  - name: stm32f401cd
```
ROMemories, line 1424:
```suggestion
  - name: stm32f401cdux
```

These laze context names should match the values passed as `--cfg` just below; we'll later (soon?) rework this following #958 to shorten these, but for now we do need that redundancy.

laze-project.yml
```diff
@@ -1413,6 +1421,18 @@ builders:
...
+    parent: stm32f401cx
+    env:
+      RUSTFLAGS:
+        - --cfg context=\"stm32f401cdux\"
+
+  - name: stm32f401ce
```
ROMemories, line 1430:
```suggestion
  - name: stm32f401ceux
```

src/ariel-os-stm32/src/lib.rs
```diff
@@ -78,6 +78,29 @@ pub fn init() -> OptionalPeripherals {
...
+        config.rcc.hse = Some(Hse {
+            freq: embassy_stm32::time::Hertz(25_000_000),
+            mode: HseMode::Oscillator,
+        });
+        config.rcc.pll_src = PllSource::HSE;
+        config.rcc.pll = Some(Pll {
+            prediv: PllPreDiv::DIV25,  // 1Hz
+            mul: PllMul::MUL336,       // 336Hz
+            divp: Some(PllPDiv::DIV4), // sysclk 84MHz
+            divq: Some(PllQDiv::DIV6), // USB 48MHz
```
ROMemories, line 94: It might also be possible to reduce the pre-divider value, which is usually better as this reduces jitter and helps the PLL lock faster (we also do have plans for generating these clock configurations automatically later):
```suggestion
            prediv: PllPreDiv::DIV25,  // 1MHz
            mul: PllMul::MUL168,       // 168MHz
            divp: Some(PllPDiv::DIV2), // sysclk 84MHz
            divq: Some(PllQDiv::DIV3), // USB 48MHz
```
(There may be better clock configurations, not claiming this is optimal either.)

laze-project.yml
ROMemories, line 1: Laze contexts for the *boards* you want to support also need to be added in the list of `builders` in this file.

ROMemories:
Thanks for the PR! Left some inline comments for this initial support: on top of the laze contexts you've added, we also need to add laze contexts for each *board* you want to add support for.

laze-project.yml
```diff
@@ -1413,6 +1421,18 @@ builders:
   - name: st-nucleo-f401re
     parent: stm32f401retx
 
+  - name: stm32f401cd
+    parent: stm32f401cx
+    env:
+      RUSTFLAGS:
+        - --cfg context=\"stm32f401cdux\"
```
ROMemories, line 1428: These are actually unneeded since #855, #959 is updating the porting guide to reflect that:
```suggestion
```

laze-project.yml
```diff
@@ -1413,6 +1421,18 @@ builders:
...
+
+  - name: stm32f401ce
+    parent: stm32f401cx
+    env:
+      RUSTFLAGS:
+        - --cfg context=\"stm32f401ceux\"
```
ROMemories, line 1434:
```suggestion
```

PR #960, opened 2025-04-02
refactor(stm32): move the SWI selection to the boards, from the MCUs

ROMemories:
# Description

This moves the definitions of `CONFIG_SWI` to the board laze contexts where they belong, as already documented in the book.

## Issues/PRs references

Depends on #959.

## Change checklist

### DIFF
--- a/book/src/adding-board-support.md
+++ b/book/src/adding-board-support.md
@@ -8,6 +8,10 @@ Feel free to report anything that is unclear or missing!
 > This guide requires working on your own copy of Ariel OS.
 > You may want to fork the repository to easily upstream your changes later.
 
+> Unless documented in the User Guide, please expect the module and context names that are defined in the `laze-project.yml` file to change.
+> We're still figuring out a proper naming scheme.
+> You've been warned.
+
 ## Adding Support for a Board
 
 The more similar a board is to one that is already supported, the easier.
@@ -36,6 +40,11 @@ builders:
   # ...
   - name: st-nucleo-f401re
     parent: stm32f401re
+    provides:
+      - has_swi
+    env:
+      CARGO_ENV:
+        - CONFIG_SWI=USART2
 ```
 
 ## Adding Support for an MCU from a Supported MCU family
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -396,8 +396,6 @@ contexts:
     parent: stm32
     selects:
       - cortex-m0-plus
-    provides:
-      - has_swi
     disables:
       # TODO: embedded-test currently hard-codes 16k for its thread
       - embedded-test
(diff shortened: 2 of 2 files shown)


book/src/adding-board-support.md
```diff
@@ -8,6 +8,8 @@ Feel free to report anything that is unclear or missing!
 > This guide requires working on your own copy of Ariel OS.
 > You may want to fork the repository to easily upstream your changes later.
 
+> Identifiers used in `laze-project.yml` are considered internal unless stated otherwise.
```
ROMemories, line 11: Don't want to commit to `has_*` laze module names just yet, before we bikeshed these when officially adding support for out-of-tree boards.

kaspar030: Can you drop that sentence from this PR? I honestly can't say what it would mean. And it is unrelated to "refactor(stm32): move the SWI selection to the boards, from the MCUs" in anycase.

ROMemories: We can definitely rephrase this, but it is related to this PR because `has_swi` appears in the documentation now, and I just wanted to make it clear that this doesn't imply us committing to that module name.

kaspar030: IMO if we make committing opt-out, we'll forget opting out and be braking commitments every other day. I'd rather make commitments opt-in.

This phrase gives the impression that everything else is *not* considered internal, which implicitly commits to *everything else*, and b/c it is part of the book, across the book. I'm not ready to commit to that, so be not having this phrase here, we're actually committing to *much* less.

book/src/adding-board-support.md
```diff
@@ -8,6 +8,8 @@ Feel free to report anything that is unclear or missing!
 > This guide requires working on your own copy of Ariel OS.
 > You may want to fork the repository to easily upstream your changes later.
 
+> Even though laze modules and contexts defined in the `laze-project.yml` file might be usable in your application, we strongly recommend not to use them yet unless they are otherwise documented in the User Guide, as we might rename or drop any of them at any time.
```
kaspar030, line 11:
```suggestion
> Unless documented in this User Guide, please expect the module and context names that are defined in this file to change. We're still figuring out a proper naming scheme. You've been warned.
```

bergzand:
```suggestion
> Even though laze modules and contexts defined in the `laze-project.yml` file might be usable in your application, we strongly recommend not to use them yet unless they are otherwise documented in the User Guide, as the names have not been stabilized yet and are subject to change.
```

book/src/adding-board-support.md
bergzand: I think it is good for transparency to users to document these things. I would rather phrase it as not yet stabilized, but that's nitpicking here.

kaspar030 (approved):

book/src/adding-board-support.md
ROMemories: Went with @kaspar030 suggestion, just changed "this User Guide" to "the User Guide" because *this* is not the User Guide, it's the Developer Guide, and additionally clarified what "this file" is.

PR #961, opened 2025-04-02
 refactor(stm32)!: remove unneeded info from laze context names 

ROMemories:
# Description

Following #958, this removes the package type and temperature range from the STM32 laze contexts, as they are not needed anymore (pin count is encoded separately from the package type). This reduces the number of required laze contexts to 1400, down from 2207.

There is no remaining occurrences of `stm32c031c6tx`, `stm32f401retx`, `stm32h755zitx`, `stm32wb55rgvx`, or `stm32wba55cgux`.

BREAKING CHANGE: this is a breaking change as we document the existence of these MCU contexts

## Testing

Tested the `hello-world` example successfully on st-nucleo-c031c6, st-nucleo-h755zi-q, st-nucleo-wb55.
The following builds successfully:
```sh
laze build -g -b st-nucleo-c031c6 -b st-nucleo-f401re -b st-nucleo-h755zi-q -b st-nucleo-wb55 -b st-nucleo-wba55
```

## Change checklist


src/ariel-os-stm32-mapping/stm32_mapping.csv
ROMemories, line 1: This file now only has 1400 lines, instead of 2207.

kaspar030 (approved):
lgtm

PR #964, opened 2025-04-03
feat(rt): use `global_asm!()` to define isr stack

kaspar030:
# Description

This uses `global_asm!()` to define the isr stack, which fixes the `.isr_stack` ELF flags. That in turn makes `size` count the ISR stack correctly as `bss`.

## Change checklist


src/ariel-os-rt/src/lib.rs
```diff
@@ -63,10 +63,13 @@ mod isr_stack {
...
 
-    // SAFETY: a linker script inserts this section in RAM.
-    #[unsafe(link_section = ".isr_stack")]
-    #[used(linker)]
-    static ISR_STACK: [u8; ISR_STACKSIZE] = [0u8; ISR_STACKSIZE];
+    core::arch::global_asm!(
+    "
+        .section .isr_stack, \"wa\"
+        .skip {size}
+    ",
```
ROMemories, line 70: optional: indentation could be improved, and using raw strings would avoid the need for escaping those quotation marks.

ROMemories, line 70: [The documentation of `.skip`](https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_7.html#SEC125) does mention that it initializes the skipped bytes to zero by default. :+1:

ROMemories (approved):
Successfully tested in hardware on st-nucleo-c031c6.
Running `size` on the `blinky` example on that board now prints:
```
  17548      32    6568   24148    5e54 build/bin/st-nucleo-c031c6/blinky/blinky.elf
```

instead of:
```
  21644      32    2472   24148    5e54 build/bin/st-nucleo-c031c6/blinky/blinky.elf
```
A difference of 4096 bytes for the `.text` and `.bss` sections, which is indeed the size of the `.isr_stack` section as printing by `size -A`:
```
.isr_stack                 4096   536870912
```

I suppose it would to useful to document that the size of that section now needs to be taken into account on top of that of the `.bss` section to estimate the "base amount" of RAM needed for an application and maybe provide a laze task to compute that.

src/ariel-os-rt/src/lib.rs
kaspar030: +1 for the raw strings. What's the desirec indentation?

kaspar030: ... which we don't do, the section is marked as `NO_LOAD` and the cortex-m init code only zeroes bss.

kaspar030: I'll fix indent in a folo, so I can rebase #950 now.

src/ariel-os-rt/src/lib.rs
```diff
@@ -63,10 +63,13 @@ mod isr_stack {
...
 
-    // SAFETY: a linker script inserts this section in RAM.
-    #[unsafe(link_section = ".isr_stack")]
-    #[used(linker)]
-    static ISR_STACK: [u8; ISR_STACKSIZE] = [0u8; ISR_STACKSIZE];
+    core::arch::global_asm!(
+    r#"
+        .section .isr_stack, "wa"
+        .skip {size}
+    "#,
```
ROMemories, line 70: Desired indentation:

```suggestion
        r#"
        .section .isr_stack, "wa"
        .skip {size}
        "#,
```

kaspar030: ah ok, will do

kaspar030: done

ROMemories (approved):

PR #967, opened 2025-04-03
fix(esp): increase esp32c3 stacksize default

kaspar030:
# Description

All but the simplest examples need this.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -342,6 +342,8 @@ contexts:
       PROBE_RS_CHIP: esp32c3
       PROBE_RS_PROTOCOL: jtag
       CARGO_TARGET_PREFIX: CARGO_TARGET_RISCV32IMC_UNKNOWN_NONE_ELF
+      # c3 needs a bit more isr stack.
+      isr_stacksize_required_default: "4096"
 
   - name: esp-c3-01m
     parent: esp32c3


laze-project.yml
```diff
@@ -342,6 +342,8 @@ contexts:
       PROBE_RS_CHIP: esp32c3
       PROBE_RS_PROTOCOL: jtag
       CARGO_TARGET_PREFIX: CARGO_TARGET_RISCV32IMC_UNKNOWN_NONE_ELF
+      # c3 needs a bit more isr stack.
+      isr_stacksize_required_default: "4096"
```
ROMemories, line 346: 4096 is still not enough for the `http-client` example, 32768 isn't either, but 65536 is. It does work if we set this here, but maybe we should set that for the biggest examples only (which include `http-server` and maybe others)?

```suggestion
      isr_stacksize_required_default: "65536"
```

ROMemories (changes requested):
The need for this PR follows from #786.

The ESP32C6 also requires the same, but this can be addressed in a follow-up.

laze-project.yml
kaspar030: huh, I thought I checked this, was wondering why only c3 needs more. I'll do some more testing, maybe it is the wifi ISRs or stack that needs more memory, we can selectively increase that also.

kaspar030: done

ROMemories (approved):

- `examples/http-client` and `examples/gpio` on `espressif-esp32-c3-lcdkit` (as a fake `ai-c3`)
- `examples/http-client` on `espressif-esp32-c6-devkitc-1`

PR #969, opened 2025-04-03
chore: use defmt 1.0

chrysn:
# Description

This enables using defmt CBOR output (but that's a separate change)

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #971, opened 2025-04-03
feat(coap): use CBOR defmt debugging

chrysn:
# Description

With defmt updated, a lot of places where CBOR debugging was already planned are now unblocked.

<del>Draft: I'll look around whether there are more places where using this makes sense.</del>

Note that this is only usable for people who <del>have a manually updated prob-rs (https://github.com/probe-rs/probe-rs/pull/3206) until they merge and</del> installed probe-rs from their master branch or their next release; for everyone else, the debug output changes subtly from the previously manually set 2-digit hex output to the fallback comma-separated decimal integers.

## Issues/PRs references

Builds on #969
Depends on #972.

## Open Questions

* [x] When do I stop?

## Change checklist

[edit: probe-rs merged my PR]

### DIFF
--- a/examples/coap-client/src/main.rs
+++ b/examples/coap-client/src/main.rs
@@ -21,7 +21,7 @@ async fn run_client_operations() {
         .with_request_payload_slice(b"This is Ariel OS")
         .processing_response_payload_through(|p| {
             info!(
-                "Uppercase response is {}",
+                "Uppercase response is {:?}",
                 core::str::from_utf8(p).map_err(|_| "not Unicode?")
             );
         });
--- a/examples/device-metadata/src/main.rs
+++ b/examples/device-metadata/src/main.rs
@@ -9,7 +9,7 @@ async fn main() {
     info!("Available information:");
     info!("Board type: {}", ariel_os::buildinfo::BOARD);
     if let Ok(id) = ariel_os::identity::device_id_bytes() {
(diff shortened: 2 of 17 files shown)


chrysn:
… and before we click "merge" here: All of this is unusable when `info!` is backed by `log`.

So, what do? Create a thin wrapper that prints things nicely on defmt and falls back to hex output on log? (I guess that's what makes most sense.)

chrysn:
For the purpose of this pull request, I'll just ignore builds with log: https://github.com/ariel-os/ariel-os/pull/972 shows that so much *is already* broken that unless there comes a conscious decision to fix that, going on here won't make it worse.

chrysn:
As there *was* feedback to go ahead with #972, this is now rebased onto that.

Note that there is now a mix of two implementations of the `Cbor()` newtype wrapper: ariel-os-coap uses Ariel's now-built-in wrapper, whereas coapcore has no dependnecies into Ariel and uses `defmt_or_log`, and thus ships its own wrapper that builds on that rather than the equivalent Ariel functionality. I *could* PR this into `defmt_or_log`, but that wouldn't reduce the duplication unless Ariel as a whole switched to `defmt_or_log`, so I suggest we leave it as is and consider PR'ing or switching to `defmt_or_log` at a later time.

chrysn:
This is on hold until #1025 is sorted out.

chrysn:
This was replaced by #1112 and #1202.

Issue #625, opened 2024-12-12
Log levels debug and trace not typechecked


chrysn:
### Summary

I just found long-broken code in a `debug!()` statement where the type requirements of Format were not fulfilled. The code has passed build in https://github.com/ariel-os/ariel-os/pull/588, but as those run with `LOG=info`, debug and trace statements do not get built.

Worse, we use defmt_or_log (explicitly in coapcore, and a manual version thereof in `ariel_os::debug::log`), which means type checking only happens for one of them instead of both (tracked for defmt_or_log in https://github.com/t-moe/defmt-or-log/issues/4), so we should

* run CI, at least the `cargo check` step, with `LOG=trace`
* run it twice, once with defmt logging, once with stdout

… or introduce a feature/flag (similar level as debug_assert) that runs both versions by the compiler but makes it dead code.

### Would you like to work on fixing this bug?

Yes

chrysn:
This is not *just* CI (as labelled): These checks would be very valuable also to local developers, who otherwise carry around broken code until they submit to CI.

chrysn:
With #972 merged, the most common type check error (defmt'isms in the default messages) is being checked against.

We should do more still, in particular check non-default log levels (no issues right now but not CI enforced) and check in the other direction (format_args'isms that don't work in defmt, which would come in on boards that default to log), but the latter really only works if we can pull in defmt as an unconditional dependency (which I'd appreciate).

PR #972, opened 2025-04-04
fix: ensure defmt log statements also work with log

chrysn:
# Description

This adds checks to defmt builds to avoid defmt-isms (such as formatting a range or slice with {:x}, or formatting results in {}) don't creep up unchecked, and adds the tools to smooth over inconsistencies between the languages.

## Testing

```
laze build -g -b st-nucleo-wb55 -D LOG=debug -s log -d defmt
```

previously failed. Once this PR is through, not only should that not fail any more, but any case that does break should just as well break the developer's build that doesn't select log over defmt.

## Issues/PRs references

Contributes to #625 (doesn't fix it completely, that'd require checking also the currently unused log levels, and doing the same exercise also in the other direction; both is harder in that it'll require extra checking to see whether that doesn't introduce needless defmt interning).

## Testing

A good test is `laze build -b st-nucleo-wb55 -C examples/device-metadata -s log run`, which should still show the device ID in a way similar to defmt (possibly prettier, but that's not implemented right now to keep things simple).

## Open Questions

* <del>What do we do with the harder parts? Right now, the fixes just throw away a bit of formatting details. We can start having our custom wrappers a la `info!("The data is {}", HexSlice(&my_slice))`, but that makes at least defmt a small bit less efficient than its own version `info!("The data is {:02x}", my_slice)` is, and at some point we should be sure that this is a trade-off (not just in runtime but also mental-load and build complexity) we really want to make for not going full defmt.</del> (resolved by doing this, like embassy does)

  (Not that going full-defmt doesn't come with its own problems that we haven't even explored -- like, doing the full formatting for a device that does want to output to serial even when it has all the defmt strings is not trivial.)

## Change checklist

### DIFF
--- a/src/ariel-os-debug-log/src/lib.rs
+++ b/src/ariel-os-debug-log/src/lib.rs
@@ -48,6 +48,10 @@ pub mod log {
 // NOTE: log macros are defined within private modules so that `doc_auto_cfg` does not produce
 // "feature flairs" on them.
 // The macros are still exported even though they are defined "within" private modules.
+//
+// The `if true` conditionals ensure that the arguments are also valid under format_args!; this is
+// requires because a laze switch can just make the back-end switch over. The actual and the unused
+// formatting need to be in different branches to avoid trouble due to arguments being moved.
 #[cfg(feature = "defmt")]
 mod log_macros {
     /// Logs a message at the trace level.
@@ -55,7 +59,11 @@ mod log_macros {
     macro_rules! trace {
         ($($arg:tt)*) => {{
             use $crate::defmt::hidden::defmt;
-            defmt::trace!($($arg)*);
+            if true {
+                defmt::trace!($($arg)*);
+            } else {
+                drop(format_args!($($arg)*));
+            }
         }};
     }
 
@@ -64,7 +72,11 @@ mod log_macros {
     macro_rules! debug {
         ($($arg:tt)*) => {{
             use $crate::defmt::hidden::defmt;
-            defmt::debug!($($arg)*);
+            if true {
+                defmt::debug!($($arg)*);
+            } else {
+                drop(format_args!($($arg)*));
+            }
         }};
     }
 
@@ -73,7 +85,11 @@ mod log_macros {
     macro_rules! info {
         ($($arg:tt)*) => {{
             use $crate::defmt::hidden::defmt;
-            defmt::info!($($arg)*);
+            if true {
+                defmt::info!($($arg)*);
+            } else {
+                drop(format_args!($($arg)*));
+            }
         }};
     }
 
@@ -82,7 +98,11 @@ mod log_macros {
     macro_rules! warn {
         ($($arg:tt)*) => {{
             use $crate::defmt::hidden::defmt;
-            defmt::warn!($($arg)*);
+            if true {
+                defmt::warn!($($arg)*);
+            } else {
+                drop(format_args!($($arg)*));
+            }
         }};
     }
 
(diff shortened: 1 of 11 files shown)


chrysn:
embassy has [Bytes](https://github.com/embassy-rs/embassy/blob/main/embassy-executor/src/fmt.rs#L245-L270) sprinkled around its formatting; we'll use the same approach with `Hex` and `Cbor`. (The latter is not actively used here, but in https://github.com/ariel-os/ariel-os/pull/971 that is to be rebased onto this).

src/ariel-os-debug-log/src/lib.rs
```diff
@@ -55,7 +55,11 @@ mod log_macros {
     macro_rules! trace {
         ($($arg:tt)*) => {{
             use $crate::defmt::hidden::defmt;
-            defmt::trace!($($arg)*);
+            if true {
+                defmt::trace!($($arg)*);
+            } else {
+                drop(format_args!($($arg)*));
+            }
```
ROMemories, line 62: If I'm not mistaken, I think you can achieve the same thing by simply adding an `if false`, which makes this easier to document:

```suggestion
            defmt::trace!($($arg)*);
            if false {
                drop(format_args!($($arg)*));
            }
```

ROMemories: Can you also add a `NOTE` comment next to the one on this module to document these weird `if`s? The only documentation currently is in the commit message, which makes this a bit difficult to discover.

src/ariel-os-debug-log/src/lib.rs
```diff
@@ -172,3 +192,52 @@ mod log_macros {
         ($($arg:tt)*) => {};
     }
 }
+
+/// A newtype around byte slices used for all of Ariel OS's logging facades that turns the bytes
+/// into some hex output.
+///
+/// Its preferred output `00 11 22 33`, but log facades may also produce something like `[00, 11,
+/// 22, 33]` (eg. while that is cheaper on defmt).
```
ROMemories, line 200: I think this sentence is missing a word.

src/ariel-os-debug-log/src/lib.rs
```diff
@@ -172,3 +192,52 @@ mod log_macros {
...
+impl<T: AsRef<[u8]>> defmt::Format for Hex<T> {
+    fn format(&self, f: defmt::Formatter) {
+        ::defmt::write!(f, "{=[u8]:02x}", self.0.as_ref())
+    }
+}
+
+/// A newtype around byte slices used for all of Ariel OS's logging facades that prefers
+/// interpreting the data as CBOR.
+///
+/// Its preferred outptu is CBOR Diagnostic Notation (EDN), but showing hex is also acceptable.
```
ROMemories, line 222:
```suggestion
/// Its preferred output is CBOR Diagnostic Notation (EDN), but showing hex is also acceptable.
```

ROMemories (commented):
I'm not yet sure of what I think of the `Hex` and `Cbor` wrappers, even though they clearly solve a real problem; I need to think about it for a bit.

ROMemories:
Can fix the extra word in the third commit message as well?

src/ariel-os-debug-log/src/lib.rs
chrysn: We can't have only an if-false branch because there are a few cases in which $arg is moving.

The comment is now in there in a fixup.

chrysn:
Thanks, all addressed. All force-pushed in a squashed way b/c only that way I could address the commit message error.

> I'm not yet sure of what I think of the Hex and Cbor wrappers, even though they clearly solve a real problem; I need to think about it for a bit.

Has anything come of that? I don't see any alternatives if we want to use formatting in a cross-formatter way.

chrysn:
> This does fail for me on the http-client and storage examples (possibly for a reason that's orthogonal to this PR).

There's more going wrong than just this one, we have defmt'isms in other examples too. This does not fix all defmt'isms, but it ensures that they are not in the log lines.

kaspar030 (approved):

PR #929, opened 2025-03-21
feat(boards): add support for the Nordic Thingy:91

ROMemories:
# Description

## TODO

- [ ] Specify the LED and button pins for examples
- [ ] Use the on-board I2C sensors for the I2C test
- [ ] Use the on-board SPI sensors for the SPI test

## Issues/PRs references

Depends on #926.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #974, opened 2025-04-07
feat(boards): add support for the Nordic Thingy:91 X

ROMemories:
# Description

This adds support for the nRF9151 MCU and the Nordic Thingy:91 X board.

This is not a duplicate of #929, which is for the older, non-X variant.

## Testing

- `examples/gpio`
- `examples/storage`
- `examples/threading`
- `tests/i2c-controller` (using an on-board sensor)
- `tests/spi-main` (using an on-board sensor)

There is no user USB support on the nRF9151, and we don't yet support other network links, so couldn't test networking.

Additionally, `laze build -g -b nordic-thingy-91-x-nrf9151` completes successfully.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -242,14 +242,23 @@ contexts:
 
   - name: nrf91
     parent: nrf
+    selects:
+      - cortex-m33f
 
   - name: nrf9160
     parent: nrf91
     provides:
       - has_storage_support
-    selects:
-      - cortex-m33f
     env:
+      PROBE_RS_CHIP: nRF9160_xxAA
+
+  - name: nrf9151
+    parent: nrf91
+    provides:
+      - has_storage_support
+    env:
+      # FIXME: probe-rs does not support the nRF9151_xxAA yet, because there is
+      # no CMSIS-Pack available
       PROBE_RS_CHIP: nRF9160_xxAA
 
   - name: rp
@@ -1394,6 +1403,10 @@ builders:
   # TODO: there also is a companion nrf52840 on this board
   - name: nrf9160dk
     parent: nrf9160
+
+  # TODO: there also is a companion nrf5340 on this board
+  - name: nordic-thingy-91-x-nrf9151
+    parent: nrf9151
 
   - name: st-nucleo-f401re
     parent: stm32f401re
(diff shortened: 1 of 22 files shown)


laze-project.yml
```diff
@@ -1395,6 +1404,10 @@ builders:
   - name: nrf9160dk
     parent: nrf9160
 
+  # TODO: there also is a companion nrf5340 on this board
+  - name: nordic-thingy-91-x-nrf9151
```
ROMemories, line 1408: The technical name is chosen based on the manufacturer URL (`thingy-91` instead of `thingy91`).
That board has two target MCU, so we agreed on that naming scheme.

tests/spi-main/src/main.rs
```diff
@@ -27,8 +27,16 @@ use ariel_os::{
...
+// WHO_AM_I register of the sensor
+#[cfg(not(context = "nordic-thingy-91-x-nrf9151"))]
 const WHO_AM_I_REG_ADDR: u8 = 0x0f;
+#[cfg(context = "nordic-thingy-91-x-nrf9151")]
+const WHO_AM_I_REG_ADDR: u8 = 0x00;
+
+#[cfg(not(context = "nordic-thingy-91-x-nrf9151"))]
+const DEVICE_ID: u8 = 0x33;
+#[cfg(context = "nordic-thingy-91-x-nrf9151")]
+const DEVICE_ID: u8 = 0x24;
```
ROMemories, line 39: This will need to be refactored so we can more cleanly add support for other boards in these tests, but I'd rather do this in a follow-up.

laze-project.yml
```diff
@@ -242,16 +242,25 @@ contexts:
 
   - name: nrf91
     parent: nrf
+    selects:
+      - cortex-m33f
```
ROMemories, line 246: Based on [Nordic's comparison table](https://www.nordicsemi.com/-/media/Publications/WQ-Product-guide/Nordic-Product-Comparison-table-WQ-Dec-2024.pdf), this seems to me to be a property of the whole nRF91 series.

kaspar030 (approved):

PR #977, opened 2025-04-08
feat(rt): add initial stack measuring & painting

kaspar030:
# Description

This adds a module that allows measuring stack usage. Currently, a stack can only report for "itself".

An additional commit enables use of the psplim register on `armv8m` (Cortex-M33).

## Open Questions

- [ ] handle multi core isr stack
- [ ] allow access from outside the active stack, e.g., for "remote" reporting of thread stacks
- [ ] risc-v
- [ ] esp
- [ ] add some way of reporting this for tests (e.g., at thread end?)

## Change checklist

### DIFF
--- a/src/ariel-os-rt/src/lib.rs
+++ b/src/ariel-os-rt/src/lib.rs
@@ -4,6 +4,8 @@
 #![allow(incomplete_features)]
 // - const_generics
 
+pub mod stack;
+
 #[cfg(feature = "threading")]
 mod threading;
 
@@ -28,9 +30,13 @@ cfg_if::cfg_if! {
         compile_error!("no runtime is defined for this MCU family");
     } else {
         // Provide a default implementation, for arch-independent tooling
+        #[cfg_attr(not(context = "ariel-os"), allow(dead_code))]
         mod arch {
-            #[cfg_attr(not(context = "ariel-os"), allow(dead_code))]
+            use crate::stack::Stack;
+
             pub fn init() {}
+            pub fn sp() -> usize { 0 }
+            pub fn stack() -> Stack { Stack::default() }
         }
     }
 }
@@ -66,6 +72,27 @@ mod isr_stack {
         "#,
         size = const ISR_STACKSIZE
     );
+
+    pub fn limits() -> (usize, usize) {
+        // ISR stack
+        // TODO: multicore!
+        unsafe extern "C" {
+            static _stack_bottom: u32;
+            static _stack_start: u32;
+        }
+
+        let bottom = &raw const _stack_bottom as usize;
+        let top = &raw const _stack_start as usize;
+        (bottom, top)
+    }
+
+    pub fn init() {
+        let stack = crate::stack::Stack::get();
+        crate::debug!("ariel-os-rt: ISR stacksize: {}", stack.size());
+
+        // initial stack paint
+        stack.repaint();
+    }
 }
 
 #[cfg(feature = "_panic-handler")]
@@ -96,7 +123,7 @@ fn startup() -> ! {
     debug!("ariel_os_rt::startup()");
 
     #[cfg(any(context = "cortex-m", context = "riscv"))]
-    debug!("ariel_os_rt: ISR_STACKSIZE={}", isr_stack::ISR_STACKSIZE);
+    crate::isr_stack::init();
 
     #[cfg(feature = "alloc")]
     // SAFETY: *this* is the only place alloc should be initialized.
--- a/src/ariel-os-rt/src/stack.rs
+++ b/src/ariel-os-rt/src/stack.rs
@@ -1,0 +1,105 @@
+//! Stack usage helpers.
+use core::{marker::PhantomData, ptr::write_volatile};
+
+use crate::arch::sp;
+
+/// Struct representing the currently active stack.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Stack {
+    /// Lowest stack address
+    pub bottom: usize,
+    /// Highest stack address
+    pub top: usize,
+
+    /// We'll be implementing nasty stuff on this Struct that requires it to
+    /// not get sent across stacks.
+    _not_send: PhantomData<*const ()>,
+}
+
+impl Stack {
+    /// Gets a handle for the currently active stack.
+    pub fn get() -> Self {
+        let sp = sp();
+        let stack = crate::arch::stack();
+        if stack.size() > 0 {
+            assert!(stack.top >= stack.bottom);
+
+            // TODO: verify bounds (are they inclusive?)
+            assert!(stack.bottom <= sp && stack.top >= sp);
+        }
+        stack
+    }
+
+    pub const fn default() -> Self {
+        Self {
+            bottom: 0,
+            top: 0,
+            _not_send: PhantomData,
+        }
+    }
+
+    pub(crate) const fn new(bottom: usize, top: usize) -> Self {
+        Self {
+            bottom,
+            top,
+            _not_send: PhantomData,
+        }
+    }
+
+    /// Returns the total size of the current stack.
+    pub fn size(&self) -> usize {
+        self.top - self.bottom
+    }
+
+    /// Returns the amount of currently free stack space.
+    pub fn free(&self) -> usize {
+        self.size() - self.used()
+    }
+
+    /// Returns the amount of currently used stack space.
+    pub fn used(&self) -> usize {
+        self.top - sp()
+    }
+
+    /// Returns the minimum free stack space since last repaint.
+    ///
+    /// This re-calculates and thus runs in `O(n)`!
+    pub fn free_min(&self) -> usize {
+        let mut free = 0usize;
+        for pos in self.bottom..self.top {
+            // Safety: dereferencing ptr to valid memory, read only
+            if unsafe { *(pos as *const u8) } == 0xCC {
+                free += 1;
+            }
+        }
+        free
+    }
+
+    /// Returns the maximum stack space used since last repaint.
+    ///
+    /// Equivalent to `size() - free_min()`.
+    ///
+    /// This re-calculates and thus runs in `O(n)`!
+    pub fn used_max(&self) -> usize {
+        self.size() - self.free_min()
+    }
+
+    /// Repaints the stack.
+    pub fn repaint(&self) {
+        let sp = crate::arch::sp();
+        if self.size() == 0 {
+            return;
+        }
+
+        // sanity check
+        assert!(self.bottom <= sp && sp <= self.top);
+
+        // Safety: writing to inactive part of active is fine.
+        unsafe {
+            for pos in self.bottom..sp {
+                write_volatile(pos as *mut u8, 0xCC);
+            }
+        }
+    }
+}
--- a/src/ariel-os-threads/src/thread.rs
+++ b/src/ariel-os-threads/src/thread.rs
@@ -24,6 +24,11 @@ pub struct Thread {
     /// Core affinity of the thread.
     #[cfg(feature = "core-affinity")]
     pub core_affinity: crate::CoreAffinity,
+
+    /// Lowest stack address
+    pub stack_bottom: usize,
+    /// Highest stack address
+    pub stack_top: usize,
 }
 
 /// Possible states of a thread
@@ -60,6 +65,8 @@ impl Thread {
             tid: ThreadId::new(0),
             #[cfg(feature = "core-affinity")]
             core_affinity: crate::CoreAffinity::no_affinity(),
+            stack_top: 0,
+            stack_bottom: 0,
         }
     }
 }
--- a/src/ariel-os/src/lib.rs
+++ b/src/ariel-os/src/lib.rs
@@ -40,6 +40,8 @@ pub use ariel_os_power as power;
 #[cfg(feature = "random")]
 #[doc(inline)]
 pub use ariel_os_random as random;
+#[doc(inline)]
+pub use ariel_os_rt as rt;
 #[cfg(feature = "storage")]
 #[doc(inline)]
 pub use ariel_os_storage as storage;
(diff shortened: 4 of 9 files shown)


src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// Returns the minimum free stack space since last repaint.
+    ///
+    /// This re-calculates and thus runs in `O(n)`!
+    pub fn free_min(&self) -> usize {
+        let mut free = 0usize;
+        for pos in self.bottom..self.top {
+            // Safety: dereferencing ptr to valid memory, read only
+            if unsafe { *(pos as *const u8) } == 0xCC {
```
ROMemories, line 72: I suppose `0xCC` is the paint color? Can you define a constant for it?

ROMemories, line 72: Is it enough to rely on a single byte? Isn't there quite a high risk of collision with the data actually in the stack?

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub fn repaint(&self) {
+        let sp = crate::arch::sp();
+        if self.size() == 0 {
+            return;
+        }
+
+        // sanity check
+        assert!(self.bottom <= sp && sp <= self.top);
+
+        // Safety: writing to inactive part of active is fine.
```
ROMemories, line 98: I think this sentence is missing a couple of words.

ROMemories, line 98: This comment is not sufficient: it needs at least to provide arguments regarding data races: the method should take `self` as `&mut` and/or provide reason about the type being `!Sync`, and making this clearer.

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.top - sp()
+    }
+
+    /// Returns the minimum free stack space since last repaint.
+    ///
+    /// This re-calculates and thus runs in `O(n)`!
+    pub fn free_min(&self) -> usize {
+        let mut free = 0usize;
+        for pos in self.bottom..self.top {
+            // Safety: dereferencing ptr to valid memory, read only
```
ROMemories, line 71: There needs to be arguments regarding *why* the pointer is valid, especially regarding its lifetime: I think this relies on the type being `!Send`, but this needs to be mention explicitly here (and *also* when writing to paint the stack below), *and* when making the type `!Send` ("nasty stuff" is not precise enough), so that the type is not mistakenly made `Send` in the future when this evolves.

src/ariel-os-rt/src/lib.rs
```diff
@@ -66,6 +72,27 @@ mod isr_stack {
         "#,
         size = const ISR_STACKSIZE
     );
+
+    pub fn limits() -> (usize, usize) {
+        // ISR stack
+        // TODO: multicore!
+        unsafe extern "C" {
+            static _stack_bottom: u32;
+            static _stack_start: u32;
```
ROMemories, line 81: I understand that these come from here:

https://github.com/ariel-os/ariel-os/blob/39b7746e2503628b44f8dc247cb0f9ceb6a7bbcf/src/ariel-os-rt/isr_stack.ld.in#L16-L17

But I think these should be `usize`:

```suggestion
            static _stack_bottom: usize;
            static _stack_start: usize;
```

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::arch::sp;
+
+/// Struct representing the currently active stack.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Stack {
+    /// Lowest stack address
+    pub bottom: usize,
+    /// Highest stack address
+    pub top: usize,
```
ROMemories, line 13: Have you considered using a `Range` instead? They're not perfect (mostly because they're not `Copy`), but they provide some nice semantics, for instance it's very clear whether the range is inclusive or not.

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,105 @@
...
+        if stack.size() > 0 {
+            assert!(stack.top >= stack.bottom);
+
+            // TODO: verify bounds (are they inclusive?)
+            assert!(stack.bottom <= sp && stack.top >= sp);
+        }
+        stack
+    }
+
+    pub const fn default() -> Self {
```
ROMemories, line 34: Does this need to be public?

src/ariel-os/src/lib.rs
```diff
@@ -40,6 +40,8 @@ pub use ariel_os_power as power;
 #[cfg(feature = "random")]
 #[doc(inline)]
 pub use ariel_os_random as random;
+#[doc(inline)]
+pub use ariel_os_rt as rt;
```
ROMemories, line 44: The `INIT_FUNCS` statics needs to be made `doc(hidden)`, otherwise it appears in user docs.

ROMemories (commented):
Left a couple of drive-by comments; will try to review more thoroughly later.

src/ariel-os-rt/src/stack.rs
kaspar030: hmm. not sure where. If I add a constant, I need to add it in at least two crates. `0xCC` is basically `11001100` in binary.

kaspar030: This is writing byte wise (just so I don't need to check for proper alignment here), but *every* byte. As a result, the stack is painted with alternating `11` and `00`.

This is checked from the bottom to the top, so any data consisting of only `0xCCCCCCCC` at the bottom of the actually stack would get counted as "unused".
While this might happen, it is unlikely, it also doesn't matter in practice - we use the stack sizes only for informational purposes.

src/ariel-os-rt/src/stack.rs
kaspar030: I've reworded and added context. There's no `&mut self` or `self` needed, only `!Send`.

ROMemories: Why don't we also need `!Sync`?

src/ariel-os/src/lib.rs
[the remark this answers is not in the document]
kaspar030: Yes. Polished enough for a first iteration (so, low standard.)

src/ariel-os-rt/src/lib.rs
kaspar030: type is irrelevant, linker passes values as *addresses* of symbols.

ROMemories: Indeed!

src/ariel-os-rt/src/stack.rs
kaspar030: I hope the (huge) "Assumptions" text above is covering this now ...

src/ariel-os-rt/src/stack.rs
kaspar030: Ah. Yeah I guess we actually do, as the methods take `&self`. But this is a doc & reasoning issue, as `PhantomData<*const()>` is both `!Send` and `!Sync`, right?

src/ariel-os-rt/src/stack.rs
kaspar030: Not considered. I did consider a type with named fields (as `(usize, usize)` is prone to mix up bottom and top order). Sometimes I want the range to be inclusive, sometimes not. E.g., some of the sanity check check for equality, which is the case e.g., when both values are zero.

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Both 1. and 2. are the case on cortex-m, risc-v and xtensa, as an ISR could technically do so at any time
+///    anyways.
+/// 3. is true on Ariel.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Stack {
+    /// Lowest stack address
+    pub bottom: usize,
+    /// Highest stack address
+    pub top: usize,
```
ROMemories, line 29: Assuming the stack grows towards lowest addresses, these fields' names are very confusing to me: the bottom of the stack is the first thing that is written to it (at the highest address).

src/ariel-os-rt/src/stack.rs
kaspar030: ... added `!Sync` to the reasoning.

src/ariel-os-rt/src/stack.rs
kaspar030: you got a point there.

kaspar030: So, would "lowest" and "highest" be less confusing?

ROMemories: Yes! Also that would be more correct because I think the calculcations then assume that anyway.

kaspar030: Renamed them all!

src/ariel-os-threads/src/thread.rs
```diff
@@ -73,6 +93,6 @@ mod tests {
         // `ThreadData` is arch-specific, and is replaced with a dummy value in tests; its size is
         // non-zero otherwise.
         assert_eq!(size_of::<ThreadData>(), 0);
-        assert_eq!(size_of::<Thread>(), size_of::<ThreadData>() + 24);
+        assert_eq!(size_of::<Thread>(), size_of::<ThreadData>() + 40);
```
kaspar030, line 96: This is 16b larger, with two usize added. So of course host-size usize.

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,180 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Stack {
+    /// Lowest stack address
+    lowest: usize,
+    /// Highest stack address
+    highest: usize,
+
+    /// Basically we need to ensure that `lowest` and `highest` precisely correspond
+    /// to the currently active stack. `!Send+!Sync` ensures that the object will
+    /// be sent to another thread (or ISR), which implies it stays on the same stack.
```
ROMemories, line 49:
```suggestion
    /// to the currently active stack. `!Send` ensures that the instance will
    /// not be sent to another thread (or ISR), which implies it stays on the same stack.
```

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,180 @@
...
+///
+/// The machinery for stack painting has a couple of assumptions:
+///
+/// 1. It is safe for an active stack to *overwrite* unused stack space from its limit (lowest
+///    address, including) to its stack pointer (not including) through raw pointers.
+/// 2. It is safe to *read* unused stack space below the raw stack pointer down to its limit (lowest address).
+/// 3. The limits of an active stack never change.
+/// 4. It is fine to specify zero for both `lowest` and `highest`, in which case usage data is invalid
+///    (always zero), but no unsafety arises.
+/// 5. `!Send + !Sync` keeps `Stack` on the stack it was created for.
```
ROMemories, line 34: That's not assumption, just how the language works:
```suggestion
/// 5. `!Send` keeps `Stack` on the stack it was created for.
```

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,180 @@
...
+/// # Note
+///
+/// The machinery for stack painting has a couple of assumptions:
+///
+/// 1. It is safe for an active stack to *overwrite* unused stack space from its limit (lowest
+///    address, including) to its stack pointer (not including) through raw pointers.
+/// 2. It is safe to *read* unused stack space below the raw stack pointer down to its limit (lowest address).
+/// 3. The limits of an active stack never change.
+/// 4. It is fine to specify zero for both `lowest` and `highest`, in which case usage data is invalid
+///    (always zero), but no unsafety arises.
```
ROMemories, line 33:
```suggestion
///    (always zero), but no unsoudness arises.
```

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,180 @@
...
+        // makes any following `used_max()` call include whatever the ISR wrote on this stack.
+        unsafe {
+            for pos in self.lowest..sp {
+                write_volatile(pos as *mut u8, 0xCC);
+            }
+        }
+    }
+
+    /// Get this stack object's `highest` address
+    pub fn highest(&self) -> usize {
```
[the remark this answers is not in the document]
kaspar030: I re-added them while debugging. IMO as soon as you start looking this deep, they become useful. Let's re-visit when we make the API public.

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///   the architecture dependent code don't make sense)
+    #[must_use]
+    pub fn get() -> Self {
+        let sp = sp();
+        let stack = crate::arch::stack();
+        if !stack.is_empty() {
+            assert!(stack.highest >= stack.lowest);
+
+            // TODO: verify bounds (are they inclusive?)
+            assert!(stack.lowest <= sp && stack.highest >= sp);
```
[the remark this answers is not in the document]
kaspar030: Just did. SP can legally point to both the lowest and the highest address, so the bounds are inclusive.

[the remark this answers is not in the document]
kaspar030: ups, that was the answer to the todo question. did reorder the bounds also.

src/ariel-os-rt/src/stack.rs
```diff
@@ -0,0 +1,180 @@
...
+///    during execution.
+///
+/// Note that this technique only provides a lower bound of stack usage, as the values stored in
+/// the stack may "collide" with the paint values.
+/// In the current implementation, and assuming the stack data follows a uniform distribution, this
+/// is unlikely to result in an underestimation of more than one byte.
+///
+/// # Note
+///
+/// The machinery for stack painting has a couple of assumptions:
```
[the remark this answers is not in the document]
kaspar030: Hm. Personally I prefer "this is how it works" explanations to be more visible. Did remove the slash anyways.

PR #982, opened 2025-04-09
chore(deps): switch to embassy-executor(-macros) upstream main

kaspar030:
# Description

This is cut out from #951, updating our embassy fork to include https://github.com/embassy-rs/embassy/pull/4020.

## Issues/PRs references

#951 

## Change checklist

### DIFF
--- a/examples/http-client/src/main.rs
+++ b/examples/http-client/src/main.rs
@@ -1,6 +1,5 @@
 #![no_main]
 #![no_std]
-#![feature(impl_trait_in_assoc_type)]
 
 use ariel_os::{
     config,
--- a/examples/http-client/src/main.rs
+++ b/examples/http-client/src/main.rs
@@ -1,7 +1,6 @@
 #![no_main]
 #![no_std]
 #![feature(impl_trait_in_assoc_type)]
-#![feature(type_alias_impl_trait)]
 
 use ariel_os::{
     config,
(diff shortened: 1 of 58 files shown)


examples/http-client/src/main.rs
```diff
@@ -1,7 +1,5 @@
 #![no_main]
 #![no_std]
-#![feature(impl_trait_in_assoc_type)]
-#![feature(type_alias_impl_trait)]
```
ROMemories, line 4: This line doesn't seem to be in the right commit.

ROMemories:
The goal of this PR is to drop unstable features *in applications* that would be required by Ariel OS itself (keeping `type_alias_impl_trait` in `http-server` is fine because I believe the need for it comes from picoserve).

examples/http-client/src/main.rs
kaspar030: pff then be it. we'll live.

src/ariel-os-macros/Cargo.toml
```diff
@@ -17,9 +17,7 @@ quote = "1.0.35"
 syn = { version = "2.0.47", features = ["full"] }
 
 [dev-dependencies]
-embassy-executor = { workspace = true, features = [
-  "nightly",
-] } # FIXME: embassy_executor::task requires embassy_executor to be imported at crate-level
+embassy-executor = { workspace = true }
```
ROMemories, line 20:
```suggestion
```
I think you can drop the dependency entirely now.

ROMemories (approved):

- `examples/http-client` on `espressif-esp32-c6-devkitc-1` (with stacksize adjusted)
- `examples/usb-keyboard` on `nrf52840dk`
- `examples/udp-echo` on `rpi-pico`
- `examples/udp-echo` on `st-nucleo-wb55`

PR #983, opened 2025-04-09
feat(boards): add support for the FireBeetle 2 ESP32-C6

ROMemories:
# Description

This adds support for the [FireBeetle 2 ESP32-C6](https://web.archive.org/web/20250214091848/https://wiki.dfrobot.com/SKU_DFR1075_FireBeetle_2_Board_ESP32_C6). This doesn't update the support matrix as that board is untested for now.

## Change checklist

### DIFF
--- a/examples/blinky/src/pins.rs
+++ b/examples/blinky/src/pins.rs
@@ -6,6 +6,9 @@ ariel_os::hal::define_peripherals!(LedPeripherals {
     led: P0_21,
 });
 
+#[cfg(context = "dfrobot-firebeetle2-esp32-c6")]
+ariel_os::hal::define_peripherals!(LedPeripherals { led: GPIO15 });
+
 #[cfg(context = "nordic-thingy-91-x-nrf9151")]
 ariel_os::hal::define_peripherals!(LedPeripherals { led: P0_29 });
 
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1344,6 +1344,9 @@ builders:
   - name: bbc-microbit-v2
     parent: nrf52833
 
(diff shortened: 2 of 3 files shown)


laze-project.yml
```diff
@@ -1344,6 +1344,9 @@ builders:
   - name: bbc-microbit-v2
     parent: nrf52833
 
+  - name: dfrobot-firebeetle2-esp32-c6
```
ROMemories, line 1347: Maybe we'd want a dash between `firebeetle` and `2` because there's a space between them, but I went with consistency with `rpi-pico2`.

laze-project.yml
```diff
@@ -1344,6 +1344,9 @@ builders:
...
+    parent: esp32c6
```
ROMemories, line 1348: This doesn't introduce an intermediate laze context as for the `espressif-esp32-c6-devkitc-1`, because this board doesn't use an off-the-shelf hardware module.

examples/blinky/src/pins.rs
```diff
@@ -6,6 +6,9 @@ ariel_os::hal::define_peripherals!(LedPeripherals {
     led: P0_21,
 });
 
+#[cfg(context = "dfrobot-firebeetle2-esp32-c6")]
+ariel_os::hal::define_peripherals!(LedPeripherals { led: GPIO15 });
```
ROMemories, line 10: There *is* an onboard LED on this board.

kaspar030: nice!

kaspar030 (approved):

PR #987, opened 2025-04-10
feat(build): default to `stable` build

kaspar030:
# Description

This PR changes Ariel's default to use the stable Rust toolchain by default.

## Open Questions

- [x] what is our actual MSRV? -> arbitrarily defined 1.85
- [ ] `examples/http-server`
- [ ] docs
- [x] CI toolchains, install both pinned nightly and MSRV stable? :+1:
- [x] CI: build only stable or add nightly? => building nightly at nightly build time

## Change checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -52,6 +52,8 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ steps.get_toolchain.outputs.toolchain }}
+          components: rust-src
+          targets: riscv32imac-unknown-none-elf
 
       - name: rust cache
         uses: Swatinem/rust-cache@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -238,6 +238,8 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ steps.get_toolchain.outputs.toolchain }}
+          targets: riscv32imac-unknown-none-elf
+          components: rust-src
 
       - name: rust cache
         uses: Swatinem/rust-cache@v2


README.md
```diff
@@ -29,8 +29,7 @@ See the [Getting Started page][getting-started-mdbook] in the User Guide to get
 
 ## Minimum Supported Rust Version (MSRV)
 
-Ariel OS makes use of selected Rust unstable features. For the time being, it is
-recommended to use the nightly version pinned by [`rust-toolchain.toml`](rust-toolchain.toml).
+Ariel OS currently requires at least Rust version 1.85.
```
ROMemories, line 32: What we want is an MSRV *policy*: which versions are supported, when/on which conditions do we increase the MSRV (see [this discussion](https://github.com/rust-lang/api-guidelines/discussions/231) for more). We could simply go with "We support the latest stable Rust version" and that would be it (and I think we should do that actually, at least for now).

src/ariel-os/Cargo.toml
```diff
@@ -4,6 +4,9 @@ version = "0.1.0"
 license.workspace = true
 edition.workspace = true
 
+# This should be the Ariel OS MSRV.
+rust-version = "1.85"
```
ROMemories, line 8: [Cargo documentation for reference](https://doc.rust-lang.org/cargo/reference/rust-version.html)

README.md
```diff
@@ -27,10 +27,11 @@ See the [Getting Started page][getting-started-mdbook] in the User Guide to get
 
 ![Terminal screencast of compiling and flashing the hello-world example](./book/src/hello-world_render.svg)
 
-## Minimum Supported Rust Version (MSRV)
+## Minimum Supported Rust Version (MSRV) and Policy
 
-Ariel OS makes use of selected Rust unstable features. For the time being, it is
-recommended to use the nightly version pinned by [`rust-toolchain.toml`](rust-toolchain.toml).
+Ariel OS currently requires at least Rust version 1.85.
```
ROMemories, line 32:
```suggestion
```
I think we should remove this line: it's redundant with the `rust-version` key in `ariel-os`'s manifest and I'd rather have that as a single source of truth, on top of not requiring us to touch the README when we bump this.
Additionally, thinking about this some more, it's not obvious to me that all crates of the project are going to have the same MSRV: a future `ariel-os-sensors` crate, which independently-published sensor driver crates will have to depend on, may actually having a lower MSRV.

README.md
```diff
@@ -27,10 +27,11 @@ See the [Getting Started page][getting-started-mdbook] in the User Guide to get
...
+We support the latest stable Rust version. Updating the MSRV is considered a
```
ROMemories, line 33: What's the reasoning for making this a breaking change? [Cargo guidelines](https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust) lean on making this a minor change instead, but this is pretty nuanced:

> It is generally recommended to treat this as a minor change, rather than as a major change, for various reasons.

ROMemories:
I think the `install-toochain` laze task needs updating to take into account the `CARGO_TOOLCHAIN` variable: I suppose simply adding `${CARGO_TOOLCHAIN}` after `rustup` should be enough.

laze-project.yml
```diff
@@ -616,6 +616,9 @@ modules:
           - --cfg context=\"cortex-m33f\"
 
   - name: xtensa
+    context:
+      - esp32
+      - esp32s3
```
ROMemories, line 621: Why do we need this?

kaspar030: I added this so we can do `laze build --select xtensa` and only build for xtensa boards. That was before I removed the "--select stable". So I think now it is not needed anymore in this PR.

ROMemories:
I suppose we also need to set the `release` of Ferrocene to stable.

laze-project.yml
ROMemories: So can we remove these or do we still need them?

ROMemories (approved):
Successfully tested the `hello-world` (compiles on stable) and `http-client` (requires nightly) on espressif-esp32-c6-devkitc-1.

scripts/rust-toolchain.sh
```diff
@@ -2,4 +2,4 @@
 
 # script is used by github workflows
 
-grep '^channel =' rust-toolchain.toml | sed 's/.*= "\(.*\)"/\1/'
+grep 'CARGO_TOOLCHAIN: .*nightly' laze-project.yml | grep -o 'nightly-[0-9-]\+'
```
chrysn, line 5: This looks a bit fragile, what's the best route here?

* Allow laze to source variables from a file? `env: global: CARGO_TOOLCHAIN: < our-toolchain.txt` (YAML may have something to offer, probably not file access but distinguishing an action from a string)
* Allow laze to export a variable? (`laze dump-var global CARGO_TOOLCHAIN`?)
* Just hope that this never breaks in unexpected non-CI-failing ways?

PR #998, opened 2025-04-16
docs: fix paths to the networking setup instructions

Ollrogge:
# Description
Fix linking to the [networking](https://github.com/ariel-os/ariel-os/tree/main/examples#networking) section in tests and examples READMEs.


ROMemories (approved):
Thanks!

PR #1000, opened 2025-04-17
fix(laze): Remove unsupported targets from context of coap-blinky test

Ollrogge:
# Description
Remove targets from the `coap-blinky` laze context that currently do not compile.

E.g. for `st-nucleo-f401re` I got:

```bash
 0: "coap-blinky" cannot resolve "coap-server"
    1: "coap-server" cannot resolve "coap"
    2: "coap" cannot resolve "random"
    3: "random" cannot resolve "hwrng"
    4: "hwrng" cannot resolve "has_hwrng"
    5: "has_hwrng" cannot resolve "doc-only"
    6: module "doc-only" not found
```

And for `stm32u083c-dk`

```bash
 0: "coap-blinky" cannot resolve "coap-server"
    1: "coap-server" cannot resolve "coap"
    2: "coap" cannot resolve "network"
```

### DIFF
--- a/tests/coap-blinky/laze.yml
+++ b/tests/coap-blinky/laze.yml
@@ -4,15 +4,12 @@ apps:
       - coap-server
       - ?coap-server-config-demokeys
     context:
-      # list of contexts that have an entry in `pins.rs`
-      - bbc-microbit-v2
       - esp
       - nrf52840dk
       - nrf5340dk
       - nrf9160dk-nrf9160
       - particle-xenon
       - rp
-      - st-nucleo-f401re
       - st-nucleo-h755zi-q
       - st-nucleo-wb55
       - stm32u083c-dk


kaspar030:
Happy #1000 everyone! 🥳

ROMemories (changes requested):
This list of contexts only *restricts* which contexts this test could possibly be compiled for, *provided the other requirements are also met* (in case `coap-server` is also required, which itself requires `network` and `random`). This list in itself does not imply that it is possible to compile this test on all the listed contexts.

So for instance `nrf5340dk` and `nrf9160dk-nrf9160` should definitely stay here because this test should be immediately usable as soon as we add support for the `hwrng` and `network` on these respectively.

We could however possibly remove `bbc-microbit-v2` and `st-nucleo-f401re` from this list, as there is no way to add support for networking on these anyway.

ROMemories:
Closing as contexts don't need to be manually listed anymore in this test since https://github.com/ariel-os/ariel-os/commit/beff30d136b34b018b39abe6df0a6fed421aa6f3 from #1282.

PR #1010, opened 2025-05-01
docs: add CHANGELOG.md

kaspar030:
# Description

Adds an initial `CHANGELOG.md` in preparation for the next release.
This is about the format, not the content, we'll track that elsewhere.

The format comes directly from [cargo-release](https://github.com/crate-ci/cargo-release/blob/master/docs/faq.md#maintaining-changelog)'s suggestions, and (apart from actual entries) is kept automatically.

## Change checklist


CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- next-header -->
+
+## [Unreleased] - ReleaseDate
+
+### Added
+
+- feat(build): default to `stable` build (#987)
```
bergzand, line 14: The `feat` and 'Added' header looks duplicate to me. Is there a case where a `feat` type of PR is not an additional feature?

kaspar030: It is, this is also the unmodified PR subject. With the huge amount of entries we'll have, maybe not having to edit every single one to drop the "feat" saves time. It also creates some consistency - at the cost of some redundancy.

bergzand: I think a regex is sufficient to handle this, no need to edit every single line manually.

> It also creates some consistency - at the cost of some redundancy.

To me it looks like extra visual clutter in the way of reading the actual release notes.

CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+<!-- next-header -->
+
+## [Unreleased] - ReleaseDate
+
+### Added
+
+- feat(build): default to `stable` build (#987)
```
bergzand, line 14: Can we teach cargo-release to add the URLs here?

```suggestion
- feat(build): default to `stable` build ([#987](https://github.com/ariel-os/ariel-os/pull/987))
```

kaspar030: Yeah, I pushed something.

CHANGELOG.md
kaspar030: ok. we'd need to keep the `build` though, do `- feat(foo): ` -> `- foo: ` I guess?

kaspar030: ... to me it doesn't add much clutter btw., and where it does it, it adds  also the recognition that this is the PR title ...

ROMemories: > ... to me it doesn't add much clutter btw., and where it does it, it adds also the recognition that this is the PR title ...

Agree, I'd also prefer we keep it.

bergzand: > ok. we'd need to keep the `build` though, do `- feat(foo): ` -> `- foo: ` I guess?

Exactly what I had in mind, maybe add bold markers around the 'foo' while we're at it.

> > ... to me it doesn't add much clutter btw., and where it does it, it adds also the recognition that this is the PR title ...
> 
> Agree, I'd also prefer we keep it.

Ok, let's see how it looks when there's a huge amount of entries.

bergzand:
I noticed you added content to this PR with 88860867931c5b50167b5ce35bd6be49415ac6ac, did you change your mind in keeping this PR only about the format?

kaspar030:
> I noticed you added content to this PR with [8886086](https://github.com/ariel-os/ariel-os/commit/88860867931c5b50167b5ce35bd6be49415ac6ac), did you change your mind in keeping this PR only about the format?

yes :)

src/ariel-os/Cargo.toml
```diff
@@ -7,6 +7,19 @@ edition.workspace = true
 # This should be the Ariel OS MSRV.
 rust-version = "1.85"
 
+
```
ROMemories, line 10:
```suggestion
```

CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
...
+- feat(boards): add support for the ST NUCLEO-F411RE (#1002)
+- fix(log): add support for `log` on architectures without atomics (#990)
```
bergzand, line 16: This should be under a `### Fixed` header

kaspar030:
If we're fine with the list, I'd let cargo release munge in the PR links.

bergzand (approved):
Ack, still disagree with keeping the commit type in the list, but not going to block on that.

PR #1013, opened 2025-05-01
build(laze): add a `cargo-expand` task

Ollrogge:
# Description
As Ariel Os makes extensive use of macros it can be helpful to see what the code looks like after macro expansion.
Therefore I added an `expand` task which runs `cargo expand`.


ROMemories (approved):
Note that since #951, this requires manually enabling the `nightly` laze module (so that the `ariel-os-cargo.toml` gets picked up through `.cargo/config.toml`), but the same is needed for the already existing (but currently undocumented) `clippy` laze task and this seems useful, so approving.

Tested successfully—after installing [`cargo-expand`](https://github.com/dtolnay/cargo-expand)—by running:

```sh
laze -C examples/hello-world build -s nightly -b rpi-pico-w expand 
```

PR #1020, opened 2025-05-07
fix(stm32): increase eth-stm32 executor stacksize requirement

kaspar030:
# Description

This increases the stacksize requirements for stm32 ethernet.
Tested `tcp-echo`, `udp-echo`, `http-server`, `http-client` with both thread and interrupt executors.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -979,6 +979,8 @@ modules:
       - network_device
     env:
       global:
+        executor_stacksize_required:
+          - "32768"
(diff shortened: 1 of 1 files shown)


ROMemories (approved):
Successfully tested in hardware on `st-nucleo-h755zi-q`:
- `tcp-echo`
- `udp-echo`
- `http-server`
- `http-client`

using Ethernet as the network link.

PR #1022, opened 2025-05-09
fix: remove defmt'isms

chrysn:
# Description

After #972, there were still examples left of defmt'isms -- just not in the format strings any more, but around them where #972 doesn't catch it.

## Testing

`laze build -g -b st-nucleo-wb55 -D LOG=debug -s log -d defmt`

(Or let's finally add support for boards where the default logging facade is not through defmt, or let's admit that defmt is the way to go and that even serial-out devices will produce defmt there.)

## Issues/PRs references

<del>Blocked by: #972 (so please just view the top 3 commits)</del>

## Open Questions

* [ ] The fix for the storage example is not pretty -- what *are* applications that desire to be portable but want to debug custom types supposed to do? Once more, I recommend we just make defmt a dependency (esp. now that it has a 1.0), and derive Format unconditionally, rather than having ifdef hell. Just b/c it's a dependency and we derive Format doesn't mean we have to emit any of its code.

  Depending on the outcome of this, we may want to pub use Format in the log module unconditionally top-level, and if they're transparent -- and then we can also drop my new types and just pub use defmt's, because they are already suitably transparent.

## Change checklist

### DIFF
--- a/examples/storage/src/main.rs
+++ b/examples/storage/src/main.rs
@@ -4,7 +4,7 @@
 
 use ariel_os::debug::{
     ExitCode, exit,
-    log::{Hex, defmt, info},
+    log::{Debug2Format, Hex, info},
 };
 
 // Imports for using [`ariel_os::storage`]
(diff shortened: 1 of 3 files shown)


chrysn:
This is on hold until #1025 is sorted out.

examples/storage/src/main.rs
```diff
@@ -13,7 +14,9 @@ use serde::{Deserialize, Serialize};
 /// Example object.
 ///
 /// The serde Serialize / Deserialize traits are required for storage
-#[derive(Serialize, Deserialize, Debug, defmt::Format)]
+#[derive(Serialize, Deserialize, Debug)]
+// FIXME: How should a user's application that may or may not run on defmt but has a custom type
```
ROMemories, line 18: By introducing a dedicated laze module, as was done in #1999.

ROMemories:
I believe this has now been entirely addressed by #1485 and #1999. Feel free to re-open if I've missed something.

PR #1024, opened 2025-05-12
build(laze): add `info-boards` task

kaspar030:
# Description

This adds a new task that just echoes the builder.

Useful in combination with `-m` for listing the boards that an application can be built for, e.g., `laze build -m info-boards | grep -v ^laze:`.

(laze needs a better story for machine parsing of task output. starting already here with reducing noise: https://github.com/kaspar030/laze/pull/716. I kept free-coding "info-boards` more than once, so maybe it already usefule as is.)

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -177,6 +177,11 @@ contexts:
         cmd:
           - rust-objdump -S ${out}
 
+      info-boards:
+        build: false
+        cmd:
+          - echo ${builder}
+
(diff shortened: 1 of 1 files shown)


ROMemories (approved):
Maybe there should be quotation marks around `${builder}`, but builders cannot really have spaces, and I don't use we have these in other places anyway.

PR #1025, opened 2025-05-13
feat(log): make defmt re-exports available unconditionally

chrysn:
# Description

As discussed in today's weekly, let's go a step more toward defmt -- user applications that want to be portable and log any own data structure already need to depend on defmt to derive Format on it, and it's 1.0 now, and it allows us to get rid of a lot of "ifdef hell".

## Issues/PRs references

Replaces https://github.com/ariel-os/ariel-os/pull/1022 (that'd would have added workarounds)

## Open Questions

* [ ] Will it build? (It's a whack-a-mole right now)
* [ ] Did I file all cases right into "this upstream module does an own defmt-or-log and thus needs to know" vs "this upstream module provides defmt impls and thus will get defmt set all the time"? (Really we should set it anyway -- unless they do defmt-or-log in a non-additive way: logging to defmt without the logging turned on should be no-ops).

## Change checklist

### DIFF
--- a/src/ariel-os-debug-log/Cargo.toml
+++ b/src/ariel-os-debug-log/Cargo.toml
@@ -12,13 +12,13 @@ workspace = true
 logging-facade.xor = { features = ["defmt", "log"] }
 
 [dependencies]
-defmt = { workspace = true, optional = true }
+defmt = { workspace = true }
 featurecomb = { workspace = true }
 log = { workspace = true, optional = true }
 
 [features]
 ## Enables `defmt` as logging facade.
-defmt = ["dep:defmt"]
+defmt = []
 ## Enables `log` as logging facade.
 log = ["dep:log"]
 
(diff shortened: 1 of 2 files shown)


chrysn:
There's a bit of a road bump here: Some (mainly embassy) crates take the "defmt" feature to mean both "impl Format for your types" but also "use defmt to panic!" -- kind of makes sense, but only until crates start implementing (but not using) defmt all over the place. See also [matrix discussion with dirbaio](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$wfjQ1CQ20yUk3Plzz199aUc5fOBek_WzGC7qERfeVEo?via=matrix.org&via=catircservices.org&via=mozilla.org), tests pending at https://github.com/chrysn/cbor-diag-py/pull/14.

This doesn't endanger the whole endeavor, but means that for crates that actually make use of defmt, we won't get the Format impls unconditionally (which is already bad for #972).

PR #1028, opened 2025-05-13
refactor(storage): use `if` in place of `match` with an empty branch

jkr-ariel:
# Description

Small refactoring of the `init` function in [ariel-os-storage/src/lib.rs](https://github.com/ariel-os/ariel-os/blob/92d3aeac74ba33b3d9ed3dd575ae91a3a227678f/src/ariel-os-storage/src/lib.rs#L93-L101) to use an `if` statement instead of a `match` with an empty branch.

## Change checklist

### DIFF
--- a/src/ariel-os-storage/src/lib.rs
+++ b/src/ariel-os-storage/src/lib.rs
@@ -90,14 +90,9 @@ pub async fn init(p: &mut OptionalPeripherals) {
     embassy_time::block_for(embassy_time::Duration::from_millis(10));
 
     // Use a marker to ensure that this storage is initialized.
-    match get::<u8>(MARKER_KEY).await {
-        Ok(Some(val)) if val == MARKER_VALUE => {
-            // all good
-        }
-        _ => {
-            ariel_os_debug::log::info!("storage: initializing");
-            erase_all().await.unwrap();
-        }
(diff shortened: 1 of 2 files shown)


ROMemories (approved):

src/ariel-os-storage/src/lib.rs
```diff
@@ -90,14 +90,9 @@ pub async fn init(p: &mut OptionalPeripherals) {
...
     // Use a marker to ensure that this storage is initialized.
-    match get::<u8>(MARKER_KEY).await {
-        Ok(Some(val)) if val == MARKER_VALUE => {
-            // all good
-        }
-        _ => {
-            ariel_os_debug::log::info!("storage: initializing");
-            erase_all().await.unwrap();
-        }
+    if Ok(Some(MARKER_VALUE)) != get::<u8>(MARKER_KEY).await {
```
ROMemories, line 93: It seems that rustdoc is actually not too happy about this and complains with:
```
the trait `PartialEq` is not implemented for `FlashError`
```
I guess you'll have to use the [`matches!` macro](https://doc.rust-lang.org/stable/core/macro.matches.html) then.

ROMemories (approved):
Successfully tested the `storage` example on nrf52840dk.

Issue #1008, opened 2025-04-29
Support debug output over UART


ROMemories:
Currently the debug output supports either RTT (through the [`rtt-target`](https://docs.rs/rtt-target/latest/rtt_target/) crate) or JTAG-Serial with [`esp-println`](https://docs.espressif.com/projects/rust/esp-println/0.13.1/esp_println/index.html) (on ESP32).

We want to additionally support UART as a transport for the debug output, used for `print!()` and logging (with `info!()`, `debug!`, etc.) both through `defmt` and `log`.

## UART support

- [ ] First, we need a [UART abstraction](https://github.com/ariel-os/ariel-os/issues/830) so that we can access the UART peripheral in a uniform manner across all our supported MCUs.

## Providing a debug output over UART

- [ ] We will want to provide `print!` and `println!` macros that print over UART, which we may need to implement manually.
- [ ] This requires obtaining a UART peripheral along with the required pins, early enough to set up the debug output.
- [ ] We may also need to do something about the `debug::exit()` function which will not work over UART alone, as it is semihosting-powered and requires a debugging interface (e.g., SWD).

## `defmt`

- [ ] For `defmt`, we need to provide a [global_logger](https://defmt.ferrous-systems.com/global-logger) that works with UART drivers.
[`defmt-serial`](https://github.com/gauteh/defmt-serial) seems like a candidate, which provides such a `global_logger` for things that implement `embedded_io::Write`. We may need to rework it a bit, and add support for `embedded-io-async` as our UART abstraction will likely not support the blocking variant.

- [ ] As the data sent over UART will be defmt-encoded, we will need to integrate [`defmt-print`](https://github.com/knurling-rs/defmt/tree/main/print) to decode it and print it on the host side.

## `log`

- [ ] As we already provide our own logger implementation for `log` that stands on top of `println!()`, this should come for free as soon as we have a `println!` implementation.

chrysn:
Discussion in RIOT has shown that we may want to take some care as to how to handle flow control -- not because any physical UART we likely have would have this wired up (seems nobody does this) but mainly because when used through USB, this *may* (as in: some research needed, my tests are inconclusive) make the difference between startup messages going into the void (host enumerates, embedded device sends, Linux device is unconnected, data is lost) and the device keeping the to-be-sent messages buffered until that buffer either overflows or an actual client connects and pulls data out.

Issue #830, opened 2025-02-17
feat: add a UART abstraction


ROMemories:
Similarly to #421 and #440, we need a UART API that abstracts over differences within an MCU family and across MCU families.

Pointers:

- The `ariel-os-embassy-common` crate provides items that need to be shared between the `ariel-os-embassy` crate and the Ariel OS HAL crates (`ariel-os-esp`, `ariel-os-nrf`, `ariel-os-rp`, and `ariel-os-stm32`).
- The `ariel-os-embassy` crate provides the API that's user-visible, in its `api` module, whose items are re-exported at root level by the `ariel-os` crate.
- The Ariel OS HAL crates provide HAL/MCU-specific implementations, which are used by the `ariel-os-embassy` crate to provide a unified API.

The file structure inside each crate should match as much as possible for this kind of API.

ROMemories:
I think this has been completed by #1365, #1366, #1367, #1368, and #1369.

PR #1029, opened 2025-05-13
feat(debug-uart): add support for debug output over UART 

ROMemories:
# Description

This provides an additional debug output backend, on top of the existing RTT and `esp-println` ones. It is currently limited to STM32 and nRF, and only compatible with `log`, not `defmt`, and autoselects `log` when enabled. There is no ringbuffer or similar, and we do not try very hard to not drop debug lines during startup/before a UART consumer is connected.

The choice of peripherals is currently hard-coded and targets [IoT-LAB](https://www.iot-lab.info/).
This new functionality is not yet documented on purpose, as it's almost only useful in one use case, which we do not want to commit to just yet.

## How to review this PR

- This PR is better reviewed commit by commit.
- This PR can be tested in hardware using the following:
  ```sh
  laze -C examples/log build -s debug-uart -b nrf52840dk
  ```

## Testing

Successfully tested the `log` example on ST B-L475E-IOT01A, nRF52840-DK and nRF52-DK on IoT-LAB, and on STM32U083C-DK and ST-NUCLEO-WB55 locally.

## Issues/PRs references

Works towards #1008.
Could use #830.

## Change checklist


src/ariel-os-debug/src/lib.rs
```diff
@@ -115,6 +105,56 @@ mod backend {
...
+            }
+
+            Ok(())
+        }
+    }
+
+    pub fn init() {
+        crate::logger::init();
+    }
+
```
ROMemories, line 133:
```suggestion

// Based on <https://blog.m-ou.se/format-args/>.
```

laze-project.yml
```diff
@@ -1357,6 +1357,17 @@ modules:
         FEATURES:
           - ariel-os/rtt-target
 
+  - name: debug-uart
```
ROMemories, line 1360: Should we call this `uart-debug` instead?

laze-project.yml
```diff
@@ -1402,7 +1413,6 @@ modules:
       # There is an issue with multi-core on esp.
       - ?esp-single-core
       - sw/threading
-      - rtt-target
```
ROMemories, line 1405: Is RTT actually necessary for `embedded-test`?

laze-project.yml
nponsard: I prefer `debug-uart`, it has 'debug' at the start similar to `debug-logging-facade` already in the file.

src/ariel-os-embassy/src/debug_uart.rs
```diff
@@ -0,0 +1,108 @@
...
+    let _ = ariel_os_debug::DEBUG_UART_WRITE_FN.init(write_debug_uart);
+}
+
+fn write_debug_uart(buffer: &[u8]) {
+    #[cfg(context = "stm32")]
+    use embedded_io::Write as _;
+    #[cfg(context = "nrf")]
+    use embedded_io_async::Write as _;
+
+    // FIXME: do not unwrap
```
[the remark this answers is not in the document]
ROMemories: Went with returning a proper error from `ariel-os-embassy`, and then discarding it in `ariel-os-debug`.

laze-project.yml
```diff
@@ -923,7 +923,7 @@ modules:
     help: use probe-rs as runner
     selects:
       - ?debug-console
-      - rtt-target
+      - ?rtt-target
```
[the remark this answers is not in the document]
kaspar030: This means that probe-rs-run is now used even for boards that don't have rtt, which means, no output on those. I guess that's fine (as we currently don't have those)?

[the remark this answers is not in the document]
kaspar030: let's go with it and refine later.

src/ariel-os-debug/src/lib.rs
```diff
@@ -113,6 +103,59 @@ mod backend {
...
+    #[doc(hidden)]
+    pub static DEBUG_UART_WRITE_FN: OnceLock<fn(&[u8]) -> Result<(), Error>> = OnceLock::new();
+
+    struct DebugUart;
+
+    impl core::fmt::Write for DebugUart {
+        fn write_str(&mut self, s: &str) -> core::fmt::Result {
+            let bytes = s.as_bytes();
+
+            if let Some(debug_uart_write_fn) = DEBUG_UART_WRITE_FN.try_get() {
```
kaspar030, line 127: This we'll need to revisit. It basically means that ISR output will be lost if a thread is already printing, which is likely to happen given uart writes with only 1 character per ms. We'll need to go ring-buffer, make it line based (so output won't mingle), ...

kaspar030: Ah, this is a `OnceLock` not a mutex. sorry for the noise.

laze-project.yml
kaspar030: `semihosting` is, IIUC. Which in Ariel means, probe-rs, actually.

kaspar030 (approved):
Let's get this in.

It feels a bit rough. Also, points our nose to the fact that many of our examples that `exit()` actually reboot, without a debugger attached. E.g., `hello-world` on the nrf52dk keeps printing its wisdom in a loop. :/

Let's take it as "first iteration" :innocent:.

laze-project.yml
ROMemories: Then this should select the existing `semihosting` laze module we have, shouldn't it?

ROMemories: We decided this was out of scope for now.

PR #1034, opened 2025-05-15
feat(boards): add support for ST B-L475E-IOT01A

nponsard:
# Description

This PR adds support to the ST B-L475E-IOT01A board and its STM32L475VG MCU 

## Open Questions

~~Do we support USB for this board ? The clock would need to be generated by the MSI (calibrated using the LSE)~~
USB will be supported on this board, I was confused on how to change `MSIPLLEN` but embassy does it automatically. 

## Testing

Successfully tested on hardware through the IoT-LAB platform: 

- `examples/gpio`
- `examples/log`
- `examples/random`
- `examples/storage`
- `examples/usb-serial`
- `tests/gpio-interrupt-stm32`

I also ran some other tests that didn't pass because it needs some external hardware connected, the errors I got were because the hardware was not connected: 

- `tests/i2c-controller`
- `tests/spi-loopback`
- `tests/spi-main`

Finally I ran `examples/http-server` and got the network interface to show up on my computer, but the link stayed down with the `NO-CARRIER` status.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -473,6 +473,19 @@ contexts:
         - --cfg capability=\"hw/stm32-hash-rng\"
         - --cfg capability=\"hw/stm32-usb-synopsis\"
 
+  - name: stm32l475vg
+    parent: stm32
+    selects:
+      - cortex-m4f
+    provides:
+      - has_hwrng
+      - has_storage_support
+    env:
+      PROBE_RS_CHIP: STM32L475VG
+      RUSTFLAGS:
+        - --cfg capability=\"hw/stm32-rng\"
+        - --cfg capability=\"hw/stm32-usb-synopsis\" # Needs an external 32.768 kHz crystal to have a stable clock
+
   - name: stm32u083mc
     parent: stm32
     selects:
@@ -1548,6 +1561,16 @@ builders:
   # TODO: there also is a companion nrf5340 on this board
   - name: nordic-thingy-91-x-nrf9151
     parent: nrf9151
+
+  - name: st-b-l475e-iot01a
+    parent: stm32l475vg
+    provides:
+      - has_swi
+      - has_usb_device_port
+    env:
+      CARGO_ENV:
+        # This ISR is unused on a naked board. Configured here for testing.
+        - CONFIG_SWI=UART5
 
   - name: st-nucleo-f401re
     parent: stm32f401re
--- a/src/ariel-os-stm32/src/lib.rs
+++ b/src/ariel-os-stm32/src/lib.rs
@@ -80,7 +80,47 @@ pub fn init() -> OptionalPeripherals {
 }
 
 // TODO: find better place for this
+#[expect(clippy::too_many_lines)]
 fn board_config(config: &mut Config) {
+    #[cfg(context = "st-b-l475e-iot01a")]
+    {
+        use embassy_stm32::rcc::*;
+
+        // This board has an LSE clock, we can use it to calibrate the MSI clock
+        config.rcc.ls = LsConfig {
+            rtc: RtcClockSource::LSE,
+            lsi: false,
+            lse: Some(LseConfig {
+                frequency: embassy_stm32::time::Hertz(32768),
+                mode: LseMode::Oscillator(LseDrive::MediumHigh),
+            }),
+        };
+        config.rcc.hsi = false;
+        // Setting the MSI range to 48 MHz crashes the system. If the source of the issue is found,
+        // we can use MSI as the clock source for the usb peripheral directly and avoid using more PLLs.
+        config.rcc.msi = Some(MSIRange::RANGE8M);
+        config.rcc.pll = Some(Pll {
+            source: PllSource::MSI,
+            prediv: PllPreDiv::DIV1, // 8 Mhz
+            mul: PllMul::MUL20,      // 160 MHz
+            divp: None,
+            divq: None,
+            divr: Some(PllRDiv::DIV2), // sysclk 80Mhz (8 / 1  * 20 / 2)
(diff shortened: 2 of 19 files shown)


kaspar030:
Please add a (temporary) commit that adds the board to the limited build list in the github workflow

nponsard:
> Please add a (temporary) commit that adds the board to the limited build list in the github workflow

I cannot find the limited build list you're referring to, I added a commit to temporarly run clippy on the board I'm adding.

ROMemories:
> I cannot find the limited build list you're referring to, I added a commit to temporarly run clippy on the board I'm adding.

That list:

https://github.com/ariel-os/ariel-os/blob/c032ead82c80a26a69b3ae4c6499e8d071e24a6d/.github/workflows/build.yml#L131-L134

nponsard:
I added the board to the list of builders (last commit) and was able to do some tests on hardware using the IoT-LAB platform (getting logs via UART using #1029).

src/ariel-os-stm32/src/lib.rs
```diff
@@ -80,7 +80,48 @@ pub fn init() -> OptionalPeripherals {
...
+            rtc: RtcClockSource::LSE,
+            lsi: false,
+            lse: Some(LseConfig {
+                frequency: embassy_stm32::time::Hertz(32768),
+                mode: LseMode::Oscillator(LseDrive::Low),
+            }),
+        };
+        config.rcc.hsi = false;
+
+        config.rcc.msi = Some(MSIRange::RANGE8M);
```
ROMemories, line 100: The reference manual says the following:

> At 48 MHz, the MSI in PLL-mode can
be used for the USB OTG FS device, saving the need of an external high-speed crystal.

It seems to me like if would be preferable to only use one PLL if possible (and only for SYSCLK) and use the MSI output directly for USB.

```suggestion
        config.rcc.msi = Some(MSIRange::RANGE48M);
```

src/ariel-os-stm32/src/lib.rs
```diff
@@ -80,7 +80,48 @@ pub fn init() -> OptionalPeripherals {
...
+            source: PllSource::MSI,
+            prediv: PllPreDiv::DIV1,
+            mul: PllMul::MUL12, // 8 MHz MSI * 12 = 96 MHz
+            divp: None,
+            divq: Some(PllQDiv::DIV2), // USB 48 MHz (8 / 1 * 12 / 2)
+            divr: None,
+        });
+
+        // With a 32.768 kHz LSE, the MSI clock will be calibrated and considered accurate enough
+        // Embassy automatically enables MSIPLLEN if the LSE is configured.
```
ROMemories, line 121: :+1:
Seems to come [from here](https://github.com/embassy-rs/embassy/blob/3ad9b73dd52d21cab8357f551fb82e281ff2aaa6/embassy-stm32/src/rcc/l.rs#L194-L198) indeed; not sure it's in the documentation however?

laze-project.yml
```diff
@@ -467,6 +467,19 @@ contexts:
...
+    selects:
+      - cortex-m4f
+    provides:
+      - has_hwrng
+      - has_storage_support
+    env:
+      PROBE_RS_CHIP: STM32L475VG
+      RUSTFLAGS:
+        - --cfg capability=\"hw/stm32-rng\"
+        - --cfg capability=\"hw/stm32-usb-synopsis\" # USB OTG, needs an LSE (32.768 kHz crystal) to have a stable clock
```
ROMemories, line 481:
```suggestion
        - --cfg capability=\"hw/stm32-usb-synopsis\" # Needs an external 32.768 kHz crystal to have a stable clock
```

laze-project.yml
```diff
@@ -1518,6 +1531,18 @@ builders:
...
+    parent: stm32l475vg
+    provides:
+      - has_swi
+      # USB port only used for OTG FS, need MSI clock to be calibrated
+      - has_usb_device_port
+
+    env:
+      CARGO_ENV:
+        # This ISR is unused on a naked board. Configured here for testing.
+        - CONFIG_SWI=UART5
```
ROMemories, line 1544: :+1:

ROMemories (changes requested):
Great work, almost there; the :+1: I left are just things I've actually checked, they don't require any action from your side :slightly_smiling_face:

src/ariel-os-stm32/src/lib.rs
nponsard: I haven't seen any mention of it in the documentation either.

ROMemories: That's fine, good investigation then :slightly_smiling_face:

src/ariel-os-stm32/src/lib.rs
nponsard: Ah indeed I misunderstood the reference manual.

ROMemories (approved):
When this is tested on hardware again, you can fix the `rustfmt` error, autosquash, and remove the commit adding this to the list of devices for the limited build (that was just for testing for this PR, but we don't want to pay the CI price for all the other ones).

nponsard:
I've hit a roadblock where setting the MSI frequency to 48MHz would crash the processor.
With @ROMemories we decided to settle on a configuration with a lower MSI frequency, that means we need to use PLL1 and PLLSAI1 to get 80MHz SYSCLK and 48MHz USB clock (meaning an higher power consumption).

I also had the occasion to have a physical access to a board so I was able to test the usb-serial and gpio capabilities. I also tried an example with usb networking but couldn't get it to work (`NO-CARRIER` status).

ROMemories (approved):

PR #1037, opened 2025-05-19
feat: make heap requirements configurable

kaspar030:
# Description

This PR gives apps/modules the option to specify how much heap they'll need, similar to how we configure stack sizes.

On Cortex-M the leftover memory is used for heap in any case, and this is just a compile time check that requirements are met. On esp we can't take all leftover memory (yet), so this helps to specify that e.g., esp-wifi needs at least 72k of heap.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -94,11 +94,13 @@ contexts:
       PROBE_RS_PROTOCOL:
         start: --protocol=
 
-      # Executor / ISR stacksize helpers
+      # Executor / ISR stacksize and heapsize helpers
       executor_stacksize_required:
         joiner: ", "
       isr_stacksize_required:
         joiner: ", "
+      heapsize_required:
+        joiner: " + "
 
     rules:
       - name: LINK
@@ -667,9 +669,18 @@ modules:
 
   ## System functionality modules
   - name: alloc
+    help: Enables a heap, a global allocator and the `alloc` crate.
     context: ariel-os
     env:
       global:
+        # *Append* to this array to increase the heapsize. The sum of all entries will
+        # be used. By default, all unused memory is used for the heap, and the sum
+        # is used to check if the amount of heap is actually available.
+        # On esp, it always allocates the sum of `heapsize_required`, but at least 2KiB.
+        heapsize_required:
+          - "0"
+        CARGO_ENV:
+          - CONFIG_HEAPSIZE=$(${heapsize_required})
         FEATURES:
           - ariel-os/alloc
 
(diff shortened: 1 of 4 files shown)


laze-project.yml
```diff
@@ -667,9 +669,18 @@ modules:
...
     env:
       global:
+        # *add* to this array to increase the heapsize. Sum of all entries will
+        # be used. By default, all unused memory is used for the heap, and the sum
+        # is used to check if the amount of heap is actually available.
+        # Xtensa always allocates the sum of `heapsize`, but at least 2KiB.
+        heapsize:
+          - "0"
+        CARGO_ENV:
+          - CONFIG_HEAPSIZE=$(${heapsize})
```
ROMemories, line 683: Do we need these?

```suggestion
          - CONFIG_HEAPSIZE=${heapsize}
```

laze-project.yml
```diff
@@ -94,11 +94,13 @@ contexts:
       PROBE_RS_PROTOCOL:
         start: --protocol=
 
-      # Executor / ISR stacksize helpers
+      # Executor / ISR stacksize and heapsize helpers
       executor_stacksize_required:
         joiner: ", "
       isr_stacksize_required:
         joiner: ", "
+      heapsize:
```
kaspar030, line 102: hm, seeing this next to each other, maybe `heap_required` or `heapsize_required` would be more consistent?

src/ariel-os-alloc/src/lib.rs
```diff
@@ -59,6 +62,10 @@ mod alloc {
             size, start
         );
 
+        // No `const { assert!(..) }` here unfortunately due to the use of linker
+        // values.
+        assert!(size >= CONFIG_HEAPSIZE);
```
ROMemories, line 67: Shouldn't this assert be *above* the debug statement?

laze-project.yml
ROMemories: I suppose `heapsize_required` would be better for consistency.

laze-project.yml
kaspar030: Yes, `heapsize_required` has ` + ` as "joiner". The `$(...)` is an expression, so multiple entries get summed up. e.g.,

```
heapsize_required:
- 1
- 2
```

... will become `CONFIG_HEAPSIZE=$(1 + 2)`, and the expression evaluates to `3`.

ROMemories: Ok that makes sense

ROMemories (approved):
Successfully tested the `http-client` example on espressif-esp32-c6-devkitc-1 on Wi-Fi.

PR #1040, opened 2025-05-19
feat(stm32): Add support for STM32 GPIO v1 speed flags

therealprof:


src/ariel-os-stm32/src/gpio.rs
```diff
@@ -94,7 +94,10 @@ impl From<Speed> for embassy_stm32::gpio::Speed {
         match speed {
             Speed::Low => Self::Low,
             Speed::Medium => Self::Medium,
+            #[cfg(gpio_v2)]
```
kaspar030, line 97: but, where does this cfg come from?

therealprof: Magically injected by `embassy-stm32`s `build.rs` is my best explanation.

therealprof: https://github.com/embassy-rs/embassy/blob/ef32187ed7349f3883d997b6f1590e11dbc8db81/embassy-stm32/src/gpio.rs#L276

kaspar030: I think the `gpio_v2` cfg is set in `embassy-stm32`'s `build.rs` [here](https://github.com/embassy-rs/embassy/blob/ef32187ed7349f3883d997b6f1590e11dbc8db81/embassy-stm32/build.rs#L38-L43), so it shouldn't be available in ariel's stm32. It is set via emitting `cargo::rustc-cfg=foo`, which isn't passed to dependees.

I guess we could replicate generating those cfgs in `ariel-os-stm32` directly from `stm32_metapac::metadata::METADATA`. @ROMemories wdyt?

kaspar030: well, without the cfgs set at all, the default (`not(gpio_v2)`) is used, which now maps `Speed::High` to `Self::VeryHigh` instead of `Self::High`. I'd guess this compiles but selects the wrong speed in the previous case.

therealprof: Whoops, you're right... this doesn't actually work. 😞 Sorry for the brainfart...

PR #1045, opened 2025-05-20
feat(stm32): provide stm32 peripheral `cfg` metadata

kaspar030:
# Description

This uses `stm32-metadata` to provide per-MCU `cfg` based on actual peripherals of each mcu.

## Issues/PRs references

Needed to make sth like #1040 work.

## Change checklist

### DIFF
--- a/src/ariel-os-stm32/Cargo.toml
+++ b/src/ariel-os-stm32/Cargo.toml
@@ -30,6 +30,9 @@ ariel-os-embassy-common = { workspace = true }
 ariel-os-random = { workspace = true, optional = true }
 static_cell = { workspace = true }
 
+[build-dependencies]
+stm32-metapac = "16.0.0"
+
 [features]
 ## Enables GPIO interrupt support.
 external-interrupts = [
--- a/src/ariel-os-stm32/build.rs
+++ b/src/ariel-os-stm32/build.rs
@@ -24,4 +24,24 @@ fn main() {
 
         println!("cargo::rerun-if-env-changed=CONFIG_SWI");
     }
+
+    peripheral_cfg_from_metapac();
+}
+
+// Enable peripheral `cfg` flags, data taken from `stm32-metapac`.
+// Similar to https://github.com/embassy-rs/embassy/blob/ef32187ed7349f3883d997b6f1590e11dbc8db81/embassy-stm32/build.rs#L38-L43
+fn peripheral_cfg_from_metapac() {
+    use std::collections::HashSet;
+    fn cfg_only_once(seen: &mut HashSet<String>, cfg: &str) {
+        if seen.insert(cfg.to_string()) {
+            println!("cargo::rustc-cfg={cfg}");
+        }
+    }
+    let mut seen = HashSet::new();
+    for p in stm32_metapac::metadata::METADATA.peripherals {
+        if let Some(r) = &p.registers {
+            cfg_only_once(&mut seen, r.kind);
+            cfg_only_once(&mut seen, &format!("{}_{}", r.kind, r.version));
+        }
+    }
 }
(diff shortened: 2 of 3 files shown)


kaspar030:
This produces e.g., for the `st-nucleo-f411re`:

```
--cfg adc --cfg adc_v2 --cfg adccommon --cfg adccommon_v2 --cfg crc --cfg crc_v1 --cfg dbgmcu --cfg dbgmcu_f4 --cfg dma --cfg dma_v2 --cfg exti
 --cfg exti_v1 --cfg flash --cfg flash_f4 --cfg gpio --cfg gpio_v2 --cfg i2c --cfg i2c_v1 --cfg iwdg --cfg iwdg_v1 --cfg pwr --cfg pwr_f4 --cfg rcc --cfg rcc_f
4 --cfg rtc --cfg rtc_v2f4 --cfg sdmmc --cfg sdmmc_v1 --cfg spi --cfg spi_v1 --cfg syscfg --cfg syscfg_f4 --cfg timer --cfg timer_v1 --cfg uid --cfg uid_v1 --c
fg usart --cfg usart_v2 --cfg otg --cfg otg_v1 --cfg wwdg --cfg wwdg_v1
```

(formatting is off, I got this by adding `-v` to the `LINK` rule cargo call, and c&p from the looong output)

src/ariel-os-stm32/Cargo.toml
```diff
@@ -30,6 +30,9 @@ ariel-os-embassy-common = { workspace = true }
 ariel-os-random = { workspace = true, optional = true }
 static_cell = { workspace = true }
 
+[build-dependencies]
+stm32-metapac = "*"
```
kaspar030, line 34: We have another dependency on `embassy-stm32` that choses a version. `*` here means `don't care`, so Cargo will chose the one `embassy-stm32` is fine with. Not sure if being more specific here makes any sense.

ROMemories: > `*` here means `don't care`, so Cargo will chose the one `embassy-stm32` is fine with

I'm not sure of that, it at least depends on the resolver; also [crates.io doesn't accept these](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#wildcard-requirements).

```suggestion
stm32-metapac = "16.0.0"
```

src/ariel-os-stm32/build.rs
```diff
@@ -24,4 +24,24 @@ fn main() {
 
         println!("cargo::rerun-if-env-changed=CONFIG_SWI");
     }
+
+    // generate stm32 peripheral cfg from metapac
```
ROMemories, line 28: This doesn't really generate anything, it *enables* the relevant `cfg` flags. Could you move this to a separate function in this build script?

src/ariel-os-stm32/build.rs
```diff
@@ -24,4 +24,24 @@ fn main() {
...
+
+        fn cfg_only_once(seen: &mut HashSet<String>, cfg: &str) {
+            if seen.insert(cfg.to_string()) {
+                println!("cargo::rustc-cfg={cfg}");
+            }
+        }
+
+        let mut seen = HashSet::new();
+
+        for p in stm32_metapac::metadata::METADATA.peripherals {
```
ROMemories, line 40: Maybe this could reference the following snippets so that we know where this comes from: https://github.com/embassy-rs/embassy/blob/ef32187ed7349f3883d997b6f1590e11dbc8db81/embassy-stm32/build.rs#L38-L43

ROMemories (approved):
The dependency version needs fixing, and the rest is optional (but I'd still like to see it); the output you've shown seems right, but I'm not sure how to reproduce it.

ROMemories (approved):

ROMemories:
Passing `-DCARGO_ARGS+=-v` to the laze command allows to see the generated list of `--cfg` flags.

PR #1046, opened 2025-05-20
fix(build): model `xtensa` being `nightly`

kaspar030:
# Description

b/c dependency resolution ordering reasons, it wasn't possible to build `examples/http-client` for xtensa boards.

This PR

- makes all three toolchain-providing modules (esp, stable, nightly) provide_unique `cargo-toolchain` (implicitly conflicting with each other)
- make esp set `--cfg nightly`
- make the ariel-os context try `?xtensa` before `?stable`.
- actually limits `xtensa` to be available only in the `esp32` and `esp32s3` contexts

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -25,6 +25,9 @@ contexts:
       - ?hw/device-identity
       - ?lto
       - ?semihosting
+
+      # Selecting default toolchain here. try xtensa, stable, nightly.
+      - ?xtensa
       - ?stable
       - ?nightly
     env:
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -626,14 +629,23 @@ modules:
           - --cfg context=\"cortex-m33f\"
 
   - name: xtensa
+    context:
+      - esp32
+      - esp32s3
```
ROMemories, line 634: I would have hoped we wouldn't need to do that anymore (and only select `xtensa` from `esp32` and `esp32s3`), but I guess that would require defining a couple of extra modules, so fine like that.

ROMemories (approved):

laze-project.yml
kaspar030: There'll be more options in the future - I hope to get to improving laze dependency resolution. Also, xtensa won't need a special toolchain anymore, in the mid- to long-term. Let's see what arrives first. :innocent:

PR #1048, opened 2025-05-21
docs(udp-example): use smaller buffers

ROMemories:
# Description

Shrink the buffers used in the UDP echo example, and add the reasoning for their sizing, as it is not currently documented upstream.
This is a first step in being able to compile the networking examples for MCUs with little RAM.

## Change checklist

### DIFF
--- a/examples/udp-echo/src/main.rs
+++ b/examples/udp-echo/src/main.rs
@@ -4,15 +4,19 @@
 use ariel_os::{debug::log::*, net, reexports::embassy_net};
 use embassy_net::udp::{PacketMetadata, UdpSocket};
 
+// UDP datagrams with payloads larger than this will be dropped and ignored, both when receiving
+// and sending, so the size of the three buffers needs to be the same in this echo example.
+const BUFFER_SIZE: usize = 128;
+
 #[ariel_os::task(autostart)]
 async fn udp_echo() {
     let stack = net::network_stack().await.unwrap();
 
-    let mut rx_meta = [PacketMetadata::EMPTY; 16];
-    let mut rx_buffer = [0; 4096];
-    let mut tx_meta = [PacketMetadata::EMPTY; 16];
-    let mut tx_buffer = [0; 4096];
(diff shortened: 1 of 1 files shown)


examples/udp-echo/src/main.rs
```diff
@@ -4,15 +4,19 @@
 use ariel_os::{debug::log::*, net, reexports::embassy_net};
 use embassy_net::udp::{PacketMetadata, UdpSocket};
 
+// UDP datagrams with payloads larger than this will be dropped and ignored, both when receiving
+// and sending, so the size of the three buffers needs to be the same in this echo example.
+const BUFFER_SIZE: usize = 128;
```
ROMemories, line 9: This limits the size of the UDP datagrams to 128 bytes, which seems to be enough for this example (and allows for consistency with the future shrinking of the TCP example buffers).

examples/udp-echo/src/main.rs
```diff
@@ -4,15 +4,19 @@
...
 #[ariel_os::task(autostart)]
 async fn udp_echo() {
     let stack = net::network_stack().await.unwrap();
 
-    let mut rx_meta = [PacketMetadata::EMPTY; 16];
-    let mut rx_buffer = [0; 4096];
-    let mut tx_meta = [PacketMetadata::EMPTY; 16];
-    let mut tx_buffer = [0; 4096];
-    let mut buf = [0; 4096];
+    let mut rx_meta = [PacketMetadata::EMPTY; 1];
```
ROMemories, line 15: My understanding is that this limits the number of packets in the ring buffer, but I couldn't trigger any issue when manually testing.

examples/udp-echo/src/main.rs
kaspar030: Using netcat I can confirm that larger "lines" are just ignored. Fine for this example.

kaspar030 (approved):
Tried to break it on the st-nucleo-f767zi, large packets are ignored as advertized. so, LGTM.

PR #1050, opened 2025-05-21
feat: add support stm32f042k6 mcu and nucleo board

therealprof:
# Description

This adds support for Cortex-M0 MCUs, the STM32F042K6 MCU in particular and the associated Nucleo-32 board

### DIFF
--- a/examples/gpio/src/pins.rs
+++ b/examples/gpio/src/pins.rs
@@ -61,6 +61,13 @@ ariel_os::hal::define_peripherals!(Peripherals {
     btn1: PC13
 });
 
+#[cfg(context = "st-nucleo-f042k6")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    led1: PB3,
+    // Note: the board only has one physical button which is hardwired to NRST so this uses D2
+    btn1: PA12
+});
+
 #[cfg(any(context = "st-nucleo-f401re", context = "st-nucleo-f411re"))]
 ariel_os::hal::define_peripherals!(Peripherals {
     led1: PA5,
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -209,8 +209,8 @@ chips:
       debug_output: supported
       ethernet_over_usb: not_available
       hwrng: not_available
-      i2c_controller: supported
-      spi_main: supported
+      i2c_controller: not_currently_supported
+      spi_main: not_currently_supported
       logging: supported
       storage:
         status: not_currently_supported
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -202,6 +202,23 @@ chips:
       user_usb: not_available
       wifi: not_available
 
+  stm32f042k6:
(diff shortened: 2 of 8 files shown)


examples/gpio/src/pins.rs
```diff
@@ -61,6 +61,14 @@ ariel_os::hal::define_peripherals!(Peripherals {
...
 });
 
+#[cfg(context = "st-nucleo-f042k6")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    led1: PB3,
+    // Note: the board only has one physical button which is hardwired to NRST so this uses D2
+    btn1: PA12
+});
+
+
```
kaspar030, line 71:
```suggestion
```

kaspar030:
hm, adding support for this 6kb RAM board kinda dips our noses into the fact that we assumed more in most examples, and that we don't have a way to filter builds by actually used RAM/flash.

E.g., setting the "isr_stacksize_required_default" to the low values will probably make more examples compile, but they might then fail at runtime. :(

Let me think about how to model this.

kaspar030:
I got something, do you mind if I push to this branch?

therealprof:
@kaspar030 Please go ahead. I also explicitly allowed maintainers to make changes.

kaspar030:
> @kaspar030 Please go ahead. I also explicitly allowed maintainers to make changes.

I somehow can't rebase the branch. would you? :)

therealprof:
> > @kaspar030 Please go ahead. I also explicitly allowed maintainers to make changes.
> 
> I somehow can't rebase the branch. would you? :)

Done.

kaspar030:
I've run a full `laze build -b st-nucleo-f042k6 -g` locally successfully.

kaspar030:
The chip needs an entry in `support_matrix.yml`:

    Error:   × invalid chip name `stm32f042k6` for board `ST NUCLEO-F042K6`

therealprof:
FWIW, I do have probably 60 different STM32 chips on even more boards here myself and I don't think I want to load that work on me to support them all and you to review all of them (and the surprisingly slow CI to greenlight them)...

kaspar030:
> I added it but I think this will need a better solution to support whole families... I think there're thousands of models in the STM32 range and that really doesn't scale.

Agreed. The current support matrix was our initial step for not having to edit the book table by hand. We'll automate as much as possible at some point. I have already some scripting that extracts the support status from our build system (which e.g., knows that the stm32f042k6 doesn't have i2c support wired up, yet).

therealprof:
I can add I2C and SPI support soon... but I noticed that embassy-stm32 doesn't implement `i2c_v2` properly yet... 🙈

https://github.com/embassy-rs/embassy/blob/edceb0dc7da32848fd31b525b4c247671a14eb2d/embassy-stm32/src/i2c/v2.rs#L732

therealprof:
> LGTM. trusting your testing on gpio.

To quote my kids: Why is that weird thing on the window sill blinking like crazy? 😬

PR #1052, opened 2025-05-22
feat(debug)!: drop support for print

kaspar030:
# Description

We were not using that (internally) at all. It required us to keep the extra rtt channel in the defmt case. Without that we instantly save 1Kib RAM, which is absolutely necessary for the smaller MCUs.

This is a breaking change obviously.

## Change checklist

### DIFF
--- a/src/ariel-os-debug-log/src/lib.rs
+++ b/src/ariel-os-debug-log/src/lib.rs
@@ -117,6 +117,19 @@ mod log_macros {
             }
         }};
     }
+
+    /// Prints to the debug output, with a newline.
+    #[macro_export]
+    macro_rules! println {
+        ($($arg:tt)*) => {{
+            use $crate::defmt::hidden::defmt;
+            if true {
+                defmt::println!($($arg)*);
+            } else {
+                drop(format_args!($($arg)*));
+            }
+        }};
+    }
 }
 
 #[cfg(feature = "log")]
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -64,7 +64,11 @@ pub fn exit(code: ExitCode) {
 
 #[cfg(all(feature = "debug-console", feature = "rtt-target"))]
 mod backend {
-    pub use rtt_target::{rprint as print, rprintln as println};
+    #[cfg(not(feature = "defmt"))]
+    pub use rtt_target::rprintln as println;
+
+    #[cfg(not(feature = "defmt"))]
+    pub use defmt::println;
 
     #[doc(hidden)]
     pub fn init() {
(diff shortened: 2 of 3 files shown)


src/ariel-os-debug/src/lib.rs
```diff
@@ -80,32 +84,27 @@ mod backend {
 
         #[cfg(feature = "defmt")]
         {
-            use rtt_target::ChannelMode::{NoBlockSkip, NoBlockTrim};
+            use rtt_target::ChannelMode::NoBlockSkip;
+            const DEFMT_BUFFER_SIZE: usize = 1024;
```
kaspar030, line 88: ups, pulling out `DEFMT_BUFFER_SIZE` is an unrelated change. Unless it bothers you my dear reviewer, lets keep it. I'll soon open a PR that makes this configurable.

src/ariel-os-debug-log/src/lib.rs
```diff
@@ -117,6 +117,19 @@ mod log_macros {
             }
         }};
     }
+
+    /// Prints a line of test.
```
chrysn, line 121:
```suggestion
    /// Prints a line of text.
```

chrysn (approved):
LGTM.

We could make this non-breaking if we instead deprecated print() and changed its behavior to be a println -- but we'll go through a bump anyway, and I'd reckon the compiler will even have a suitable suggestion anyway based on the textual similarity from print to println.

src/ariel-os-debug-log/src/lib.rs
ROMemories, line 121: That's the doc comment we have in `ariel-os-debug`:
```suggestion
    /// Prints to the debug output, with a newline.
```

ROMemories (commented):
So this drops the RTT channel that was dedicated to `print!()`ing, and because defmt only seems to provide `defmt::println!()` and no `defmt::print!()`, we need to drop `print!()` from our API; is that correct? Is that an inherent limitation of `defmt` or would it be possible to add support for `print!()` upstream later?

kaspar030:
> So this drops the RTT channel that was dedicated to `print!()`ing, and because defmt only seems to provide `defmt::println!()` and no `defmt::print!()`, we need to drop `print!()` from our API; is that correct? Is that an inherent limitation of `defmt` or would it be possible to add support for `print!()` upstream later?

That's correct. `defmt` is inherently line based IIUC, so it is unlikely that we can add it later.

kaspar030:
> We could make this non-breaking if we instead deprecated print() and changed its behavior to be a println -- but we'll go through a bump anyway, and I'd reckon the compiler will even have a suitable suggestion anyway based on the textual similarity from print to println.

hm. Let's just break it. If we turn `print!()` into `println!()`, all valid uses of it produce broken output.

ROMemories (approved):

PR #1053, opened 2025-05-22
feat(build): introduce `ram-tiny` laze module

kaspar030:
# Description

This PR adds a module bundling tiny RAM requirements.
I started this for #1050, in combination with this ram-tiny, the st-nucleo-f042k6 builds fine.

As we currently don't have a way to filter builds by actually available / used RAM, this is just an arbitrary set of `small ram settings`.

E.g., when not using log but defmt, the stack requirements go way down.
Or, currently there's no network stack / driver combo that works with little ram devices.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -745,6 +745,20 @@ modules:
           - -Clink-arg=-Tkeep-stack-sizes.x
           - -Z emit-stack-sizes
 
+  - name: ram-tiny
+    help: This module marks boards with very little RAM (currently, <10KiB)
+    conflicts:
+      # disabling alloc until we have a usable use case
+      - alloc
+      # disabling as the default stack sizes are too small for log (defmt works)
+      - log
+      # disabling until we do have a driver/stack combo that works
+      - network
+    env:
+      global:
+        isr_stacksize_required_default: "512"
+        executor_stacksize_required_default: "512"
+
   - name: stable
     help: build with stable Rust
     provides_unique:
--- a/examples/threading-channel/laze.yml
+++ b/examples/threading-channel/laze.yml
@@ -2,3 +2,7 @@ apps:
   - name: threading-channel
     selects:
       - sw/threading
+    conflicts:
+      # This example uses five threads with the default stack size of 2KiB,
+      # which is too much for the smallest MCUs.
+      - ram-tiny
--- a/examples/threading-event/laze.yml
+++ b/examples/threading-event/laze.yml
@@ -4,3 +4,7 @@ apps:
       - sw/threading
       - "context::stm32c031c6":
           - too-little-memory
+    conflicts:
+      # This example uses five threads with the default stack size of 2KiB,
+      # which is too much for the smallest MCUs.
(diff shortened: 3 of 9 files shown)


laze-project.yml
```diff
@@ -745,6 +745,20 @@ modules:
           - -Clink-arg=-Tkeep-stack-sizes.x
           - -Z emit-stack-sizes
 
+  - name: ram-tiny
+    help: This module marks boards with very little RAM (currently, <10KiB)
```
kaspar030, line 749: The 10KiB are an arbitrary choice. Gotta start somewhere.

laze-project.yml
```diff
@@ -745,6 +745,20 @@ modules:
...
+    conflicts:
```
kaspar030, line 750: These are general restrictions for those small devices, at this time ...
I hope to unlock some of them in the future. E.g., a really small network stack is not unthinkable.

laze-project.yml
```diff
@@ -745,6 +745,20 @@ modules:
...
+      # disabling alloc until we have a usable use case
+      - alloc
+      # disabling until we do have a driver/stack combo that works
+      - network
+      # disabling as the default stack sizes are too small for log (defmt works)
+      - log
+    env:
+      global:
+        isr_stacksize_required_default: "512"
```
kaspar030, line 759: This is a slightly dangerous thing to do. Setting these too low will still compile but then fail at runtime. The non-tiny defaults are 2Kib (ISR) and 8KiB (Executor) ...

therealprof:
The 512 was an arbitrary choice. I don't think it allowed me to compile any examples I couldn't compile with a larger stack. Also: I'm curious on whether rustc/linker couldn't/wouldn't warn if the sizes were too small to actually fit the stack contents... the stack requirements are always known at compile time, no?

laze-project.yml
```diff
@@ -745,6 +745,20 @@ modules:
...
 
+  - name: ram-tiny
+    help: This module marks boards with very little RAM (currently, <10KiB)
+    conflicts:
+      # disabling alloc until we have a usable use case
+      - alloc
+      # disabling until we do have a driver/stack combo that works
+      - network
+      # disabling as the default stack sizes are too small for log (defmt works)
+      - log
```
ROMemories, line 756: optional: I'd rather we stuck to lexicographical order, unless laze requires a different order because of dependencies:

```suggestion
      # disabling as the default stack sizes are too small for log (defmt works)
      - log
      # disabling until we do have a driver/stack combo that works
      - network
```

ROMemories (approved):
We don't currently have a supported MCU to test this on, do we? The first one should be #1053, right?

kaspar030:
> There's [cargo-call-stack](https://github.com/japaric/cargo-call-stack?rgh-link-date=2025-05-22T13%3A57%3A46Z) that makes use of that to figure out stack use of a call stack.

I think @Dirbaio's fork is the one that got the most tuning for our use case: https://github.com/Dirbaio/cargo-call-stack

PR #1054, opened 2025-05-22
ci(cargo): pass `--locked` to Cargo through `laze build` 

### LINKED ISSUE #890: ci: enforce that the Cargo lockfile is always up to date
Our Clippy jobs do use `--locked` to reject an outdated `Cargo.lock`, however this does not check our examples and tests, for which the lockfile is still sometimes out of date.

ROMemories:
# Description

This makes sure that the Cargo lockfile is up to date, even for in-tree examples and tests.

## Issues/PRs references

Fixes #890.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,7 +138,7 @@ jobs:
         run: |
           sccache --start-server || true # work around https://github.com/ninja-build/ninja/issues/2052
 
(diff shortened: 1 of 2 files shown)


ROMemories:
Test build in CI (bumped an arbitrary crate and "forgot" to commit the resulting lockfile) failed as expected with the following error:

```sh
error: the lock file /home/runner/work/ariel-os/ariel-os/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

which is exactly what we want.

PR #1068, opened 2025-05-27
feat(nrf): Add support for BBC micro:bit v1

therealprof:
# Description

This adds support for the venerable BBC micro:bit v1.

## Issues/PRs references

None

## Open Questions

None

## Change checklist

- [X] I have cleaned up my commit history and squashed fixup commits.
- [X] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [X] I have performed a self-review of my own code.
- [X] I have made corresponding changes to the documentation.

### DIFF
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -53,8 +53,8 @@ functionalities:
 
 # Encodes support status for each chip.
 chips:
-  nrf51822:
-    name: nRF51822
+  nrf51822-xxaa:
+    name: nRF51822_xxAA
     support:
       gpio: supported
       debug_output: supported
@@ -358,7 +358,7 @@ boards:
   bbc-microbit-v1:
     name: BBC micro:bit V1
     url: https://web.archive.org/web/20250109121140/https://microbit.org/get-started/features/overview/#original-micro:bit
-    chip: nrf51822
+    chip: nrf51822-xxaa
     support:
       user_usb: not_available
       ethernet_over_usb: not_available
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -64,6 +64,8 @@ chips:
       logging: supported
       storage: not_currently_supported
       wifi: not_available
+      user_usb: not_available
+      ethernet_over_usb: not_available
 
   nrf52832:
     name: nRF52832
@@ -361,8 +363,6 @@ boards:
     chip: nrf51822-xxaa
     tier: "3"
     support:
-      user_usb: not_available
-      ethernet_over_usb: not_available
 
   bbc-microbit-v2:
     name: BBC micro:bit V2
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -58,7 +58,7 @@ chips:
     support:
       gpio: supported
       debug_output: supported
-      hwrng: not_currently_supported
+      hwrng: supported
       i2c_controller: not_currently_supported
       spi_main: not_currently_supported
       logging: supported
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -53,6 +53,18 @@ functionalities:
 
 # Encodes support status for each chip.
 chips:
+  nrf51822:
+    name: nRF51822
+    support:
+      gpio: supported
+      debug_output: supported
+      hwrng: not_currently_supported
+      i2c_controller: not_currently_supported
+      spi_main: not_currently_supported
+      logging: supported
+      storage: not_currently_supported
+      wifi: not_available
+
   nrf52832:
     name: nRF52832
     support:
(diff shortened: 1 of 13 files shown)


doc/support_matrix.yml
```diff
@@ -53,6 +53,18 @@ functionalities:
...
+  nrf51822:
+    name: nRF51822
+    support:
+      gpio: supported
+      debug_output: supported
+      hwrng: not_currently_supported
+      i2c_controller: not_currently_supported
+      spi_main: not_currently_supported
+      logging: supported
+      storage: supported
```
ROMemories, line 65:
```suggestion
      storage: not_currently_supported
```
To add support for storage, the `ariel-os-storage` build script needs to be modified and the `has_storage_support` laze module needs to be provided by the `nrf51` laze context.

doc/support_matrix.yml
```diff
@@ -312,6 +324,14 @@ chips:
 # Encodes support status for each board.
 # Boards inherit support statuses of their chip, but can also override them.
 boards:
+  bbc-microbit-v1:
+    name: BBC micro:bit V1
+    url: https://web.archive.org/web/20250109121140/https://microbit.org/get-started/features/overview/#original-micro:bit
+    chip: nrf51822
+    support:
+      user_usb: not_available
+      ethernet_over_usb: not_available
```
ROMemories, line 333: These could be moved to the chip info, as I believe they do not support USB at all.

src/ariel-os-nrf/build.rs
```diff
@@ -6,7 +6,9 @@ fn main() {
         return;
     }
 
-    let (ram, flash) = if is_in_current_contexts(&["nrf52832"]) {
+    let (ram, flash) = if is_in_current_contexts(&["nrf51822"]) {
+        (16, 256)
```
ROMemories, line 10: The nRF51822 comes in different variants with different sizes of RAM and flash, so the laze context name needs to be more specific (`nrf51822-xxaa` in this case)

doc/support_matrix.yml
```diff
@@ -53,6 +53,18 @@ functionalities:
 
 # Encodes support status for each chip.
 chips:
+  nrf51822:
+    name: nRF51822
+    support:
+      gpio: supported
+      debug_output: supported
+      hwrng: not_currently_supported
```
ROMemories, line 61: Can you test the `random` in hardware? The HWRNG should actually be supported.

laze-project.yml
```diff
@@ -205,6 +205,11 @@ contexts:
     provides:
       - has_hwrng
 
+  - name: nrf51822
```
ROMemories, line 208:
```suggestion
  - name: nrf51822-xxaa
```

ROMemories (changes requested):
Thanks for the PR! Could you additionally update the tests in the `tests` directory, by grepping for `"nrf52"` and adding what's required for `"nrf51"`?

laze-project.yml
therealprof: Uhm, I was looking into doing that and I couldn't help but notice that all the other nRF targets don't include their variants either, except when it comes to the PROBE_RS_CHIP variable...  It's a bit weird to just do it for the new kid on the block, isn't it?

src/ariel-os-nrf/build.rs
therealprof: I agree in principle but at least some of the other nRF targets also have multiple variants, like the 52832 which are currently not considered.

kaspar030: Only nrf51822 has multiple variants regarding ram/flash size, the others don't. (compare https://docs.nordicsemi.com/bundle/struct_nrf52/page/struct/nrf52.html).

(all variants come with differing GPIO pin count, we'll have to add the variants there as well at some point ...)

therealprof:
> Thanks for the PR! Could you additionally update the tests in the `tests` directory, by grepping for `"nrf52"` and adding what's required for `"nrf51"`?

How do I actually build the tests? Also I'd note that the current "pin" setup is mostly useless for generic nRF MCUs since any pin can be used for anything so the choice of pins is rather arbitrary.

src/ariel-os-nrf/build.rs
kaspar030: So in this case, please add the variant already, so I don't have to look up the variant again when I fix the nrf5832. :)

doc/support_matrix.yml
therealprof: Okay, I tested it, works just fine.

laze-project.yml
```diff
@@ -205,6 +205,13 @@ contexts:
     provides:
       - has_hwrng
 
+  - name: nrf51822-xxaa
+    parent: nrf51
+    selects:
+      - ram-tiny
```
ROMemories, line 211: Marked this MCU as `ram-tiny`, as many threading examples currently do not have enough RAM to run. We may revisit this later.

ROMemories:
@therealprof Thanks for this PR! I took the liberty of appending some minor changes to the branch (as you've said in another PR that you were fine with this), and marked the MCU as `ram-tiny`.

Issue #560, opened 2024-11-21
feat: integrate trouble BLE stack


kaspar030:
It should be a matter of just using the "ble" feature on `ariel_os`, and have our build system deal with the differences of the [upstream examples](https://github.com/embassy-rs/trouble/tree/main/examples).

chrysn:
From the list of good project goals, BT Home might make a good initial application.

ROMemories:
Similarly to Wi-Fi or USB, we want to add support for BLE, providing an abstraction over the different MCU families.
[trouBLE](https://github.com/embassy-rs/trouble) provides such an abstraction, with different controllers for each supported MCU family.

We will need to:

- [ ] Introduce a `ble` laze module that enables the to-be-introduced `ble` Cargo feature on the `ariel-os` crate.
- [ ] Expose a BLE API, which will be exposed from `ariel-os-embassy`. This may be as simple as re-exporting the relevant items trouBLE, or may want to introduce our own layer on top of it.
- [ ] Initialize BLE in `ariel-os-embassy` when enabled, similarly to USB for instance. 
- [ ] Initialize the controllers specific to each MCU family in the Ariel OS HAL crates (e.g., `ariel-os-nrf`, `ariel-os-rp`), and expose these to the abstraction from `ariel-os-embassy`.

We are first interested in adding support for nRF (using the Softdevice controller) and Pico W; we will take care of the other MCU families later.

For testing, we will first want to be able to send BLE advertising packets (and detect these with another device), and then be able to scan for these.

chrysn:
What this plan is missing AFAICT is that we'll have to account for the softdevice in additional ways:

* [ ] It has to be flashed (our flash layouts usually grab the whole chip with no concern for the softdevice sitting eg. at 0x1000 on an nRF52840)
* [ ] The softdevice is a tad picky when it comes to critical sections (whatever embassy does there *probably* is good already, but we do muck with interrupts ourselves, esp. around scheduling).

> or may want to introduce our own layer on top of it

I would not; trouBLE is already a good dispatch between backends, and I doubt we have the BLE expertise in the room to come up with any better abstraction.

> For testing

I have an application largely written on the softdevice that I'd be happy to port once the basics are in.

jamessizeland:
Trouble doesn't require flashing softdevice separately, so that should be okay. I've not used Ariel yet, so I'll try building a template first and see about adding in the BLE parts after that.

chrysn:
[...]

PR #1069, opened 2025-05-28
feat(ble): add public interface for BLE functionality

nponsard:
# Description

This PR adds the public interface to access the TrouBLE stack, an example is provided to show how to use it

## Issues/PRs references

This is a step towards #560. 

## Open Questions

Does the book page about BLE need to be in this PR or should it be a separate one ? 

At this time the public API lacks some features:

- [x] ~~There needs to be a way to switch between peripheral and central mode.~~
- [x] ~~A way to change the address.~~
- A way to customize the MTU, though 27 is probably fine.

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,6 +109,7 @@ jobs:
             --verbose
             --locked
             --features "
+                ble,
                 coap,
                 csprng,
                 dns,
@@ -252,6 +253,7 @@ jobs:
                 -p coapcore \
                 --features "
                     bench,
+                    ble,
                     coap,
                     core-affinity,
                     csprng,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,6 +147,8 @@ rtt-target = { version = "0.6.0" }
 rp-pac = { version = "7.0", default-features = false }
 serde = { version = "1.0.197", default-features = false }
 static_cell = { version = "2.0.0", features = ["nightly"] }
+trouble-host = "0.1"
+bt-hci = { version = "0.2.0" }
 
 [workspace.lints.rust]
 # rustc lints are documented here: https://doc.rust-lang.org/rustc/lints/listing/index.html
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1082,6 +1082,41 @@ modules:
         FEATURES:
           - ariel-os/usb-ethernet
 
+  - name: ble
+    selects:
+      - hw/ble
+    env:
+      global:
+        FEATURES:
+          - ariel-os/ble
+
+  - name: hw/ble
+    help: provided if a device has a BLE capability
+    selects:
+      - has_ble
+    env:
+      global:
+        RUSTFLAGS:
+          - --cfg capability=\"hw/ble\"
+
+  - name: ble-peripheral
+    help: Enables BLE peripheral functionality
+    selects:
+      - ble
+    env:
+      global:
+        FEATURES:
+          - ariel-os/ble-peripheral
+
+  - name: ble-central
+    help: Enables BLE central functionality
+    selects:
+      - ble
+    env:
+      global:
+        FEATURES:
+          - ariel-os/ble-central
+
   - name: usb
     selects:
       - hw/usb-device-port
--- a/src/ariel-os-embassy-common/Cargo.toml
+++ b/src/ariel-os-embassy-common/Cargo.toml
@@ -18,6 +18,7 @@ embassy-time = { workspace = true }
 embedded-hal = { workspace = true }
 embedded-hal-async = { workspace = true }
 const-sha1 = { version = "0.3.0", default-features = false }
+trouble-host = { workspace = true, optional = true }
 
 [features]
 ## Enables GPIO interrupt support.
@@ -34,3 +35,5 @@ defmt = ["dep:defmt", "fugit?/defmt"]
 executor-thread = []
 
 _test = ["i2c", "spi", "external-interrupts"]
+
+ble = ["dep:trouble-host"]
--- a/src/ariel-os-embassy-common/src/ble.rs
+++ b/src/ariel-os-embassy-common/src/ble.rs
@@ -1,0 +1,18 @@
+//! Common BLE types to be used across different HALs.
+
+/// Configuration for the BLE stack.
+///
+/// You can customize it using the `ble-config-override` feature.
+pub struct Config {
+    /// The address of the BLE device.
+    pub address: trouble_host::Address,
+}
+
+/// Default address, this one is used in the trouBLE examples, needs to be changed/randomized in production.
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            address: trouble_host::Address::random([0xff, 0x8f, 0x1a, 0x05, 0xe4, 0xff]),
+        }
+    }
+}
--- a/src/ariel-os-embassy/src/lib.rs
+++ b/src/ariel-os-embassy/src/lib.rs
@@ -22,6 +22,9 @@ pub mod spi;
 #[cfg(feature = "usb")]
 pub mod usb;
 
+#[cfg(feature = "ble")]
+pub mod ble;
+
 #[cfg(feature = "net")]
 pub mod net;
 
@@ -53,6 +56,8 @@ pub mod api {
         pub use embassy_time::{Delay, Duration, Instant, TICK_HZ, Timer};
     }
 
+    #[cfg(feature = "ble")]
+    pub use crate::ble;
     #[cfg(feature = "i2c")]
     pub use crate::i2c;
     #[cfg(feature = "net")]
@@ -65,6 +70,8 @@ pub mod api {
 
 // These are made available in `ariel_os::reexports`.
 pub mod reexports {
+    #[cfg(feature = "ble")]
+    pub use ariel_os_embassy_common::ble;
     #[cfg(feature = "net")]
     pub use embassy_net;
     #[cfg(feature = "time")]
@@ -208,6 +215,10 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     hal::usb::init();
 
     // Move out the peripherals required for drivers, so that tasks cannot mistakenly take them.
+
+    #[cfg(feature = "ble")]
+    let ble_peripherals = hal::ble::Peripherals::new(&mut peripherals);
+
     #[cfg(feature = "usb")]
     let usb_peripherals = hal::usb::Peripherals::new(&mut peripherals);
 
@@ -215,6 +226,12 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     // configure the drivers using hooks.
     for task in EMBASSY_TASKS {
         task(spawner, &mut peripherals);
+    }
+
+    #[cfg(feature = "ble")]
+    {
+        let config = ble::config();
+        hal::ble::driver(ble_peripherals, spawner, config);
     }
 
     #[cfg(feature = "usb")]
--- a/src/ariel-os-hal/Cargo.toml
+++ b/src/ariel-os-hal/Cargo.toml
@@ -33,6 +33,11 @@ embedded-storage-async = { workspace = true }
 
 ariel-os-embassy-common = { workspace = true }
 
+# For ble dummy implementation
+bt-hci = { workspace = true, optional = true }
+embedded-io = { workspace = true, optional = true }
+trouble-host = { workspace = true, optional = true }
+
 [features]
 external-interrupts = [
   "ariel-os-esp/external-interrupts",
@@ -61,6 +66,10 @@ usb = [
   "ariel-os-rp/usb",
   "ariel-os-stm32/usb",
 ]
+
+ble = ["dep:bt-hci", "dep:embedded-io", "dep:trouble-host"]
+ble-peripheral = []
+ble-central = []
 
 hwrng = [
   "ariel-os-esp/hwrng",
--- a/src/ariel-os-hal/src/dummy/ble.rs
+++ b/src/ariel-os-hal/src/dummy/ble.rs
@@ -1,0 +1,77 @@
+use bt_hci::{
+    cmd,
+    controller::{ControllerCmdAsync, ControllerCmdSync},
+};
+use embassy_executor::Spawner;
+
+pub struct Peripherals {}
+
+impl Peripherals {
+    pub fn new(_peripherals: &mut crate::OptionalPeripherals) -> Self {
+        unimplemented!();
+    }
+}
+
+pub async fn ble_stack() -> &'static trouble_host::Stack<'static, DummyController> {
+    async { unimplemented!() }.await
+}
+
+pub fn driver(_p: Peripherals, _spawner: Spawner, _config: ariel_os_embassy_common::ble::Config) {
+    unimplemented!();
+}
+
+pub struct DummyController {}
+
+#[derive(Debug)]
+pub struct DummyError {}
+
+impl embedded_io::Error for DummyError {
+    fn kind(&self) -> embedded_io::ErrorKind {
+        unimplemented!()
+    }
+}
+
+impl embedded_io::ErrorType for DummyController {
+    type Error = DummyError;
+}
+impl bt_hci::controller::Controller for DummyController {
+    async fn write_acl_data(
+        &self,
+        _packet: &bt_hci::data::AclPacket<'_>,
+    ) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
+
+    async fn write_sync_data(
+        &self,
+        _packet: &bt_hci::data::SyncPacket<'_>,
+    ) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
+
+    async fn write_iso_data(
+        &self,
+        _packet: &bt_hci::data::IsoPacket<'_>,
+    ) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
+
+    async fn read<'a>(
+        &self,
+        _buf: &'a mut [u8],
+    ) -> Result<bt_hci::ControllerToHostPacket<'a>, Self::Error> {
+        unimplemented!()
+    }
+}
+
+impl<C: cmd::AsyncCmd + ?Sized> ControllerCmdAsync<C> for DummyController {
+    async fn exec(&self, _cmd: &C) -> Result<(), cmd::Error<Self::Error>> {
+        unimplemented!();
+    }
+}
+
+impl<C: cmd::SyncCmd + ?Sized> ControllerCmdSync<C> for DummyController {
+    async fn exec(&self, _cmd: &C) -> Result<C::Return, cmd::Error<Self::Error>> {
+        unimplemented!();
+    }
+}
--- a/src/ariel-os-hal/src/dummy/mod.rs
+++ b/src/ariel-os-hal/src/dummy/mod.rs
@@ -24,6 +24,10 @@ pub mod peripheral {
 }
 
 #[doc(hidden)]
+#[cfg(feature = "ble")]
+pub mod ble;
+
+#[doc(hidden)]
 #[cfg(feature = "hwrng")]
 pub mod hwrng;
 
--- a/src/ariel-os-macros/src/config.rs
+++ b/src/ariel-os-macros/src/config.rs
@@ -9,10 +9,11 @@
 ///
 /// - The name of the driver the constant provides configuration for.
 ///
-/// | Driver    | Expected type                  | Cargo feature to enable   |
-/// | --------- | ------------------------------ | ------------------------- |
-/// | `network` | `embassy_net::Config`          | `network-config-override` |
-/// | `usb`     | `embassy_usb::Config`          | `override-usb-config`     |
+/// | Driver    | Expected type                     | Cargo feature to enable   |
+/// | --------- | --------------------------------- | ------------------------- |
+// | `ble`     | `ariel_os::reexport::ble::Config` | `ble-config-override`     |
+/// | `network` | `embassy_net::Config`             | `network-config-override` |
+/// | `usb`     | `embassy_usb::Config`             | `override-usb-config`     |
 ///
 /// # Note
 ///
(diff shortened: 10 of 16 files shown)


nponsard:
This PR is ready to be reviewed, I added the ble feature to the relevant CI jobs.

examples/ble-advertise/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This application is demonstrating basic BLE peripheral advertisement using the TrouBLE stack.
+
+## Running
+
+In this directory, run
+
+    laze build -b nrf52840dk run
+
+You should see using [BLE Radar](https://f-droid.org/packages/f.cking.software/) a device named "Ariel OS BLE".
```
ROMemories:
```suggestion
You should see using [BLE Radar](https://github.com/Semper-Viventem/MetaRadar) a device named "Ariel OS BLE".
```

nponsard:
While writing and testing the BLE scan example I encountered an issue where the controller in the return type of `ble_stack()` needed to implement `Controller + ControllerCmdSync<LeSetScanParams>`.

This made me realize that the `Controller` trait does not implement everything we need for all use cases planned. One solution would be to add all the possibles commands to the return types in `src/ariel-os-embassy/src/ble.rs` but that would block the cases where the controller doesn't implement some commands and the current use case doesn't need it.

I've chosen the second option: re-exporting a function that every HAL supporting BLE should export.

.github/workflows/main.yml
```diff
@@ -124,6 +124,7 @@ jobs:
                 udp,
                 usb,
                 usb-ethernet,
+                ble,
```
ROMemories: Please preserve the lexicographical order here.

.github/workflows/main.yml
```diff
@@ -273,7 +274,8 @@ jobs:
                     udp,
                     usb,
                     usb-hid,
-                    coapcore/_nightly_docs
+                    coapcore/_nightly_docs,
+                    ble,
```
ROMemories: Please preserve the lexicographical order here.

.github/workflows/build-deploy-docs.yml
```diff
@@ -90,7 +90,8 @@ jobs:
                   udp,
                   usb,
                   usb-hid,
-                  coapcore/_nightly_docs
+                  coapcore/_nightly_docs,
+                  ble,
```
ROMemories: Actually, unlike what I said before, we don't want to enable this feature here, as we don't want this to be documented yet (we *do* want to keep it in the rustdoc check in the `main.yml` however).

src/ariel-os-hal/src/dummy/mod.rs
```diff
@@ -50,6 +50,10 @@ pub mod storage;
 #[cfg(feature = "usb")]
 pub mod usb;
 
+#[doc(hidden)]
+#[cfg(feature = "ble")]
+pub mod ble;
```
ROMemories: Please preserve the lexicographical order here, moving it just before `hwrng` (first *feature-gated* module).

src/ariel-os-macros/src/config.rs
```diff
@@ -9,10 +9,11 @@
...
 ///
-/// | Driver    | Expected type                  | Cargo feature to enable   |
-/// | --------- | ------------------------------ | ------------------------- |
-/// | `network` | `embassy_net::Config`          | `network-config-override` |
-/// | `usb`     | `embassy_usb::Config`          | `override-usb-config`     |
+/// | Driver    | Expected type                     | Cargo feature to enable   |
+/// | --------- | --------------------------------- | ------------------------- |
+/// | `network` | `embassy_net::Config`             | `network-config-override` |
+/// | `usb`     | `embassy_usb::Config`             | `override-usb-config`     |
+/// | `ble`     | `ariel_os::reexport::ble::Config` | `ble-config-override`     |
```
ROMemories: Please preserve the lexicographical order here.

src/ariel-os-macros/src/config.rs
```diff
@@ -66,6 +67,10 @@ pub fn config(args: TokenStream, item: TokenStream) -> TokenStream {
             format_ident!("__ariel_os_usb_config"),
             quote! {#ariel_os_crate::reexports::embassy_usb::Config<'static>},
         ),
+        Some(ConfigKind::Ble) => (
```
ROMemories: Please preserve the lexicographical order here.

src/ariel-os-macros/src/config.rs
```diff
@@ -136,13 +141,15 @@ mod config_macro {
     pub enum ConfigKind {
         Network,
         Usb,
+        Ble,
```
ROMemories: Please preserve the lexicographical order here.

src/ariel-os-macros/src/config.rs
```diff
@@ -136,13 +141,15 @@ mod config_macro {
...
     }
 
     impl ConfigKind {
         pub fn as_name(&self) -> &'static str {
             match self {
                 Self::Network => "network",
                 Self::Usb => "usb",
+                Self::Ble => "ble",
```
ROMemories: Please preserve the lexicographical order here.

src/ariel-os-macros/tests/ui/config/misspelled_config_kind.stderr
```diff
@@ -1,4 +1,4 @@
-error: unsupported parameter (`network`, `usb` are supported)
+error: unsupported parameter (`network`, `usb`, `ble` are supported)
```
ROMemories: `ble` will be moved in first position when reordering the `ConfigKind` enum variants.

src/ariel-os/Cargo.toml
```diff
@@ -114,6 +114,8 @@ usb-hid = ["ariel-os-embassy/usb-hid"]
 network-config-override = ["ariel-os-embassy/network-config-override"]
 ## Enables custom USB configuration.
 override-usb-config = ["ariel-os-embassy/override-usb-config"]
+## Enables custom BLE configuration.
+ble-config-override = ["ariel-os-embassy/ble-config-override"]
```
ROMemories: Please preserve the lexicographical order here.

src/ariel-os/Cargo.toml
```diff
@@ -139,6 +141,13 @@ wifi-esp = ["ariel-os-embassy/wifi-esp"]
 # Selects STM32 Ethernet
 eth-stm32 = ["ariel-os-embassy/eth-stm32"]
 
+# ## Bluetooth support
```
ROMemories, line 144: Undocumented as they should be, as they must be used through their respective laze modules :+1:

src/ariel-os/Cargo.toml
```diff
@@ -139,6 +141,13 @@ wifi-esp = ["ariel-os-embassy/wifi-esp"]
...
+ble = ["ariel-os-embassy/ble"]
+# BLE peripheral
```
```suggestion
# Enables BLE to act as a peripheral. Can be used in conjunction with `ble-central`.
```

src/ariel-os/Cargo.toml
```diff
@@ -139,6 +141,13 @@ wifi-esp = ["ariel-os-embassy/wifi-esp"]
...
+ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
+# BLE central
```
```suggestion
# Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
```

src/ariel-os-hal/src/dummy/ble.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+pub fn driver(_p: Peripherals, _spawner: &Spawner, config: ariel_os_embassy_common::ble::Config) {
+    static HOST_RESOURCES: StaticCell<trouble_host::HostResources<1, 1, 27>> = StaticCell::new();
+
+    let resources = HOST_RESOURCES.init(trouble_host::HostResources::new());
+
+    let stack = trouble_host::new(DummyController {}, resources).set_random_address(config.address);
+
+    if STACK.init(stack).is_err() {
```
ROMemories: I haven't looked at this in details, but is this really necessary? These dummy modules are really just to satisfy Clippy/rustdoc/rust-analyzer. That would also allow us to remove the extra dependencies.

src/ariel-os-hal/src/dummy/ble.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn new(_peripherals: &mut crate::OptionalPeripherals) -> Self {
+        unimplemented!();
+    }
+}
+
+pub async fn ble_stack() -> &'static trouble_host::Stack<'static, impl trouble_host::Controller> {
+    STACK.get().await
+}
+
+pub fn driver(_p: Peripherals, _spawner: &Spawner, config: ariel_os_embassy_common::ble::Config) {
```
ROMemories: `Spawner` is `Copy`, no need to pass a reference to it:

```suggestion
pub fn driver(_p: Peripherals, _spawner: Spawner, config: ariel_os_embassy_common::ble::Config) {
```

src/ariel-os-embassy/src/lib.rs
```diff
@@ -205,12 +212,21 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     #[cfg(feature = "usb")]
     let usb_peripherals = hal::usb::Peripherals::new(&mut peripherals);
 
+    #[cfg(feature = "ble")]
+    let ble_peripherals = hal::ble::Peripherals::new(&mut peripherals);
```
ROMemories, line 220: I think the peripherals should be extracted in the same order they are going to be used below; can you move this before USB just above?

src/ariel-os-embassy-common/src/ble.rs
```diff
@@ -0,0 +1,16 @@
...
+/// You can customize it using the `ble-config-override` feature.
+pub struct Config {
+    /// The address of the BLE device.
+    pub address: trouble_host::Address,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            address: trouble_host::Address::random([0xff, 0x8f, 0x1a, 0x05, 0xe4, 0xfc]),
```
ROMemories: I'm always a bit wary of having these arbitrary addressed hard-coded, but I guess we have the same for IPv4; can you add a comment mentioning where that address is coming from (I think that's from the trouBLE examples)?

src/ariel-os-embassy-common/Cargo.toml
```diff
@@ -34,3 +35,5 @@ defmt = ["dep:defmt", "fugit?/defmt"]
 executor-thread = []
 
 _test = ["i2c", "spi", "external-interrupts"]
+
+ble = ["trouble-host"]
```
ROMemories: I thought that wasn't valid anymore since the edition 2024?

```suggestion
ble = ["dep:trouble-host"]
```

examples/ble-scan/laze.yml
```diff
@@ -0,0 +1,5 @@
+apps:
+  - name: example-ble-scan
+    selects:
+      - ble
```
ROMemories: I don't think it makes sense to have the `ble` laze module selected, without also having `ble-central` or `ble-peripheral`, so I think we should only consider these two public, not `ble`:

```suggestion
```

examples/ble-advertise/laze.yml
```diff
@@ -0,0 +1,5 @@
+apps:
+  - name: example-ble-advertise
+    selects:
+      - ble
```
ROMemories:
```suggestion
```

examples/ble-advertise/src/main.rs
```diff
@@ -0,0 +1,54 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::debug::log::info;
+use embassy_futures::join::join;
+use embassy_time::{Duration, Timer};
```
ROMemories: Please use our re-exports from `ariel_os::time` instead.

examples/ble-scan/src/main.rs
```diff
@@ -0,0 +1,58 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::debug::log::info;
+use core::cell::RefCell;
+use embassy_futures::join::join;
+use embassy_time::{Duration, Timer};
```
ROMemories: Please use our re-exports from `ariel_os::time` instead.

examples/ble-advertise/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+                        adv_data: &adv_data[..len],
+                        scan_data: &[],
+                    },
+                )
+                .await;
+
+            loop {
+                info!("Still running");
+
+                Timer::after(Duration::from_secs(60)).await;
```
ROMemories:
```suggestion
                Timer::after_secs(60).await;
```

examples/ble-scan/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+        let mut config = ScanConfig::default();
+        config.active = true;
+        config.phys = PhySet::M1;
+        config.interval = Duration::from_secs(1);
+        config.window = Duration::from_secs(1);
+        let mut _session = scanner.scan(&config).await.unwrap();
+        // Scan forever
+        loop {
+            info!("scanning...");
+            Timer::after(Duration::from_secs(1)).await;
```
ROMemories:
```suggestion
            Timer::after_secs(1).await;
```

examples/ble-scan/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os" }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+
+embassy-futures = { version = "0.1.1" }
+embassy-time = { version = "0.4.0" }
```
ROMemories:
```suggestion
```

examples/ble-advertise/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os" }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+
+bt-hci = { version = "0.2.0" }
+embassy-futures = { version = "0.1.1" }
+embassy-time = { version = "0.4.0" }
```
ROMemories:
```suggestion
```

examples/ble-advertise/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os" }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+
+bt-hci = { version = "0.2.0" }
```
ROMemories: Do we need this dependency in the example?

src/ariel-os-embassy-common/src/ble.rs
```diff
@@ -0,0 +1,16 @@
+//! Common BLE types to be used across different HALs.
+
+/// Configuration for the BLE stack
+/// You can customize it using the `ble-config-override` feature.
```
ROMemories:
```suggestion
/// Configuration for the BLE stack.
///
/// You can customize it using the `ble-config-override` feature.
```

examples/ble-advertise/src/main.rs
```diff
@@ -0,0 +1,54 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::debug::log::info;
+use embassy_futures::join::join;
+use embassy_time::{Duration, Timer};
+
+use trouble_host::prelude::*;
```
ROMemories: I'm not a huge fan of preludes as they make it more difficult to know where each item is coming from; could import the required items individually?

examples/ble-scan/src/main.rs
```diff
@@ -0,0 +1,58 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::debug::log::info;
+use core::cell::RefCell;
+use embassy_futures::join::join;
+use embassy_time::{Duration, Timer};
+use heapless::Deque;
+use trouble_host::prelude::*;
```
ROMemories: Same as for the other example

examples/ble-scan/src/main.rs
```diff
@@ -0,0 +1,58 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::debug::log::info;
+use core::cell::RefCell;
```
ROMemories:
```suggestion
use core::cell::RefCell;

use ariel_os::debug::log::info;
```

src/ariel-os-hal/src/dummy/ble.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct Peripherals {}
+
+impl Peripherals {
+    pub fn new(_peripherals: &mut crate::OptionalPeripherals) -> Self {
+        unimplemented!();
+    }
+}
+
+pub async fn ble_stack() -> &'static trouble_host::Stack<'static, impl trouble_host::Controller> {
+    STACK.get().await
```
ROMemories: Same as below (which would I think allow to remove the `STACK` static above, as it is not public):
```suggestion
    unimplemented!()
```

examples/ble-scan/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This application is demonstrating basic BLE central scanning using the TrouBLE stack.
+
+## Running
+
+In this directory, run
+
+    laze build -b nrf52840dk run
+
+The discovered devices will be logged in the log output.
```
ROMemories:
```suggestion
The discovered devices will be logged to the log output.
```

nponsard:
> It feels a bit weird to have examples in tree that wouldn't be able to work; _after the next review round_ (not now), would it make sense to split that PR, moving the examples to another one (stacked onto this one)?

It would make sense to merge the examples only after a working implementation is merged yes.

examples/ble-advertise/Cargo.toml
nponsard: `trouble-host` re-exports what is needed from `bt-hci` so this dependency line is superfluous.

nponsard:
I've added the requested changes. 
I also opened a separate PR #1088 to move the examples to.
If adding the examples in a separate PR is approved, I'm gonna remove them from this PR and set #1088 as ready for review.

src/ariel-os-macros/src/config.rs
```diff
@@ -9,10 +9,11 @@
 ///
 /// - The name of the driver the constant provides configuration for.
 ///
-/// | Driver    | Expected type                  | Cargo feature to enable   |
-/// | --------- | ------------------------------ | ------------------------- |
-/// | `network` | `embassy_net::Config`          | `network-config-override` |
-/// | `usb`     | `embassy_usb::Config`          | `override-usb-config`     |
+/// | Driver    | Expected type                     | Cargo feature to enable   |
+/// | --------- | --------------------------------- | ------------------------- |
+/// | `ble`     | `ariel_os::reexport::ble::Config` | `ble-config-override`     |
```
ROMemories: We don't want it to appear in the documentation just yet:
```suggestion
// | `ble`     | `ariel_os::reexport::ble::Config` | `ble-config-override`     |
```

src/ariel-os/Cargo.toml
```diff
@@ -110,6 +110,8 @@ usb-hid = ["ariel-os-embassy/usb-hid"]
 #! specific system functionality.
 #! The features below need to be enabled so that the provided custom
 #! configuration is taken into account.
+## Enables custom BLE configuration.
```
ROMemories: We don't want it to appear in the docs just yet:
```suggestion
# Enables custom BLE configuration.
```

src/ariel-os-macros/src/config.rs
```diff
@@ -134,13 +139,15 @@ mod config_macro {
 
     #[derive(Debug, enum_iterator::Sequence)]
     pub enum ConfigKind {
+        Ble,
```
ROMemories: Now that I think of think, we cannot have that variant defined yet, otherwise the error message advertises a feature that's not supported yet:
```suggestion
        // Ble,
```
Another option would be to have `enum_iterator` skip that variant, but I could not find that from a quick glance at its documentation.

src/ariel-os-macros/src/config.rs
```diff
@@ -134,13 +139,15 @@ mod config_macro {
...
         Network,
         Usb,
     }
 
     impl ConfigKind {
         pub fn as_name(&self) -> &'static str {
             match self {
+                Self::Ble => "ble",
```
ROMemories:
```suggestion
                // Self::Ble => "ble",
```

src/ariel-os-macros/tests/ui/config/misspelled_config_kind.stderr
```diff
@@ -1,4 +1,4 @@
-error: unsupported parameter (`network`, `usb` are supported)
+error: unsupported parameter (`ble`, `network`, `usb` are supported)
```
ROMemories: That should result from the `Ble` variant being commented out:

```suggestion
error: unsupported parameter (`network`, `usb` are supported)
```

src/ariel-os-hal/Cargo.toml
```diff
@@ -33,6 +33,11 @@ embedded-storage-async = { workspace = true }
 
 ariel-os-embassy-common = { workspace = true }
 
+# For ble dummy implementation
+bt-hci = { version = "0.2.0", optional = true }
```
ROMemories: I think this should use the workspace's version?

```suggestion
bt-hci = { workspace = true, optional = true }
```

Cargo.toml
```diff
@@ -147,6 +147,8 @@ rtt-target = { version = "0.6.0" }
 rp-pac = { version = "7.0", default-features = false }
 serde = { version = "1.0.197", default-features = false }
 static_cell = { version = "2.0.0", features = ["nightly"] }
+trouble-host = "0.1"
+bt-hci = { version = "0.2.0" }
```
ROMemories, line 151: Not sure whether this has already been mentioned elsewhere, but why is this not using the latest version (v0.3.2 right now)?

ROMemories (changes requested):
Almost there! After applying the suggestions, you can remove the examples from this PR as agreed, and then it should be good to go.

Cargo.toml
nponsard: `trouble-host` v0.1.0 is using version 0.2.0 of `bt-hci` but doesn't re-export everything needed from it.

ROMemories: Unfortunate, but makes sense, thanks for the clarification.

nponsard:
I moved the examples and cleaned the fixup commits.

laze-project.yml
```diff
@@ -1082,6 +1082,41 @@ modules:
...
 
+  - name: ble
+    selects:
+      - hw/ble
+    env:
+      global:
+        FEATURES:
+          - ariel-os/ble
+
+  - name: hw/ble
```
ROMemories, line 1093: It will likely be useful later to introduce another module `has_ble_antenna` that *boards* could provide, similar to `has_usb_device_port`, but no need to do this in this PR.

ROMemories (approved):
I checked that there was no leftover in the documentation, and this seems like a good base to build the nRF and CYW43 impls on, so LGTM!

Issue #1036, opened 2025-05-19
split out `cargo fetch`


kaspar030:
Currently, cargo re-checks the remote repositories a lot:

```
    Updating git repository `https://github.com/ariel-os/embassy`
    Updating git repository `https://github.com/ariel-os/embassy`                                                                                              
    Updating git repository `https://github.com/ariel-os/embassy`                                                                                              
    Updating git repository `https://github.com/ariel-os/embassy`                                                                                              
    Updating git repository `https://github.com/ariel-os/embassy`
    Updating git repository `https://github.com/ariel-os/embassy`
    Updating git repository `https://github.com/ariel-os/embassy`
    Updating git repository `https://github.com/ariel-os/embassy`              
    Updating git repository `https://github.com/ariel-os/esp-hal`
    Updating git repository `https://github.com/seanmonstar/try-lock`                                                                                          
```

Somehow this happens for every xtensa build.

A solution would be to split the build in two, `cargo fetch` to create/update `Cargo.lock` and then `cargo --offline build/run/...`, depending on the latter. That way the check is at least centralized, and only done once when doing multiple builds.

PR #1070, opened 2025-06-02
ci: split out `cargo fetch`

kaspar030:
# Description

For the CI build workflow, split out `cargo fetch` and make the following `cargo` calls use offline mode.

## Issues/PRs references

#1036 but only for CI.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,6 +133,12 @@ jobs:
         run: |
           echo "LAZE_BUILDERS=ai-c3,espressif-esp32-devkitc,espressif-esp32-c6-devkitc-1,espressif-esp32-s3-devkitc-1,bbc-microbit-v2,nrf52840dk,nrf5340dk,rpi-pico,rpi-pico-w,st-nucleo-c031c6,st-nucleo-h755zi-q,st-nucleo-wb55" >> "$GITHUB_ENV"
 
+      - name: "cargo fetch"
+        if: steps.result-cache.outputs.cache-hit != 'true'
+        run: |
+            cargo fetch --config ariel-os-cargo.toml --locked
+            echo "CARGO_NET_OFFLINE=true" >> "$GITHUB_ENV"
+
       - name: "ariel-os compilation test (default toolchain)"
         if: steps.result-cache.outputs.cache-hit != 'true'
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -136,7 +136,8 @@ jobs:
       - name: "cargo fetch"
         if: steps.result-cache.outputs.cache-hit != 'true'
         run: |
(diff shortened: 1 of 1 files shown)


ROMemories:
CI seems to be failing for two different reasons, one of them being the following:

```
error: failed to run custom build command for `liboscore v0.2.4`

Caused by:
  process didn't exit successfully: `/home/runner/work/ariel-os/ariel-os/tests/coap/../../build/bin/ai-c3/cargo/release/build/liboscore-ff58e9f9e3736f71/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/liboscore-0.2.4/build.rs:136:10:
  Failure generating cbindgen headers for the cryptobackend: CargoMetadata("../liboscore-cryptobackend-0.2.4/Cargo.toml", Metadata(Output { status: ExitStatus(unix_wait_status(25856)), stdout: "", stderr: "\u{1b}[1m\u{1b}[31merror\u{1b}[0m\u{1b}[1m:\u{1b}[0m failed to download `autocfg v1.2.0`\n\nCaused by:\n  attempting to make an HTTP request, but --offline was specified\n" }))
```

This seems to be stemming from the fact that the `cargo fetch` step is fetching dependencies from the workspace's lockfile only. That `liboscore` build script seems however to be relying on [its own lockfile, where `autocfg` is pinned to v1.2.0](https://gitlab.com/oscore/liboscore/-/blob/b1f58a74a753b3e3c533fbb23acc7adf824bdc2e/rust/Cargo.lock#L72); that dependency not being fetched by the `cargo fetch` step, the build step fails as it is not allowed to download that new dependency.

kaspar030:
This could be 'cargo metadata' not picking up the 'aariel-os-cargo.toml'. Maybe if we select the nightly toolchain with 'RUSTUP_TOOLCHAIN' globally for the fetch.

ROMemories:
> This could be 'cargo metadata' not picking up the 'aariel-os-cargo.toml'. Maybe if we select the nightly toolchain with 'RUSTUP_TOOLCHAIN' globally for the fetch.

I don't think that's it: if I try locally to run `cargo metadata` alone, it fails with the usual error when a Cargo command is not picking up the `ariel-os-cargo.toml`, while if I pass `--config` to it using `cargo metadata --config ariel-os-cargo.toml`, it gets picked up properly. In addition, only the v1.4.0 of `autocfg` gets discovered this way, not the v1.2.0 (which only comes from `liboscore`'s lockfile).

ROMemories:
Here is a local reproducer for the CI failure:

```sh
CARGO_HOME=/tmp/foo cargo fetch
CARGO_HOME=/tmp/foo CARGO_NET_OFFLINE=true laze -C examples/coap-server build -b nrf52840dk
```

PR #1071, opened 2025-06-02
docs(examples): add a new `i2c-scanner` example

therealprof:
This will scan the configured I2C bus for connected devices and report their address.

# Description

We need more examples using peripherals so this should be a good example on how initialise and do something on the I2C bus.

## Change checklist

- [X] I have cleaned up my commit history and squashed fixup commits.
- [X] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [X] I have performed a self-review of my own code.
- [X] I have made corresponding changes to the documentation.


kaspar030:
Nice one!

You'll need to commit the Cargo.lock.

therealprof:
@kaspar030 So... does it work for you on the micro:bit v2, just returning the devices at address 25 and 30?

ROMemories:
Running this example on the micro:bit V2.2.1 I have (with an LSM303AGR), I get the following (after modifying the output to print addresses as hex):

```sh
[INFO ] Found device at address 0x19 (i2c_scanner i2c_scanner/src/main.rs:33)
[INFO ] Found device at address 0x1e (i2c_scanner i2c_scanner/src/main.rs:33)
[INFO ] Found device at address 0x70 (i2c_scanner i2c_scanner/src/main.rs:33)
[INFO ] Found device at address 0x72 (i2c_scanner i2c_scanner/src/main.rs:33)
```

It seems to be finding the [expected I2C targets](https://tech.microbit.org/software/spec-i2c-protocol/#i2c-secondary-addresses): the LSM303AGR has two addresses (`0x19` for the accelerometer and `0x1e` for the magnetometer) and the other two come from the interface MCU, with addresses `0x70` and `0x71`.

therealprof:
Great find with the document, I was looking for that... the documentation is a bit weirdly structured. Good to have confirmation that it's working just fine, I was worried about the "ghost" targets on the bus but if the others are coming from the DAP, then it's all good.

ROMemories (approved):
Refactored the example a bit, and simplified the I2C driver instantiation and usage.

Successfully tested in hardware on a micro:bit V2.2.1 with an LSM303AGR.

PR #1072, opened 2025-06-02
feat(nrf): Add basic support for nrf5340 network core

Ollrogge:
# Description
This PR introduces basic support for the network core of the `nrf5340`. Since the nrf5340 features a dual-core architecture—with each core having its own flash and RAM—Inter-Processor Communication (IPC) must be implemented to enable coordination between the application core and the network core.

I initially assumed the application core would be the default target, so I kept the name as `nrf5340`. However, it might be clearer to rename it to `nrf5340-app` to explicitly reflect that it targets the application core and avoid any ambiguity with the network core.

## Testing
According to the documentation, the network core does not start automatically and must be explicitly released by the application core using [release_network_core](https://docs.embassy.dev/embassy-nrf/git/nrf5340-app-ns/reset/fn.release_network_core.html). That said, when flashing only the network core using `probe-rs`, it starts running immediately—allowing `examples/hello_world` to function without involvement from the application core.

`probe-rs` automatically determines which core to flash based on the memory layout of the firmware binary, so flashing a binary linked for the network core will target it automatically.

For other examples like `blinky`, the application core must assign GPIOs to the network core, as demonstrated [here](https://github.com/Ollrogge/ariel-os/blob/4d2e557d24e68c161ab2a77429fa4d03d358a63d/examples/blinky/src/main.rs#L22). Since these pin assignments are specific to the `nrf5340` and would introduce hardware-dependent configuration, I chose not to include them in this PR.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -259,12 +259,22 @@ contexts:
 
   - name: nrf5340
     parent: nrf53
-    provides:
-      - has_storage_support
     selects:
       - cortex-m33f
+    provides:
+      - has_storage_support
     env:
       PROBE_RS_CHIP: nrf5340_xxAA
+
+  - name: nrf5340-net
+    parent: nrf53
+    selects:
+      - cortex-m33
+    provides:
+      - has_storage_support
+    env:
+      PROBE_RS_CHIP: nrf5340_xxAA
+    help: nRF5340 Network Core support
 
   - name: nrf91
     parent: nrf
@@ -653,6 +663,15 @@ modules:
       global:
         RUSTFLAGS:
           - --cfg context=\"cortex-m7f\"
+
+  - name: cortex-m33
+    selects:
+      - cortex-m
+      - thumbv8m.main-none-eabi
+    env:
+      global:
+        RUSTFLAGS:
+          - --cfg context=\"cortex-m33\"
 
   - name: cortex-m33f
     selects:
@@ -1626,6 +1645,11 @@ builders:
 
   - name: nrf5340dk
     parent: nrf5340
+    provides:
+      - has_usb_device_port
+
+  - name: nrf5340dk-net
+    parent: nrf5340-net
     provides:
       - has_usb_device_port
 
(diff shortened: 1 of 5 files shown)


laze-project.yml
```diff
@@ -259,6 +259,16 @@ contexts:
     env:
       PROBE_RS_CHIP: nrf5340_xxAA
 
+  - name: nrf5340-net
+    parent: nrf53
+    provides:
+      - has_storage_support
+    selects:
+      - cortex-m33f
```
0e4ef622: the net core doesn't have an FPU

kaspar030:
This is great, thanks for the PR already! I'll give it the attention it deserves next week.

So `probe-rs run` knows to flash & run on the network core. This works only with `probe-rs run` I'd guess - if probe-rs is not running the show, the app binary would need to release the core?

Ollrogge:
> This is great, thanks for the PR already! I'll give it the attention it deserves next week.
> 
> So `probe-rs run` knows to flash & run on the network core. This works only with `probe-rs run` I'd guess - if probe-rs is not running the show, the app binary would need to release the core?

Yes I think so. Also, when flashing both the network core and the application core using `probe-rs`, then the network core needs to be unlocked by the application core.

Ollrogge:
@kaspar030 rebased on main as `ariel-os-nrf/build.rs` was removed

laze-project.yml
```diff
@@ -266,6 +266,16 @@ contexts:
     env:
       PROBE_RS_CHIP: nrf5340_xxAA
 
+  - name: nrf5340-net
+    parent: nrf53
+    provides:
+      - has_storage_support
+    selects:
+      - cortex-m33
```
nponsard: Just a minor thing concerning the order: in the rest of the file we have the `selects` section before the `provides` section, it would be nice to have the same ordering here.

nponsard:
Otherwise everything else looks good to me, what do you think @kaspar030 ?

laze-project.yml
Ollrogge: Fixed :) also changed it for `nrf5340` as it also had the order mixed up

nponsard:
Thanks for your contribution!

PR #1074, opened 2025-06-03
feat(nrf-ble): add ble support for nordic MCUs

nponsard:
# Description

This uses the proprietary softdevice stack from Nordic Semiconductor thanks to the `nrf-sdc` bindings.

I'm currently struggling with an issue with the interrupt executor. I get an exception back from the MPSL: `MultiprotocolServiceLayer: 112:1311` (means error at file 112 line 1311). It works fine with the thread executor.  

## Issues/PRs references

Depends on PR #1069. Linked to #560.

## Change checklist


kaspar030:
> I'm currently struggling with an issue with the interrupt executor. I get an exception back from the MPSL: `MultiprotocolServiceLayer: 112:1311` (means error at file 112 line 1311). It works fine with the thread executor.

I think it is fine to make `ble` depend on `executor-thread` when using the sdc. *maybe* open an issue.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,156 @@
...
+    );
+
+    static RNG: StaticCell<embassy_nrf::rng::Rng<'static, RNG>> = StaticCell::new();
+    let rng = RNG.init(embassy_nrf::rng::Rng::new(p.rng, Irqs));
+
+    // Executor stack should be bigger than SDC memory size
+    static SDC_MEM: StaticCell<sdc::Mem<12288>> = StaticCell::new();
+    let sdc_mem = SDC_MEM.init(sdc::Mem::new());
+
+    let sdc = build_sdc(sdc_p, rng, mpsl, sdc_mem).unwrap();
```
Ollrogge:
```suggestion
    let sdc = build_sdc(sdc_p, rng, mpsl, sdc_mem).expect("Failed to build SDC");
```

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,156 @@
...
+        mpsl::Peripherals::new(p.rtc0, p.timer0, p.temp, p.ppi_ch19, p.ppi_ch30, p.ppi_ch31);
+    let lfclk_cfg = mpsl::raw::mpsl_clock_lfclk_cfg_t {
+        source: mpsl::raw::MPSL_CLOCK_LF_SRC_RC as u8,
+        rc_ctiv: mpsl::raw::MPSL_RECOMMENDED_RC_CTIV as u8,
+        rc_temp_ctiv: mpsl::raw::MPSL_RECOMMENDED_RC_TEMP_CTIV as u8,
+        accuracy_ppm: mpsl::raw::MPSL_DEFAULT_CLOCK_ACCURACY_PPM as u16,
+        skip_wait_lfclk_started: mpsl::raw::MPSL_DEFAULT_SKIP_WAIT_LFCLK_STARTED != 0,
+    };
+    static MPSL: StaticCell<MultiprotocolServiceLayer> = StaticCell::new();
+    let mpsl = MPSL.init(mpsl::MultiprotocolServiceLayer::new(mpsl_p, Irqs, lfclk_cfg).unwrap());
```
Ollrogge:
```suggestion
    let mpsl = MPSL.init(mpsl::MultiprotocolServiceLayer::new(mpsl_p, Irqs, lfclk_cfg).expect("Failed to initialize MPSL"));
```

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    // Executor stack should be bigger than SDC memory size
+    static SDC_MEM: StaticCell<sdc::Mem<12288>> = StaticCell::new();
+    let sdc_mem = SDC_MEM.init(sdc::Mem::new());
+
+    let sdc = build_sdc(sdc_p, rng, mpsl, sdc_mem).unwrap();
+
+    let address: Address = config.address;
+
+    static HOST_RESOURCES: StaticCell<HostResources<1, 1, 27>> = StaticCell::new();
```
Ollrogge: Maybe use constants here ?

Cargo.lock
```diff
@@ -2566,6 +2579,27 @@ dependencies = [
...
+name = "example-ble-scan"
+version = "0.0.0"
+dependencies = [
+ "ariel-os",
+ "ariel-os-boards",
+ "embassy-futures",
+ "heapless 0.8.0",
+ "trouble-host",
+]
+
```
ROMemories: I think these are leftover of testing with the examples:
```suggestion
```

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+    peripherals::{self, RNG},
+    rng,
+};
+use embassy_sync::once_lock::OnceLock;
+use nrf_sdc::{
+    self as sdc, SoftdeviceController,
+    mpsl::{self, MultiprotocolServiceLayer},
+};
+use static_cell::StaticCell;
+use trouble_host::prelude::*;
```
ROMemories: Can you import the required items manually? Preludes make it hard to get where items come from.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    let sdc = build_sdc(sdc_p, rng, mpsl, sdc_mem).expect("Failed to build SDC");
+
+    static HOST_RESOURCES: StaticCell<HostResources<MAX_CONNS, MAX_CHANNELS, L2CAP_MTU>> =
+        StaticCell::new();
+    let resources = HOST_RESOURCES.init(HostResources::new());
+
+    let address: Address = config.address;
+    if STACK
+        .init(trouble_host::new(sdc, resources).set_random_address(address))
```
ROMemories: Can you split this into two statements? And then the result of `init()` can simply be ignored with `let _ =`.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+    let sdc_p = sdc::Peripherals::new(
+        p.ppi_ch17, p.ppi_ch18, p.ppi_ch20, p.ppi_ch21, p.ppi_ch22, p.ppi_ch23, p.ppi_ch24,
+        p.ppi_ch25, p.ppi_ch26, p.ppi_ch27, p.ppi_ch28, p.ppi_ch29,
+    );
+    static SDC_MEM: StaticCell<sdc::Mem<SDC_MEM_SIZE>> = StaticCell::new();
+    let sdc_mem = SDC_MEM.init(sdc::Mem::new());
+
+    let sdc = build_sdc(sdc_p, rng, mpsl, sdc_mem).expect("Failed to build SDC");
+
+    static HOST_RESOURCES: StaticCell<HostResources<MAX_CONNS, MAX_CHANNELS, L2CAP_MTU>> =
```
ROMemories: It seems that there is the same thing in the [CYW42 implementation](https://github.com/ariel-os/ariel-os/pull/1091), can you find a way to factor that out?

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+    };
+    static MPSL: StaticCell<MultiprotocolServiceLayer> = StaticCell::new();
+    let mpsl = MPSL.init(
+        mpsl::MultiprotocolServiceLayer::new(mpsl_p, Irqs, lfclk_cfg)
+            .expect("Failed to initialize MPSL"),
+    );
+    spawner.must_spawn(mpsl_task(&*mpsl));
+
+    static RNG: StaticCell<embassy_nrf::rng::Rng<'static, RNG>> = StaticCell::new();
+    let rng = RNG.init(embassy_nrf::rng::Rng::new(p.rng, Irqs));
```
ROMemories: Can you look into whether we really need to take out the RNG peripheral for this? We need to still be able to use `ariel_os::random` when BLE is in use. It is possible that this is still working because I think the hwrng on nRF only takes an exclusive reference to the peripherals temporarily, when seeding the RNG; if so that is only working because the RNG initialization happens before the BLE initialization. In that case can you add a comment in `init_task()` that the BLE initialization must happen *after* the RNG initialization (otherwise the RNG peripheral would already have been taken out, and we'd get a runtime panic).

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+const MAX_CONNS: usize = 1;
+const MAX_CHANNELS: usize = 1;
+const L2CAP_TXQ: u8 = 20;
+const L2CAP_RXQ: u8 = 20;
+
+pub async fn ble_stack() -> &'static Stack<'static, SoftdeviceController<'static>> {
+    STACK.get().await
+}
+
+pub struct Peripherals {
```
ROMemories, line 43: Can you feature-gate this to `nrf52`? It seems that the set of required peripherals is different on nRF53 and this should then fail compilation on it.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+
+            rng: peripherals.RNG.take().unwrap(),
+        }
+    }
+}
+
+pub fn driver<'d>(p: Peripherals, spawner: Spawner, config: ariel_os_embassy_common::ble::Config) {
+    debug!("Initializing nRF BLE driver");
+    let mpsl_p =
+        mpsl::Peripherals::new(p.rtc0, p.timer0, p.temp, p.ppi_ch19, p.ppi_ch30, p.ppi_ch31);
```
ROMemories, line 108: It's unfortunate that these peripherals need to be exclusively managed by that driver, but I guess there isn't much we can do about it for now.

src/ariel-os-nrf/Cargo.toml
```diff
@@ -80,8 +86,13 @@ storage = ["dep:embassy-embedded-hal"]
 ## Enables USB support.
 usb = []
 
+## Enables BLE support.
+ble = ["hwrng", "dep:nrf-sdc", "dep:trouble-host"]
```
ROMemories: This Cargo feature now needs to be enabled for Clippy in CI.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+        rc_temp_ctiv: mpsl::raw::MPSL_RECOMMENDED_RC_TEMP_CTIV as u8,
+        accuracy_ppm: mpsl::raw::MPSL_DEFAULT_CLOCK_ACCURACY_PPM as u16,
+        skip_wait_lfclk_started: mpsl::raw::MPSL_DEFAULT_SKIP_WAIT_LFCLK_STARTED != 0,
+    };
+    static MPSL: StaticCell<MultiprotocolServiceLayer> = StaticCell::new();
+    let mpsl = MPSL.init(
+        mpsl::MultiprotocolServiceLayer::new(mpsl_p, Irqs, lfclk_cfg)
+            .expect("Failed to initialize MPSL"),
+    );
+    spawner.must_spawn(mpsl_task(&*mpsl));
```
ROMemories: This reborrow doesn't seem required:
```suggestion
    spawner.must_spawn(mpsl_task(mpsl));
```

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+
+fn build_sdc<'d, const N: usize>(
+    p: nrf_sdc::Peripherals<'d>,
+    rng: &'d mut rng::Rng<RNG>,
+    mpsl: &'d MultiprotocolServiceLayer,
+    mem: &'d mut sdc::Mem<N>,
+) -> Result<nrf_sdc::SoftdeviceController<'d>, nrf_sdc::Error> {
+    let builder = sdc::Builder::new()?;
+
+    // Order matters here if we want multirole to work.
```
ROMemories, line 165: Thanks for pointing that out :+1:

src/ariel-os-nrf/src/ble.rs
ROMemories: Maybe there would a bit of simply leveraging `ariel_os::random` instead?

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mpsl_p =
+        mpsl::Peripherals::new(p.rtc0, p.timer0, p.temp, p.ppi_ch19, p.ppi_ch30, p.ppi_ch31);
+    let lfclk_cfg = mpsl::raw::mpsl_clock_lfclk_cfg_t {
+        source: mpsl::raw::MPSL_CLOCK_LF_SRC_RC as u8,
+        rc_ctiv: mpsl::raw::MPSL_RECOMMENDED_RC_CTIV as u8,
+        rc_temp_ctiv: mpsl::raw::MPSL_RECOMMENDED_RC_TEMP_CTIV as u8,
+        accuracy_ppm: mpsl::raw::MPSL_DEFAULT_CLOCK_ACCURACY_PPM as u16,
+        skip_wait_lfclk_started: mpsl::raw::MPSL_DEFAULT_SKIP_WAIT_LFCLK_STARTED != 0,
+    };
+    static MPSL: StaticCell<MultiprotocolServiceLayer> = StaticCell::new();
```
ROMemories: I think I'd prefer these statics to be define at module-level instead; I think Clippy should already be saying yelling about defining statics (and consts) first in the scope.

src/ariel-os-hal/Cargo.toml
```diff
@@ -67,9 +67,9 @@ usb = [
   "ariel-os-stm32/usb",
 ]
 
-ble = ["dep:bt-hci", "dep:embedded-io", "dep:trouble-host"]
-ble-peripheral = []
-ble-central = []
+ble = ["ariel-os-nrf/ble", "dep:trouble-host", "dep:bt-hci", "dep:embedded-io"]
```
ROMemories: I would prefer we preserve lexicographical order, while keeping `dep:` features first (I really need to take some time to integrate [`cargo-manifest-check`](https://crates.io/crates/cargo-manifest-check) in our CI):

```suggestion
ble = ["dep:bt-hci", "dep:embedded-io", "dep:trouble-host", "ariel-os-nrf/ble"]
```

laze-project.yml
```diff
@@ -239,20 +239,26 @@ contexts:
       - cortex-m4f
     env:
       PROBE_RS_CHIP: nrf52832_xxAA
+    provides:
```
ROMemories, line 253: Can you move the `provides` below the `selects` but above `env`? The way I see it is: that module selects things, to provide other things, and sets these variables.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+#[cfg(all(feature = "ble-central", not(feature = "ble-peripheral")))]
+const SDC_MEM_SIZE: usize = 2912;
+#[cfg(all(feature = "ble-peripheral", feature = "ble-central"))]
+const SDC_MEM_SIZE: usize = 6080;
+
+// Safe value of 27, compatible with all versions.
+const L2CAP_MTU: usize = 27;
+
+// Safe defaults used in trouble_host examples
+const MAX_CONNS: usize = 1;
```
ROMemories: I know that upstream examples don't always make it obvious, but I'd like to see descriptions for these constants: what they represent, what (kind) of values are acceptable, and how/when to change them (when relevant).

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+
+#[cfg(all(feature = "ble-peripheral", not(feature = "ble-central")))]
+const SDC_MEM_SIZE: usize = 2880;
+#[cfg(all(feature = "ble-central", not(feature = "ble-peripheral")))]
+const SDC_MEM_SIZE: usize = 2912;
+#[cfg(all(feature = "ble-peripheral", feature = "ble-central"))]
+const SDC_MEM_SIZE: usize = 6080;
+
+// Safe value of 27, compatible with all versions.
+const L2CAP_MTU: usize = 27;
```
ROMemories: That should indeed work with all BLE versions; I think we should still consider making that configurable: [BLE 4.2+ allows an MTU of 251](https://embassy.dev/trouble/#_background), which I suppose can significantly increase the throughput. If possible that configuration should be HAL agnostic, but that should come for free when factoring out `HOST_RESOURCES`.

Cargo.toml
```diff
@@ -149,6 +149,7 @@ serde = { version = "1.0.197", default-features = false }
 static_cell = { version = "2.0.0", features = ["nightly"] }
 trouble-host = "0.1"
 bt-hci = { version = "0.2.0" }
+nrf-sdc = { git = "https://github.com/alexmoon/nrf-sdc.git", rev = "551a95436e999b4290b4a33383aa3d6747b63dd9" }
```
ROMemories, line 152: It's unfortunate that `nrf-sfc` isn't yet published on crates.io, that would [prevent us from publishing `ariel-os-nrf` on crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries).

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,172 @@
...
+use static_cell::StaticCell;
+use trouble_host::prelude::*;
+
+use ariel_os_debug::log::debug;
+
+use crate::irqs::Irqs;
+
+pub static STACK: OnceLock<Stack<'static, SoftdeviceController<'static>>> = OnceLock::new();
+
+// During testing central mode needed 2912 bytes, peripheral mode needed 1448 bytes. Multirole (central + peripheral) needed 6080 bytes. Allocate more here if using extended features.
```
ROMemories: style: Can you make sure to wrap comments at 100 characters? That one is a bit long.

Cargo.lock
nponsard: whoops that my bad, I thought I removed these already.

src/ariel-os-nrf/Cargo.toml
nponsard: Do you mean the [clippy for nRF](https://github.com/ariel-os/ariel-os/blob/f92d89783902c1dd759600eec1085862b57ea47c/.github/workflows/main.yml#L186) task ?

ROMemories: Yes

src/ariel-os-nrf/src/ble.rs
nponsard: I pushed a change in b23855dc7bbf66931756fcac95667544a2555f69 to move the host resources initialization in ariel-os-embassy-common, what do you think ?

src/ariel-os-nrf/Cargo.toml
nponsard: I added the `ble-central` feature in the clippy check (some crates can't build with no role selected).
Since `ble` enables `hwrng` that wasn't enabled before, I'm also getting clippy warnings/errors from there.

src/ariel-os-nrf/src/ble.rs
nponsard: I tried, but the builder requires the generator to be `Send`, but in Ariel we make the generator explicitly not `Send`. Asked chrysn about it in the public Matrix channel.

nponsard:
With the feedback I've made changes to the common code for all BLE stacks:
- Moved the creation of `HostResources` to `ariel-os-embassy-common` so all implementations use the same function. 
- Since `HostResources` relies on `MTU`, `MAX_CONNS` and `MAX_CHANNELS` constants, I also moved them to `ariel-os-embassy-common`.
- Clippy reported that the configuration passed to `driver()` didn't need to be copied, I made it a reference.

nponsard:
I found a [page in nrfxlib](https://github.com/nrfconnect/sdk-nrfxlib/blob/e380532c76d5750015bc4e53d7e2e9e27ac1f3c6/softdevice_controller/README.rst?plain=1#L15) that explains in which modes certain features are available.

nponsard:
I changed how the RNG is initialized as suggested by @chrysn, I tested on an nrf52840dk and the examples work. This allows us to leave the RNG device available after initialization.

src/ariel-os-random/src/lib.rs
```diff
@@ -89,6 +89,29 @@ impl RngCore for FastRng {
     }
 }
 
+/// Same as [`FastRng`], but can be shared across threads and tasks.
+/// This is less efficient and should only be used when necessary (e.g.) integrating with C code.
+pub struct FastRngSend {
```
ROMemories: I think we'll want to keep all these new items as `#[doc(hidden)]` for now, at least because it's not the focus of this PR. Making them part of the docs later would be trivial.

nponsard: The additions to random are now hidden.

.github/workflows/main.yml
```diff
@@ -194,8 +194,10 @@ jobs:
                 external-interrupts,
                 i2c,
                 spi,
+                ble-central,
```
ROMemories: Please preserve lexicographical order.

ROMemories: Would it make sense to also enable `ble-peripheral`?

src/ariel-os-embassy-common/src/ble.rs
```diff
@@ -1,5 +1,31 @@
...
+pub const MAX_CHANNELS: usize = 1;
+
+static HOST_RESOURCES: StaticCell<BleHostResource> = StaticCell::new();
+
+/// Alias for the BLE host resources, setting the generic parameters to the constants for convenience.
+pub type BleHostResource = HostResources<MAX_CONNS, MAX_CHANNELS, MTU>;
+
+/// Initialize the BLE host resources.
+///
+/// Call this function to get the `HostResources` instance that can be used to initialeze the trouBLE stack.
```
ROMemories:
```suggestion
/// Call this function to get the `HostResources` instance that can be used to initialize the trouBLE stack.
```

src/ariel-os-embassy-common/src/ble.rs
```diff
@@ -1,5 +1,31 @@
...
+pub const MAX_CONNS: usize = 1;
+/// Maximum number of concurrent channels to be handled by the BLE stack (not including GATT), minimum 1.
+pub const MAX_CHANNELS: usize = 1;
+
+static HOST_RESOURCES: StaticCell<BleHostResource> = StaticCell::new();
+
+/// Alias for the BLE host resources, setting the generic parameters to the constants for convenience.
+pub type BleHostResource = HostResources<MAX_CONNS, MAX_CHANNELS, MTU>;
+
+/// Initialize the BLE host resources.
```
ROMemories:
```suggestion
/// Initializes the BLE host resources.
```

src/ariel-os-embassy-common/src/ble.rs
```diff
@@ -1,5 +1,31 @@
...
+
+/// Maximum number of concurrent connections to be handled by the BLE stack, minimum 1.
+pub const MAX_CONNS: usize = 1;
+/// Maximum number of concurrent channels to be handled by the BLE stack (not including GATT), minimum 1.
+pub const MAX_CHANNELS: usize = 1;
+
+static HOST_RESOURCES: StaticCell<BleHostResource> = StaticCell::new();
+
+/// Alias for the BLE host resources, setting the generic parameters to the constants for convenience.
+pub type BleHostResource = HostResources<MAX_CONNS, MAX_CHANNELS, MTU>;
```
ROMemories: Shouldn't that be plural?

```suggestion
pub type BleHostResources = HostResources<MAX_CONNS, MAX_CHANNELS, MTU>;
```

src/ariel-os-embassy/src/lib.rs
```diff
@@ -231,7 +231,7 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     #[cfg(feature = "ble")]
     {
         let config = ble::config();
-        hal::ble::driver(ble_peripherals, spawner, config);
+        hal::ble::driver(ble_peripherals, spawner, &config);
```
ROMemories: Why pass the config by reference? It works just as well by passing it by move, is it because of its size? Fine if there's a reason, just curious.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,180 @@
...
+
+static STACK: OnceLock<Stack<'static, SoftdeviceController<'static>>> = OnceLock::new();
+static MPSL: StaticCell<MultiprotocolServiceLayer> = StaticCell::new();
+static SDC_MEM: StaticCell<sdc::Mem<SDC_MEM_SIZE>> = StaticCell::new();
+static RNG: StaticCell<ariel_os_random::CryptoRngSend> = StaticCell::new();
+
+// Memory to allocate to the SoftDevice Controller (SDC).
+//
+// During testing central mode needed 2912 bytes, peripheral mode needed 1448 bytes.
+// Multirole (central + peripheral) needed 6080 bytes. Allocate more here if using extended features.
```
ROMemories, line 24: We'll likely need to revisit this, but seems fine at first.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -0,0 +1,180 @@
...
+
+#[cfg(all(feature = "ble-peripheral", not(feature = "ble-central")))]
+const SDC_MEM_SIZE: usize = 2880;
+#[cfg(all(feature = "ble-central", not(feature = "ble-peripheral")))]
+const SDC_MEM_SIZE: usize = 2912;
+#[cfg(all(feature = "ble-peripheral", feature = "ble-central"))]
+const SDC_MEM_SIZE: usize = 6080;
+
+// Size of the TX buffer (number of packets), minimum is 1, SoftDevice default is 3 (SDC_DEFAULT_TX_PACKET_COUNT).
+const L2CAP_TXQ: u8 = 20;
```
ROMemories: So why set it to 20?

src/ariel-os-nrf/src/usb.rs
```diff
@@ -1,23 +1,10 @@
...
-    usb::{
-        self, Driver,
-        vbus_detect::{self, HardwareVbusDetect},
-    },
+    pac, peripherals,
+    usb::{Driver, vbus_detect::HardwareVbusDetect},
 };
 
-#[cfg(context = "nrf52")]
-bind_interrupts!(struct Irqs {
```
ROMemories, line 11: No need to do this for this PR, but usually it would be easier to review if the interrupt refactor was split into two commits: one that does the refactor while keeping the same behavior, and another one that builds on it.

src/ariel-os-embassy/src/lib.rs
nponsard: It was in reaction to the clippy lint, so I guess its better this way ?

src/ariel-os-nrf/src/ble.rs
nponsard: I don't know a better way to handle this. On the nRF5340 its also different sizes.

src/ariel-os-embassy/src/lib.rs
ROMemories: Which Clippy lint?

src/ariel-os-nrf/src/ble.rs
nponsard: This is leftovers from testing. 
Just tested and the default values work correctly.

.github/workflows/main.yml
nponsard: The code paths are not that different, it wouldn't hurt to enable it. I noticed that the issues concerning enabling peripheral or central mode appear at runtime.

src/ariel-os-embassy/src/lib.rs
nponsard: When running clippy on example/ble-scan:
```
warning: this argument is passed by value, but not consumed in the function body
   --> src/ariel-os-nrf/src/ble.rs:101:57
    |
101 | pub fn driver(p: Peripherals, spawner: Spawner, config: ariel_os_embassy_common::ble::Config) {
    |                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&ariel_os_embassy_common::ble::Config`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value
    = note: `-W clippy::needless-pass-by-value` implied by `-W clippy::pedantic`
    = help: to override `-W clippy::pedantic` add `#[allow(clippy::needless_pass_by_value)]`
```

ROMemories:
I think the `ble` Cargo feature of `ariel-os-embassy` also needs to enable `hwrng`, otherwise the crypto RNG doesn't get populated at startup:

```
Initialization should have populated RNG
```

I think it's fine to enable it in `ariel-os-embassy` even though technically only nRF requires it for now, because it's just kind of expected to have an RNG when doing networky stuff (maybe add a comment mentioning that not every HAL may actually require it).

src/ariel-os-embassy/src/lib.rs
ROMemories: Ok, you can just add an `expect` (for consistency with other driver initialization functions), or consume the `config` with `let _ = config;` after using `config.address` if that works.

nponsard:
> I think the `ble` Cargo feature of `ariel-os-embassy` also needs to enable `hwrng`, otherwise the crypto RNG doesn't get populated at startup:
> 
> ```
> Initialization should have populated RNG
> ```
> 
> I think it's fine to enable it in `ariel-os-embassy` even though technically only nRF requires it for now, because it's just kind of expected to have an RNG when doing networky stuff (maybe add a comment mentioning that not every HAL may actually require it).

I changed the laze project configuration to have to select hwrng when ble is selected on the nrf boards. 
The error shouldn't appear anymore, but I can still add the dependency in `ariel-os-embassy` if you want.

.github/workflows/main.yml
ROMemories: All features we can enable we should enable.

nponsard: okay I enabled both

nponsard:
I rebased and cleaned up my commits. The `SDC_MEM_SIZE` requirements for different modes seem to have changed again, weird. Now theres a warning when starting:
`WARN  Memory buffer too big. 1480 bytes needed` (this one is for central mode)
Edit: its linked to `L2CAP_TXQ` and `L2CAP_RXQ`

ROMemories (approved):
Great job on this!

Successfully tested the `ble-advertise` example on an nrf52840dk while scanning with the `ble-scan` example on an bbc-microbit-v2 from #1088.

ROMemories:
> Now theres a warning when starting:
> `WARN Memory buffer too big. 1480 bytes needed` (this one is for central mode)
> Edit: its linked to `L2CAP_TXQ` and `L2CAP_RXQ`

I can indeed reproduce this, my understanding is this this is not dangerous, it's just the buffer being *too* large, as we've reduced the size requirement as you said. Going with this for now, we can revisit these allocations later.

PR #1077, opened 2025-06-04
fix(rp235x): use the built-in TRNG as HWRNG instead of `RoscRng`

ROMemories:
# Description

Unlike the RP2040, the RP235x does have a FIPS 140-compliant TRNG instead of a best-effort HWRNG relying on the ROSC.
This PR wires up that TRNG.

## Testing

Tested the `random` example on the rpi-pico2-w and the rpi-pico-w (to make sure it still works). Commenting out the RP2040-specific block when compiling for the RP235x and vice versa allows to make sure each block is properly used.

## Changelog entry

(RP235x) The `ariel_os::random` module is now seeding its RNGs from the TRNG (which is not available on the RP2040), instead of relying on the `RoscRng`.

## Change checklist

### DIFF
--- a/src/ariel-os-rp/src/hwrng.rs
+++ b/src/ariel-os-rp/src/hwrng.rs
@@ -1,3 +1,29 @@
-pub fn construct_rng(_peripherals: &mut crate::OptionalPeripherals) {
-    ariel_os_random::construct_rng(embassy_rp::clocks::RoscRng);
+pub fn construct_rng(peripherals: &mut crate::OptionalPeripherals) {
+    #[cfg(context = "rp2040")]
+    let hwrng = {
+        let _ = peripherals; // Mark used
+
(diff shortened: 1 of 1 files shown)


src/ariel-os-rp/src/hwrng.rs
```diff
@@ -1,3 +1,29 @@
...
+
+    #[cfg(context = "rp235xa")]
+    let hwrng = {
+        embassy_rp::bind_interrupts!(struct Irqs {
+            TRNG_IRQ => embassy_rp::trng::InterruptHandler<embassy_rp::peripherals::TRNG>;
+        });
+
+        let trng = peripherals.TRNG.take().unwrap();
+
+        // Enable all entropy checks to ensure acceptable results.
```
ROMemories, line 19: Embassy now keeps these tests enabled, but that's not the case in the current version of our fork.

kaspar030 (approved):

Issue #639, opened 2024-12-19
refactor: Unify & generate linker scripts (memory.x)


chrysn:
### Description

@kaspar030 wants to reduce duplication in the linker scripts

*update (10.06.2026)*

So this is getting some traction. We're working on a [tool](https://codeberg.org/bergzand/memsolve) that'll allow modeling complex linker scripts.
We'll treat esp partitions as "the way to do flash partitioning" on esp, and memsolve will get support for that.

### Motivation

Main motivation is that for bootloaders / software updates, and storage, we want programmatic control over linker section sizes (or esp flash partitions).

### Current Implementation

- nrf linker scripts are generated in `ariel-os-rt`'s `build.rs` using [ld-memory](https://github.com/kaspar030/ld-memory)
- others are using fixed or hal-supplied memory.x

### Are you planning to do it yourself in a pull request?

No

chrysn:
One aspect of this is to update link-rs to not only allow skipping the first few KiB in a configurable way, but also the last few KiB. This will enable using the bootloader for #997, but would also help create a reliable place for persistent storage.

PR #1082, opened 2025-06-06
refactor(rt): move ld-memory handling into ariel-os-rt

kaspar030:
This moves the [ld-memory](kaspar030/ld-memory) based `memory.x` generation from `ariel-os-nrf` into `ariel-os-rt`. Additionally, this drops the MCU specific env varialble prefixes, this is now the default for all MCUs (`LDMEMORY_`).

This allows e.g., adding a bootloader offset via env variable, e.g.,

`LDMEMORY_OFFSET=16384 laze build -b nrf52840dk`.

Kept for follow-ups:

1. pass ram/flash/page-sizes via laze
2. allow setting & exporting `LDMEMORY_OFFSET` in laze
3. read ram/flash/page-sizes from stm32-metapac
4. ...

While this does move MCU specifics into `ariel-os-rt`, this is supposed to be temporary.

## Issues/PRs references

Works towards #639.

### DIFF
--- a/src/ariel-os-rt/build.rs
+++ b/src/ariel-os-rt/build.rs
@@ -78,27 +78,13 @@ fn write_memoryx() {
         panic!("please set the MCU laze context");
     };
 
-    let slot_prefix = if context("nrf51") {
-        "NRF51_FLASH"
-    } else if context("nrf52") {
-        "NRF52_FLASH"
-    } else if context("nrf5340") {
-        "NRF5340_FLASH"
-    } else if context("nrf9151") {
-        "NRF9151_FLASH"
-    } else if context("nrf9160") {
-        "NRF9160_FLASH"
-    } else {
-        unreachable!();
-    };
-
     // generate linker script
     let memory = Memory::new()
         .add_section(MemorySection::new("RAM", 0x2000_0000, ram * 1024))
         .add_section(
             MemorySection::new("FLASH", 0x0, flash * 1024)
                 .pagesize(4096)
-                .from_env_with_prefix(slot_prefix),
+                .from_env(),
         );
 
     memory.to_cargo_outdir("memory.x").expect("wrote memory.x");


src/ariel-os-rt/build.rs
```diff
@@ -56,9 +45,60 @@ fn main() {
     std::fs::copy("eheap.x", out.join("eheap.x")).unwrap();
     std::fs::copy("keep-stack-sizes.x", out.join("keep-stack-sizes.x")).unwrap();
 
+    #[cfg(feature = "ld-memory")]
+    memoryx();
+
     println!("cargo:rerun-if-changed=linkme.x");
     println!("cargo:rerun-if-changed=eheap.x");
     println!("cargo:rerun-if-changed=keep-stack-sizes.x");
+    println!("cargo:rerun-if-changed=build.rs");
```
ROMemories: If I'm reading [the docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) right, this doesn't seem necessary:

```suggestion
```

src/ariel-os-rt/build.rs
```diff
@@ -56,9 +45,60 @@ fn main() {
...
+    } else if context("nrf52833") {
+        (128, 512)
+    } else if context("nrf52840") {
+        (256, 1024)
+    } else if context("nrf5340") {
+        (512, 1024)
+    } else if context_any(&["nrf9151", "nrf9160"]).is_some() {
+        (256, 1024)
+    } else {
+        panic!("nrf52: please set MCU feature");
```
ROMemories: This string was leftover, looks like an opportunity to update it:
```suggestion
        panic!("please set the MCU laze context");
```

src/ariel-os-rt/Cargo.toml
```diff
@@ -54,6 +55,7 @@ panic-printing = []
 _panic-handler = []
 single-core = ["cortex-m/critical-section-single-core"]
 multi-core = ["embassy-rp/critical-section-impl"]
+ld-memory = ["dep:ld-memory"]
```
ROMemories: I don't the Cargo feature name should be the name of the dependency, but indicate what the feature does; `embassy-stm32` calls this `memory-x`: 
```suggestion
memory-x = ["dep:ld-memory"]
```

src/ariel-os-rt/Cargo.toml
```diff
@@ -17,6 +17,7 @@ ariel-os-utils = { workspace = true }
 
 [build-dependencies]
 rustflags = "0.1"
```
ROMemories: Looks like that dependency is removed as part of this PR: 
```suggestion
```

src/ariel-os-rt/build.rs
```diff
@@ -56,9 +45,60 @@ fn main() {
...
     println!("cargo:rerun-if-changed=linkme.x");
     println!("cargo:rerun-if-changed=eheap.x");
     println!("cargo:rerun-if-changed=keep-stack-sizes.x");
+    println!("cargo:rerun-if-changed=build.rs");
 
     println!("cargo:rustc-link-search={}", out.display());
 }
+
+#[cfg(feature = "ld-memory")]
+fn memoryx() {
```
ROMemories: Function names should be verbs, and this helps clarify that this actually writes to the FS (doesn't just return the generate `memory.x` as a string for instance):

```suggestion
fn write_memoryx() {
```

src/ariel-os-rt/build.rs
ROMemories: I think this would panic if the `memory-x`/`ld-memory` Cargo feature is enabled on this crate by any other HAL, but I understand this is a first refactoring step, so that's fine.

ROMemories (approved):
Successfully tested `hello-world` on stm32u083c-dk and nrf52840dk.

PR #1085, opened 2025-06-09
feat(threads): Fix `stack_pos` in `setup_stack`

Ollrogge:
Signed-off-by: Ollrogge <nils-ollrogge@outlook.de>

# Description
The `setup_stack` function for `cortex_m` reserved 36 bytes, even though the basic exception frame only requires 32 bytes. Also, some of the comments were misleading / confusing in my opinion.


kaspar030:
Both fixes seem totally correct. Thanks for spotting and fixing!

... makes me wonder how that worked before, looks like we didn't actually have the 8b aligned stack pointer.

Ollrogge:
> Both fixes seem totally correct. Thanks for spotting and fixing!
> 
> ... makes me wonder how that worked before, looks like we didn't actually have the 8b aligned stack pointer.

At least for exception entry, it is guaranteed that the stack pointer will be aligned to 8 bytes if `CCR.STKALIGN` is set, which it is by default on most chips I think.

kaspar030:
Thanks!

PR #1091, opened 2025-06-11
feat(rp): add BLE support for RP MCUs (CYW43 chip)

jkr-ariel:
# Description

This PR aims to expose the [trouBLE](https://github.com/embassy-rs/trouble) API for use with Raspberry Pi MCUs.

## Issues/PRs references

Depends on #1069.
Linked to #560.

## Open Questions

There is currently a limitation that comes from the fact that the CYW43 chip handles both wifi and bluetooth. With the current cargo feature setup, it is required to provide wifi credentials (even if the app will not join the network) when building an application that only uses bluetooth.
I did not want to make relatively big changes without some kind of review/discussion beforehand.

There are 3 use cases to think about:
- the user wants bluetooth AND wifi;
- the user ONLY wants bluetooth (and thus should not build wifi related code, including the wifi creds);
- the user ONLY wants wifi (this should already be okay).

## Change checklist

### DIFF
--- a/examples/ble-advertiser/laze.yml
+++ b/examples/ble-advertiser/laze.yml
@@ -3,6 +3,6 @@ apps:
     env:
       global:
         executor_stacksize_required:
-          - "32768"
+          - "16384"
     selects:
       - ble-peripheral
--- a/examples/ble-scanner/laze.yml
+++ b/examples/ble-scanner/laze.yml
@@ -3,6 +3,6 @@ apps:
     env:
       global:
         executor_stacksize_required:
-          - "32768"
+          - "16384"
     selects:
       - ble-central


src/ariel-os-rp/Cargo.toml
```diff
@@ -67,6 +72,15 @@ storage = []
 ## Enables USB support.
 usb = []
 
+## Enables Bluetooth Low Energy support.
+ble = [
```
ROMemories: This needs to be named similarly to `wifi-cyw43`: it just happens that the official devkits with the RP MCUs (the Pico W and Pico 2 W) feature a CYW43 chip, but it is not intrinsic to that MCU in any way.
```suggestion
ble-cyw43 = [
```

laze-project.yml
```diff
@@ -1595,6 +1599,7 @@ builders:
     provides:
       - has_usb_device_port
       - has_wifi_cyw43
+      - has_ble
```
ROMemories: Please preserve lexicographical order.

laze-project.yml
```diff
@@ -1606,6 +1611,7 @@ builders:
...
```
ROMemories: Please preserve lexicographical order.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -34,7 +34,7 @@ mod wifi;
 #[cfg(feature = "eth")]
 mod eth;
 
-use ariel_os_debug::log::debug;
+use ariel_os_debug::log::{debug,info};
```
ROMemories: These debugging log lines should be removed.
```suggestion
use ariel_os_debug::log::debug;
```

src/ariel-os-embassy/src/lib.rs
```diff
@@ -216,23 +217,30 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
     // Move out the peripherals required for drivers, so that tasks cannot mistakenly take them.
 
-    #[cfg(feature = "ble")]
-    let ble_peripherals = hal::ble::Peripherals::new(&mut peripherals);
-
     #[cfg(feature = "usb")]
     let usb_peripherals = hal::usb::Peripherals::new(&mut peripherals);
 
+    #[cfg(all(feature = "ble", not(context = "rp")))]
+    let ble_peripherals = hal::ble::Peripherals::new(&mut peripherals);
```
ROMemories, line 220: Please move this back above, so that the peripherals are extracted in the same order as they are used to initialize the drivers.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -216,23 +217,30 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
 
     // Move out the peripherals required for drivers, so that tasks cannot mistakenly take them.
 
-    #[cfg(feature = "ble")]
-    let ble_peripherals = hal::ble::Peripherals::new(&mut peripherals);
-
     #[cfg(feature = "usb")]
     let usb_peripherals = hal::usb::Peripherals::new(&mut peripherals);
 
+    #[cfg(all(feature = "ble", not(context = "rp")))]
```
ROMemories, line 219: I'd like us to revisit when the peripherals for the CYW43 chip are extracted, similar to USB, but that's indeed out of scope of this PR.

src/ariel-os-rp/src/lib.rs
```diff
@@ -46,6 +46,10 @@ pub mod storage;
 #[doc(hidden)]
 pub mod usb;
 
+#[cfg(feature = "ble")]
+#[doc(hidden)]
+pub mod ble;
```
ROMemories: Please move this just before `cyw43` to preserve the lexicographical order.

src/ariel-os-rp/src/lib.rs
```diff
@@ -17,7 +17,7 @@ mod picotool;
 #[cfg(feature = "wifi")]
 mod wifi;
 
-#[cfg(feature = "wifi-cyw43")]
+#[cfg(any(feature = "wifi-cyw43", feature = "ble"))]
```
ROMemories:
```suggestion
#[cfg(any(feature = "ble-cyw43", feature = "wifi-cyw43"))]
```

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -4,7 +4,12 @@
 )]
 mod rpi_pico_w;
 
+#[cfg(feature = "ble")]
+use crate::ble;
+
 use ariel_os_debug::log::info;
+#[cfg(feature = "ble")]
+use bt_hci::controller::ExternalController;
```
ROMemories: Can you reorder the new imports so that the feature-gated ones come after all the non-feature-gated ones? Also the `crate::*` ones should be a separate group from the external crate one, placed after it, similar to what the [`StdExternalCrate` configuration of the `group_imports` options of rustfmt does](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=group#StdExternalCrate%5C%3A) (unfortunately we can't use this yet).

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -87,3 +93,61 @@ pub async fn device<'a, 'b: 'a>(
 
     (net_device, control)
 }
+
+#[cfg(feature = "ble")]
+pub async fn device<'a, 'b: 'a>(
```
ROMemories: It seems that this function would be better merged with the existing one, which could make handling the Wi-Fi + BLE use that you've found easier.

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -87,3 +93,61 @@ pub async fn device<'a, 'b: 'a>(
...
+
+    control.init(clm).await;
+
+    let controller: ExternalController<_, 10> = ExternalController::new(bt_device);
+    static HOST_RESOURCES: StaticCell<trouble_host::HostResources<1, 1, 27>> = StaticCell::new();
+    let resources = HOST_RESOURCES.init(trouble_host::HostResources::new());
+    let stack = trouble_host::new(controller, resources).set_random_address(config.address);
+    if ble::STACK.init(stack).is_err() {
+        panic!("error initializing OnceLock with BLE stack");
+    }
```
ROMemories: We're fine with ignoring the result in this case: (a) if the stack was already initiliazed for whatever reason, it's fine to keep it and (b) this could reduce the binary size a bit.

```suggestion
    let _ = ble::STACK.init(stack);
```

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -87,3 +93,61 @@ pub async fn device<'a, 'b: 'a>(
...
+        cyw43::new_with_bluetooth(STATE.init_with(|| cyw43::State::new()), pwr, spi, fw, btfw)
+            .await;
+
+    // this needs to be spawned here (before using `control`)
+    spawner.spawn(wifi_cyw43_task(runner)).unwrap();
+
+    control.init(clm).await;
+
+    let controller: ExternalController<_, 10> = ExternalController::new(bt_device);
+    static HOST_RESOURCES: StaticCell<trouble_host::HostResources<1, 1, 27>> = StaticCell::new();
```
ROMemories: We also need to come up with a way for configuring the MTU.

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -87,3 +93,61 @@ pub async fn device<'a, 'b: 'a>(
...
+    let (net_device, bt_device, mut control, runner) =
+        cyw43::new_with_bluetooth(STATE.init_with(|| cyw43::State::new()), pwr, spi, fw, btfw)
+            .await;
+
+    // this needs to be spawned here (before using `control`)
+    spawner.spawn(wifi_cyw43_task(runner)).unwrap();
+
+    control.init(clm).await;
+
+    let controller: ExternalController<_, 10> = ExternalController::new(bt_device);
```
ROMemories: Please define a constant to avoid this magic number.

src/ariel-os-rp/src/cyw43.rs
ROMemories: It looks to me that this isn't HAL-specific, we should find a way to move this into a shared crate (e.g., `ariel-os-embassy` or `ariel-os-embassy-common`).

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -87,3 +93,61 @@ pub async fn device<'a, 'b: 'a>(
...
+    static HOST_RESOURCES: StaticCell<trouble_host::HostResources<1, 1, 27>> = StaticCell::new();
+    let resources = HOST_RESOURCES.init(trouble_host::HostResources::new());
+    let stack = trouble_host::new(controller, resources).set_random_address(config.address);
```
ROMemories: Looking at the docs, it seems that there is a [set_random_generator_seed()`](https://docs.embassy.dev/trouble-host/git/default/struct.Stack.html#method.set_random_generator_seed) method, we might need to use it, even if it's not in the examples.

src/ariel-os-rp/src/cyw43/firmware/43439A0.bin
ROMemories, line 1: I'll review the blobs in a later round.

ROMemories (changes requested):
Looks like a really good start! I've left some inline comments; please also address the Clippy lints.

---

We indeed want to support the three use cases you've found (Wi-Fi and/or BLE), as both the CYW43 chip and the `cyw43` driver support them. We also do want to not require users to provide a dummy SSID/password pair for a non-existent Wi-Fi network if they're only to use BLE, as you mention. I'm totally fine doing this in another round to keep the changes rather minimal for the next one.

src/ariel-os-rp/Cargo.toml
ROMemories: This Cargo feature also needs to be enabled in CI for Clippy (in `main.yml`).

ROMemories:
Can you make sure the various lint checks are passing?

src/ariel-os-rp/src/cyw43.rs
jkr-ariel: It seems like the rustfmt CI check wants these imports to be at the top (in alphabetical order), whether they are feature gated or not.

ROMemories: Yes, that's why you need to make the feature-gated imports a separate group (by inserting a blank line): rustfmt only re-orders within such groups.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -229,10 +229,27 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
+
+        #[cfg(not(context = "rp"))]
+        let (device, control) = ((), ());
+
+        (device, control)
+    };
+    // Ugly way to silence unused variables warning when using Bluetooth but not
+    // wifi with RP boards.
+    #[cfg(all(feature = "ble", not(feature = "wifi-cyw43")))]
+    let ((), ()) = (device, control);
```
ROMemories: Looks to me that this should be merged with the `wifi-cyw43` instantiation from below and simplified; that last line does not compile, you need `_` instead of `()` as an lvalue.

src/ariel-os-rp/src/cyw43.rs
```diff
@@ -14,9 +13,31 @@ use embassy_rp::{
...
+#[cfg(feature = "ble-cyw43")]
+const SLOTS: usize = 10;
+#[cfg(feature = "ble-cyw43")]
+// Max number of BLE connections supported.
+const CONNS: usize = 1;
+#[cfg(feature = "ble-cyw43")]
+// Max number of L2CAP channels supported (not including GATT).
+const CHANNELS: usize = 1;
+#[cfg(feature = "ble-cyw43")]
+// Safe default MTU value that should work everywhere.
```
ROMemories: style: We usually place comments before attributes.

src/ariel-os-rp/src/ble.rs
```diff
@@ -0,0 +1,11 @@
+use bt_hci::controller::ExternalController;
+use cyw43::bluetooth::BtDriver;
+use embassy_sync::once_lock::OnceLock;
+
+pub(crate) static STACK: OnceLock<
+    trouble_host::Stack<'static, ExternalController<BtDriver<'static>, 10>>,
```
ROMemories: What's the role of `ExternalController` here? The type of that stack seems to be different in #1074 but it's not clear to me why (except for the difference `embassy-nrf` vs cy43 difference).
cc @nponsard

If we keep it, a const needs to be defined in place of that magic number, and reused in the `cyw43.rs` module (and documented).

src/ariel-os-rp/src/cyw43/firmware/README.md
ROMemories, line 1: I checked the three binary blobs and this README, their SHA256 hashes do match the ones from [upstream](https://github.com/embassy-rs/embassy/tree/dda17cc7cb85e3818d43561102b1fc6358d059c5/cyw43-firmware).

However, [upstream changed their blob source from the official Infineon repository to some third-party one](https://github.com/embassy-rs/embassy/pull/2865), and it's not clear to me where these are now coming from.

cc @kaspar030 @chrysn

src/ariel-os-rp/src/ble.rs
nponsard: `ExternalController` corresponds to the controller part of the BLE stack, it receives commands and communicates with the hardware. I haven't found a way to abstract it out, the best way to store it right now is in an crate/file specific to the manufacturer.

jkr-ariel: The reason we need to use `ExternalController` is because the `BtDriver` type returned by the `cyw43` crate does not implement the `Controller` trait from the `bt_hci` crate, whereas `SoftDeviceController` from `nrf-sdc` does.

.github/workflows/main.yml
```diff
@@ -177,6 +177,8 @@ jobs:
                 external-interrupts,
                 i2c,
                 spi,
+                ble,
+                ble-cyw43,
```
ROMemories: Please preserve the lexicographical order.

src/ariel-os-embassy-common/src/ble.rs
ROMemories, line 1: Unless I'm mistaken, it seems that changes to this file are actually not used in this PR, can you remove them?

src/ariel-os-rp/Cargo.toml
```diff
@@ -67,6 +72,25 @@ storage = []
...
 usb = []
 
+## Enables Bluetooth Low Energy support.
+ble = ["ariel-os-embassy-common/ble"]
+
+ble-cyw43 = [
+  "dep:cyw43",
+  "dep:cyw43-pio",
+  "dep:embassy-net-driver-channel",
+  "dep:static_cell",
```
ROMemories: Could you introduce an intermediate Cargo feature to factor these out with `wifi-cyw43`? As it is functionally private, prefixing it with an underscore would it from docs.rs (e.g., `_cyw43`).

src/ariel-os-rp/src/ble.rs
```diff
@@ -0,0 +1,14 @@
+use bt_hci::controller::ExternalController;
+use cyw43::bluetooth::BtDriver;
+use embassy_sync::once_lock::OnceLock;
+
+/// Number of command slots for the Bluetooth driver
```
ROMemories:
```suggestion
/// Number of command slots for the Bluetooth driver.
```

src/ariel-os-hal/Cargo.toml
```diff
@@ -67,7 +67,13 @@ usb = [
...
-ble = ["dep:bt-hci", "dep:embedded-io", "dep:trouble-host", "ariel-os-nrf/ble"]
+ble = [
+  "dep:bt-hci",
+  "dep:embedded-io",
+  "dep:trouble-host",
+  "ariel-os-rp/ble-cyw43",
+  "ariel-os-nrf/ble"
+]
 ble-peripheral = ["ariel-os-nrf/ble-peripheral"]
 ble-central = ["ariel-os-nrf/ble-central"]
```
ROMemories:
```suggestion
ble-peripheral = ["ariel-os-nrf/ble-peripheral", "ariel-os-rp/ble-peripheral"]
ble-central = ["ariel-os-nrf/ble-central", "ariel-os-rp/ble-central"]
```

src/ariel-os-embassy/src/lib.rs
```diff
@@ -313,7 +312,9 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
         spawner.spawn(usb::usb_task(usb)).unwrap();
     }
 
-    #[cfg(feature = "wifi-cyw43")]
+    #[cfg(all(feature = "ble", context = "rp"))]
+    let (device, control) = hal::cyw43::device(&mut peripherals, &spawner, config).await;
+    #[cfg(all(feature = "wifi-cyw43", not(feature = "ble")))]
```
ROMemories: These `cfg`s should be rewritten to only depend on `_cyw43`, `ble-cyw43`, and `wifi-cyw43`.

ROMemories (commented):
Can you make sure the now-landed examples compile?

src/ariel-os-embassy/src/lib.rs
jkr-ariel: I guess the best way to do that would be to introduce a `ble-cyw43` laze module, but I don't see how to use it (in the same way that `wifi-cyw43` is used when the user selects the `network` module) when the user selects the more general `ble` laze module in their application.

laze-project.yml
```diff
@@ -1162,6 +1162,20 @@ modules:
...
+    selects:
+      - hwrng
+    env:
+      global:
+        FEATURES:
+          - ariel-os/ble-cyw43
+
+  - name: has_ble_cyw43
+    selects:
+      - ble-cyw43
```
ROMemories: Seems like that's inverted: `ble-cyw43` should select `has_ble_cyw43` instead, like `wifi-cyw43` does.

ROMemories:
Can you make sure the Build workflow passes before I review?

examples/ble-advertiser/laze.yml
```diff
@@ -1,4 +1,8 @@
 apps:
   - name: ble-advertiser
+    env:
+      global:
+        executor_stacksize_required:
+          - "32768"
```
nponsard: I'm a bit concerned that this would prevent the example from running on some MCUs with 32K or less RAM. A nrf52832 variant has only 32K of RAM and BLE capabilities.

jkr-ariel: Would reducing it to 16K be low enough?

src/ariel-os/Cargo.toml
```diff
@@ -143,6 +143,8 @@ eth-stm32 = ["ariel-os-embassy/eth-stm32"]
 
 # ## Bluetooth support
 ble = ["ariel-os-embassy/ble"]
+# Selects Wi-Fi (with the CYW43 chip).
```
```suggestion
# Selects BLE (with the CYW43 chip).
```

src/ariel-os-rp/Cargo.toml
```diff
@@ -67,22 +72,39 @@ storage = []
...
+  "hwrng",
+  "ariel-os-random/csprng",
+  "ble",
+]
+ble-peripheral = ["ble-cyw43"]
+ble-central = ["ble-cyw43"]
+
 ## Enables defmt support.
-defmt = ["dep:defmt", "embassy-rp/defmt"]
+defmt = ["dep:defmt", "cyw43/defmt", "embassy-rp/defmt"]
```
ROMemories: The question mark is required because `cyw43` is an optional dependency, otherwise [the `cyw43` *dependency* will get enabled](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) when `defmt` is enabled: 

```suggestion
defmt = ["dep:defmt", "cyw43?/defmt", "embassy-rp/defmt"]
```

src/ariel-os-rp/Cargo.toml
```diff
@@ -67,22 +72,39 @@ storage = []
...
 
+## Enables Bluetooth Low Energy support.
+ble = ["ariel-os-embassy-common/ble"]
+
+ble-cyw43 = [
+  "dep:trouble-host",
+  "dep:bt-hci",
+  "dep:embassy-sync",
+  "_cyw43",
+  "cyw43/bluetooth",
```
ROMemories: Same as below, even though in this case `_cyw43` would enable the `cyw43` dependency anyway:

```suggestion
  "cyw43?/bluetooth",
```

src/ariel-os-rp/Cargo.toml
```diff
@@ -67,22 +72,39 @@ storage = []
...
+  "hwrng",
```
ROMemories: I may be wrong, but I don't think we need to enable this here; it *has* to be enabled at a higher-level anyhow, so that the HWRNG gets seeded, and enabling this here could hide this not working:

```suggestion
```

src/ariel-os-hal/Cargo.toml
```diff
@@ -67,9 +67,11 @@ usb = [
   "ariel-os-stm32/usb",
 ]
 
-ble = ["dep:bt-hci", "dep:embedded-io", "dep:trouble-host", "ariel-os-nrf/ble"]
-ble-peripheral = ["ariel-os-nrf/ble-peripheral"]
-ble-central = ["ariel-os-nrf/ble-central"]
+ble = ["_ble", "ariel-os-nrf/ble"]
+ble-cyw43 = ["_ble", "ariel-os-rp/ble-cyw43"]
```
ROMemories: I'm not sure introducing `_ble` is needed here, why not just enable `ble` though `ble-cyw43`?
```suggestion
ble-cyw43 = ["ble", "ariel-os-rp/ble-cyw43"]
```

examples/ble-advertiser/laze.yml
nponsard: Probably, I don't have the hardware to confirm it.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -229,10 +229,9 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     }
 
     #[cfg(feature = "ble")]
-    {
-        let config = ble::config();
-        hal::ble::driver(ble_peripherals, spawner, config);
-    }
+    let config = ble::config();
```
ROMemories: As that variable is not so local anymore, it seems best to rename it:

```suggestion
    let ble_config = ble::config();
```

src/ariel-os-embassy/Cargo.toml
```diff
@@ -118,15 +118,16 @@ wifi-esp = ["ariel-os-hal/wifi-esp", "net", "wifi"]
 eth = []
 eth-stm32 = ["ariel-os-hal/eth-stm32", "net", "eth"]
 
-ble = [
-  "ariel-os-hal/ble",
+ble = ["_ble", "ariel-os-hal/ble"]
+ble-cyw43 = ["_ble", "ariel-os-hal/ble-cyw43"]
```
ROMemories: Same as below, I'm not sure introducing `_ble` is needed?

```suggestion
ble-cyw43 = ["ble", "ariel-os-hal/ble-cyw43"]
```

src/ariel-os-embassy-common/src/executor_swi.rs
ROMemories, line 1: I would prefer this change be moved to a separate PR.

laze-project.yml
```diff
@@ -1162,6 +1162,21 @@ modules:
         RUSTFLAGS:
           - --cfg capability=\"hw/ble\"
 
+  - name: ble-cyw43
+    selects:
+      - hwrng
+      - has_ble_cyw43
```
```suggestion
      - has_ble_cyw43
      - hwrng
```

examples/ble-advertiser/laze.yml
ROMemories: > Probably, I don't have the hardware to confirm it.

That can be tested on the `nrf52dk` on IoT-LAB.

nponsard: I wanted to be able to test on an nRF52832-QFAB, with 32K of RAM. The nrf52dk has a nRF52832-QFAA with 64K of RAM and the examples were working on it with the 32K stack size.

ROMemories:
Can you rebase on main, autosquash, and make sure CI passes?

Edit: when this is done, can you also list what testing you did?

jkr-ariel:
> Edit: when this is done, can you also list what testing you did?

The testing was done by running the `ble-advertiser` and `ble-scanner` examples on both a RP Pico W and a RP Pico 2 W.

examples/ble-advertiser/laze.yml
```diff
@@ -1,4 +1,8 @@
 apps:
   - name: ble-advertiser
+    env:
+      global:
+        executor_stacksize_required:
+          - "16384"
```
ROMemories, line 6: Can you test on IoT-LAB that this actually works on the smallest nRF we support? Same for the `ble-scanner` example.

jkr-ariel: The `ble-advertiser` and `ble-scanner` examples ran successfully on both a nrf52dk and a nrf52840dk on IoT-Lab.

ROMemories (approved):

- BLE/Wi-Fi working on RP+CYW43
  - Pico W
    - ✅ ble-advertiser
    - ✅ ble-scanner
    - ✅ http-client
  - Pico 2 W
    - ✅ ble-advertiser
    - ✅ ble-scanner
    - ✅ http-client
- BLE still working on nrf52840dk
  - ✅ ble-advertiser
  - ✅ ble-scanner
- BLE still working on bbc-microbit-v2
  - ✅ ble-advertiser
  - ✅ ble-scanner

PR #1096, opened 2025-06-16
feat(coapcore): facilitate implementation of traits in out-of-tree crates

KiyoLelou10:
Changed credentialErrorDetail from pub(crate) struct to pub struct and then changed
pub use credentials::CredentialError                    ->
 pub use credentials::{CredentialError, CredentialErrorDetail as CredentialErrorKind};

### DIFF
--- a/src/lib/coapcore/src/ace.rs
+++ b/src/lib/coapcore/src/ace.rs
@@ -59,6 +59,10 @@ type UnprotectedAuthzInfoPost<'a> = AceCbor<'a>;
 /// registry](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters).
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[derive(minicbor::Decode, Debug)]
+#[allow(
+    missing_docs,
+    reason = "Fields correspond 1:1 to the domain items of the same name"
+)]
 #[cbor(map)]
 #[non_exhaustive]
 pub struct HeaderMap<'a> {
@@ -233,6 +237,10 @@ struct SigStructureForSignature1<'a> {
 #[allow(
     dead_code,
     reason = "Presence of the item makes CBOR derive tolerate the item"
+)]
+#[allow(
+    missing_docs,
+    reason = "Fields correspond 1:1 to the domain items of the same name"
 )]
 #[cbor(map)]
 #[non_exhaustive]


chrysn:
Thanks, LGTM, awaiting CI checks. CI will complain about the commit messages (they appear to be auto-generated by some IDE?), but IIRC we allow squashing at commit time (as would be suitable here).

chrysn:
For context, this came out of @KiyoLelou10 using the API to implement an own storage -- where he found that the token side of things is not really implementable, as the CredentialError type can be constructed only from minicbor errors, but nothing else. This didn't hit previous users because they either used the shipped minimal config implementation, or (in Ariel OS's case) didn't add token support yet.

src/lib/coapcore/src/lib.rs
```diff
@@ -81,4 +81,4 @@ mod seccontext;
 pub use seccontext::*;
 
 mod error;
-pub use error::CredentialError;
+pub use credentials::{CredentialError, CredentialErrorDetail as CredentialErrorKind};
```
chrysn:
```suggestion
pub use error::{CredentialError, CredentialErrorDetail as CredentialErrorKind};
```

This should still work on the original position -- it will also work through credentials, but it's a re-export there.

KiyoLelou10:
Changed to error

src/lib/coapcore/src/error.rs
```diff
@@ -12,7 +12,7 @@ pub struct CredentialError {
 
 #[derive(Debug, Copy, Clone)]
 #[non_exhaustive]
```
chrysn, line 19:
```suggestion
/// Classification of a [`CredentialError`].
///
/// The variants in here are mainly used for debug output, and all signify *that*
/// the processing of the token was not successful. This type can be used to
/// construct a [`CredentialError`].
#[derive(Debug, Copy, Clone)]
#[non_exhaustive]
```

CI complained about missing documentation; let's see if there's more on the next run.

chrysn:
Squashed as mentioned in https://github.com/ariel-os/ariel-os/pull/1096#issuecomment-2976590732. I think this should run through with no further trouble. Thanks for providing those changes!

chrysn:
A brief note on future compatibility: This is making a few fields pub that might evolve into have more options (eg. HeaderMap.alg might grow a str variant) or become stricter (prep-parsed scopes). That's OK -- coapcore does not need to go to 1.0, and we don't have the experience with using those types yet, from which we'd create a more sustainable builder/accessor pattern.

(Also, this might be moved out of coapcore sooner or later anyway as the COSE parts can be reused for SUIT).

PR #1097, opened 2025-06-16
feat(sched): Add support for FPU on Cortex-M

Ollrogge:
# Description
This PR adds hard float support in the scheduler on Cortex-M devices that have a FPU.

The changes were tested for both `thumbv8m.main-none-eabihf` (cortex-m33f) and `thumbv7em-none-eabihf` (cortex-m4f). I did not test on cortex-m7f as I dont have a device using this processor.

The second commit is just a test application I used to verify that everything works. It can be removed if wanted.

## Issues/PRs references
Addresses #939 

## Changelog entry

(Cortex-M) Hard floats are now supported on this architecture: applications are now compiled with the `eabihf` variant, and floating point registers are now saved and restored by the preemptive scheduler as necessary.

## Change checklist

### DIFF
--- a/tests/threading-fpu/Cargo.toml
+++ b/tests/threading-fpu/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "threading-fpu"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os" }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
--- a/tests/threading-fpu/src/main.rs
+++ b/tests/threading-fpu/src/main.rs
@@ -1,0 +1,41 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::{debug::log::info, thread};
+
+#[ariel_os::thread(autostart)]
+fn thread0() {
+    info!("Hello from thread 0");
+    let mut a: f32 = 1.111;
+    let mut b = 2.222;
+    let mut test = 0.0;
+    for _ in 0..10 {
+        a += 1.1234;
+        b += 2.5678;
+
+        test += a + b;
+        thread::yield_same();
+    }
+
+    assert_eq!(test, 236.34601_f32);
+    info!("Thread 0 success");
+}
+
+#[ariel_os::thread(autostart)]
+fn thread1() {
+    info!("Hello from thread 1");
+    let mut a: f32 = 3.333;
+    let mut b = 4.444;
+    let mut test = 0.0;
+    for _ in 0..10 {
+        a += 3.4321;
+        b += 4.8765;
+
+        test += a * b;
+
+        thread::yield_same();
+    }
+
+    assert_eq!(test, 8324.532_f32);
+    info!("Thread 1 success");
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5271,6 +5271,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "threading-fpu"
+version = "0.0.0"
+dependencies = [
+ "ariel-os",
+ "ariel-os-boards",
+]
+
+[[package]]
 name = "threading-lock"
 version = "0.0.0"
 dependencies = [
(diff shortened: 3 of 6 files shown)


therealprof:
Sweet!

tests/threading-fpu/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use ariel_os::thread;
+
+#[ariel_os::thread(autostart)]
+fn thread0() {
+    info!("Hello from thread 0");
+    let mut a: f32 = 1.111;
+    let mut b = 2.222;
+    let mut test = 0.0;
+    for _ in 0..10 {
```
kaspar030, line 12: Do these tests need some [blackbox](https://doc.rust-lang.org/beta/core/hint/fn.black_box.html) hint? Seems to me the compiler might just pre-calculate the result.

Ollrogge: Looking at the disassembly I do see the loop so I dont think `blackbox` is required

tests/threading-fpu/src/main.rs
```diff
@@ -0,0 +1,45 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::debug::log::*;
+use ariel_os::debug::{ExitCode, exit, log::info};
+use ariel_os::gpio::{Level, Output};
+
+use ariel_os::thread;
```
ROMemories:
```suggestion
use ariel_os::{
    debug::{ExitCode, exit, log::info},
    thread,
};
```

tests/threading-fpu/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "threading-fpu"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["time"] }
```
ROMemories:
```suggestion
license.workspace = true
publish = false

[dependencies]
ariel-os = { path = "../../src/ariel-os" }
```

tests/threading-fpu/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "threading-fpu"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os" }
```
ROMemories, line 8: Even though `ariel-os-boards` currently does nothing and that example therefore works without it, we do need to keep it as a dependency for now, for not making that a breaking change later when it actually becomes functionally required:

```suggestion
ariel-os = { path = "../../src/ariel-os" }
ariel-os-boards = { path = "../../src/ariel-os-boards" }
```

src/ariel-os-threads/src/arch/cortex_m.rs
```diff
@@ -93,7 +107,9 @@ impl Arch for Cpu {
 }
 
 #[cfg(any(armv7m, armv8m))]
-global_asm!(
+macro_rules! define_pendsv_without_fpu {
```
kaspar030, line 110: I'm getting this when compiling *with* eabihf:

```
warning: unused macro definition: `define_pendsv_without_fpu`
   --> src/ariel-os-threads/src/arch/cortex_m.rs:110:14
    |                                                                          
110 | macro_rules! define_pendsv_without_fpu {                                                                                                                 
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^                         
    |   
    = note: `#[warn(unused_macros)]` on by default
```

I guess that's caused by both being defined in the eabihf case [here](https://github.com/ariel-os/ariel-os/pull/1097/files#diff-fa3a8522998e459b9038844dc42126e08d698cf0b11a22c82887d561ffc9d67cR580-R581).

Please either allow the unused macro, or adjust the cfg to sth like `all(any(armv7m, armv8m), not(any(armv7m_eabihf, armv8m,noeabihf)))`.

tests/threading-fpu/src/main.rs
kaspar030: ok!

kaspar030 (changes requested):
Good stuff! Code looks fine.

I tried breaking it, without success. :slightly_smiling_face:
Tested on nrf5340dk and rpi-pico2 (with both cores).

Just one warning to be fixed...

kaspar030 (approved):

PR #1098, opened 2025-06-17
fix(rp235x): fix non-deterministic panics at startup when using the TRNG

ROMemories:
# Description

Fixes a non-deterministic issue when using the TRNG on an RP235x, introduced when adding support for that TRNG in #1077. The issue was present upstream, but fixed since our last fork update. This backports the fix from https://github.com/embassy-rs/embassy/pull/4139 in our Embassy fork (one single commit), and use that new fork branch for `embassy-rp`.

There were some conflicts when cherry-picking the fix, as Embassy has reworked how they handle peripherals in the meantime (`Peripheral` → `Peri`).

## How to review this PR

Without this fix, the `random` example *sometimes* panics at startup on rpi-pico2(-w), with the following error (happens abount once every three runs in my testing, YMMV):

```
[ERROR] panicked at 'RNG not busy, but ehr is not valid!'
```

With this fix, that example does not panic anymore (tested successfully about two dozens of times).

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1786,7 +1786,7 @@ dependencies = [
 [[package]]
 name = "embassy-rp"
 version = "0.4.0"
-source = "git+https://github.com/ariel-os/embassy?branch=embassy-rp-v0.4.0%2Bariel-os#dab783fcee77ed024728de37a2ea4cb26de7d01c"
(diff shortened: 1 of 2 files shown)


kaspar030 (approved):
LGTM, trusting your testing.

PR #1088, opened 2025-06-10
docs(ble): BLE examples

nponsard:
# Description

This PR adds examples for using the Ariel OS BLE stack.

## Issues/PRs references

See also #560.
- Depends on #1069.
- Depends on #1074.
- Depends on #1118.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1100, opened 2025-06-18
feat(ble): support BLE on nrf5340 network core

nponsard:
# Description

This PR adds BLE support for the nrf5340 MCU, the BLE stack needs to run on the network core.

To avoid the need to upgrade embassy, I made a fork in [ariel-os/nrf-sdc, branch nrf-sdc-nrf53-backport](https://github.com/ariel-os/nrf-sdc/tree/nrf-sdc-nrf53-backport) where I selected the commits relevant to the support of the nRF5340 and adapted it to the previous embassy peripheral model.

## Issues/PRs references

- Depends on #1072
- Depends on #1074 
- Depends on #1102
- Linked to #560   

## Open Questions

How does the review process for a fork work ?

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,7 @@ serde = { version = "1.0.197", default-features = false }
 static_cell = { version = "2.1.1", default-features = false }
 trouble-host = "0.1"
 bt-hci = { version = "0.2.0" }
-nrf-sdc = { git = "https://github.com/alexmoon/nrf-sdc.git", rev = "551a95436e999b4290b4a33383aa3d6747b63dd9" }
+nrf-sdc = { git = "https://github.com/ariel-os/nrf-sdc.git", rev = "8ce4780b691ec6850b05a0382cef395846b67e02" }
 
 [workspace.lints.rust]
 # rustc lints are documented here: https://doc.rust-lang.org/rustc/lints/listing/index.html
--- a/src/ariel-os-nrf/src/ble.rs
+++ b/src/ariel-os-nrf/src/ble.rs
@@ -23,11 +23,34 @@ static RNG: StaticCell<ariel_os_random::CryptoRngSend> = StaticCell::new();
 // During testing central mode needed 2912 bytes, peripheral mode needed 1448 bytes.
 // Multirole (central + peripheral) needed 6080 bytes. Allocate more here if using extended features.
 
-#[cfg(all(feature = "ble-peripheral", not(feature = "ble-central")))]
+#[cfg(all(
+    context = "nrf52",
+    feature = "ble-peripheral",
+    not(feature = "ble-central")
+))]
 const SDC_MEM_SIZE: usize = 2880;
-#[cfg(all(feature = "ble-central", not(feature = "ble-peripheral")))]
+#[cfg(all(
+    context = "nrf52",
+    feature = "ble-central",
+    not(feature = "ble-peripheral")
+))]
 const SDC_MEM_SIZE: usize = 2912;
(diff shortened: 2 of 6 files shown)


nponsard:
Tested examples in #1088, scan and advertise works on the thingy91x (nrf5340 network core).

I noticed a weird behavior where my phone (pixel 9 pro, android 16, nrf-connect & ble radar), could not detect the advertisement packets when it could detect other boards (nrf52dk, nrf52840dk) just fine.
I tried with another android phone running nrf-connect and the thingy91x was detected correctly.

Ollrogge:
Why are you trying to avoid upgrading embassy ?

kaspar030:
> Why are you trying to avoid upgrading embassy ?

Because we have to apply the right subset of patches to individual embassy crate branches. Quite a hassle. :)

src/ariel-os-nrf/src/ble.rs
```diff
@@ -23,11 +23,34 @@ static RNG: StaticCell<ariel_os_random::CryptoRngSend> = StaticCell::new();
...
+))]
+const SDC_MEM_SIZE: usize = 4768;
+#[cfg(all(
+    context = "nrf53",
+    feature = "ble-central",
+    not(feature = "ble-peripheral")
+))]
+const SDC_MEM_SIZE: usize = 4904;
+#[cfg(all(context = "nrf53", feature = "ble-peripheral", feature = "ble-central"))]
 const SDC_MEM_SIZE: usize = 6080;
```
nponsard, line 54: This is starting to get a bit long, would an universal value be better ? It can waste memory though.

ROMemories: Would it be possible to instead compute that value based on the size of the values that are stored into the buffer, using `core::mem::size_of` (which usable in `const` contexts)? Not sure whether the values are named Rust types.

ROMemories: We can do this in a follow-up.

ROMemories (approved):
Successfully tested the `ble-advertizer` and `ble-scanner` examples on an nRF5340-DK.

Feel free to smash the green button!

Cargo.toml
```diff
@@ -150,7 +150,7 @@ serde = { version = "1.0.197", default-features = false }
 static_cell = { version = "2.1.1", default-features = false }
 trouble-host = "0.1"
 bt-hci = { version = "0.2.0" }
-nrf-sdc = { git = "https://github.com/alexmoon/nrf-sdc.git", rev = "551a95436e999b4290b4a33383aa3d6747b63dd9" }
+nrf-sdc = { git = "https://github.com/ariel-os/nrf-sdc.git", rev = "8ce4780b691ec6850b05a0382cef395846b67e02" }
```
ROMemories, line 153: We'll need to reconsider this before we make the BLE support official so hopefully we don't have to somewhat commit to that fork existing.

ROMemories:
> How does the review process for a fork work ?

I had a look at the version from the lock file, it seemed fine :slightly_smiling_face:

ROMemories:
I think you'll need to rebase on main to account for the newly introduced `check-labels` CI check.

Cargo.toml
nponsard: We can wait for the embassy upgrade in Ariel, then we won't need the fork.

src/ariel-os-nrf/src/ble.rs
nponsard: I looked in nrf-sdc, the needed memory is returned by `raw::sdc_cfg_set()` from the nrf binary that can't be determined at build time.

PR #1102, opened 2025-06-18
feat(nrf53-net): add hwrng support

nponsard:
# Description

This adds hwrng support for the nrf5340-net core. I don't have the hardware to test it out.
It seems that the nrf5340 app core doesn't have a (regular) random number generator available.

No errors when building with `laze build -g -b nrf5340dk-net` and `laze build -g -b nrf5340dk`   

## Issues/PRs references

~~Depends on #1072~~

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -271,8 +271,9 @@ contexts:
     selects:
       - cortex-m33
     provides:
-    # Currently hard-faults.
-    # - has_storage_support
+      - has_hwrng
+      # Currently hard-faults.
+      # - has_storage_support
     disables:
       # USB peripheral available only on app core
       - usb
--- a/src/ariel-os-nrf/src/hwrng.rs
+++ b/src/ariel-os-nrf/src/hwrng.rs
@@ -6,7 +6,7 @@ pub fn construct_rng(peripherals: &mut crate::OptionalPeripherals) {
     cfg_if::cfg_if! {
         // The union of all contexts that wind up in a construct_rng should be synchronized
         // with laze-project.yml's hwrng module.
-        if #[cfg(any(context = "nrf51", context = "nrf52"))] {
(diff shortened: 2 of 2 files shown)


ROMemories:
@kaspar030 Do we want to automatically send the HWRNG value through the IPC channel so that it can be used transparently from the application core?

nponsard:
Maybe there's a way to use the TRNG in the CryptoCell on the application core ?

kaspar030:
> Maybe there's a way to use the TRNG in the CryptoCell on the application core ?

That would be great!

kaspar030:
> @kaspar030 Do we want to automatically send the HWRNG value through the IPC channel so that it can be used transparently from the application core?

Also an option. I guess that opens a can of missing features:

1. laze support for building & flashing both cores
2. implement generic AMP IPC channel (traits), implement for nrf5340 peripheral
3. wire up net core hwrng to that
4. wire up our hwrng abstraction to the ipc provided one

Or did you mean something simpler? :slightly_smiling_face:

nponsard:
> > Maybe there's a way to use the TRNG in the CryptoCell on the application core ?
> 
> That would be great!

Though it seems that we won't have much documentation about it, there's a disclaimer in the documentation:

> The CRYPTOCELL subsystem is recommended for use with the libraries in the Nordic Semiconductor ASA SDK.

With more text saying that the register description may not be accurate.

ROMemories:
> Or did you mean something simpler? 🙂

I did mean exactly that :slightly_smiling_face:

---

Fine without that for now; needs some hardware testing before we merge this.

ROMemories (approved):
Successfully tested the `random` example on an nRF5340-DK, using the `nrf5340dk-net` builder. This also confirms that that firmware indeed runs on the network core, as the application core does not have such HWRNG.

PR #1105, opened 2025-06-20
fix(executor): change `executor_swi` macro to take executor as param

jkr-ariel:
# Description

This PR adds another parameter to the `executor_swi` macro to specify the executor to use. This change is needed to fix a clippy lint error in CI.

## Change checklist


kaspar030:
> This change is needed to fix a clippy lint error in CI.

What's the error? As this clutters our (internal but still) API, maybe an exception is preferable?

jkr-ariel:
The error is from the CI of [#1091](https://github.com/ariel-os/ariel-os/actions/runs/15708548665/job/44260187662?pr=1091)
```
Error: error: `crate` references the macro call's crate
  --> src/ariel-os-embassy-common/src/executor_swi.rs:39:22
   |
39 |             unsafe { crate::EXECUTOR.on_interrupt() }
   |                      ^^^^^ help: to reference the macro definition's crate, use: `$crate`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#crate_in_macro_def
   = note: `-D clippy::crate-in-macro-def` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::crate_in_macro_def)]`
   ```

**Comment by kaspar030** [comment]
Ok, in this case I think `#[allow(clippy::crate_in_macro_def)]` it is - `EXECUTOR` is expected to be exported at crate root level, and used by `ariel-os-embassy`. No need to clutter API.

File: src/ariel-os-embassy-common/src/executor_swi.rs [hunk 1]
```diff
@@ -18,15 +18,15 @@
...
+///     unsafe { EXECUTOR.on_interrupt() }
 /// }
 /// ```
 ///
 /// Note: this expects the `interrupt` to be present (e.g., "used") and that it contains the ISR
 /// type.
 #[macro_export]
 macro_rules! executor_swi {
-    ($swi:ident) => {
+    ($exe:ident, $swi:ident) => {
```
ROMemories: The type needs to be more generic, otherwise it wouldn't be possible to pass `crate::EXECUTOR` for instance; I think `expr` is the right one here:

```suggestion
    ($exe:expr, $swi:ident) => {
```

ROMemories (changes requested):
Good job updating the documentation! There is another occurrence of `executor_swi!` that needs to be updated in the `ariel-os-stm32` crate (in the build script), and then it should be good to go after some hw testing.

ROMemories:
> Ok, in this case I think `#[allow(clippy::crate_in_macro_def)]` it is - `EXECUTOR` is expected to be exported at crate root level, and used by `ariel-os-embassy`. No need to clutter API.

@kaspar030 This doesn't "clutter the API", this makes the macro more hygienic, especially as that macro is used across crate boundaries.

kaspar030:
> @kaspar030 This doesn't "clutter the API", this makes the macro more hygienic, especially as that macro is used across crate boundaries.

There's four usage sites, with potentially more to come. They all now have the identical extra `EXECUTOR` first argument. Maybe "clutter the API" is the wrong wording, but this is *not* an improvement.

clippy exceptions exist for a reason. IMO this is one...

ROMemories:
> but this is _not_ an improvement.

If macros were actually hygienic, that wouldn't even work in the first place (especially not across crates).
That also makes the macro definition easier to read and understand, as currently you're left wondering where that `crate::EXECUTOR` thing is coming from, and whether that's actually a typo for `$crate::EXECUTOR` (it's not).

kaspar030:
> That also makes the macro definition easier to read and understand, as currently you're left wondering where that `crate::EXECUTOR` thing is coming from, and whether that's actually a typo for `$crate::EXECUTOR` (it's not).

The macro exists so we don't have to duplicate the ISR handler definitions.
It is pretty simple, and the doc comment pretty much explains what the macro does.
If this is still too hard to understand, improve the doc comment above.

This is "across crates" b/c we had to split out `ariel-os-embasssy-common`. Unfortunately there's no such thing as a "crate group" so visibility can be reduced to specific downstream crates.

With `EXECUTOR` in the arg list, one has to wonder where *that* comes from at every call site.
That is again not an improvement.

kaspar030:
May I suggest to fix the linter error first (by adding the `allow` statement) to unblock dependent builds, and keep this discussion orthogonal?

ROMemories:
> With `EXECUTOR` in the arg list, one has to wonder where _that_ comes from at every call site.

Except it would be part of the same crate, so it'd be trivial to find it.

---

> May I suggest to fix the linter error first (by adding the allow statement) to unblock dependent builds, and keep this discussion orthogonal?

Anyway, fiiine, but make it an `expect` at least, and provide a `reason` to explain `crate::EXECUTOR` comes from downstream crates.

PR #1109, opened 2025-06-24
feat(build): pass through task arguments where useful

kaspar030:
# Description

This wires up passing through task arguments in our laze tasks.

This worked before laze 0.1.36 b/c any extra args were implicitly passed through, which was an undocumented laze feature. Current laze 0.1.37 now has proper (but still undocumented :innocent:) handlilng for this, allowing multi-cmd tasks to only pass through the extra args where desired.

This requires updating laze to at least version 0.1.37.

Try with e.g., `laze build -b nrf52840dk size`, then the same but appending `-A` to change the output mode of `size`.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1,4 +1,4 @@
-laze_required_version: 0.1.33
+laze_required_version: 0.1.37
 
 contexts:
   # base context that all other contexts inherit from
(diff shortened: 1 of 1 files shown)


ROMemories (approved):

```
laze -C examples/blinky build -b nrf52840dk size -A
laze -C examples/blinky build -b nrf52840dk tree -i heapless
```

Had also a look for possible other candidate tasks to add `$@` to, but didn't find any.

---

However, I'm not sure this needs to be a breaking change? We don't treat MSRV bumps of the Rust toolchain as such. If it really is, it would be good to use the [`BREAKING CHANGE` footer](https://www.conventionalcommits.org/en/v1.0.0/) to explain the reason, so it can be machine-parsed if needed.

kaspar030:
> However, I'm not sure this needs to be a breaking change?

Well, this now fails building unless laze 0.1.37 is installed. So it might need some action on the user's end.

> We don't treat MSRV bumps of the Rust toolchain as such.

Indeed ... Ok, I remove the `!` from the PR description and we keep the commit, saving the CI joules.

ROMemories:
> Indeed ... Ok, I remove the `!` from the PR description and we keep the commit, saving the CI joules.

You could have renamed the commit as well: as that doesn't modify the tree the Build workflow would just hit the cache and take less than a minute.

PR #1111, opened 2025-06-25
fix(coapcore): restore functionality of `log` feature

chrysn:
# Description

Logging in coapcore had a few defmt'isms in it and thus failed to build with the `log` feature.

Note that unlike most other places where we had these issues, coapcore goes through defmt-or-log rather than through Ariel's re-implementation thereof.

[edit:] Note that this is a stop-gap measure. Comments have been updated to a more proper solution, but right now, this fixes breakage at the acceptable cost of reducing prettiness of debug output everywhere.

## Issues/PRs references

See-Also: https://github.com/t-moe/defmt-or-log/issues/4 (which would catch this)
See-Also: https://github.com/ariel-os/ariel-os/pull/972 (which catches this in the rest of the code base)

CC @KiyoLelou10 who'll need that to use coapcore on RIOT.

## Change checklist


chrysn:
Closing: Overtaken by #1112 where upstream was super fast to accept https://github.com/t-moe/defmt-or-log/pull/8 and release.

PR #1116, opened 2025-07-02
feat(random): add RNG that can be shared between threads

nponsard:
# Description

Adds `crypto_rng_send()` and `fast_rng_send()` functions, this allows to get a random number generator that can be shared between threads, this is needed for the nRF BLE driver (SDC).

## Issues/PRs references

The need was discovered in #1074 

## Change checklist

### DIFF
--- a/src/ariel-os-random/src/lib.rs
+++ b/src/ariel-os-random/src/lib.rs
@@ -89,6 +89,31 @@ impl RngCore for FastRng {
     }
 }
 
+/// Same as [`FastRng`], but can be shared across threads and tasks.
+/// This is to differentiate from [`FastRng`] that could get optimized in a future version to use thread-local storage.
+/// Should be used only when sharing between threads is necessary (e.g. integrating with C code).
+#[doc(hidden)]
+pub struct FastRngSend {
+    inner: rand_pcg::Pcg32,
+}
+
+// Re-implementing the trait rather than Deref'ing into inner: This avoids leaking implementation
+// details to users who might otherwise come to depend on platform specifics of the FastRng.
+impl RngCore for FastRngSend {
+    fn next_u32(&mut self) -> u32 {
+        self.inner.next_u32()
+    }
+    fn next_u64(&mut self) -> u64 {
+        self.inner.next_u64()
+    }
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        self.inner.fill_bytes(dest);
+    }
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand_core::Error> {
+        self.inner.try_fill_bytes(dest)
+    }
+}
+
 /// The OS provided cryptographically secure random number generator.
 ///
 /// Such an RNG can be requested by any component, and will always be seeded appropriately.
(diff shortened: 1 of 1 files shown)


src/ariel-os-random/src/lib.rs
```diff
@@ -126,6 +150,43 @@ mod csprng {
...
+        fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand_core::Error> {
+            self.inner.try_fill_bytes(dest)
+        }
+    }
+
+    impl rand_core::CryptoRng for super::CryptoRngSend {}
+
+    /// Asserts that [`SelectedRngSend`] is [`CryptoRng`], justifying the implementation above.
+    trait _AssertCryptoRng: rand_core::CryptoRng {}
+    impl _AssertCryptoRng for SelectedRng {}
```
ROMemories:
```suggestion
```

Alternatively, this should instead assert `rand_chacha::ChaCha20Rng` (the inner type of `CryptoRngSend`).

src/ariel-os-random/src/lib.rs
```diff
@@ -89,6 +89,30 @@ impl RngCore for FastRng {
     }
 }
 
+/// Same as [`FastRng`], but can be shared across threads and tasks.
+/// This is less efficient and should only be used when necessary (e.g.) integrating with C code.
```
ROMemories: I don't think this is less efficient right now, it just precludes future evolution which could be more so.

ROMemories (changes requested):

src/ariel-os-random/src/lib.rs
nponsard: I don't really understand this trait assertion, it was in the module initializing `CryptoRng`, I copied it as-is.

ROMemories: It only makes sense to impl `CryptoRng` (the trait) on `CryptoRng` (the type) when `SelectedRng`, which is decoupled from that struct, also impls `CryptoRng`. This assert makes sure of that.

However in this send implementation we don't need it and it is redundant.

src/ariel-os-random/src/lib.rs
nponsard: I changed the documentation text to specify that it's a differentiation that will allow us to optimize the non-Send one

src/ariel-os-random/src/lib.rs
nponsard: Alright I removed it

ROMemories (approved):
All of these new items are hidden from the documentation, as they should.

Please autosquash!

src/ariel-os-random/src/lib.rs
chrysn: The squashed version has the original text again – is that intentional?

(GitHub's interface is being weird; it showed me that you changed it to something "so that we could optimize", but the latest squashed commit da3da176baf637087f5338dd4fb52e88cea8e6ad does have the "This is less efficient" version).

By the way, I like the original version; to address @ROMemories's comment, it could say "This could be less efficient", but retain the "and should only be used when (that property) is necessray".

nponsard: I forgot to change it in the struct doc... should be good now.
I also added the mention that it should be used only when `Send` is necessary.

PR #1118, opened 2025-07-02
fix(nrf-ble): restrict the `has_nrf_ble` laze module to `nrf`

ROMemories:
# Description

This makes sure the `has_nrf_ble` laze module introduced in #1074 can only be enabled on `nrf`, instead of being enabled by any laze module that merely selects it.

## Testing

When combined with #1088, this reduces the number of tasks ran by laze with:

```
laze -C examples/ble-scanner build -m
```

from 24 to 8.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1144,6 +1144,8 @@ modules:
 
   - name: has_nrf_ble
     help: Marks an incompatibility of nrf devices with the interrupt executor (MPSL crashing)
+    context:
+      - nrf52
     selects:
       - hwrng
     provides:


nponsard:
I got confused to why the ai-c3 board considered compatible with bluetooth. Isn't laze supposed to restrict to the contexts that provides `has_ble` and `has_nrf_ble` ?

ROMemories:
> Isn't laze supposed to restrict to the contexts that provides `has_ble` and `has_nrf_ble` ?

Thing is, laze basically works *top-down*, so what happens is that the dependency on `ble` selects `hw/ble`, which selects `has_ble`, which selects `has_nrf_ble`, as nothing prevents that. There is two solutions to this: either introducing the following module:

```yaml
- name: has_nrf_ble
  selects:
    - doc-only
```

where the `doc-only` module does *not* exist and then to *provide* that `has_nrf_ble` module in each laze context where it is indeed available (e.g., `nrf52`), or to use the `context` key in that module to restrict where it is available, which is less flexible but simpler in this case, so I went with that.

nponsard:
Thanks for the explanation. Wouldn't the version using `doc-only` be better for extending the support?

ROMemories:
> Wouldn't the version using `doc-only` be better for extending the support?

In this case we don't really need this because (a) we don't need to be able to allow more contexts *out-of-tree* and (b) it's just shorter and simpler to list the 2–3 contexts we'll have there.

nponsard:
Couldn't figure out how to use the `doc-only` trick in this situation, your solution seems simpler.

nponsard (approved):

PR #1119, opened 2025-07-02
fix(threads): fix second core stack initialization

kaspar030:
# Description

Multicore startup currently always crashes due to a double `take()` on a `ConstStaticCell`. This was probably introduced in #977 already.

## Change checklist


ROMemories (approved):
Successfully tested the `threading` example on a rpi-pico-w.

Before this, it failed with:

```
`ConstStaticCell` is already taken, it can't be taken twice
```

PR #1120, opened 2025-07-02
fix(examples): fix `threading-event` priorities for multi-core

kaspar030:
# Description

The current priorities expect one of the threads with priority 0 to be launched last. On multicore with two idle threads, that thread actually never gets scheduled.

This PR increments all test thread priorities by one, mitigating the issue.

(Let's find a more general solution elsewhere.)

## Change checklist

### DIFF
--- a/examples/threading-event/src/main.rs
+++ b/examples/threading-event/src/main.rs
@@ -22,27 +22,27 @@ fn waiter() {
     }
 }
 
-#[ariel_os::thread(autostart, priority = 0)]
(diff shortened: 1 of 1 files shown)


kaspar030:
For testing this on actual multi-core, this depends on #1119.

ROMemories (approved):
Successfully tested the `threading-event` example on an rpi-pico-w (by merging with #1119). The example does *now* terminate properly. Also tested on an espressif-esp32-c6-devkitc-1 even if it worked properly before this as well (as it is single-core).

PR #1121, opened 2025-07-02
fix(doc): Fix minor spelling mistakes

Ollrogge:
# Description
Small fixes to spelling in various places.

## Open Questions
Maybe it makes sense to add something like `codespell` to CI similar to how RIOT [does](https://github.com/RIOT-OS/RIOT/blob/master/dist/tools/codespell/check.sh) it ?

## Change checklist

### DIFF
--- a/src/ariel-os-rp/src/cyw43/firmware/README.md
+++ b/src/ariel-os-rp/src/cyw43/firmware/README.md
@@ -6,4 +6,4 @@ Licensed under the [Infineon Permissive Binary License](./LICENSE-permissive-bin
 
 ## Changelog
 
-* 2023-07-28: synced with `ad3bad0` - Update 43439 fw from 7.95.55 ot 7.95.62
+* 2023-07-28: synced with `ad3bad0` - Update 43439 fw from 7.95.55 to 7.95.62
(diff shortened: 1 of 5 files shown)


src/ariel-os-rp/src/cyw43/firmware/README.md
```diff
@@ -6,4 +6,4 @@ Licensed under the [Infineon Permissive Binary License](./LICENSE-permissive-bin
 
 ## Changelog
 
-* 2023-07-28: synced with `ad3bad0` - Update 43439 fw from 7.95.55 ot 7.95.62
+* 2023-07-28: synced with `ad3bad0` - Update 43439 fw from 7.95.55 to 7.95.62
```
ROMemories, line 9: That one shouldn't have been fixed as this file is copied verbatim from upstream. No need to undo this however, as #1091 will update this file anyway.

ROMemories:
> Maybe it makes sense to add something like `codespell` to CI similar to how RIOT [does](https://github.com/RIOT-OS/RIOT/blob/master/dist/tools/codespell/check.sh) it ?

We do already have [typos](https://github.com/crate-ci/typos) in CI, which is designed to minimize false positives, but clearly it overlooked these. I'll try to report them upstream so they don't happen again in the future.

PR #1126, opened 2025-07-04
fix(stm32): add usb-ethernet enumeration workaround

kaspar030:
# Description

This PR adds a workaround for https://github.com/embassy-rs/embassy/issues/2376.
Without it, usb-ethernet doesn't enumerate on stm32 synopsis USB ip.

This is not perfect, e.g., after plugging/reconnecting the USB cable, this won't re-enumerate again. Thus I don't mark `usb-ethernet` as supported on the affected boards.

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -565,10 +565,6 @@ contexts:
     parent: stm32
     selects:
       - cortex-m33f
-    disables:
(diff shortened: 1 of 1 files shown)


ROMemories (approved):

```
laze -C examples/udp-echo build -s usb-ethernet -b st-nucleo-h755zi-q run
```

It does now show up after 3 s as expected.

PR #1127, opened 2025-07-07
feat(esp): add BLE support for ESP MCUs

jkr-ariel:
# Description

Add BLE capabilities for supported ESP devices by exposing the [trouble](https://github.com/embassy-rs/trouble) API.

## Issues/PRs references

- Depends on https://github.com/ariel-os/ariel-os/pull/1069
- Linked to https://github.com/ariel-os/ariel-os/issues/560
- Depends on #1328

## Open Questions

It seems like there is an issue when running the ble examples where `esp_wifi::init()` never returns...

## Change checklist

### DIFF
--- a/src/ariel-os-esp/src/scheduler.rs
+++ b/src/ariel-os-esp/src/scheduler.rs
@@ -97,6 +97,13 @@ impl SchedulerImplementation for ArielScheduler {
     ) -> ThreadPtr {
         trace!("task_create()");
 
+        // TODO: report this bug, the task `btController` doesn't respect `max_task_priority`
+        let priority = if priority <= self.max_task_priority() {
+            priority
+        } else {
+            self.max_task_priority()
+        };
+
         // upstream uses 16b as minimum alignment on both architectures
         let stack_slice = alloc_aligned_leaked_buffer(task_stack_size, 16);
 


ROMemories:
Can you rebase on `main` as we've now landed the examples (and then update the lockfile to try to make the CI turn green)?

ROMemories:
Now that #1091 is in, can you rebase on main to simplify this PR a bit?

src/ariel-os-esp/Cargo.toml
```diff
@@ -110,15 +114,31 @@ log = ["esp-hal/log", "esp-hal-embassy/log", "esp-wifi?/log"]
...
+]
+ble-central = ["ble-esp"]
+ble-peripheral = ["ble-esp"]
+
 ## Enables Wi-Fi support.
 wifi = []
 
 ## Enables built-in Wi-Fi hardware.
-wifi-esp = ["dep:embassy-time", "dep:esp-alloc", "dep:esp-wifi", "wifi"]
+wifi-esp = ["_esp", "esp-wifi/wifi", "wifi"]
```
ROMemories:
```suggestion
wifi-esp = ["_esp", "esp-wifi?/wifi", "wifi"]
```

ROMemories:
Could you try to rebase this on top of #1328 to see where that gets us?

Cargo.lock
ROMemories, line 1: Can you make sure this only modifies the lockfile as necessary? This shouldn't be updating other crates.

jkr-ariel:
Trying to run the `ble-advertiser` or `ble-scanner` examples on a DFRobot FireBeetle 2 ESP32-C6 board results in `espflash` being stuck in an infinite loop:
```
$ laze -C examples/ble-advertiser build -b dfrobot-firebeetle2-esp32-c6 run
[...]
ESP-ROM:esp32c6-20220919
Build:Sep 19 2022
rst:0x15 (USB_UART_HPSYS),boot:0x1c (SPI_FAST_FLASH_BOOT)
Saved PC:0x4080053c
SPIWP:0xee
mode:DIO, clock div:2
load:0x4086c410,len:0xd48
load:0x4086e610,len:0x2d68
0x4086e610 - ariel_os_alloc::alloc::init_esp_alloc::HEAP
    at ??:??
load:0x40875720,len:0x1800
0x40875720 - ariel_os_alloc::alloc::init_esp_alloc::HEAP
    at ??:??
entry 0x4086c410
I (23) boot: ESP-IDF v5.1-beta1-378-gea5e0ff298-dirt 2nd stage bootloader
I (23) boot: compile time Jun  7 2023 08:02:08
I (24) boot: chip revision: v0.1
I (28) boot.esp32c6: SPI Speed      : 40MHz
I (33) boot.esp32c6: SPI Mode       : DIO
I (37) boot.esp32c6: SPI Flash Size : 4MB
I (42) boot: Enabling RNG early entropy source...
I (48) boot: Partition Table:
I (51) boot: ## Label            Usage          Type ST Offset   Length
I (58) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (66) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (73) boot:  2 factory          factory app      00 00 00010000 003f0000
I (81) boot: End of partition table
I (85) esp_image: segment 0: paddr=00010020 vaddr=42000020 size=06e1ch ( 28188) map
I (99) esp_image: segment 1: paddr=00016e44 vaddr=40808000 size=00014h (    20) load
I (102) esp_image: segment 2: paddr=00016e60 vaddr=42006e60 size=4de5ch (319068) map
I (175) esp_image: segment 3: paddr=00064cc4 vaddr=40808014 size=08a50h ( 35408) load
E (184) esp_image: invalid segment length 0x7596
E (185) boot: Factory app partition is not bootable
E (185) boot: No bootable app partitions in the partition table
[...]
```

The `blinky` and `hello-world` examples can run just fine.

jkr-ariel:
> Trying to run the `ble-advertiser` or `ble-scanner` examples on a DFRobot FireBeetle 2 ESP32-C6 board results in `espflash` being stuck in an infinite loop:

Same behaviour on the ESP32-C6-DevKitC-1.

ROMemories:
> Trying to run the `ble-advertiser` […] examples on a DFRobot FireBeetle 2 ESP32-C6 board results in `espflash` being stuck in an infinite loop

I can indeed reproduce that on an espressif-esp32-c6-devkitc-1: it does flash but then seemingly furiously bootloops with the following errors:

```sh
E (183) esp_image: invalid segment length 0x7616
E (184) boot: Factory app partition is not bootable
E (184) boot: No bootable app partitions in the partition table
```

ROMemories:
The above error seems to stem from using an esp-flash old enough to not support the partitioning brought by #1328. Updating to espflash v4.3.0 does seem to fix that issue.

@kaspar030 Now I instead get the following panic:

```sh
panicked at [redacted]/ariel-os/src/ariel-os-runqueue/src/runqueue.rs:282:13:
index out of bounds: the len is 16 but the index is 29
```

kaspar030:
eps32-c6 (and -c3) is currently broken - something wrong with the ~~interrupt handling~~ task switching

src/ariel-os-esp/src/ble.rs
```diff
@@ -0,0 +1,75 @@
...
+        .get_mut_async()
+        .await
+        .expect("Stack needs to be taken from main executor")
+}
+
+register_scheduler_implementation!(static SCHEDULER: ArielScheduler = ArielScheduler{});
+register_wait_queue_implementation!(ArielWaitQueue);
+register_semaphore_implementation!(CompatSemaphore);
+register_timer_implementation!(CompatTimer);
+register_queue_implementation!(CompatQueue);
```
nponsard: This should be factored out so the macros are called once when BLE and WIFI are both enabled. 

I think the way to go should be to create a `radio.rs` module that gets included when either wifi or ble is needed.

boards/espressif-esp32-c6-devkitc-1.yaml
```diff
@@ -1,6 +1,8 @@
 boards:
   espressif-esp32-c6-devkitc-1:
     chip: esp32-c6-wroom-1
+    flags:
+      - has_ble_esp
```
nponsard: BLE should be available at a chip level, it is present on esp32c3, esp32c6, esp32, esp32s3 (only the esp32s2 doesn't have it in the MCUs we support)

src/ariel-os-esp/src/radio.rs
```diff
@@ -0,0 +1,15 @@
...
+    register_semaphore_implementation, register_timer_implementation,
+    register_wait_queue_implementation, semaphore::CompatSemaphore, timer::CompatTimer,
+};
+
+use crate::scheduler::ArielScheduler;
+use crate::wait_queue::ArielWaitQueue;
+
+register_scheduler_implementation!(static SCHEDULER: ArielScheduler = ArielScheduler{});
+register_wait_queue_implementation!(ArielWaitQueue);
+register_semaphore_implementation!(CompatSemaphore);
```
nponsard, line 14: This needs to use `crate::semaphore::CompatSemaphore`

jkr-ariel:
Tested by running the `ble-scanner` and `ble-advertiser` examples on the ESP32 C6 dekvit C1 board.

jkr-ariel:
The last commit implements a workaround that fixes a [bug](https://github.com/esp-rs/esp-hal/issues/5062) in esp-radio which doesn't respect `Scheduler::max_task_priority`.

src/ariel-os-esp/src/scheduler.rs
```diff
@@ -97,6 +97,13 @@ impl SchedulerImplementation for ArielScheduler {
     ) -> ThreadPtr {
         trace!("task_create()");
 
+        // TODO: report this bug, the task `btController` doesn't respect `max_task_priority`
```
ROMemories, line 100: It seems that this has already been [fixed upstream](https://github.com/esp-rs/esp-hal/issues/5062) after the report, but no need to change this comment.

src/ariel-os-esp/Cargo.toml
```diff
@@ -148,5 +173,7 @@ wifi-esp = [
 ## Enables the interrupt executor.
 executor-interrupt = ["embassy-executor/executor-interrupt"]
 
+_esp = ["dep:embassy-time", "dep:esp-alloc", "dep:esp-radio"]
+
```
ROMemories: This seems redundant with `ble-esp`, and doesn't seem used by another feature?
```suggestion
```

src/ariel-os-esp/Cargo.toml
```diff
@@ -130,6 +135,24 @@ log = ["esp-hal/log-04", "esp-radio-rtos-driver?/log-04", "esp-radio?/log-04"]
 # Enables USB support.
 usb = []
 
+## Enables Bluetooth support.
```
ROMemories: That's important because Bluetooth Classic is only supported on the ESP8266, which we don't support:
```suggestion
## Enables BLE support.
```

laze-project.yml
```diff
@@ -1639,6 +1644,27 @@ modules:
...
+      - alloc
+      - has_ble_esp
+      - hwrng
+    provides:
+      - has_ble
+    env:
+      global:
+        # esp-wifi needs a lot of ISR stack.
+        isr_stacksize_required: "32768"
+        # esp-wifi needs 72KiB
```
ROMemories: It's been renamed in the meantime:
```suggestion
        # esp-radio needs a lot of ISR stack.
        isr_stacksize_required: "32768"
        # esp-radio needs 72KiB
```

laze-project.yml
```diff
@@ -1639,6 +1644,27 @@ modules:
...
+        heapsize_required:
+          - $(72*1024)
```
ROMemories: To align this with `wifi-esp`:
```suggestion
          - ${esp_wifi_heapsize_required}
```

laze-project.yml
ROMemories, line 445: The ESP32-S2 indeed doesn't have BLE (it's Wi-Fi-only).

src/ariel-os-esp/src/ble.rs
```diff
@@ -0,0 +1,60 @@
...
+use embassy_sync::{
+    blocking_mutex::raw::CriticalSectionRawMutex, mutex::Mutex, once_lock::OnceLock,
+};
+use esp_radio::ble::controller::BleConnector;
+use static_cell::StaticCell;
+use trouble_host::prelude::DefaultPacketPool;
+
+use ariel_os_embassy_common::cell::SameExecutorCell;
+
+/// Number of command slots for the Bluetooth driver.
```
ROMemories:
```suggestion
/// Number of command slots for the BLE driver.
```

src/ariel-os-esp/src/wifi/esp_wifi.rs
```diff
@@ -1,22 +1,12 @@
-mod scheduler;
-mod semaphore;
-mod wait_queue;
-
 use ariel_os_debug::log::{debug, info};
 use core::ffi::c_void;
```
ROMemories: This seems now unused.
```suggestion
```

src/ariel-os-esp/src/wifi/esp_wifi.rs
```diff
@@ -1,22 +1,12 @@
...
 };
-use esp_radio_rtos_driver::{
-    queue::CompatQueue, register_queue_implementation, register_scheduler_implementation,
-    register_semaphore_implementation, register_timer_implementation,
-    register_wait_queue_implementation, timer::CompatTimer,
-};
 
-use scheduler::ArielScheduler;
-use wait_queue::ArielWaitQueue;
+use crate::semaphore;
```
ROMemories: This also seems unused now; I think these were expected by the macros below?
```suggestion
```

src/ariel-os-esp/src/ble.rs
```diff
@@ -0,0 +1,60 @@
...
+pub const SLOTS: usize = 10;
```
ROMemories, line 13: I'm not exactly sure what this does/restricts, but this can be improved in a follow-up (if there is even anything to do about it).

src/ariel-os-esp/src/ble.rs
```diff
@@ -0,0 +1,60 @@
...
+    DefaultPacketPool,
+>;
+
+pub(crate) static STACK: StaticCell<SameExecutorCell<BleStack>> = StaticCell::new();
+// The stack can effectively only be taken by a single application; once taken, the Option is None.
+pub(crate) static STACKREF: OnceLock<
+    Mutex<CriticalSectionRawMutex, Option<&'static mut SameExecutorCell<BleStack>>>,
+> = OnceLock::new();
+
+pub async fn init(
```
ROMemories, line 27: It seems to me this could be unified with nRF's by calling this `driver` and changing the parameter order, but we'll do this in a follow-up if it makes sense.

nponsard:
I added networking to the `ble-scanner` and `ble-advertiser` examples (merging `tcp-client` into them) and tested on `esp32c6` + `esp32s3`: it works !

I had to increase the interval to 2 seconds in `ScanConfig` for `ble-scanner` (keeping the window to 1 second) otherwise the antenna would be used 100% of the time by BLE.

PR #1130, opened 2025-07-09
docs(book): update the installation command for `espup`

ROMemories:
# Description

Using `--locked` for installing `espup` is now the [recommended way](https://github.com/esp-rs/espup/pull/488), and installation currently fails without it.

**Note:** this PR doubles as a test for the `ci-build:skip` PR label, so do *not* merge it for now.

## Issues/PRs references

Depends on #1135.

## Change checklist


ROMemories:
Now that #1135 is in, we can the Build workflow only took 3 min 30 s, which each job taking around 20 s at most.

PR #1131, opened 2025-07-09
refactor(threads): use pointer for thread function

kaspar030:
# Description

Use function pointers in API and move cast to register sized integer into arches.
Also, make `arg` an Option.

https://doc.rust-lang.org/stable/core/primitive.fn.html#casting-to-and-from-integers
https://doc.rust-lang.org/stable/core/ptr/index.html#provenance

## Issues/PRs references

#647 

## Change checklist

### DIFF
--- a/src/ariel-os-threads/src/arch/cortex_m.rs
+++ b/src/ariel-os-threads/src/arch/cortex_m.rs
@@ -50,7 +50,7 @@ impl Arch for Cpu {
     /// |   PC    |
     /// |   PSR   |
     /// +---------+
-    fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: usize, arg: usize) {
+    fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: fn(), arg: Option<usize>) {
         let stack_start = stack.as_ptr() as usize;
 
         // 1. The stack starts at the highest address and grows downwards.
@@ -62,13 +62,13 @@ impl Arch for Cpu {
         let stack_pos = (stack_highest - 32) as *mut usize;
 
         unsafe {
-            write_volatile(stack_pos.offset(0), arg); // -> R0
+            write_volatile(stack_pos.offset(0), arg.unwrap_or_default()); // -> R0
             write_volatile(stack_pos.offset(1), 1); // -> R1
             write_volatile(stack_pos.offset(2), 2); // -> R2
             write_volatile(stack_pos.offset(3), 3); // -> R3
             write_volatile(stack_pos.offset(4), 12); // -> R12
             write_volatile(stack_pos.offset(5), cleanup as usize); // -> LR
-            write_volatile(stack_pos.offset(6), func); // -> PC
+            write_volatile(stack_pos.offset(6), func as usize); // -> PC
             write_volatile(stack_pos.offset(7), 0x01000000); // -> APSR
         }
 
--- a/src/ariel-os-threads/src/arch/riscv.rs
+++ b/src/ariel-os-threads/src/arch/riscv.rs
@@ -27,15 +27,15 @@ impl Arch for Cpu {
 
     /// On RISC-V (ESP32), the stack doesn't need to be set up with any register values since
     /// they are restored from the stored [`TrapFrame`].
-    fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: usize, arg: usize) {
+    fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: fn(), arg: Option<usize>) {
         let stack_start = stack.as_ptr() as usize;
         // 16 byte alignment.
         let stack_pos = (stack_start + stack.len()) & 0xFFFFFFE0;
         // Set up PC, SP, RA and first argument for function.
         thread.data.sp = stack_pos;
-        thread.data.a0 = arg;
+        thread.data.a0 = arg.unwrap_or_default();
         thread.data.ra = cleanup as usize;
-        thread.data.pc = func;
+        thread.data.pc = func as usize;
 
         thread.stack_lowest = stack_start;
         thread.stack_highest = stack_pos;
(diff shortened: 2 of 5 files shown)


src/ariel-os-threads/src/lib.rs
```diff
@@ -578,15 +578,18 @@ unsafe impl<T: Sync + Sized> Arguable for &'static T {
...
     prio: u8,
     core_affinity: Option<CoreAffinity>,
-) -> ThreadId {
+) -> ThreadId
+where
+    T: Arguable + Send,
+{
     let arg = arg.into_arg();
-    unsafe { create_raw(func as usize, arg, stack, prio, core_affinity) }
+    unsafe { create_raw(core::mem::transmute(func), arg, stack, prio, core_affinity) }
```
ROMemories: I don't see how this is sound; `*const fn()` is a pointer *to* a function pointer, right?

Transmuting `A` to `B` would require `A` and `B` to be [ABI compatible](https://doc.rust-lang.org/stable/core/primitive.fn.html#abi-compatibility) for the eventual call to the function to not be UB, and `fn(T)` and `*const fn()` are not (~~both~~ because of the extra indirection ~~and because of the different parameter list~~).

Edit: ~~`fn()` and `fn(T)` are [actually ABI-compatible](https://doc.rust-lang.org/stable/core/primitive.fn.html#abi-compatibility), however the *call* after transmutation would be UB.~~

src/ariel-os-threads/src/arch/cortex_m.rs
```diff
@@ -68,7 +68,7 @@ impl Arch for Cpu {
             write_volatile(stack_pos.offset(3), 3); // -> R3
             write_volatile(stack_pos.offset(4), 12); // -> R12
             write_volatile(stack_pos.offset(5), cleanup as usize); // -> LR
-            write_volatile(stack_pos.offset(6), func); // -> PC
+            write_volatile(stack_pos.offset(6), func as usize); // -> PC
```
ROMemories, line 71: I believe this is correct, but I'd like to see a reference that specifies that ABI. I think the AAPCS Section 5.1 would do? It seems Arm calls function pointers "code pointers" (which makes sense as the hardware has no concept of "functions").

src/ariel-os-threads/src/arch/riscv.rs
```diff
@@ -27,15 +27,15 @@ impl Arch for Cpu {
...
+    fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: *const fn(), arg: usize) {
         let stack_start = stack.as_ptr() as usize;
         // 16 byte alignment.
         let stack_pos = (stack_start + stack.len()) & 0xFFFFFFE0;
         // Set up PC, SP, RA and first argument for function.
         thread.data.sp = stack_pos;
         thread.data.a0 = arg;
         thread.data.ra = cleanup as usize;
-        thread.data.pc = func;
+        thread.data.pc = func as usize;
```
ROMemories, line 38: I believe this is correct, but I'd like to see a reference that specifies that ABI.

src/ariel-os-threads/src/arch/xtensa.rs
```diff
@@ -23,7 +23,12 @@ impl Arch for Cpu {
         crate::smp::schedule_on_core(crate::core_id())
     }
 
-    fn setup_stack(thread: &mut crate::thread::Thread, stack: &mut [u8], func: usize, arg: usize) {
+    fn setup_stack(
+        thread: &mut crate::thread::Thread,
+        stack: &mut [u8],
+        func: *const fn(),
```
ROMemories: (Can't add a comment on the line using `func` below.) I believe this is correct, but I'd like to see a reference that specifies that ABI.

src/ariel-os-threads/src/arch/cortex_m.rs
kaspar030: This is what is was before, `fn() as usize`. 

(I skimmed AAPCS Section 5.1 and don't find anything that justifies this better.)

kaspar030:
I don't feel like researching why writing the `fn() as usize` into the registers is fine on our arches. We did it like that forever. :) It is also IMO not the significant change here.

src/ariel-os-threads/src/arch/cortex_m.rs
ROMemories: Yes I know, because that `as usize` cast was factored out. But now that we're moving these casts into arch-specific modules (for the better), we can add ABI-specific justifications.

kaspar030: But that's "while you're at it" scope extension. Let's do the justification another time.

src/ariel-os-threads/src/lib.rs
```diff
@@ -578,15 +578,26 @@ unsafe impl<T: Sync + Sized> Arguable for &'static T {
...
+    T: Arguable + Send,
+{
+    let arg = Some(arg.into_arg());
+
+    // Convert `fn(T)` into `fn()`, must go through *const().
+    //
+    // SAFETY:
+    // https://doc.rust-lang.org/stable/std/primitive.fn.html#casting-to-and-from-integers
+    // "Transmuting between raw pointers and function pointers (i.e., two pointer types) is fine."
+    let func = unsafe { core::mem::transmute::<*const (), fn()>(func as *const ()) };
```
ROMemories: In general I would prefer to only have as little as possible in the `unsafe` block (*casting* to `*const ()` is safe). But fine to not incur another CI run.

ROMemories (approved):
Successfully tested the `threading` example on espressif-esp32-c6-devkitc-1, nrf52840dk, rpi-pico-w, rpi-pico2-w, stm32u083c-dk.

src/ariel-os-threads/src/lib.rs
kaspar030: I changed it, this is not *that* urgent.

ROMemories (approved):

PR #1134, opened 2025-07-09
docs(thread-macro): document the `affinity` macro parameter 

ROMemories:
# Description

This explains how to actually pin threads, which wasn't exactly documented, and I don't believe this was an undocumented feature on purpose.

## Issues/PRs references

Depends on #1130 (doesn't actually depend on it semantically, this is just to test the `ci-build:skip` PR label on that PR)

## Change checklist

### DIFF
--- a/book/src/multithreading.md
+++ b/book/src/multithreading.md
@@ -49,8 +49,9 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 
 ### Core Affinity
 
-Core affinity, also known as core pinning, is optionally configurable for each thread.
+Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
 It allows to restrict the execution of a thread to a specific core and prevent it from being scheduled on another one.
+See the [`threading-multicore` example][threading-multicore-example-repo] for a usage example.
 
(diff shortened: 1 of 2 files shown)


ROMemories:
Even when stacked over #1130, which has the `ci-build:skip` PR label attached, the result value of the `should-skip` CI step of *this PR* was indeed `false` as it should be, as it had the `ci-build:small` PR label.

bergzand (commented):
Thanks for adding this.

book/src/multithreading.md
```diff
@@ -49,7 +49,7 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 
 ### Core Affinity
 
-Core affinity, also known as core pinning, is optionally configurable for each thread.
+Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
```
bergzand, line 52: How about adding a link to `examples/threading-multicore` in here, so users immediately have an example of how the affinity should be supplied to the macro

ROMemories: I thought we didn't have links to the examples in the book, but we actually do so I added one, good idea. :slightly_smiling_face:

bergzand (approved):
LGTM

PR #1135, opened 2025-07-10
ci(build): skip the Build when the `ci-build:skip` label is attached

ROMemories:
# Description

This follows #843 and actually acts on the value obtained by the `should-skip` CI workflow step, based on the cache hit and whether the `ci-build:skip` label is attached on the PR. Stacked PRs #1130 and #1134 helped test that that value was indeed correctly obtained, even in the case of stacked PRs (at least for CI runs before merging).

This PR actually implements skipping the main steps of the `Build` CI workflow, to make CI much faster when only updating the documentation.

## How to review this PR

Look at the logs of CI runs of #1130 and #1134 and check the `result` value of the step named "Whether to skip next steps". Only the one labeled with `ci-build:skip` should be `true`.

## Issues/PRs references

Follows #843.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,15 +109,15 @@ jobs:
           echo "${result}" >> $GITHUB_OUTPUT
 
       - name: Run sccache-cache
-        if: steps.result-cache.outputs.cache-hit != 'true'
+        if: steps.should-skip.outputs.result != 'true'
         uses: mozilla-actions/sccache-action@v0.0.8
 
       - id: get_toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -109,15 +109,15 @@ jobs:
           echo "${result}" >> $GITHUB_OUTPUT
```
ROMemories, line 109: There is no remaining occurrences of `result-cache` or `cache-hit` in this file below this line.

PR #1138, opened 2025-07-11
my dev branch base

kaspar030:
This PR splits the CI labels logic out of the build job.

Signed-off-by: Kaspar Schleiser <kaspar@schleiser.de># Description

## Change checklist


kaspar030:
With 'ci-build:skip' set, this now passes, not launching any build step.

kaspar030:
Still correctly fails if more than one label is set.

kaspar030:
> Still correctly fails if more than one label is set.

... or maybe not. The "Final Results" is still queued. Waiting...

ROMemories:
Can you document the benefits of this? I think I see them but it would be better if they were listed in the description.

kaspar030:
> ... or maybe not. The "Final Results" is still queued. Waiting...

So "Final Results" succeeds. With this merged, the label check needs to be marked as "required". Now testing full build.

kaspar030:
> With this merged, the label check needs to be marked as "required".

With that I think I'll also make "Final Results" skip if the label check failed.

kaspar030:
Note: once merged, the "check-labels" check *must* be marked "required" in the branch rules!

.github/workflows/build.yml
```diff
@@ -104,7 +117,7 @@ jobs:
       - name: Whether to skip next steps
         id: should-skip
         run: |
-          result="result=${{ steps.result-cache.outputs.cache-hit == 'true' || steps.get-pr-labels.outputs.result == 'ci-build:skip' }}"
+          result="result=${{ steps.result-cache.outputs.cache-hit == 'true' || needs.check-labels.outputs.labels == 'ci-build:skip' }}"
```
ROMemories: Do we still need to the labels here? I think if the label is `ci-build:skip` is attached the `build` job is already skipped by the `if` condition on the whole job. However fine to keep it and refactor later.

.github/workflows/build.yml
```diff
@@ -14,32 +14,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  check-labels:
     runs-on: ubuntu-24.04
-
+    outputs:
+      labels: ${{ steps.get-pr-labels.outputs.result }}
```
ROMemories: There is only at most one label:

```suggestion
      label: ${{ steps.get-pr-labels.outputs.result }}
```

.github/workflows/build.yml
```suggestion
          result="result=${{ steps.result-cache.outputs.cache-hit == 'true' || needs.check-labels.outputs.label == 'ci-build:skip' }}"
```

.github/workflows/build.yml
ROMemories: Can be done in a `ci-build:skip` follow-up.

ROMemories (approved):

.github/workflows/build.yml
kaspar030: no, I had left it in but removed it now.

ROMemories (approved):

ROMemories:
With this, it's now required to "Re-run all jobs", not just "Re-run all failed jobs" when changing the labels.

PR #1152, opened 2025-07-15
fix(debug): actually compare log level in `log` backend

kaspar030:
# Description

Seems like the log backend didn't actually compare to the configured log level.

## Change checklist

### DIFF
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -195,7 +195,7 @@ pub use backend::*;
 #[doc(hidden)]
 #[cfg(feature = "log")]
 mod logger {
-    use log::{Level, LevelFilter, Metadata, Record};
+    use log::{LevelFilter, Metadata, Record};
 
     static LOGGER: DebugLogger = DebugLogger;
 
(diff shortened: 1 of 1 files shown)


src/ariel-os-debug/src/lib.rs
```diff
@@ -259,7 +259,7 @@ mod logger {
 
     impl log::Log for DebugLogger {
         fn enabled(&self, metadata: &Metadata) -> bool {
-            metadata.level() <= Level::Info
```
ROMemories, line 262: Seems like the `Level` import is now unused and can be removed.

ROMemories (commented):
I am not sure how to test this tbh; I tried:

```sh
laze -C examples/log build -s log -DLOG=warn -b nrf52840dk run
```

and the `INFO` line is not shown, as expected, but that is also working *without* that change.

I also tried using the [`log_enabled!`](https://docs.rs/log/latest/log/macro.log_enabled.html) macro, as it is mentioned in the [docs of `Log::enabled()`](https://docs.rs/log/latest/log/trait.Log.html#tymethod.enabled), but it does also return the expected value even without this change.

---

Not saying that this change is not needed, but I'm not sure how to test this.

src/ariel-os-debug/src/lib.rs
kaspar030: fixed, also fixed the lifetime lints

kaspar030:
> and the `INFO` line is not shown, as expected, but that is also working _without_ that change.

Try with `-DLOG=trace`, without this change, no messages get shown that are more verbose than `info`.

ROMemories (approved):
Successfully tested with:

```sh
laze -C examples/log build -s log -DLOG=trace -b nrf52840dk run
```

Issue #1115, opened 2025-06-27
Support the modem on nRF9151


nponsard:
### Description

Add support to the modem on the nRF9151 sip. This modem is only officially usable via the nordic SDK and includes a closed source binary blob. 

There is one project aiming to support the modem for Rust: [nrf-modem](https://github.com/diondokter/nrf-modem).
There is also another project centered around the nrf9160 that doesn't seem to get much updates as the company's website is offline now: [nrfxlib](https://github.com/nrf-rs/nrfxlib).

According to the documentation, using the modem will require the application core to run in NS mode, meaning isolation features will be enabled. Ariel OS doesn't seem to be able to boot in NS mode currently.

### Motivation

We have some upcoming projects using the thingy91x (nRF9151) that will need access to the GPS and LTE-M features.

### Requirements

1. Access to GPS features to obtain a position
2. Access to LTE-M features to do networking

### Are you planning to do it yourself in a pull request?

Yes

nponsard:
I managed to go a bit forward in the initialization of the modem in NS mode, I get some logs back (about the TX region, RX region, Trace region) but then the initialization fails with the error ` NrfError(-116)`. This seems to be similar to https://github.com/diondokter/nrf-modem/issues/21.

Also it seems that the executor-thread doesn't work in NS mode currently, the firmware crashes in `ariel-os-rt` at `threading::start();` with the red led of the thingy91x lighting up.

diondokter:
> There is one project aiming to support the modem for Rust: [nrf-modem](https://github.com/diondokter/nrf-modem).
> There is also another project centered around the nrf9160 that doesn't seem to get much updates as the company's website is offline now: [nrfxlib](https://github.com/nrf-rs/nrfxlib).

To give a little bit of background: First there was nrfxlib and Jonathan made things work. But then it sort of got abandoned. I think I updated it once or so, but then decided this shiny new async thing was very interesting. So I made my own version which is nrf-modem. nrf-modem is very much inspired from nrfxlib.

nrfxlib is no longer supported (and is quite old w.r.t. the blob version it uses). Nrf-modem is sort of supported by me, passively. That means I'm not really doing active work on it, but will respond to issues and merge new work.

Nrf-modem is made to work with bogstandard embassy. But I'd be open to changes to make it work in more places!

nponsard:
Thanks for the clarification! 
I'm trying to track down where the problems I'm having come from, once I figure this out I'll open an issue/PR if changes need to be made.

[...]

PR #1156, opened 2025-07-17
feat(gnss): add GNSS public interface

nponsard:
# Description

This PR adds the public interface for users to access GPS information.

## Issues/PRs references

Related to #1115.

## Open Questions

I used an `embassy_sync::watch::Watch` to allow the user to access it once or to async poll the changes. I'm open to suggestions for a better interface. 

Concerning the configuration, currently only the operating mode (periodic, continuous, single fix) can be configured but nrf-modem can expose more settings to indicate the wanted accuracy, and if the module should download almanacs.

Ublox modules don't seem to have much accuracy-altering configuration available though, should more settings be exposed ? Meaning if a module doesn't have the option it will be ignored.

The two most interesting settings to expose for nrf-modem would be:
- Elevation angle: Using the default value of 5 degrees over the horizon should be fine for most use cases, but some may want to increase this value to increase precition
- Low precition mode: would allow to get a fix faster

## Change checklist

### DIFF
--- a/src/ariel-os-embassy-common/src/gnss.rs
+++ b/src/ariel-os-embassy-common/src/gnss.rs
@@ -668,23 +668,12 @@ impl core::fmt::Display for GnssVelocity {
     }
 }
 
-// Based on NMEA RMC message
-/// Represents date and time information from GNSS. All fields are in UTC.
+/// Represents date and time information from GNSS.
 #[derive(Debug, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct GnssDateTime {
-    /// Year
-    pub year: u16,
-    /// Month (1-12)
-    pub month: u8,
-    /// Day of the month (1-31)
-    pub day: u8,
-    /// Hour (0-23)
-    pub hour: u8,
-    /// Minute (0-59)
-    pub minute: u8,
-    /// Second (0-59)
-    pub second: u8,
+    /// Seconds since Ariel epoch (2024-01-01T00:00:00Z)
+    pub seconds: u32,
     /// Milliseconds (0-999)
     pub milliseconds: u16,
 }


src/ariel-os-hal/Cargo.toml
```diff
@@ -71,6 +71,10 @@ ble = ["dep:bt-hci", "dep:embedded-io", "dep:trouble-host", "ariel-os-nrf/ble"]
 ble-peripheral = ["ariel-os-nrf/ble-peripheral"]
 ble-central = ["ariel-os-nrf/ble-central"]
 
+gps = []
+
+modem = []
```
ROMemories: Possibly unrelated?

chrysn:
Reiterating from chat: I don't think that GPS deserves special treatment compared to any other sensor:
* It can be composite from multiple data sources -- so is the spacial orientation of a device when an IMU combines magnetic and accelerometer readings.
* Its power consumption depends on the spacial (data-wise) or temporal resolution requested -- as is the case for any ADC readout based sensor.

ROMemories:
After some discussion with @kaspar030, we do think having a dedicated interface for GNSS makes sense. This does not preclude also having a GNSS sensor through the abstraction later.

@nponsard Can you use "[GNSS](https://en.wikipedia.org/wiki/Satellite_navigation)" instead of "[GPS](https://en.wikipedia.org/wiki/Global_Positioning_System)"? It is more generic, GPS refers to the US system specifically.

src/ariel-os-macros/src/config.rs
```diff
@@ -140,6 +145,7 @@ mod config_macro {
     #[derive(Debug, enum_iterator::Sequence)]
     pub enum ConfigKind {
         // Ble,
+        // Gnss,
```
ROMemories, line 154: Now that #1194 is in, it will need to be updated for the new `Gnss` variant.

src/ariel-os-embassy/src/gnss.rs
```diff
@@ -0,0 +1,67 @@
...
+pub fn get_receiver<'a>() -> Option<GnssDataReceiver<'a>> {
+    let config = config();
+
+    if matches!(config.operation_mode, GnssOperationMode::SingleShot(_)) {
+        return None;
+    }
+    WATCH.receiver()
+}
+
+/// Request a GNSS fix.
```
ROMemories:
```suggestion
/// Requests a GNSS fix.
```

src/ariel-os-embassy/src/gnss.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+// Initialize the GNSS with the provided configuration and spawner.
+#[allow(dead_code, reason = "false positive during builds outside of laze")]
+pub(crate) async fn init_gnss(spawner: Spawner) {
+    let config = config();
+    let sender: GnssDataSender<'_> = WATCH.sender();
+    crate::hal::gnss::init_gnss(spawner, sender, config).await;
+}
+
+/// Get an `embassy_sync::watch::Receiver` to get updates on GNSS data.
```
ROMemories:
```suggestion
/// Returns an [`embassy_sync::watch::Receiver`] to get updates on GNSS data.
```

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+            operation_mode: GnssOperationMode::Continuous,
+        }
+    }
+}
+
+/// Represents position data from GNSS.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct GnssPosition {
+    /// Latitude in degrees
```
ROMemories: All doc comments should end with a period.
```suggestion
    /// Latitude in degrees.
```

laze-project.yml
```diff
@@ -1214,6 +1214,27 @@ modules:
...
+  - name: gnss
+    selects:
+      - hw/gnss
+    env:
+      global:
+        FEATURES:
+          - ariel-os/gnss
+
+  - name: hw/gnss
+    help: provided if a device has a gnss receiver
```
```suggestion
    help: provided if a device has a GNSS receiver
```

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Positive values indicate north, negative values indicate south.
+    pub latitude: f64,
```
ROMemories: Do we *really* need double-precision floats here (and for the rest)? Even hard floats are a luxury, and the sensor API made a point of not relying on them.

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+pub struct GnssVelocity {
+    /// Speed in m/s
+    pub speed: f64,
+    /// Speed accuracy in m/s
+    pub speed_accuracy: f64,
+    /// Vertical speed in m/s
+    pub vertical_speed: f64,
+    /// Vertical speed accuracy in m/s
+    pub vertical_speed_accuracy: f64,
+    /// Heading in degrees (0-360)
```
ROMemories:
```suggestion
    /// Heading in degrees (0–360)
```

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Altitude in meters above sea level
+    pub altitude: f64,
+    /// Accuracy of the position in meters
+    pub accuracy: f64,
+    /// Altitude accuracy in meters
+    pub altitude_accuracy: f64,
+}
+
+/// Represents velocity data from GNSS.
+#[derive(Debug, Clone)]
```
ROMemories: Other traits should be derived when it makes sense, see [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits). Also I think *adding* these later would [be a breaking change](https://cheats.rs/#api-stability), because it changes the *nature* of the type.

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct GnssData {
+    /// The position data, if available.
+    pub position: Option<GnssPosition>,
+    /// The velocity data, if available.
+    pub velocity: Option<GnssVelocity>,
+    /// The date and time information, if available.
+    pub datetime: Option<GnssDateTime>,
+    /// The time when the data was recorded (from start of the MCU).
+    pub recorded_at: embassy_time::Instant,
```
ROMemories: Is there a use case for that field? An `embassy_time::Instant` is internally a `u64`, that costs 8 bytes.

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/// Represents GNSS data that can be received from the GNSS module.
+///
+/// A field can be `None` if the GNSS module did not provide that information.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct GnssData {
+    /// The position data, if available.
+    pub position: Option<GnssPosition>,
```
ROMemories: Is wrapping these in `Option`s really useful—i.e., can this reasonably happen in practice (genuinely asking)? Because of packing and alignment, I believe they cost 4 bytes in this specific struct.

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub heading: f64,
+    /// Heading accuracy in degrees
+    pub heading_accuracy: f64,
+}
+
+// Based on NMEA RMC message
+/// Represents date and time information from GNSS.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct GnssDateTime {
```
ROMemories, line 674: Is this the right way of representing this? This results in a lot of integer fragmentation, where a single `u64` could likely do the job.

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+///
+/// You can customize it using the `gnss-config-override` feature.
+#[derive(Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config {
+    /// The mode of GNSS to use.
+    pub operation_mode: GnssOperationMode,
+}
+
+impl Default for Config {
```
ROMemories, line 56: It would possibly be useful to have *`const`* `new()` constructor (which `default()` would be calling).

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+pub type GnssDataSender<'a> = Sender<'a, CriticalSectionRawMutex, GnssData, MAX_WATCH_RECEIVERS>;
+
+/// Operation modes for the GNSS module.
+#[derive(Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssOperationMode {
+    /// Always keep the GNSS module active.
+    Continuous,
+    /// Update the GNSS fix periodically.
+    Periodic(u16), // Period in seconds
```
ROMemories: If this type is public, these comments should likely be moved in the doc comments.

src/ariel-os-embassy-common/src/gnss.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Try to get a GNSS fix only when requested, you won't be able to get updates using `get_receiver`.
+    SingleShot(u16), // Timeout in seconds
```
ROMemories: Would it make sense to shrink these to `u8`s? That's already more than 4 minutes.

src/ariel-os-embassy/src/gnss.rs
```diff
@@ -0,0 +1,67 @@
+use embassy_executor::Spawner;
+
+use ariel_os_embassy_common::gnss::{
+    Config, GnssData, GnssDataReceiver, GnssDataSender, GnssDataWatch, GnssOperationMode,
+};
+
+static WATCH: GnssDataWatch = GnssDataWatch::new();
```
ROMemories, line 7: I'll have a look later whether we can make do without a watcher.

src/ariel-os-embassy-common/src/gnss.rs
nponsard: My idea was to be able to use `recorded_at` and `datetime` to be able to calibrate a clock with more precision.

src/ariel-os-embassy-common/src/gnss.rs
nponsard: based it on the nrf implementation that returns f64 for latitude and longitude. I guess we can convert them to f32 to save space.

src/ariel-os-embassy-common/src/gnss.rs
nponsard: If I understand correctly, many GNSS sensors communicate using [NMEA sentences](https://aprs.gids.nl/nmea/#zda). The `GPZDA` sentence uses a format with individual numbers for year, month, day, ... 
The conversion to an amount of seconds since an EPOCH could indeed be done by the driver.

src/ariel-os-embassy-common/src/gnss.rs
nponsard: The [nrf-modem api](https://github.com/nrfconnect/sdk-nrfxlib/blob/c96750005cc5d89d3152e454164a79c57d08948c/nrf_modem/include/nrf_modem_gnss.h#L1076) takes an u16 with a default value of 60 seconds, in some examples 180 seconds is used. [On wikipedia](https://en.wikipedia.org/wiki/Time_to_first_fix) and discussed on forums its suggested that some slower hardware can take more than 10 minutes to get a fix while in a factory state.

src/ariel-os-embassy-common/src/gnss.rs
nponsard: I'm adding `Copy` and `Display`, do we need more than those ?

src/ariel-os-embassy-common/src/gnss.rs
nponsard: The idea was to set the properties to `None` when no value is available (the current nrf_modem implementation returns default values instead). For example the GNSS module may not have a position fix already but already determined the date and time (this happens when only 2 satellites are in view), the idea would be to set `position` and `velocity` to `None` with `datetime` populated to reflect the current state.

src/ariel-os-macros/src/config.rs
nponsard: The only change is that I need to add an entry to the `variants` in `parse()` right ?

ROMemories: cc @jkr-ariel

src/ariel-os-embassy-common/src/gnss.rs
ROMemories: I suppose `PartialEq` makes sense as well; we can't have `Eq` with floats though.

nponsard:
Concerning the configuration, currently only the operating mode (periodic, continuous, single fix) can be configured but nrf-modem can expose more settings to indicate the wanted accuracy, and if the module should download almanacs.

Ublox modules don't seem to have much accuracy-altering configuration available though, should more settings be exposed ? Meaning if a module doesn't have the option it will be ignored.


The two most interesting settings to expose for nrf-modem would be:
- Elevation angle: Using the default value of 5 degrees over the horizon should be fine for most use cases, but some may want to increase this value to increase precition
- Low precition mode: would allow to get a fix faster

nponsard:
@ROMemories I added a commit to use fixed point numbers for latitude and longitude like you suggested, `fixed` types don't implement `defmt::Format` so I have to manually implement the formatting of the struct. 
I'm also not sure of what are the benefits of using a fixed point number, I couldn't find documentation explaining the benefits of fixed point numbers versus floating point.

ROMemories:
> I'm also not sure of what are the benefits of using a fixed point number, I couldn't find documentation explaining the benefits of fixed point numbers versus floating point.

Although you started this interface in the context of the nRF9151 modem, it is intended to be more generic and should work well with every MCU we support, and they do not all have an FPU. If you use floats without an FPU, the compiler will provide a software implementation of them, which will be much slower than fixed-point arithmetic, and will inflate the binary.

Even when you do have an FPU, fixed-point arithmetic can get you more precision when the range of values is smaller. It can also make context switching faster, because then the preemptive scheduler does not have to save/restore FP registers (see https://github.com/ariel-os/ariel-os/pull/1097).

There are other benefits of fixed-point arithmetic over floats that are not relevant in the context of this interface, which are determinism of operations, and interoperability with other systems (FP operations do not necessarily produce exactly the same results on different platforms).

What this means in the context of this interface is that we need to remove all floats and only use fixed-point numbers.

---

Some of these points are taken from [this StackOverflow post](https://stackoverflow.com/questions/67225433/how-to-decide-when-to-use-fixed-point-arithmetic-over-float).

nponsard:
Thanks for the clarification, should I work on finding if more data elements should be converted to a fixed point number ?

ROMemories:
> should I work on finding if more data elements should be converted to a fixed point number ?

All floats need to be removed yes; we may also want to consider simply using `i32`s with an appropriate scaling factor.

ROMemories:
So I have thought about this some more and I think the best move here for things that are currently floats is to create semantic, dedicated newtypes (e.g., `Latitude`/`DecimalDegrees`, `Velocity`, `Heading`) that would internally use a fixed-point representation, provide custom implementations for `Debug`/`Display`/`defmt::Format`, allow access to the integer and fractional parts, provide a `as_be_bytes()` method (or similar) so it can easily be turned into a `fixed` type (through `from_be_bytes()`) if users need to do fixed-point arithmetic, and possibly provide an `as_f32()` (or `as_f64()` if `f32` don't have enough precision) method (if that method is not used, I expect the software impl of floats not to be included; this would need to be verified).

See [C-CONV](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) for why I think these methods should be `as_` methods.

src/ariel-os-macros/src/config.rs
jkr-ariel: > The only change is that I need to add an entry to the `variants` in `parse()` right ?

Yes.

src/ariel-os-embassy-common/src/gnss.rs
ROMemories: As we discussed offline, a `u64` would actually result in worse memory usage than the three words `GnssDataTime` is currently using.

chrysn: It's not just either u64 or the fragmented integers; u32 is a sensible choice for date-in-seconds.

Considering that at some point we'll want to have generic time anyway (see also https://github.com/ariel-os/ariel-os/discussions/642#discussioncomment-14153832), we can use a u32 seconds from a custom epoch (say, Age of Ariel that starts 2024-01-01, or at the timestamp of the last release; anything that's in the recent past and not as volatile as to break reproducible builds) in a newtype that hides all effects of the epoch shifting other than the range of expressible values. That'll last >130 years, way longer than a single device would ever go without a firmware update.

nponsard: In NMEA sentences also exposes the milliseconds, if we want to keep this information we would need an u64.

ROMemories: Unless actually we use a `u32` for seconds *and* a `u16` for milliseconds, then we get both of both worlds I think (hadn't that idea when we discussed this).

nponsard: Sounds good, I've added a commit to change that.

nponsard:
@ROMemories 
I made changes to use a custom fixed-point implementation like we discussed. Tell me if it was what you were thinking about.

ROMemories:
Superseded by #1330.

PR #1157, opened 2025-07-17
feat(native): wire up hwrng

kaspar030:
# Description

This adds hwrng to the native board.

## Issues/PRs references

Depends on #647.

## Change checklist

### DIFF
--- a/src/ariel-os-native/src/hwrng.rs
+++ b/src/ariel-os-native/src/hwrng.rs
@@ -1,0 +1,4 @@
+/// Constructs the hardware random number generator (RNG).
+pub fn construct_rng(_peripherals: &mut crate::OptionalPeripherals) {
+    ariel_os_random::construct_rng(rand::rngs::OsRng::default());
+}
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1876,6 +1876,7 @@ builders:
       - infini-core
       - ?debug-console
     provides:
+      - has_hwrng
       - sw/benchmark
     disables:
       - semihosting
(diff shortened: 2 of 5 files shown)


src/ariel-os-native/src/hwrng.rs
```diff
@@ -0,0 +1,11 @@
+/// Constructs the hardware random number generator (RNG).
+///
+pub fn construct_rng(_peripherals: &mut crate::OptionalPeripherals) {
+    cfg_if::cfg_if! {
+        if #[cfg(context = "native")] {
+            ariel_os_random::construct_rng(rand::rngs::OsRng::default());
```
chrysn: Why is this gated? It's in ariel-os-native, which is pulled in only on native anyway.

kaspar030: I had wasm in mind, which probably needs a different implementation.

kaspar030: I can remove this now and re-add when we actually need it.

chrysn: Yes I think it's better to cross that bridge when we're there.

src/ariel-os-native/src/hwrng.rs
```diff
@@ -0,0 +1,5 @@
+/// Constructs the hardware random number generator (RNG).
+///
```
chrysn:
```suggestion
```

stylistically, but only if there's any other reason to re-run CI -- otherwise, please dismiss this and merge as is.

ROMemories (approved):
Successfully tested with:

```sh
laze -C examples/random build -b native run
```

PR #1153, opened 2025-07-16
ci(build-pr-labels): introduce more labels to control the set of devices

ROMemories:
# Description

Following #843 and #1135, this introduces more fine-grained PR labels to control the set of devices the Build workflow builds for.

## How to review this PR

This PR is better reviewed commit by commit.
Additionally, Enabling "Hide whitespace" in the diff view makes it obvious the existing `echo "LAZE_BUILDERS=[…]` line has only been indented.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1159, opened 2025-07-18
ci(caching): fix the cache key for limited builds

ROMemories:
# Description

Following #1153, the set of builders also have to be taken into account in the cache key, otherwise changing the build label of a PR whose CI already passed would hit the cache and skip the Build quickly.

This technically uses the build label string in the cache key (instead of the actual set of builders), but that is fine because the association between these is done in-tree, and changing it would invalidate the cache anyway as the file tree hash is part of the cache key. We could instead hash the actual set of builders, but using the build label string is more readable for humans.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,7 +112,7 @@ jobs:
         id: result-cache
         with:
           path: .tree-hash
-          key: success-${{ steps.get-tree-hash.outputs.hash }}-${{ matrix.partition }}-${{ github.event_name == 'schedule' && 'full' || 'limited' }}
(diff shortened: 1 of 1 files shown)


ROMemories:
Tested by running with `ci-build:rp`, then with `ci-build:small` and checked that the cache was indeed *not* reused, as the cache keys were different as expected: `success-4635db176f836fd699100355f491c542ce5d5980-1/20-ci-build:small` vs `success-4635db176f836fd699100355f491c542ce5d5980-1/20-ci-build:rp`.

After switching back to `ci-build:rp`, the cache was reused as expected.

PR #1162, opened 2025-07-18
fix(threads): fix `current_tid()`

kaspar030:
# Description

#647 broke the scheduler in single-core mode by using a wrong comparison and insufficient clippy fallback.

## Change checklist

### DIFF
--- a/src/ariel-os-threads/src/lib.rs
+++ b/src/ariel-os-threads/src/lib.rs
@@ -168,7 +168,7 @@ impl Scheduler {
     #[allow(clippy::unnecessary_wraps, reason = "only unnecessary in clippy path")]
     fn current_tid(&self) -> Option<ThreadId> {
         cfg_if::cfg_if! {
-            if #[cfg(context = "single-core")] {
+            if #[cfg(feature = "single-core")] {
                 self.current_thread
             }
             else if #[cfg(feature = "multi-core")] {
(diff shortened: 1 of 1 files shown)


src/ariel-os-threads/src/lib.rs
```diff
@@ -29,6 +29,13 @@
 #![allow(clippy::indexing_slicing)]
 #![expect(clippy::cast_possible_truncation)]
 
+#[cfg(not(any(
```
kaspar030: this is unrelated actually, but IMO useful. and yeah, a debug leftover.

kaspar030: lemme know if we can keep it. :)

kaspar030: ok, it broke the tests, so I dropped it.

src/ariel-os-threads/src/lib.rs
```diff
@@ -178,9 +178,8 @@ impl Scheduler {
                 ThreadData::ID.get()
             }
             else {
-                // makes clippy happy
                 let _ = self;
-                Some(ThreadId::new(0))
+                unreachable!();
```
kaspar030, line 182: This is so it now fails at runtime should if ever be executed.

nponsard (approved):
Tested on the thingy91X (nrf9151, nrf5340-app, nrf5340-net). This fixes the runtime error that is currently on main.

PR #1163, opened 2025-07-18
fix(nrf): enable flash cache

kaspar030:
# Description

The NRFs have a flash cache that improves performance quite a bit, but is not enabled by default by the nrf pac.

Performance wise, the `bench_sched_flags` result goes from 1574 to 1305 ticks on nrf52840dk, and from ~2300 to ~1100 on nrf5340dk.

*edit* ~~I just realized that as is this only works on nrf52840dk, looking into the pac now.~~ I got nrf52*, nrf5340, nrf5340-net and nrf91.
Tested on nrf52840dk and nrf5340dk and -net.

## Issues/PRs references

Depends on #1162.

## Open Questions

- [x] nrfs other than nrf52840

## Change checklist

### DIFF
--- a/src/ariel-os-nrf/src/lib.rs
+++ b/src/ariel-os-nrf/src/lib.rs
@@ -72,6 +72,27 @@ pub static EXECUTOR: Executor = Executor::new();
 #[doc(hidden)]
 #[must_use]
 pub fn init() -> OptionalPeripherals {
+    enable_flash_cache();
+
     let peripherals = embassy_nrf::init(Config::default());
     OptionalPeripherals::from(peripherals)
+}
+
+fn enable_flash_cache() {
+    // (no flash cache on nrf51)
+    cfg_if::cfg_if! {
+        if #[cfg(any(
+                context = "nrf52",
+                context = "nrf5340-net",
+                context = "nrf91"
+            ))] {
+            embassy_nrf::pac::NVMC
+                .icachecnf()
+                .write(|w| w.set_cacheen(true));
+        }
+        else if #[cfg(context = "nrf5340")] {
+            embassy_nrf::pac::CACHE_S
+                .enable().write(|w| w.set_enable(true));
+        }
+    }
 }


nponsard:
Tested this on the nrf9151 of the thingy91X. 
`bench_sched_flags` crashes on main commit 26d3ea428bc64ae5977baf22cef04b387cfa05f9 .
I got 1637 ticks per iteration on tag v0.2.1, with this PR I get 1067.

kaspar030:
> `bench_sched_flags` crashes on main commit [26d3ea4](https://github.com/ariel-os/ariel-os/commit/26d3ea428bc64ae5977baf22cef04b387cfa05f9) .

probably due to missing #1162.

src/ariel-os-nrf/src/lib.rs
```diff
@@ -72,6 +72,26 @@ pub static EXECUTOR: Executor = Executor::new();
...
     OptionalPeripherals::from(peripherals)
 }
+
+fn flash_cache_enable() {
+    cfg_if::cfg_if! {
+        if #[cfg(any(
+                context = "nrf52",
+                context = "nrf5340-net",
+                context = "nrf91"
+                ))] {
```
ROMemories:
```suggestion
            ))] {
```

ROMemories:
What's the tradeoff here? Why is it not enabled by the HAL in all cases, or at least part of the HAL's `Config`?

src/ariel-os-nrf/src/lib.rs
```diff
@@ -72,6 +72,26 @@ pub static EXECUTOR: Executor = Executor::new();
 #[doc(hidden)]
 #[must_use]
 pub fn init() -> OptionalPeripherals {
+    flash_cache_enable();
+
     let peripherals = embassy_nrf::init(Config::default());
     OptionalPeripherals::from(peripherals)
 }
+
+fn flash_cache_enable() {
```
ROMemories:
```suggestion
fn enable_flash_cache() {
```

nponsard:
> What's the tradeoff here? Why is it not enabled by the HAL in all cases, or at least part of the HAL's `Config`?

The nrf9151's datasheet says its mainly about power consumption (part 4.4.7): 
> If the reduction in average current due to reduced flash accesses is larger than the
cache power requirement, the average current to execute the program code is reduced.

src/ariel-os-nrf/src/lib.rs
ROMemories: Can you add a comment explaining there is no I-cache on the nRF51 (I think)?

ROMemories (commented):
The tradeoff seems to be power consumption (§ 4.3.8 of the nRF52840 datasheet):

> Enabling the cache can increase CPU performance and reduce power consumption by reducing the number of wait cycles and the number of flash accesses. This will depend on the cache hit rate. Cache will use some current when enabled. If the reduction in average current due to reduced flash accesses is larger than the cache power requirement, the average current to execute the program code will decrease.

I am not against enabling this, but I'd like to understand why the HAL doesn't do this all the time if it is so beneficial first (maybe it's just an oversight?).

I think we should also keep in mind that, even though the results on the `bench_sched_flags` benchmark are impressive (1.5–2× speedup depending on the MCU/board), it also seems to be the ideal case, where the same (possibly small) set of  instructions is just re-executed over and over.

kaspar030:
Zephyr seems to enable the cache by default on all platforms. It is configurable though.

ROMemories:
Removed `ci-build:small` to spare the CI, I hope we can have #1166 in to test this PR.

ROMemories:
Can you rebase on main and use the `ci-build:nrf` label?

PR #1164, opened 2025-07-18
feat(native): add hardware identity support

chrysn:
# Description

This implements the `has_device_identity` / `hw/identity` feature for native based on the [standard `/etc/machine-id`](https://www.freedesktop.org/software/systemd/man/latest/machine-id.html) file.

As by the standard, we don't use the machine ID 1:1 (as that'd allow third parties to correlate our use of the ID with another application's use), but hash it with an application specific identifier.

As by the docs of ariel-os-identity (which BTW I think we should have built somewhere if only to link from here), this identifier stays constant across using different applications -- neither the application name nor the build path are factored in. This can be a bit weird because there *can* be multiple instances of a native application running at the same time, but then, if those are networked, the user will have to do manual assignment of the devices to tap interfaces etc as well.

This PR does not introduce any means for the user to override the machine ID (just as there is currently no override for the RNG in the form of a seed). I envision that at some point per-running-program configuration will be needed (most importantly, assigning a tap device), either in the form of an environment variable or a command line argument, and possibly through a config file, but there is no urgent need to do this here, so let's do this properly when we need it.

## Change checklist

  *will the feature table update automatically? or scream in CI?*

### DIFF
--- a/src/ariel-os-native/Cargo.toml
+++ b/src/ariel-os-native/Cargo.toml
@@ -17,6 +17,7 @@ embassy-time = { workspace = true, default-features = false, features = [
   "std",
 ] }
 embedded-hal-async = { workspace = true }
+ariel-os-buildinfo = { workspace = true }
 ariel-os-debug = { workspace = true, features = ["std"] }
 ariel-os-embassy-common = { workspace = true }
 ariel-os-random = { workspace = true, optional = true }
--- a/src/ariel-os-native/src/identity.rs
+++ b/src/ariel-os-native/src/identity.rs
@@ -13,7 +13,8 @@ pub struct DeviceId([u8; SHA256LEN]);
 impl ariel_os_embassy_common::identity::DeviceId for DeviceId {
     fn get() -> Result<Self, impl std::error::Error> {
         let mut hash = sha2::Sha256::new();
-        hash.update(b"Ariel OS device identity");
+        hash.update(ariel_os_buildinfo::OS_NAME);
+        hash.update(b"device identity");
         // We *could* decode the hex value, but why -- the format of that file is fixed anyway.
         hash.update(std::fs::read("/etc/machine-id")?);
         let hash = hash.finalize();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -402,6 +402,7 @@ dependencies = [
 name = "ariel-os-native"
 version = "0.2.0"
 dependencies = [
+ "ariel-os-buildinfo",
  "ariel-os-debug",
  "ariel-os-embassy-common",
  "ariel-os-random",


src/ariel-os-hal/src/lib.rs
```diff
@@ -33,7 +33,7 @@ cfg_if::cfg_if! {
     if #[cfg(context = "native")] {
         mod dummy;
         pub use ariel_os_native::*;
-        pub use dummy::{gpio, identity, peripheral};
+        pub use dummy::{gpio, peripheral};
```
chrysn, line 36: I've seen warnings fly out of local builds around the dummy peripheral. They also occurred on main, so we may need to increase our coverage on such warnings, but that wouldn't belong in here..

src/ariel-os-native/src/identity.rs
```diff
@@ -0,0 +1,33 @@
...
+//!
+//! [machine-id]: https://www.freedesktop.org/software/systemd/man/latest/machine-id.html
+
+use sha2::Digest as _;
+
+const SHA256LEN: usize = 32;
+
+pub struct DeviceId([u8; SHA256LEN]);
+
+extern crate std;
```
chrysn: If this looks weird to you, remember that the only reason why the RNG implementation doesn't need to do this is because it happens in a two-stage process: ariel-os-native depends on rand::getrandom, and that pulls in the rand crate's std feature.

ROMemories:
`ci-build:*` labels are not *currently* additive, I've removed `ci-build:native` as it is currently not taken into account at all.

src/ariel-os-native/Cargo.toml
```diff
@@ -23,6 +23,7 @@ ariel-os-random = { workspace = true, optional = true }
 rand = { workspace = true, default-features = false, optional = true, features = [
   "getrandom",
 ] }
+sha2 = { version = "0.10.8", default-features = false }
```
ROMemories, line 27: Any reason not to use `0.10.9`?

src/ariel-os-native/src/identity.rs
```diff
@@ -0,0 +1,33 @@
+//! Provides a Device ID based on [machine-id]
```
ROMemories:
```suggestion
//! Provides a Device ID based on [machine-id].
```

laze-project.yml
```diff
@@ -1878,6 +1878,7 @@ builders:
     provides:
       - has_hwrng
       - sw/benchmark
+      - has_device_identity
```
ROMemories: As the order of these has meaning, they should be kept in lexicographical order unless there is a reason not to:

```suggestion
      - has_device_identity
      - has_hwrng
      - sw/benchmark
```

src/ariel-os-hal/src/lib.rs
ROMemories: I did mention these dead code warnings in [my review](https://github.com/ariel-os/ariel-os/pull/647#pullrequestreview-3029608354), but it's not so much an issue of coverage (they do show up in the CI logs) than it is that we just don't fail the builds on them (yet).

src/ariel-os-native/src/identity.rs
ROMemories: That should be moved as the first item in this file (above the `use`s), similarly to `mod` declarations.

ROMemories: What actually looks weird to me now is why the `ariel-os-native` crate is actually `no_std`, and deciding whether to link against std conditionally, based to a Cargo feature.

kaspar030: instead of this, you could also remove the `no_std` in `lib.rs`, right?

src/ariel-os-native/src/identity.rs
```diff
@@ -0,0 +1,33 @@
...
+const SHA256LEN: usize = 32;
+
+pub struct DeviceId([u8; SHA256LEN]);
+
+extern crate std;
+
+impl ariel_os_embassy_common::identity::DeviceId for DeviceId {
+    fn get() -> Result<Self, impl std::error::Error> {
+        let mut hash = sha2::Sha256::new();
+        hash.update(b"Ariel OS device identity");
```
ROMemories: optional: to avoid hard-coding the OS name here, you could use the one from `ariel-os-buildinfo` (it needs to be added as a dependency, but that shouldn't lead to cyclic dependencies later):
```suggestion
        hash.update(ariel_os_buildinfo::OS_NAME);
```

kaspar030:
> As by the docs of ariel-os-identity (which BTW I think we should have built somewhere if only to link from here), this identifier stays constant across using different applications -- neither the application name nor the build path are factored in. 

ok

> This can be a bit weird because there _can_ be multiple instances of a native application running at the same time, but then, if those are networked, the user will have to do manual assignment of the devices to tap interfaces etc as well.

So I guess once there is networking, we'll need to maybe hash the interface path into this. The device identity is used to calculate the tap's Ariel side MAC address, so it needs to be unique.

src/ariel-os-native/Cargo.toml
chrysn: Yes, I didn't want to have too much churn on the Cargo.toml, that just causes merge conflicts.

(The dependency should generally reflect what is intended and tested; I did test with 0.10.8, so it's fine, although I'm pretty sure I don't depend on anything beyond 0.10.0)

src/ariel-os-native/src/identity.rs
chrysn: Yes, making it std would be the right thing, will change this here.

chrysn:
> So I guess once there is networking, we'll need to maybe hash the interface path into this. The device identity is used to calculate the tap's Ariel side MAC address, so it needs to be unique.

I'm softly leaning toward factoring that in only into the eui48 calculation then, because that'd allow other identifiers (such as a claimed name in a cryptographic credential) to persist across changing the network interfaces in a build.

chrysn:
There were a few weirdly non-interactable comments in the PR view. I tried whether a rebase-force-push helps GitHub make up its mind about them, but apparently it doesn't. @ROMemories, can you mark them as resolved?

src/ariel-os-native/Cargo.toml
ROMemories: > I didn't want to have too much churn on the Cargo.toml

I think you mean the `Cargo.lock`? We don't currently have a direct dependency on `sha2`.

Otherwise makes sense, even though that dependency could have been defined in the workspace's manifest now that I think about it, it's likely not going to be the only crate using it.

ROMemories:
> There were a few weirdly non-interactable comments in the PR view. I tried whether a rebase-force-push helps GitHub make up its mind about them, but apparently it doesn't. @ROMemories, can you mark them as resolved?

To me all the threads show up as resolved by you, so I have nothing to do.

Why did you remove the `ci-build:small` label, is this not pretty much ready to merge?

chrysn:
I removed it temporarily while I was accepting changes, then sending my own changes addressing some comments, and didn't want to cancel to many CI jobs keeping others up.

Re-enabled now, and ready from my PoV. (Nothing to do on the documentation side, as native is not yet in the hardware matrix).

PR #1158, opened 2025-07-18
ci(build-pr-labels): introduce even more build labels

ROMemories:
# Description

Following #1153, this introduces a `ci-build:*` label for every HAL we support, and for the new native builder introduced by #647.

## How to review this PR

This PR is better reviewed commit by commit.
Check the CI logs for the `LAZE_BUILDERS` value, and the set of builders actually built for, for each relevant CI run.

## Issues/PRs references

Depends on #1159.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1167, opened 2025-07-18
ci(build): install the Xtensa toolchain only when needed

ROMemories:
# Description

This relies on the set of builders from `LAZE_BUILDERS` to determine whether installing the Xtensa toolchain is needed. Installing it takes a least one minute in CI, for each job, and sometimes takes much longer due to network fluctuations, so this should improve the performance and reliability of the CI a bit.

## How to review this PR

This PR is much better reviewed commit by commit.

## Issues/PRs references

Made possible by #1158.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -178,7 +178,7 @@ jobs:
           components: rust-src, rustfmt
 
       - name: Install Rust for Xtensa
-        if: steps.should-skip.outputs.result != 'true'
+        if: steps.should-skip.outputs.result != 'true' && (env.LAZE_BUILDERS == null || contains(env.LAZE_BUILDERS, 'esp32-s3'))
         # No tag yet after https://github.com/esp-rs/xtensa-toolchain/pull/38 (will probably be @v1.5.5)
         uses: esp-rs/xtensa-toolchain@main
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -150,7 +178,7 @@ jobs:
           components: rust-src, rustfmt
 
       - name: Install Rust for Xtensa
-        if: steps.should-skip.outputs.result != 'true'
+        if: steps.should-skip.outputs.result != 'true' && (env.LAZE_BUILDERS == null || contains(env.LAZE_BUILDERS, 'esp32-s3'))
```
ROMemories, line 181: [Documentation of `contains()`](https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#contains)

This is a bit of a crude check, but:
- in case of a false positive, this just costs an extra minute of CI (and it would be weird for a builder to have `esp32-s3` in its name and not actually use an ESP32-C3) and,
- in case of a false negative (the Xtensa toolchain not being installed when it was needed), compilation would just fail, which would be annoying but not dangerous (this also cannot happen with the *current* way of building `LAZE_BUILDERS`).

ROMemories, line 181: `env.LAZE_BUILDERS == null` is needed to handle the full build, be it triggered by `ci-build:full` or the scheduled build.

ROMemories:
Tested the `ci-build:esp` and `ci-build:full` labels, they both triggered the Xtensa toolchain installation. Installation is skipped as expected when using `ci-build:native` for instance.

kaspar030 (approved):

kaspar030:
undraft?

PR #1168, opened 2025-07-21
fix(nrf-ble): set executor interrupt priority

### LINKED ISSUE #1165: nrf-ble: investigate why the nrf-sdc crate doesn't work with the interrupt executor
Currently using BLE on an nRF MCU disables the interrupt executor because it causes an error in the MSPL.

After chatting a bit in Embassy's Matrix channel it sounds like it's an issue with irq priorities.

The objective is to be able to have BLE work on nRF boards when using the interrupt executor.

nponsard:
# Description

This allows to use ble with the interrupt executor on nRF boards.

Tested on nrf5340dk-net, nrf52dk.

## Issues/PRs references

Fixes #1165 

## Open Questions

Seems like usb on stm32 also needs the priority to be changed, should we make P1 the default ?

## Changelog entry

The `executor-interrupt` flavor is now using the lowest interrupt priority on STM32 and nRF MCUs. This allows using BLE in combination with that executor flavor on nRF.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1262,7 +1262,6 @@ modules:
       - doc_only
 
   - name: has_nrf_ble
-    help: Marks an incompatibility of nrf devices with the interrupt executor (MPSL crashing)
     context:
       - nrf52
       - nrf53
@@ -1270,8 +1269,6 @@ modules:
       - hwrng
     provides:
       - has_ble
-    conflicts:
-      - executor-interrupt
 
   - name: ble-peripheral
     help: Enables BLE peripheral functionality
--- a/src/ariel-os-embassy/src/lib.rs
+++ b/src/ariel-os-embassy/src/lib.rs
@@ -152,6 +152,12 @@ pub(crate) fn init() {
     debug!("ariel-os-embassy::init(): using interrupt mode executor");
     let p = hal::init();
 
+    #[cfg(any(context = "nrf", context = "stm32"))]
+    {
+        use crate::hal::interrupt::{InterruptExt, Priority};
+        hal::SWI.set_priority(Priority::P1);
+    }
+
     #[cfg(any(context = "nrf", context = "rp", context = "stm32"))]
     {
         hal::EXECUTOR.start(hal::SWI);
--- a/src/ariel-os-nrf/src/irqs.rs
+++ b/src/ariel-os-nrf/src/irqs.rs
@@ -25,7 +25,7 @@ bind_interrupts!(pub(crate) struct Irqs {
     #[cfg(all(feature = "ble", context = "nrf52"))]
     EGU1_SWI1 => nrf_sdc::mpsl::LowPrioInterruptHandler;
     #[cfg(all(feature = "ble", context = "nrf53"))]
-    EGU0 => nrf_sdc::mpsl::LowPrioInterruptHandler;
+    SWI0 => nrf_sdc::mpsl::LowPrioInterruptHandler;
 
     #[cfg(feature = "ble")]
     RADIO => nrf_sdc::mpsl::HighPrioInterruptHandler;


src/ariel-os-embassy/src/lib.rs
```diff
@@ -140,6 +140,13 @@ pub(crate) fn init() {
 
     #[cfg(any(context = "nrf", context = "rp", context = "stm32"))]
     {
+        // nrf-sdc doesn't like when the SWI is priority 0.
+        #[cfg(all(context = "nrf", feature = "ble"))]
```
kaspar030: I think we should have this on P1 even if ble is not used, so all the testing applies everywhere.

```suggestion
        #[cfg(all(context = "nrf"))]
```

kaspar030: ... just reading the stm32 question. let's try making P1 default!

nponsard: Okay I'll try some examples on the boards I have to see if something breaks.

nponsard: This change prevents `examples/tcp-echo/` to function correctly on `rpi-pico-w`, no log output appears where on main I get log output from the example.

kaspar030: > This change prevents `examples/tcp-echo/` to function correctly on `rpi-pico-w`, no log output appears where on main I get log output from the example.

Can you try rpi-pico-w with the change, but build with `--select executor-thread`?
Maybe `PIO0_IRQ_0` needs to be higher than SWI?

nponsard: With executor thread it works fine, the change is not built into the final binary (only affects `executor-interrupt`)

kaspar030: [here](https://github.com/ariel-os/ariel-os/blob/aaefefb260aebccfa67187ad725de51f9f3d8e48/src/ariel-os-rp/src/cyw43.rs#L65) would be the point to set a different `PIO0_IRQ_0` priority. The default is P0 (highest priority) IIUC, so if setting SWI (the embassy executor prio) to P1, maybe setting `PIO0_IRQ_0` to P1 as well helps?

nponsard: I'll try that thanks

nponsard: Apparently the default priority for the SWI on the rpi pico is P3

kaspar030: ah: https://github.com/ariel-os/ariel-os/blob/f4b1b49e77f21502c4806f0fc9bd6d0755b35f18/src/ariel-os-rp/src/lib.rs#L77-L79

That might explain why overriding the prio to P1 doesn't work.

nponsard: Ahh yeah, should I make the interrupt priority change only on stm32 and nrf ?

src/ariel-os-nrf/src/lib.rs
```diff
@@ -55,7 +55,14 @@ ariel_os_embassy_common::executor_swi!(SWI0);
 ariel_os_embassy_common::executor_swi!(EGU0_SWI0);
 
 #[cfg(feature = "executor-interrupt")]
-#[cfg(any(context = "nrf53", context = "nrf91"))]
+#[cfg(context = "nrf5340-net")]
```
kaspar030: this seems unrelated

nponsard: `EGU0` is used by `nrf-sdc` for ble on nrf5340. There was no conflict before because BLE on nRF MCUs only used the thread executor.
Now we need to change the SWI for BLE to still work on nrf5340.

kaspar030: can we just default to using EGU1 for the nrfs, for use as SWI? less special cases ...

nponsard: I just checked I can change the interrupt used for the BLE, I'll go for that

laze-project.yml
```diff
@@ -1231,17 +1231,16 @@ modules:
     selects:
       - doc_only
 
-  - name: has_nrf_ble
-    help: Marks an incompatibility of nrf devices with the interrupt executor (MPSL crashing)
-    context:
-      - nrf52
-      - nrf53
+  - name: nrf-ble
```
kaspar030: these seem unrelated. remove and merge?

PR #1171, opened 2025-07-21
fix(deps): use revision instead of branch

### LINKED ISSUE #1170: build: reduce re-checking of repositories
Currently when building Ariel OS, Cargo re-checks our remote repose (e.g., the crates.io patches from `ariel-os-cargo.toml`) multiple times.

I suspect that Cargo just re-checks branches on every call. Maybe we can reduce this by specifying the concrete revisions, using `rev` instead of branch?

So,

1. replace all branch specifiers in `ariel-os-cargo.toml` with commit hashes
2. keep the branch names in a comment
3. check whether Cargo does less git activity

nponsard:
# Description

This avoids cargo doing requests to check for updates on the branch.

Checked using: 

```sh 
CARGO_HOME=/tmp/foo cargo fetch                                                                                                                                                                                                                     
CARGO_HOME=/tmp/foo CARGO_NET_OFFLINE=true laze build -g -b nrf52840dk
```

This would error out when using the branch attribute, now the build step doesn't need to connect to the internet.

I get a weird error when trying the same for `espressif-esp32-s3-devkitc-1`:
```
error: no matching package named `rustc-literal-escaper` found
location searched: crates.io index
required by package `proc_macro v0.0.0 (/home/nponsard/.rustup/toolchains/esp/lib/rustlib/src/rust/library/proc_macro)`
    ... which satisfies path dependency `proc_macro` (locked to 0.0.0) of package `sysroot v0.0.0 (/home/nponsard/.rustup/toolchains/esp/lib/rustlib/src/rust/library/sysroot)`
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without `--offline`.
```

Trying without offline mode works fine.

*edit by @kaspar030* some numbers:

On main:

```
❯ hyperfine "laze -Cexamples/hello-world build -b espressif-esp32-s3-devkitc-1"
Benchmark 1: laze -Cexamples/hello-world build -b espressif-esp32-s3-devkitc-1
  Time (mean ± σ):      3.481 s ±  0.062 s    [User: 0.681 s, System: 0.446 s]
  Range (min … max):    3.416 s …  3.629 s    10 runs
```

This branch:
```
❯ hyperfine "laze -Cexamples/hello-world build -b espressif-esp32-s3-devkitc-1"
Benchmark 1: laze -Cexamples/hello-world build -b espressif-esp32-s3-devkitc-1
  Time (mean ± σ):     998.2 ms ±  13.6 ms    [User: 614.7 ms, System: 412.9 ms]
  Range (min … max):   979.0 ms … 1023.5 ms    10 runs
```

## Issues/PRs references

Fixes #1170.

## Open Questions

<!-- Unresolved questions, if any. -->

## Change checklist

<!--
We don't enforce a strict convention for commit messages,
but please make sure that:
- the commit history is clear and informative.
- the Developer Certificate of Origin (DCO) Sign-off is present
  in your commits. 
  - See https://github.com/ariel-os/ariel-os/blob/main/CONTRIBUTING.md#developer-certificate-of-origin
-->

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1725,7 +1725,7 @@ dependencies = [
 [[package]]
 name = "embassy-embedded-hal"
 version = "0.3.0"
-source = "git+https://github.com/ariel-os/embassy?branch=embassy-embedded-hal-v0.3.0%2Bariel-os#21147bdceb85f55e375973f30e47c17036063689"
+source = "git+https://github.com/ariel-os/embassy?rev=21147bdceb85f55e375973f30e47c17036063689#21147bdceb85f55e375973f30e47c17036063689"
 dependencies = [
  "defmt 0.3.100",
  "embassy-futures",
@@ -1741,7 +1741,7 @@ dependencies = [
 [[package]]
 name = "embassy-executor"
 version = "0.7.0"
-source = "git+https://github.com/ariel-os/embassy?branch=embassy-executor-v0.7.0%2B04.04.25%2Bariel-os#b552a87578479a92b3027bd208c712445499589c"
+source = "git+https://github.com/ariel-os/embassy?rev=b552a87578479a92b3027bd208c712445499589c#b552a87578479a92b3027bd208c712445499589c"
 dependencies = [
  "cortex-m",
  "critical-section",
@@ -1752,7 +1752,7 @@ dependencies = [
 [[package]]
 name = "embassy-executor-macros"
 version = "0.6.2"
-source = "git+https://github.com/ariel-os/embassy?branch=embassy-executor-macros-v0.6.2%2B04.04.25#1b08496e441cfd2b2e9660fe0cc589c3bc5f0824"
+source = "git+https://github.com/ariel-os/embassy?rev=1b08496e441cfd2b2e9660fe0cc589c3bc5f0824#1b08496e441cfd2b2e9660fe0cc589c3bc5f0824"
 dependencies = [
  "darling",
  "proc-macro2",
@@ -1772,7 +1772,7 @@ dependencies = [
 [[package]]
 name = "embassy-hal-internal"
 version = "0.2.0"
-source = "git+https://github.com/ariel-os/embassy?branch=embassy-hal-internal-v0.2.0%2Bariel-os#ed7d22c1881b64f9bdd2e69a5777b002279033a6"
+source = "git+https://github.com/ariel-os/embassy?rev=ed7d22c1881b64f9bdd2e69a5777b002279033a6#ed7d22c1881b64f9bdd2e69a5777b002279033a6"
 dependencies = [
  "cortex-m",
  "critical-section",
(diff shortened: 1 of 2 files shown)


kaspar030:
> rustc-literal-escaper

that's probably due to esp being built with `-Zbuild-std`. let me check....
yes, seems like Cargo doesn't fetch those deps, there are some others.
Let's go with this as is!

kaspar030:
Thanks!

ROMemories:
After testing this, I am not 100% this makes a difference?

`CARGO_HOME=/tmp/foo CARGO_NET_OFFLINE=true laze build -g -b nrf52840dk` still errors out on this PR with the same error as in #1070 (that is expected), but I cannot reproduce this command erroring out differently on main.

nponsard:
> After testing this, I am not 100% this makes a difference?
> 
> `CARGO_HOME=/tmp/foo CARGO_NET_OFFLINE=true laze build -g -b nrf52840dk` still errors out on this PR with the same error as in #1070 (that is expected), but I cannot reproduce this command erroring out differently on main.

On stable have you tried with 

```sh 
CARGO_HOME=/tmp/foo cargo fetch --config ariel-os-cargo.toml                                                                                                                                                                                            
CARGO_HOME=/tmp/foo CARGO_NET_OFFLINE=true laze build -g -b nrf52840dk
```
?

Edit: I just tried, indeed its the same state between main and this PR when using stable.

kaspar030:
Anyhow, this was not meant to make building with "--offline" work, but to reduce network requests.

I'm trying to get numbers, but `CONFIG_WIFI_PASSWORD=foo CONFIG_WIFI_NETWORK=bar laze build -g -b espressif-esp32-s3-devkitc-1 -d http-server` seems much faster on this branch vs. main.

On this branch, (after the first fetch ofc), the "updating repository..." lines just fly by, whereas on main, each takes a bit (< 1sec each but still).

kaspar030:
> Anyhow, this was not meant to make building with "--offline" work, but to reduce network requests.
> 
> I'm trying to get numbers, but `CONFIG_WIFI_PASSWORD=foo CONFIG_WIFI_NETWORK=bar laze build -g -b espressif-esp32-s3-devkitc-1 -d http-server` seems much faster on this branch vs. main.
> 
> On this branch, (after the first fetch ofc), the "updating repository..." lines just fly by, whereas on main, each takes a bit (< 1sec each but still).

I got a quicker reproducer: `laze -Cexamples/hello-world build -b espressif-esp32-s3-devkitc-1`

On this branch, repeat builds take less than a sec on my system, whereas on main, it takes ~3.5 sec, which is mostly spent on "Updating git repository ...".

nponsard:
perviously I was getting errors like this on main with nightly:

```
error: failed to load source for dependency `embedded-test`

Caused by:
  Unable to update https://github.com/ariel-os/embedded-test?branch=v0.6.2%2Bariel-os

Caused by:
  failed to lookup reference in preexisting repository, and can't check for updates in offline mode (--offline)

Caused by:
  failed to find branch `v0.6.2+ariel-os`

Caused by:
  cannot locate remote-tracking branch 'origin/v0.6.2+ariel-os'; class=Reference (4); code=NotFound (-3)
```

Switching the dependency to a specific commit seemed to fix it.
But now I can't replicate the error on the main branch.

nponsard:
> I got a quicker reproducer: `laze -Cexamples/hello-world build -b espressif-esp32-s3-devkitc-1`
> 
> On this branch, repeat builds take less than a sec on my system, whereas on main, it takes ~3.5 sec, which is mostly spent on "Updating git repository ...".

I observe the same results

ROMemories:
> On this branch, repeat builds take less than a sec on my system, whereas on main, it takes ~3.5 sec, which is mostly spent on "Updating git repository ...".

I can reproduce the "Updating git repository" lines being faster :rocket:

Issue #1182, opened 2025-07-22
Hardware support documentation redesign


ROMemories:
As our list of supported hardware is growing and we want to provide more and more fine-grained information, we will soon need to redesign our documentation of hardware support.

In particular, we want to allow for the following:

-  [x] Documenting support for MCUs that do not have off-the-shelf boards available; this is especially useful for STM32 MCUs (e.g, #1183)
- [x] Introducing board pages, so that we can document their particularities—e.g., how to flash them #876
- [x] Rendering the `comments` we provide in the (likely to evolve) `support_matrix.yml` file
- [ ] Introducing non-orthogonal labels for boards: e.g., devboard, breakout board, module, gadget (exact names TBD)
- [ ] Possibly introducing user-visible kebab-case feature names—similar to [what Zephyr does](https://docs.zephyrproject.org/latest/boards/nordic/nrf52840dk/doc/index.html)—so we can introduce more fine-grained features if needed

These may require completely reworking how we generate that documentation, especially as we will need to be able to generate multiple pages, and we will to come up with a way to write these board pages, factoring out common notes as reusable snippets.

ROMemories:
We'll eventually want to generate one page per board, so we may need to go back to the setup we had before we switched to `mdbook-cmdrun` in #795.

PR #1183, opened 2025-07-23
feat(stm32): add support for the STM32U073KC

ROMemories:
# Description

This adds support for the STM32U073KC MCU: the STM32U073 is the same as the STM32U083 we already support, the only difference being it does not have an AES accelerator. Adding support for it is useful because it is slightly cheaper and, maybe more importantly for now, is more stocked by PCBA manufacturers.

There is no official development kit with that MCU (the closest one is the STM32U083C-DK which we do support), so this does not add a board in the support matrix (nor to the CI), only the MCU. How we render support for MCUs without a board is part of #1182.

## Change checklist

### DIFF
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -325,6 +325,21 @@ chips:
           - removing items not supported
       wifi: not_available
 
+  stm32u073kc:
+    name: STM32U073KC
+    support:
+      gpio: supported
+      debug_output: supported
+      hwrng: supported
+      i2c_controller: needs_testing
+      spi_main: needs_testing
+      logging: supported
+      storage:
+        status: supported_with_caveats
+        comments:
+          - removing items not supported
+      wifi: not_available
+
   stm32u083mc:
     name: STM32U083MC
     support:
--- a/src/ariel-os-stm32/src/lib.rs
+++ b/src/ariel-os-stm32/src/lib.rs
@@ -222,7 +222,7 @@ fn board_config(config: &mut Config) {
         config.rcc.mux.spi123sel = mux::Saisel::PLL1_Q; // Reset value
     }
 
-    #[cfg(context = "stm32u083mc")]
+    #[cfg(any(context = "stm32u073kc", context = "stm32u083mc"))]
     {
         use embassy_stm32::rcc::*;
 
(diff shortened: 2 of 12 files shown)


doc/support_matrix.yml
```diff
@@ -325,6 +325,21 @@ chips:
...
       wifi: not_available
 
+  stm32u073kc:
+    name: STM32U073KC
+    support:
+      gpio: supported
+      debug_output: supported
+      hwrng: supported
+      i2c_controller: needs_testing
+      spi_main: needs_testing
```
ROMemories, line 335: Should we consider these `supported` because they are in the `stm32u083mc` and these MCUs are almost exactly the same (except for the AES accelerator)?

src/ariel-os-storage/build.rs
```diff
@@ -7,21 +7,26 @@ fn main() {
...
-        } else if is_in_current_contexts(&["stm32h755zi"]) {
-            (256 * KIBIBYTES, 128 * KIBIBYTES)
-        } else if !is_in_current_contexts(&["ariel-os"]) {
-            // Dummy value for platform-independent tooling.
-            (8 * KIBIBYTES, 4 * KIBIBYTES)
-        } else {
-            panic!("MCU not supported");
-        };
+    let (storage_size_total, flash_page_size) = if is_in_current_contexts(&[
+        "stm32u073kc",
```
ROMemories, line 11: The only change is adding this line, the rest is auto-reformatting.

src/ariel-os-stm32/src/lib.rs
```diff
@@ -222,7 +222,7 @@ fn board_config(config: &mut Config) {
         config.rcc.mux.spi123sel = mux::Saisel::PLL1_Q; // Reset value
     }
 
-    #[cfg(context = "stm32u083mc")]
+    #[cfg(any(context = "stm32u073kc", context = "stm32u083mc"))]
```
ROMemories, line 225: This doesn't seem to conflict with #1388, I tested it locally.

doc/support_matrix.yml
kaspar030: not unless you've tested them on some board I'd say

kaspar030: "needs_testing" means it compiles, we assume it working but have never seen it work ourselves or by proxy

src/ariel-os-storage/build.rs
kaspar030: I very much like "hide whitespace" in GHs diff options for these.

Issue #942, opened 2025-03-25
Guide to using Visual Studio Code for Ariel OS


emmanuelsearch:
### Description

Provide a short how-to support for coding with this editor.

### Motivation

Visual Studio Code is an extremely popular editor...

### Requirements

Provide a short how-to support for coding with this editor.

### Are you planning to do it yourself in a pull request?

No

nponsard:
I've started taking notes on my different vscode configurations here: https://pad.riot-os.org/vbeZBdF4Sg2t1L814xCBfg?view

I plan on automating the generation of the configurations and documenting the useage in the book.

nponsard:
The next step is to have support for developing inside the ariel-os repository.

nponsard:
I'm stumbling on an issue where some parts of rust-analyzer still use the resolver version 1, meaning all the target-specific features get enabled, for example in `ariel-os-nrf`:

```toml
[target.'cfg(context = "nrf52840")'.dependencies]
embassy-nrf = { workspace = true, features = ["nfc-pins-as-gpio", "nrf52840"] }
nrf-sdc = { workspace = true, features = ["nrf52840"], optional = true }

[target.'cfg(context = "nrf5340-net")'.dependencies]
embassy-nrf = { workspace = true, features = ["nrf5340-net"] }
nrf-sdc = { workspace = true, features = ["nrf5340-net"], optional = true }
```

With only `--cfg context="nrf5340-net"` in the `RUSTFLAGS`, rust analyzer will still consider that the features `nrf5340-net` and `nrf52840` are enabled on `nrf-sdc` and `embassy-nrf`, leading to errors like "expected 10 arguments, found 6" in `ariel-os/src/ariel-os-nrf/src/ble.rs`.

This issue seems to be tracked in https://github.com/rust-lang/rust-analyzer/issues/15233 . I also created a forum post about it : https://users.rust-lang.org/t/rust-analyzer-seems-to-use-resolver-1/132235/

PR #1186, opened 2025-07-23
fix(vscode-config): copy RUSTFLAGS

nponsard:
# Description

This copies the RUSTFLAGS environment variable into the environment of rust-analyzer, without the target prefix. This solves a problem where some of the tools used by rust-analyzer do not read the RUSTFLAGS of the target, resulting in situations where jumping to definition would lead to the dummy implementation.

## Issues/PRs references

Related to #942 

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -199,7 +199,7 @@ contexts:
           - CARGO_ARGS
           - CARGO_ENV
         cmd: # We have to obfuscate the variables to avoid cargo interpreting them since we're running a rust script
-          - _CARGO_TOOLCHAIN='${CARGO_TOOLCHAIN}' ${relroot}/${root}/scripts/vscode.rs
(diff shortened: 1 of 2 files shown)


anlavandier:
Do you have an example of something that is fixed by this so that I can check it on my side ?

nponsard:
> Do you have an example of something that is fixed by this so that I can check it on my side ?

Here's how I tested it on my side on this project: https://github.com/nponsard/ble-sniffer-ariel

```
laze build -b nrf5340dk-net vscode-config
```

Check in `src/main.rs` if going to the definition of `ariel_os::ble::ble_stack()` correctly leads to `imports/ariel-os/src/ariel-os-nrf/src/ble.rs`.

Previously it was leading to `imports/ariel-os/src/ariel-os-hal/src/dummy/ble.rs`

anlavandier (approved):
Tested with your example and it worked.

PR #1188, opened 2025-07-24
refactor(coap): use `UdpSocket::endpoint()` instead of `.with()` patch

ROMemories:
# Description

This uses the upstream `UdpSocket::endpoint()` instead of `UdpSocket::with()` that is only only made accessible by our patch `embassy-net-v0.6.0+ariel-os` in https://github.com/embassy-rs/embassy/commit/7f577b5a7506f9e5e41ec11dff6f44cea589c894.

I am however unsure why `.with()` was used in the first place, so it is possible that this is not actually equivalent, even though [`UdpSocket::endpoint()` *seems* to me to be simply delegating to `smoltcp::socket::udp::Socket::endpoint()`](https://github.com/embassy-rs/embassy/blob/953288f5c831e7048d38479db8f43cd8126807dd/embassy-net/src/udp.rs#L317-L319).

@chrysn If this substitution does not actually make sense, feel free to just close this PR.

---

This would allow us to basically do without our `embassy-net` patch, as the only other thing it does is printing the IP address at `info` level, which we could do some other way.

## Testing

I *only* tested this by removing our `embassy-net` patch, making sure it did not compile, then applying this change and making sure it compiled again (the IP address not being printed as expected when the log level is set as `info`). I am not sure how to test this at runtime.

## Change checklist

### DIFF
--- a/src/ariel-os-coap/src/udp_nal/mod.rs
+++ b/src/ariel-os-coap/src/udp_nal/mod.rs
@@ -141,7 +141,7 @@ impl nal::UnconnectedUdp for UnconnectedUdp<'_> {
         // information, so the underlying layers won't even have a *chance* to care if we don't
         // check here.
         debug_assert!(
-            local.port() == 0 || local.port() == self.socket.with(|s, _| s.endpoint().port),
+            local.port() == 0 || local.port() == self.socket.endpoint().port,
             "Port of local address, when given, must match bound port."
         );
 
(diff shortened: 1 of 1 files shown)


chrysn:
Honestly my best guess is that I used `.with()` because I originally developed this for use in embassy-net, then things got slow and I moved the code in here, and then I missed that for the very functionality I needed `.with()` for, there was a public version.

Thanks for spotting this; I'll run an actual test (running the CoAP demo), but I don't see how things could break – and then mark as approved.

ROMemories:
> I take it you prefer to remove the patch in a separate PR?

Yes, will do later, thanks!

PR #1191, opened 2025-07-25
docs(sensors): add an example demonstrating homogeneous sensor access

ROMemories:
# Description

*This is very WIP.* At least another sensor driver needs to be added for this example to be able to demonstrate accessing multiple sensors.

This PR also contains patches to the sensor API, which will need to be split out.

## Change checklist

### DIFF
--- a/examples/sensors-debug/Cargo.toml
+++ b/examples/sensors-debug/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "sensors-debug"
+license.workspace = true
+edition.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["i2c", "sensors", "time"] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os-builtin-sensors = { path = "../../src/ariel-os-builtin-sensors" }
+ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
+cfg-if = { workspace = true }
+embassy-sync = { workspace = true }
+embedded-hal = { workspace = true }
+embedded-hal-async = { workspace = true }
+linkme = { workspace = true }
+once_cell = { workspace = true }
--- a/examples/sensors-debug/laze.yml
+++ b/examples/sensors-debug/laze.yml
@@ -1,0 +1,4 @@
+apps:
+  - name: sensors-debug
+    context:
+      - stm32u083c-dk
--- a/examples/sensors-debug/src/main.rs
+++ b/examples/sensors-debug/src/main.rs
@@ -1,0 +1,119 @@
+//! This example is merely to illustrate and test raw bus usage.
+//!
+//! Please use [`ariel_os::sensors`] instead for a high-level sensor abstraction that is
+//! HAL-agnostic.
+//!
+//! This example requires an onboard sensor or an external LIS3DH/LSM303AGR sensor (3-axis
+//! accelerometer).
+#![no_main]
+#![no_std]
+
+mod pins;
+mod sensors;
+
+use ariel_os::{
+    asynch::Spawner,
+    debug::log::{debug, error, info},
+    hal,
+    i2c::controller::{I2cDevice, Kilohertz, highest_freq_in},
+    sensors::{REGISTRY, Reading, sensor::Accuracy},
+    time::Timer,
+};
+use ariel_os_builtin_sensors::stts22h;
+use embassy_sync::mutex::Mutex;
+
+pub static I2C_BUS: once_cell::sync::OnceCell<
+    Mutex<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, hal::i2c::controller::I2c>,
+> = once_cell::sync::OnceCell::new();
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    let mut i2c_config = hal::i2c::controller::Config::default();
+    i2c_config.frequency = const { highest_freq_in(Kilohertz::kHz(100)..=Kilohertz::kHz(400)) };
+    debug!("Selected frequency: {:?}", i2c_config.frequency);
+
+    let i2c_bus = pins::SensorI2c::new(peripherals.i2c_sda, peripherals.i2c_scl, i2c_config);
+
+    let _ = I2C_BUS.set(Mutex::new(i2c_bus));
+
+    let i2c_device = I2cDevice::new(I2C_BUS.get().unwrap());
+
+    let spawner = Spawner::for_current_executor().await;
+
+    sensors::STTS22H_I2C
+        .init(
+            spawner,
+            stts22h::Peripherals {},
+            i2c_device,
+            stts22h::Config::default(),
+        )
+        .await;
+
+    spawner.spawn(sensors::stts22h_i2c_runner()).unwrap();
+
+    loop {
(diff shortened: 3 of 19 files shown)


examples/sensors-debug/src/pins.rs
```diff
@@ -0,0 +1,104 @@
...
+    i2c_scl: PIN_13,
+});
+
+#[cfg(context = "st-steval-mkboxpro")]
+pub type SensorI2c = i2c::controller::I2C1;
+#[cfg(context = "st-steval-mkboxpro")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    i2c_sda: PB7,
+    i2c_scl: PB6,
+});
```
ROMemories, line 48: This file is identical to [`tests/i2c-controller/src/pins.rs`](https://github.com/ariel-os/ariel-os/blob/404387ec4ffe90af79bbc2afdc4afe4bded487fc/tests/i2c-controller/src/pins.rs), except this board is added here. These will become unnecessary as soon as we're able to declare these pins in the SBD files.

examples/sensors-debug/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    let mut i2c_config = hal::i2c::controller::Config::default();
+    i2c_config.frequency = const { highest_freq_in(Kilohertz::kHz(100)..=Kilohertz::kHz(400)) };
+    debug!("Selected frequency: {:?}", i2c_config.frequency);
+
+    let i2c_bus = pins::SensorI2c::new(peripherals.i2c_sda, peripherals.i2c_scl, i2c_config);
+    let _ = I2C_BUS.set(Mutex::new(i2c_bus));
+    let i2c_device = I2cDevice::new(I2C_BUS.get().unwrap());
```
nponsard: I don't understand why i2c is initialized here since it's not used in the example. In my opinion this should be added when the first sensor needing i2c is added to Ariel OS.

examples/sensors-debug/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        Timer::after_secs(2).await;
+    }
+}
+
+fn print_sample(sensor: &dyn Sensor, sample: Sample, reading_channel: ReadingChannel) {
+    let display_name = sensor.display_name().unwrap_or("unknown");
+    let label = sensor.label().unwrap_or("no label");
+
+    let value = match sample.value() {
```
nponsard, line 68: Opaque samples should be displayed differently (currently there's no check for it).

examples/sensors-debug/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+  "sensors",
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
+embassy-sync = { workspace = true }
+embedded-hal = { workspace = true }
+embedded-hal-async = { workspace = true }
+once_cell = { workspace = true }
+
```
nponsard, line 17: A dependency on `linkme` will be needed here when we add the first Sensor.

ROMemories: Actually we don't even need it, I've updated #1412 to use the `#[linkme]` attribute and make it use the reexport, avoiding the need for the direct dependency. Well spotted nonetheless!

examples/sensors-debug/src/main.rs
ROMemories: I guess initially I was seeing this example as a more low-level debug tool, but I agree it's actually better to demonstrate that here (and the three lines can always be commented out for testing if needed). Could you try it out on #1330?

examples/sensors-debug/src/main.rs
ROMemories: I'd rather not have to duplicate this in the five WIP sensor drivers PRs we have that use I2C, as that increases the likelihood of merge conflicts between these. I don't think having such temporarily unused piece of code is really an issue. This is already tentative anyway as that depends on the declarative configuration/codegen we'll have for this in the future.

examples/sensors-debug/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "i2c",
+  "sensors",
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
+embassy-sync = { workspace = true }
+once_cell = { workspace = true }
```
ROMemories, line 16: I've also removed the dependency on `embedded-hal` and `embedded-hal-async` as they were actually not used (even when bringing in sensor drivers in the example).

examples/sensors-debug/Cargo.toml
nponsard: Oh didn't know about that, added it to the GNSS sensor declaration and it works :+1:

examples/sensors-debug/src/main.rs
nponsard: Opaque samples correctly get ignored, since its a debug program maybe we can print out the raw value using a lower log level (and a description saying its opaque) ?

ROMemories: > since its a debug program maybe we can print out the raw value using a lower log level (and a description saying its opaque) ?

I like the idea of logging them as `debug` in principle, but the six existing `info!` statements would need to handle that and without an dynamic allocator I don't think that composes too well :/

nponsard: Could just be before the return: 
```rust
        debug!("Opaque value : {:?}", sample.value());
```
Its for debugging and with a lower level, in my opinion it doesn't need to be pretty.

examples/sensors-debug/src/main.rs
ROMemories: After some offline discussion with @kaspar030, we've decided to move this initialization to a dedicated module file, and clearly indicate it is also meant to be codegened.

examples/sensors-debug/src/sensors.rs
```diff
@@ -0,0 +1,8 @@
+//! This module is intended to contain the auto-@generated instantiation and registration of sensor
+//! drivers.
+
+// This example does not currently register any sensor drivers, they will be added later.
+
+pub async fn init() {
+    // Sensor driver instances are to be initialized here.
+}
```
ROMemories, line 8: I've also introduced this init function, so that the initialization of sensor driver instances can be moved out of `main`.

examples/sensors-debug/src/main.rs
ROMemories: I've added a `debug!` statement with a bit more details, the output looks like this (faked an opaque channel on another sensor driver for testing):

```sh
DEBUG 6-axis IMU (onboard): Ok(-57378) (Opaque)
```

examples/sensors-debug/laze.yml
```diff
@@ -0,0 +1,2 @@
+apps:
+  - name: sensors-debug
```
nponsard, line 2: The build context needs to be restricted otherwise the nightly build will try this example on every board including those where `i2c` is not supported.

ROMemories: Good catch! Copied the list of contexts from `i2c-controller`, and added `st-steval-mkboxpro` to it as we do provide pins for it in this example.

PR #1197, opened 2025-08-05
feat: cellular networking using nrf-modem

nponsard:
# Description

Connects the modem to `embassy-net`.

## Issues/PRs references

- depends on #1398
- depends on #1955 

Documentation PR: #1795 

## Testing

The nrf91 series doesn't have a supported `hwrng`, use #1690: 

```
laze -C examples/tcp-client build -b nrf9160dk-nrf9160 run 
```

To test IPv6, change the tcp-client example:

```diff
--- a/examples/tcp-client/src/main.rs
+++ b/examples/tcp-client/src/main.rs
@@ -32,7 +32,7 @@ async fn tcp_echo() {
 
         info!("Connecting...");
         // Connect to https://tcpbin.com/ without using DNS
-        let host_addr = embassy_net::Ipv4Address::from_str("45.79.112.203").unwrap();
+        let host_addr = embassy_net::Ipv6Address::from_str("2600:3c01::f03c:91ff:feab:f98b").unwrap();
         if let Err(e) = socket.connect((host_addr, 4242)).await {
             warn!("connect error: {:?}", e);
             Timer::after_secs(10).await;
```

and then run:

```sh
laze -C examples/tcp-client build -s ipv6 -d ipv4 -b nrf9160dk-nrf9160 -DLOG=debug run    
```

## Changelog Entry

Support for LTE-M has been added for the nrf91 series of MCUs.

## Change checklist

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -99,6 +99,11 @@ who = "chrysn <chrysn@fsfe.org>"
 criteria = "safe-to-deploy"
 delta = "0.0.4 -> 0.0.8"
 
+[[audits.nrf-modem]]
+who = "Nils Ponsard <nils.ponsard@inria.fr>"
+criteria = "safe-to-deploy"
+delta = "0.10.1 -> 0.10.2"
+
 [[audits.nrfxlib-sys]]
 who = "chrysn <chrysn@fsfe.org>"
 criteria = "safe-to-deploy"
@@ -131,6 +136,12 @@ delta = "0.4.8 -> 0.4.11"
 who = "chrysn <chrysn@fsfe.org>"
 criteria = "safe-to-deploy"
 version = "0.1.0"
+
+[[audits.tinyrlibc]]
+who = "Nils Ponsard <nils.ponsard@inria.fr>"
+criteria = "safe-to-deploy"
+delta = "0.5.0 -> 0.5.1"
+notes = "Documentation changes, no functional changes."
 
 [[audits.try-lock-portable-atomic]]
 who = "chrysn <chrysn@fsfe.org>"


src/ariel-os-embassy/src/lib.rs
```diff
@@ -383,6 +392,23 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
+        #[cfg(feature = "ltem-nrf-modem")]
+        {
+            // TODO: user configuration of APN settings
+
+            // let apn_config = hal::ltem::Config {
+            //     apn: b"eapn1.net",
+            //     auth_prot: hal::ltem::AuthProt::Pap,
+            //     auth: Some((b"NordicSe", b"NordicSe")),
+            //     pin: None,
+            // };
```
nponsard: Left as a TODO as I don't have the tools to try the configuration options, I found that with the SIM card I have, providing no configuration allows it to connect to the cell network. Adding a manual configuration never worked.

examples/tcp-client/src/main.rs
ROMemories, line 1: Can you split out this example into a separate PR? This should be testable without LTE-M, right?

nponsard: Yes this is not specific to LTE-M, I'm splitting this out.

src/ariel-os/Cargo.toml
```diff
@@ -156,6 +156,8 @@ wifi-esp = ["ariel-os-embassy/wifi-esp"]
 eth-stm32 = ["ariel-os-embassy/eth-stm32"]
 # Selects the network backend that goes through tun/tap (native only).
 tuntap = ["ariel-os-embassy/tuntap"]
+# Selects LTE-M on nrf91xx SiP (nrf91-modem).
```
```suggestion
# Selects LTE-M on nRF SiPs (currently only available on nRF91 SiPs).
```

src/ariel-os-nrf/Cargo.toml
```diff
@@ -111,6 +115,15 @@ ble-central = ["ble", "nrf-sdc/central"]
 ## Enables modem support.
 nrf91-modem = ["dep:cortex-m", "dep:nrf-modem", "dep:tinyrlibc"]
 
+## Enables ltem networking support for nrf91 SiP.
```
ROMemories:
```suggestion
## Enables LTE-M networking support on nRF SiPs.
```

src/ariel-os-embassy-common/src/sim_card.rs
```diff
@@ -0,0 +1,37 @@
+//! Common SIM card types to be used across different HALs.
+
+/// Configuration to use the SIM to connect to the network
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config<'a> {
+    /// Access Point Name, usually a domain given by the provider
+    pub apn: &'a str,
+    /// Witch protocol to authentify with the provider
```
ROMemories:
```suggestion
    /// Which protocol to use to authenticate with the provider.
```

src/ariel-os-embassy-common/Cargo.toml
```diff
@@ -40,5 +40,7 @@ _test = ["external-interrupts", "i2c", "spi", "uart"]
 
 ble = ["dep:static_cell", "dep:trouble-host"]
 
+sim-card = []
```
ROMemories: This feature, and maybe others from this PR, needs to be added to the Clippy CI (which will trigger some new warnings).

src/ariel-os-embassy/src/lib.rs
```diff
@@ -360,13 +370,17 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
         )))]
         // The creation of `device` is not organized in such a way that they could be put in a
         // cfg-if without larger refactoring; relying on unused variable lints to keep the
         // condition list up to date.
         let device: NetworkDevice = net::new_dummy();
 
+        #[cfg(not(feature = "ltem-nrf-modem"))]
         let config = net::config();
+        #[cfg(feature = "ltem-nrf-modem")]
+        let config = embassy_net::Config::default();
```
ROMemories: Shouldn't this be moved inside `net::config()` instead?

src/ariel-os-embassy/Cargo.toml
```diff
@@ -123,13 +124,24 @@ storage = ["dep:ariel-os-storage", "ariel-os-hal/storage", "time"]
 
 debug-uart = []
 
+# Enables SIM card configuration
```
ROMemories:
```suggestion
# Enables SIM card configuration.
```

book/src/networking.md
ROMemories, line 1: Can you move these docs into a separate `ci-build:skip` PR?

src/ariel-os-embassy/src/lib.rs
```diff
@@ -338,6 +343,11 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
 
     #[cfg(feature = "tuntap")]
     let device = crate::hal::tuntap::create();
+    #[cfg(feature = "ltem-nrf-modem")]
+    let (device, control) = hal::ltem::init(spawner).await;
+
+    #[cfg(feature = "sim-card")]
```
ROMemories: Is this outside the `net` block below so that it would give us some build-time error in case the provided SIM config was incorrect, while having the `net` module disabled? If so this could use a comment, it not this could be moved closer to the `control_task()` call below.

src/ariel-os-embassy/src/sim_card.rs
```diff
@@ -0,0 +1,58 @@
...
+                credentials,
+                pin,
+            })
+        } else {
+            panic!(
+                "If you set CONFIG_SIM_APN you also need to set at least CONFIG_SIM_AUTHENTICATION_PROTOCOL",
+            );
+        }
+    } else {
+        None
```
ROMemories: In which cases is a SIM config *not* required?

src/ariel-os-embassy-common/src/sim_card.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/// Credentials to authenticate to the network provider
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct SimCredentials<'a> {
+    /// Username
+    pub username: &'a str,
+    /// Password
+    pub password: &'a str,
```
ROMemories: Should this password be given some extra care? For instance [`secrecy`](https://docs.rs/secrecy/latest/secrecy/) provides wrapper types so these secrets don't end up in logs, and makes a best-effort attempt at wiping these from memory on drop (not sure how this point is very relevant here).

src/ariel-os-embassy-common/src/sim_card.rs
```diff
@@ -0,0 +1,37 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config<'a> {
+    /// Access Point Name, usually a domain given by the provider
+    pub apn: &'a str,
+    /// Witch protocol to authentify with the provider
+    pub authentication_protocol: AuthenticationProtocol,
+    /// Credentials, if necessary
+    pub credentials: Option<SimCredentials<'a>>,
+    /// A string containing the PIN, if set.
+    pub pin: Option<&'a str>,
```
ROMemories: How secret is the PIN expected to remain? (What's the attack model about it?)

src/ariel-os-embassy-common/src/sim_card.rs
```diff
@@ -0,0 +1,37 @@
+//! Common SIM card types to be used across different HALs.
+
+/// Configuration to use the SIM to connect to the network
```
ROMemories: Can't be too clear about which network we're talking about:
```suggestion
/// Configuration to use the SIM to connect to the cellular network
```

src/ariel-os-embassy-common/src/sim_card.rs
ROMemories, line 1: Is this the right name for this module? I may be wrong, but I think the APN/auth protocol are not properties of the SIM card, are they? *If* that makes sense, I'd go with something like `cell_network_auth`/`cell_net_auth` (and rename the Cargo feature and env variables accordingly).

src/ariel-os-embassy-common/src/sim_card.rs
```diff
@@ -0,0 +1,37 @@
...
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config<'a> {
```
ROMemories: I think these end up in an `nrf_modem::embassy_net_modem::context::Config` in which the fields are mostly byte slices: could it be too restrictive to only use *UTF-8* strings?

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    control
+        .run(|status| {
+            stack.set_config_v4(status_to_config(status));
+        })
+        .await
+        .unwrap();
+}
+
+/// Create an embassy-net  IPv4 config from an modem status update.
```
ROMemories:
```suggestion
/// Creates an embassy-net IPv4 config from a modem status update.
```

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,109 @@
...
+///
+/// # Panics
+/// Panics if the modem returns an invalid IPv4 or too many dns servers are returned.
```
ROMemories:
```suggestion
/// Panics if the modem returns an invalid IPv4 address or too many DNS servers are returned.
```

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,109 @@
...
+#[must_use]
+fn status_to_config(status: &Status) -> embassy_net::ConfigV4 {
```
ROMemories: Why is this IPv4-only?

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,109 @@
...
+/// # Panics
+/// If the modem task cannot be spawned.
+#[must_use]
+pub async fn init<'a>(spawner: Spawner) -> (NetworkDevice, &'a context::Control<'a>) {
+    let (driver, control, runner) =
+        nrf_modem::embassy_net_modem::new(LTEM_STATE.init(State::new())).await;
+
+    spawner.spawn(modem_task(runner)).unwrap();
+
+    let control = LTEM_CONTROL.init(context::Control::new(control, 0).await);
```
ROMemories: Can you introduce a variable for that `0`, I had to look at the docs to know what that was? Also, what's a CID?

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    let gateway = match status.gateway {
+        Some(IpAddr::V4(addr)) => Some(addr),
+        _ => None,
+    };
+
+    let mut dns_servers = Vec::new();
+    for dns in &status.dns {
+        if let IpAddr::V4(ip) = dns {
+            dns_servers.push(*ip).unwrap();
```
ROMemories: This `unwrap` needs some justification (and/or a const `assert`) of why it can't panic.

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    embassy_net::ConfigV4::Static(embassy_net::StaticConfigV4 {
+        address: Ipv4Cidr::new(addr, 32),
```
ROMemories: Why is the prefix length always 32? This at least needs a constant.

ROMemories (commented):
It would likely be useful that @chrysn looks at this before the next round, especially to check things related to terminology and security.

src/ariel-os-embassy-common/src/sim_card.rs
nponsard: It seems like in many cases the password is the same as the username, I don't think this has to be protected more than a wifi password.

src/ariel-os-embassy/src/sim_card.rs
nponsard: I don't know, I'm struggling to find the specifications for this. The SIM I'm using for examples works with no configuration.

src/ariel-os-embassy-common/src/sim_card.rs
nponsard: AT commands are text-based, I don't think characters outside of UTF-8 will be needed.

src/ariel-os-embassy-common/src/sim_card.rs
nponsard: At the end the PIN is sent as plaintext to the modem using an AT command. I think we should take the same care as the WiFi password.

src/ariel-os-nrf/src/ltem.rs
nponsard: I changed this function and added a command explaining the CIDR, is a constant still needed ?

nponsard:
Thanks for the review, I realized I didn't understand everything that I was adding. I reworked a good bit and it should be better documented.

src/ariel-os-embassy-common/Cargo.toml
ROMemories: I understand that Clippy now runs on this as part of the Clippy run for the nRF91, but if that makes sense `cellular-networking` should still be added to the common Clippy run, so it is used for linting `ariel-os-embassy` and `ariel-os-embassy-common` directly.

src/ariel-os-embassy-common/src/cellular_networking.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// No authentication.
+    None,
+    /// PAP.
+    Pap,
+    /// CHAP.
+    Chap,
+}
+
+/// Which type of communication happens on this PDP.
+#[derive(Clone, Copy, PartialEq, Debug)]
```
ROMemories: Why is this not `Eq` as well? There may be reasons for not making a `PartialEq` type not-`Eq` even when it can currently be (e.g., if there is a possibility of later having floats in it), but generally there should be a comment about it.

src/ariel-os-embassy-common/src/sim_card.rs
ROMemories: AT commands are indeed ASCII-only, so UTF-8 is not going to prevent from doing anything. Should there be checks for the username, password (and others) to make sure they are indeed ASCII-only? Possibly using [`const_str::is_ascii!`](https://docs.rs/const-str/latest/const_str/macro.is_ascii.html) if needed.

src/ariel-os-embassy-common/src/sim_card.rs
ROMemories: I don't think the attack model is the same as Wi-Fi but I guess the conclusion is indeed the same.

nponsard:
> We'll need to wait for the patch to land in `nrf-modem` or replicate your patch into an `ariel-os`-owned fork, wont we?

Yes, usually Dion is quite fast on merging changes. If https://github.com/diondokter/nrf-modem/pull/54 isn't merged when we want to merge this PR we can host a fork on the ariel-os org.

src/ariel-os-embassy-common/src/cellular_networking.rs
```diff
@@ -0,0 +1,62 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PdnAuth<'a> {
+    /// Which protocol to use to authenticate with the provider.
+    pub authentication_protocol: AuthenticationProtocol,
+    /// Credentials, if necessary.
+    pub credentials: Option<PdnCredentials<'a>>,
+}
+
+/// Authentication protocol to authenticate to the network provider.
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
```
ROMemories: Not sure why you've changed the order of these last minute, we always use the `Debug, Copy, Clone, PartialEq, Eq` order elsewhere.

nponsard: It was to stay consistent with `PdpType`, I chose the wrong one to be consistent with.

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,199 @@
...
+
+// Packet Data Protocol context id, range 0-10. On nrf9160 only cid 0 is already allocated.
+const PDP_CONTEXT_ID: u8 = 0;
+
+#[embassy_executor::task]
+async fn modem_task(runner: Runner<'static>) -> ! {
+    runner.run().await
+}
+
+fn convert_cellular_networking_config(
```
chrysn, line 51: Why are those distinct types in the first place?

(Or: Why is a config passed around, when the nrf_modem::…::PdConfig creating function could just as well tap into the environment? Having config objects would work if there's the option to switch at runtime, but that doesn't seem to be implemented right now, as there can be only one control task and that is auto-started.)

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,199 @@
...
+    let mut v4_dns_servers = Vec::new();
+    #[cfg(feature = "ipv6")]
+    let mut v6_dns_servers = Vec::new();
+    for dns in &status.dns {
+        #[allow(unused, reason = "conditional compilation")]
+        match dns {
+            IpAddr::V4(ip) => {
+                // `embassy_net::StaticConfigV4` stores up to 3 DNS addresses,
+                // `nrf_modem::embassy_net_modem::context::Status` contains a maximum of 2.
+                // PANICS: This will never panic
```
chrysn: That depends on a size setup that sounds like it might change any time, and all of a sudden that panics on user (OK, network operator, but they're even worse) controlled data.

Suggestions:
- Add a const assert that one size is smaller than the other.
- Ignore the extra server on overflow (with a debug level log entry).

(Same below for v6)

src/ariel-os-nrf/src/ltem.rs
nponsard: My idea was to have an unique configuration struct that can be used in multiple drivers that use cellular networking. We could go the same route as the WiFi credentials where each driver reads the environment variables but this doesn't look like the way forward, The `TODO` in `ariel-os-esp/src/wifi.rs` seems to say the same thing.

src/ariel-os-nrf/src/ltem.rs
nponsard: Added a const check for it, you can try to trigger it by adding a line 
```
    let mut vec: Vec<core::net::Ipv4Addr, 1> = can_contain(&status.dns);
```

in `status_to_config`.

And building an example using LTE-M:

```
laze -C examples/tcp-client/ build -b nrf9160dk-nrf9160
```

src/ariel-os-nrf/src/ltem.rs
chrysn: We may want to do some unification on that type of configuration later, but for now, this makes sense then. (I'd personally prefer a separation along the lines of "we do it easy in the PR that does the ground work" and "we make it well configurable" later, but no point in splitting anything up at this stage).

src/ariel-os-nrf/src/ltem.rs
chrysn: Nice solution, thanks.

src/ariel-os-embassy-common/src/sim_card.rs
chrysn: There are now is_ascii checks in the latest version, resolving.

src/ariel-os-embassy-common/src/cellular_networking.rs
chrysn: Is now as suggested, resolving.

src/ariel-os-embassy-common/src/cellular_networking.rs
chrysn: Is now Eq, resolving.

src/ariel-os-embassy-common/src/sim_card.rs
chrysn: Bearer tokens don't deserve security ;-)

More seriously: Those are configured on the host, placed into the source code, pass by the compiler … none of those are designed to deal with high-value secrets, nor are PIN and WiFi password such.

As we enhance our onboarding stories, we can gradually up our game around those, but for the time being, those get all the protection we have (which is just little, due to how they go in).

src/ariel-os-embassy-common/src/cellular_networking.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Setting this to `None` will keep the modem's default.
+    pub pdn_auth: Option<PdnAuth<'a>>,
+    /// Packet Domain Protocol type.
+    pub pdp_type: PdpType,
+}
+
+/// Authentication parameters for the Packet Data Network (PDN).
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PdnAuth<'a> {
```
chrysn: This could be an enum right away, to avoid inconsistency like "it's PAP/CHAP but has no credentials" or "it's None but has credentials".

But for as far as I am concerned, that can also be changed when we go for anything more user-/runtime configurable.

src/ariel-os-embassy/src/cellular_networking.rs
```diff
@@ -0,0 +1,99 @@
...
+#[cfg(all(feature = "ipv4", feature = "ipv6"))]
+const PDP_TYPE: PdpType = PdpType::IpV4V6;
+#[cfg(all(feature = "ipv4", not(feature = "ipv6")))]
+const PDP_TYPE: PdpType = PdpType::Ip;
+#[cfg(all(not(feature = "ipv4"), feature = "ipv6"))]
+const PDP_TYPE: PdpType = PdpType::IpV4V6;
+#[cfg(not(any(feature = "ipv4", feature = "ipv6")))]
+const PDP_TYPE: PdpType = PdpType::NonIp;
+
+const fn auth_protocol_from_str(str: &str) -> Option<AuthenticationProtocol> {
```
chrysn: This could be an impl on AuthenticationProtocol (just a suggestion; it's internal and really only stylistic).

src/ariel-os-embassy/src/cellular_networking.rs
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    PdConfig {
+        pdn_auth,
+        apn,
+        pdp_type: PDP_TYPE,
+    }
+};
+
+/// Returns the configuration to authenticate to the cell network.
```
chrysn, line 108: In terms of lifetime, this will need to change, probably even at the PdConfig level because it won't be able to reference into 'static memory any more.

But it's an internal function, and we're not afraid of breaking changes anyway, so this is good for now.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -387,6 +395,29 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
+                        control,
+                        cellular_networking_config,
+                        sim_pin,
+                        stack,
+                    ))
+                    .unwrap();
+            }
+
+            // Use the config so clippy doesn't complain
+            #[cfg(not(any(feature = "ltem-nrf-modem")))]
```
chrysn: Not sure in which configuration it would complain; would it help to have a block above like

```rust
#[cfg(not(any(feature = "lte-nrf-modem")))]
{
    let (_, _) = (cellular_networking_config, sim_pin);
    // maybe not(doc)
    compile_error!("No cellular modem configured");
}
```

so that we don't get things unused silently.

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,213 @@
+use core::net::IpAddr;
+
+use embassy_executor::Spawner;
+#[cfg(feature = "ipv6")]
+use embassy_net::Ipv6Cidr;
+use embassy_net::{Ipv4Cidr, Stack};
```
chrysn: Is this not unused in IPv6-only setups?

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,213 @@
...
+    PdConfig {
+        apn,
+        pdn_auth,
+        pdp_type,
+    }
+}
+
+/// Task responsible of maintaining the connection status up to date.
+/// Also configures the modem when starting (if a `config` is provided)
+/// # Panics
```
chrysn, line 84:
```suggestion
///
/// # Panics
```

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,213 @@
...
+            N_CONTAINER >= N_CONTAINED,
+            "embassy config cannot contain the maximum number of DNS servers"
+        )
+    };
+    Vec::new()
+}
+
+/// Creates an embassy-net config from a modem status update.
+///
+/// # Panics
```
chrysn: If clippy says that this needs to have a Panics section but we know it can't, I'd rather have the lint overidden with a reason than its internals in the docs.

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,213 @@
...
+        });
+    }
+    config
+}
+
+/// Initializes the modem for LTE-M networking.
+/// The control task needs to be spawned using [`control_task`].
+///
+/// # Panics
+/// If the modem task cannot be spawned.
```
chrysn: I guess that's the only reason this could not be spawned:

```suggestion
///
/// … when called more than once (because it spawns a task that does not terminate).
```

src/ariel-os-embassy/src/lib.rs
nponsard: This is to avoid the "unused" error when running clippy on the `cellular-networking` feature without the `ltem-nrf-modem` feature.

src/ariel-os-nrf/src/ltem.rs
nponsard: Indeed, good catch !

src/ariel-os-embassy/src/cellular_networking.rs
nponsard: I prefer to keep all the parsing in `ariel-os-embassy` for now. Moving it would add more complexity of the error handling (instead of just using panic! at build time).

src/ariel-os-embassy-common/src/cellular_networking.rs
nponsard: Refactored to be an enum.

chrysn:
Running with v6 after crudely working around the issue above like this:

```patch
diff --git a/src/ariel-os-embassy/src/net.rs b/src/ariel-os-embassy/src/net.rs
index 08b40b39..90b15c63 100644
--- a/src/ariel-os-embassy/src/net.rs
+++ b/src/ariel-os-embassy/src/net.rs
@@ -252 +252 @@ fn __ariel_os_network_config() -> embassy_net::Config {
-            dns_servers: heapless::Vec::new(),
+            dns_servers: Default::default(),
```

I got the build-time errors about static IP config being required. That's odd, given that during review I've seen code that receives address, gateway and DNS server info from the modem and sets the config accordingly.

So:
* How should IPv6 be tested there?
* Why is static configuration (which IMO should *never* be default on any system because it's not how you do IPv6) selected here?

nponsard:
> Can you assess whether that's related to this PR or an unrelated problem?

It's an unrelated problem #1827

chrysn:
How should the IPv6 code in this PR be tested?

nponsard:
> How should the IPv6 code in this PR be tested?

I don't know, I don't have a SIM from a carrier that can provide me an IPv6 stack.

nponsard:
Concerning the static IPv6 configuration, I changed it so `ltem-nrf-modem` is marked as providing one in laze. Otherwise laze would select the only one available that is the static configuration.

chrysn:
So none of the IPv6 code in there ever ran?

The two shipped SIM cards (onomondo and conexa) did appear not to connect up in v6-only mode (and when requesting both, only got v4 addresses). My private SIM (Austrian bob) does seem to get an IPv4 and IPv6 address -- but I can only tell that from it crashing around the time it should get the addresses up (both in v6 and in mixed mode; trace from v6-only):

```
[TRACE] AT -> AT+CGATT?                                                                                                                                                                                            14:17:36 [0/1999]
 (nrf_modem nrf-modem-0.10.0/src/at.rs:168)                                                                                    
[TRACE] Modem log <4>: AT+CGATT?                                                                                               
 (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)                                                                                                                                                  
[TRACE] Modem log <4>: +CGATT: 1 (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)                                              
[TRACE] AT <- +CGATT: 1                                        
OK                                                                                                                             
 (nrf_modem nrf-modem-0.10.0/src/at.rs:36)                                                                                     
[TRACE] AT -> AT+CGATT?                                                                                                        
 (nrf_modem nrf-modem-0.10.0/src/at.rs:168)                 
[TRACE] Modem log <4>: AT+CGATT?                                                                                               
 (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)                                                                                   
[TRACE] Modem log <4>: +CGATT: 1 (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)                                                                                                                  
[TRACE] AT <- +CGATT: 1                                                                                                        
OK                                                                                                                                                                                            
 (nrf_modem nrf-modem-0.10.0/src/at.rs:36)                                                                                     
[TRACE] AT -> AT+CGPADDR=0                                                                                                     
 (nrf_modem nrf-modem-0.10.0/src/at.rs:168)                                                                                    
[TRACE] Modem log <4>: AT+CGPADDR=0                                                                                                                                                           
 (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)                   
[TRACE] Modem log <4>: +CGPADDR: 0,"2001:4BB8:0262:2730:0000:0000:57A5:1E93" (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)
[TRACE] AT <- +CGPADDR: 0,"2001:4BB8:0262:2730:0000:0000:57A5:1E93"                                                                                                                                                                 
OK                                                                                                                                                                                            
 (nrf_modem nrf-modem-0.10.0/src/at.rs:36)                                                                                     
[TRACE] AT -> AT+CGCONTRDP=0                                                                                                                                                                  
 (nrf_modem nrf-modem-0.10.0/src/at.rs:168)                                                                                    
[TRACE] Modem log <4>: AT+CGCONTRDP=0                                                                                                                                                         
 (nrf_modem nrf-modem-0.10.0/src/ffi.rs:705)                                                                                   
[TRACE] AT <- +CGCONTRDP: 0,,"bob.at","","","10.122.6.68","10.122.8.126"                                                                                                                      
+CGCONTRDP: 0,,"bob.at","","","FDA1:3FB1:0000:010F:0000:0000:0000:0001","FDA1:3FB1:0000:0110:0000:0000:0000:0004",,,,,1280     
OK                                                                                                                                                                                            
 (nrf_modem nrf-modem-0.10.0/src/at.rs:36)                                                                                                                                                    
panicked at src/ariel-os-nrf/src/ltem.rs:95:10:                                                                                                                                               
called `Result::unwrap()` on an `Err` value: AtParseError                                                                      
Core 0                                                                                                                         
    Frame 0: syscall_readonly @ 0x0000cfe4 inline                                                                                                                                             
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/semihosting-0.1.20/src/sys/arm_compat/syscall/arm.rs:88:9                                                                                                  
    Frame 1: sys_exit_extended @ 0x000000000000cfe2 inline                                                                     
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/semihosting-0.1.20/src/sys/arm_compat/mod.rs:191:9    
    Frame 2: exit @ 0x000000000000cfd8 inline                                                                                  
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/semihosting-0.1.20/src/sys/arm_compat/mod.rs:160:5    
    Frame 3: exit @ 0x000000000000cfd8                                                                                         
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/semihosting-0.1.20/src/process.rs:57:5                
    Frame 4: exit @ 0x00001090                                                                                                 
       /home/chrysn/git/crates/ariel-os/src/ariel-os-debug/src/lib.rs:55:5                                                     
    Frame 5: panic @ 0x00007fc2                                                                                                
       /home/chrysn/git/crates/ariel-os/src/ariel-os-rt/src/lib.rs:140:5                                                       
    Frame 6: panic_fmt @ 0x00008642                                                                                            
       /rustc/254b59607d4417e9dffbc307138ae5c86280fe4c/library/core/src/panicking.rs:80:14                                     
    Frame 7: unwrap_failed @ 0x00009c2a                                                                                        
       /rustc/254b59607d4417e9dffbc307138ae5c86280fe4c/library/core/src/result.rs:1867:5                                       
    Frame 8: Result<(), nrf_modem::embassy_net_modem::context::Error>::unwrap @ 0x00005830 inline                             
       /home/chrysn-cache/dot-cache/rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1233:23                                                                                       
    Frame 9: {async_fn#0} @ 0x0000000000005810 inline                                                                          
       /home/chrysn/git/crates/ariel-os/src/ariel-os-nrf/src/ltem.rs:95:10                                                     
    Frame 10: TaskStorage<ariel_os_nrf::ltem::__control_task_task::__control_task_task_inner_function::{async_fn_env#0}>::poll @ 0x0000000000004f06                                                                                 
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.9.1/src/raw/mod.rs:236:22          
    Frame 11: {closure#0} @ 0x00004446 inline                                                                                  
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.9.1/src/raw/mod.rs:461:13          
    Frame 12: <unknown function @ 0x00004446> @ 0x0000000000004444 inline                                                                                                                     
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.9.1/src/raw/run_queue_atomics.rs:85:13                                                            
    Frame 13: <unknown function @ 0x00004446> @ 0x0000000000004430 inline                                                                                                                     
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.9.1/src/raw/mod.rs:454:24                                                                         
    Frame 14: <unknown function @ 0x00004446> @ 0x0000000000004430 inline                                                                                                                     
       /home/chrysn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.9.1/src/raw/mod.rs:564:20                                                                         
    Frame 15: __cortex_m_rt_EGU0 @ 0x0000000000004430 inline                                                                                                                                  
       /home/chrysn/git/crates/ariel-os/src/ariel-os-embassy-common/src/executor_swi.rs:43:38                                                                                                 
    Frame 16: __cortex_m_rt_EGU0_trampoline @ 0x0000000000004430                                                                                                                              
       /home/chrysn/git/crates/ariel-os/src/ariel-os-embassy-common/src/executor_swi.rs:35:9                                                                                                  
    Frame 17: External interrupt #27 @ 0x2000a450                                                                                                                                             
                                                                                                                                                                                              
    Frame 18: ariel_os_embassy::EMBASSY_TASKS::DUPCHECK @ 0x2000a450> @ 0x000000002000a450                                                                                                    
                                                                                                                                                                                              
    Frame 19: panic @ 0x00007fc2                                                                                                                                                              
       /home/chrysn/git/crates/ariel-os/src/ariel-os-rt/src/lib.rs:140:5                                                                                                                      
    Frame 20: __cortex_m_rt_main @ 0x00007fd2                                                                                                                                                 
       /home/chrysn/git/crates/ariel-os/src/ariel-os-rt/src/cortexm.rs:194:5                                                                                                                  
    Frame 21: __cortex_m_rt_main_trampoline @ 0x00007fca                                                                                                                                      
       /home/chrysn/git/crates/ariel-os/src/ariel-os-rt/src/cortexm.rs:192:1                                                                                                                  
    Frame 22: Reset @ 0x0000017e> @ 0x000000000000017e                                                                                                                                        
                                                                                                                                                                                              
    Frame 23: Reset @ 0x0000017e> @ 0x000000000000017e
```

I ran with those patches:

```diff
diff --git a/examples/tcp-client/src/main.rs b/examples/tcp-client/src/main.rs
index 2f800edc..15a69fe7 100644
--- a/examples/tcp-client/src/main.rs
+++ b/examples/tcp-client/src/main.rs
@@ -32,7 +32,7 @@ async fn tcp_echo() {
 
         info!("Connecting...");
         // Connect to https://tcpbin.com/ without using DNS
-        let host_addr = embassy_net::Ipv4Address::from_str("45.79.112.203").unwrap();
+        let host_addr = embassy_net::Ipv6Address::from_str("2600:3c01::f03c:91ff:feab:f98b").unwrap();
         if let Err(e) = socket.connect((host_addr, 4242)).await {
             warn!("connect error: {:?}", e);
             Timer::after_secs(10).await;
diff --git a/src/ariel-os-embassy/src/net.rs b/src/ariel-os-embassy/src/net.rs
index 08b40b39..90b15c63 100644
--- a/src/ariel-os-embassy/src/net.rs
+++ b/src/ariel-os-embassy/src/net.rs
@@ -249,7 +249,7 @@ fn __ariel_os_network_config() -> embassy_net::Config {
 
         config.ipv6 = embassy_net::ConfigV6::Static(embassy_net::StaticConfigV6 {
             address: embassy_net::Ipv6Cidr::new(ipaddr, PREFIX_LEN),
-            dns_servers: heapless::Vec::new(),
+            dns_servers: Default::default(),
             gateway: Some(gw_addr),
         });
     }
```

and this command line:

```
laze -C examples/tcp-client build -b nrf9151-dk -s ipv6 -d ipv4 -D LOG=trace -d network-config-override run
```

nponsard:
The LTEM part of the modem can now be enabled/disabled manually, you can test it by editing the `tcp-client` example: 

```diff
diff --git a/examples/tcp-client/src/main.rs b/examples/tcp-client/src/main.rs
index 2f800edc..b63e6b81 100644
--- a/examples/tcp-client/src/main.rs
+++ b/examples/tcp-client/src/main.rs
@@ -10,6 +10,7 @@ use embedded_io_async::Write as _;
 use ariel_os::time::Timer;
 use ariel_os::{
     debug::log::{info, warn},
+    hal::ltem,
     net,
     reexports::embassy_net,
     time::Duration,
@@ -27,6 +28,8 @@ async fn tcp_echo() {
     stack.wait_config_up().await;
 
     loop {
+        ltem::enable();
+        stack.wait_link_up().await;
         let mut socket = embassy_net::tcp::TcpSocket::new(stack, &mut rx_buffer, &mut tx_buffer);
         socket.set_timeout(Some(Duration::from_secs(10)));
 
@@ -51,7 +54,7 @@ async fn tcp_echo() {
             info!("txd: {}", core::str::from_utf8(msg).unwrap());
             Timer::after_secs(1).await;
         }
-
+        ltem::disable();
         Timer::after_secs(4).await;
     }
 }
```

src/ariel-os-nrf/src/ltem.rs
```diff
@@ -0,0 +1,201 @@
...
+pub fn disable() {
+    CONTROL_SIGNAL.signal(Command::Disable);
+}
+/// Enable the LTE-M link.
+///
+/// Will be replaced by the new network interface toggle API.
+#[doc(hidden)]
+pub fn enable() {
+    CONTROL_SIGNAL.signal(Command::Enable);
+}
```
nponsard, line 44: I'm using this in mobile-asset-tracker until the proper API to toggle network interfaces is merged. Marked as `#[doc(hidden)]` so we can remove it as soon as we have an alternative.

src/ariel-os-embassy/Cargo.toml
```diff
@@ -42,6 +42,9 @@ ariel-os-storage = { workspace = true, optional = true }
 ariel-os-threads = { path = "../ariel-os-threads", optional = true }
 ariel-os-utils = { workspace = true }
 
+const-str = { workspace = true, optional = true }
+heapless = "0.9.1"
+once_cell = { workspace = true }
```
ROMemories: Drive-by comment: I don't think these are used?


```suggestion
```

nponsard: Indeed, removed them and it still builds.

laze-project.yml
```diff
@@ -1228,6 +1228,8 @@ modules:
   - name: network-config-ipv4-default
     help: use default network configuration method
     selects:
+      # The cellular modem provides it's own network configuration.
```
```suggestion
      # The cellular modem provides its own network configuration.
```

laze-project.yml
```diff
@@ -1281,6 +1283,8 @@ modules:
   - name: network-config-ipv6-default
     help: use default network configuration method
     selects:
+      # The cellular modem provides it's own network configuration.
```
```suggestion
      # The cellular modem provides its own network configuration.
```

laze-project.yml
```diff
@@ -1281,6 +1283,8 @@ modules:
...
+      - ?ltem-nrf-modem
```
ROMemories: My only remaining concern is whether this should have the highest priority because, unlike the other transports, this one costs money if used inadvertently. Might still be fine though.

nponsard: This is currently the only available networking on the nrf91 series (until we add a support for external wifi chips).

I had to put this here otherwise laze would choose another network configuration mode and block the build because they are conflicting:
```
       "ltem-nrf-modem" conflicts already provided "network-config-ipv4" (by "network-config-ipv4-dhcp")
```

nponsard: The issue comes from the fact that the protocol selection (and configuration method) comes before the network device selection.

nponsard: Played around with the laze configuration and aaa9a2e13768ba7465f0752c4a8a4092cd2e0fc7 seems to solve it. This should probably be in a different PR though.

nponsard:
Rebased on main with #1955 merged, the conflict isn't there anymore for the network configuration.

PR #1148, opened 2025-07-15
feat(nrf91): experimental modem support

nponsard:
# Description

This PR shows how the modem on nrf91 MCUs could be used in Ariel OS. This needs more work to abstract the communication with the modem.

## Issues/PRs references

- Depends on #1156 
- #1115 
- Depends on #1198 

## Open Questions

I'm open to suggestions on how to abstract communication with the GPS module and LTE-M capabilities.

## Change checklist

- [ ] I have cleaned up my commit history and squashed fixup commits.
- [ ] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [ ] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

PR #1198, opened 2025-08-07
fix(build): enable no-catch-reset option for probe-rs

nponsard:
This adds the `--no-catch-reset` option to probe-rs run and debug commands. This prevents recent versions of probe-rs from stopping when a soft reset happens, keeping the same behaviour as v0.27.

# Description

## Issues/PRs references

Probe-rs issue: https://github.com/probe-rs/probe-rs/issues/3469

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -349,7 +349,7 @@ contexts:
     tasks:
       debug-rs:
         cmd:
-          - probe-rs debug --chip ${PROBE_RS_CHIP} --exe ${out} $@
+          - probe-rs debug --chip ${PROBE_RS_CHIP} --exe ${out} --no-catch-reset $@
 
   - name: rp235xa
     parent: rp
@@ -1000,7 +1000,7 @@ modules:
 
     env:
       global:
-        CARGO_RUNNER: "'probe-rs run ${PROBE_RS_PROTOCOL} --chip ${PROBE_RS_CHIP} --preverify'"
+        CARGO_RUNNER: "'probe-rs run ${PROBE_RS_PROTOCOL} --chip ${PROBE_RS_CHIP} --preverify --no-catch-reset'"
 
     tasks:
       flash-erase-all:


ROMemories:
I'm looking into this, I agree that `--no-catch-reset` is the right way of restoring the previous behavior (it's listed in the [changelog of v0.28.0](https://github.com/probe-rs/probe-rs/releases/tag/v0.28.0)), but it's not yet obvious to me that we do want to restore it. Can you give an example of this happening in Ariel OS?

nponsard:
To work around errata 7 of the nrf9151 it is needed to change some registers and soft reset the core. https://github.com/embassy-rs/embassy/pull/4417

This is not yet needed in Ariel OS but will be needed to use the modem of the nrf9151.

nponsard:
One problem I see is if someone is using a version of probe-rs older than v0.28.0, the command will fail because the flag is not available.

ROMemories:
> To work around errata 7 of the nrf9151 it is needed to change some registers and soft reset the core. [embassy-rs/embassy#4417](https://github.com/embassy-rs/embassy/pull/4417)

Thanks for the pointers! I see how we can end up with soft resets now.

> This is not yet needed in Ariel OS but will be needed to use the modem of the nrf9151.

Then I suppose it's fine we wait until we get to that?

nponsard:
> Then I suppose it's fine we wait until we get to that?

Should I make #1148 depend on it then ?

ROMemories:
> Should I make #1148 depend on it then ?

Yes, sound like a good idea!

nponsard:
This is gonna cause issues for users using `probe-rs` versions prior to `v0.28`, since the option doesn't exist `probe-rs` will error out.

This may be an issue because some people use `v0.27` for compatibility reasons.

nponsard:
This is less of an issue than I thought: the hfxo fix that resets the MCU only happens once when the MCU is powered or when you flash a firmware with the fix for the first time. 
This makes it that the first `laze build run` will crash with 

```
Firmware exited unexpectedly: Exception
```

The subsequent runs will work fine.

PR #1201, opened 2025-08-14
feat(stm32): enable flash cache for stm32f4

bergzand:
# Description

This enables the flash cache for both instruction and data access on supported stm32f4 devices (the stm32f401re and the stm32f411re)

This increases the performance on the stm32f401re with the benchmarks by ≈7%. Higher clocked models benefit more from this as they need more wait states with the flash access. 

## Issues/PRs references

- The stm32f7 also has a cache that can be enabled, but it is a bit more complex. It's split between the ST ART cache on the (I|D)TCM bus and the Cortex-M7 cache on the AXI bus. From what I can see, we link our binaries so that the flash is accessed through the AXI bus and not through the TCM busses.
- The stm32u5 also has a cache but I haven't looked into that one yet

## Changelog entry

(STM32F4) The flash cache is now enabled. This increases performance on supported STM32F4 MCUs.

## Change checklist

### DIFF
--- a/src/ariel-os-stm32/src/lib.rs
+++ b/src/ariel-os-stm32/src/lib.rs
@@ -76,6 +76,8 @@ pub fn init() -> OptionalPeripherals {
     #[cfg(capability = "hw/stm32-dual-core")]
     let peripherals = embassy_stm32::init_primary(config, &SHARED_DATA);
 
+    enable_flash_cache();
+
     OptionalPeripherals::from(peripherals)
 }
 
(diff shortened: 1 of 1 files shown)


bergzand:
Looking at the reference manual of the stm32f4, the data cache updates when writing to the flash memory. It doesn't update when erasing the flash memory though. I'm not sure how that combines with sequential storage.

kaspar030 (approved):
Clock config could have been a separate PR.

PR #1205, opened 2025-08-18
ci(lint): install `taplo-cli` without relying on `cargo-binstall`

ROMemories:
# Description

Now that [`taiki-e/install-action` supports `taplo-cli` directly](https://github.com/taiki-e/install-action/blob/main/TOOLS.md), we can avoid relying on it falling back on using `cargo-binstall`.
The correct name for `taiki-e/install-action` [is however `taplo`, not `taplo-cli`](https://github.com/taiki-e/install-action/issues/1100).

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -371,7 +371,8 @@ jobs:
       - name: Install taplo
         uses: taiki-e/install-action@v2
         with:
-          tool: taplo-cli
+          tool: taplo
+          fallback: none
 
       - name: Check toml formatting
         run: taplo fmt --check


.github/workflows/main.yml
```diff
@@ -371,7 +371,8 @@ jobs:
       - name: Install taplo
         uses: taiki-e/install-action@v2
         with:
-          tool: taplo-cli
+          tool: taplo
+          fallback: none
```
ROMemories, line 375: This disables `install-action`'s fallbacks—currently `cargo-binstall`—as that would try to install the `taplo` *lib* crate, which would fail, instead of the `tapli-cli` bin crate.

bergzand (approved):
LGTM

PR #1206, opened 2025-08-18
ci(build): fix `check-labels` succeeding on PRs with many commits

ROMemories:
# Description

Fixes an issue with the `check-labels` job introduced in #843 which could arise when the PR has more commits than the pagination limit of GitHub's API (30 commits). It would then incorrectly succeed, which would result in the `Build` workflow running for the entire set of builders, as the build would then not be restricted.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -160,6 +160,11 @@ jobs:
             'ci-build:stm32')
               echo "LAZE_BUILDERS=st-b-l475e-iot01a,st-steval-mkboxpro,st-nucleo-c031c6,st-nucleo-f042k6,st-nucleo-f401re,st-nucleo-f411re,st-nucleo-f767zi,st-nucleo-h755zi-q,st-nucleo-wb55,st-nucleo-wba55,stm32u083c-dk" >> "$GITHUB_ENV"
               ;;
+            *)
+              # Can happen if `check-labels` finds no labels but still completes
+              # successfully because of a bug.
+              exit 1
+              ;;
           esac
 
       - name: Run sccache-cache
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,7 @@ jobs:
             for (const pr of items) {
               const prNumber = pr.number;
               const { data: commits } = await github.rest.pulls.listCommits({ owner: 'ariel-os', repo: 'ariel-os', pull_number: prNumber });
+              // TODO: handle pagination
               const tipCommitId = commits[commits.length - 1].sha;
               if (tipCommitId != commitId) {
                 console.log(`Wrong PR: tip is ${tipCommitId}`);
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -157,6 +160,11 @@ jobs:
             'ci-build:stm32')
               echo "LAZE_BUILDERS=st-b-l475e-iot01a,st-steval-mkboxpro,st-nucleo-c031c6,st-nucleo-f042k6,st-nucleo-f401re,st-nucleo-f411re,st-nucleo-f767zi,st-nucleo-h755zi-q,st-nucleo-wb55,st-nucleo-wba55,stm32u083c-dk" >> "$GITHUB_ENV"
               ;;
+            *)
+              # Can happen if `check-labels` finds no labels but still completes
+              # successfully because of a bug.
+              exit 1
```
ROMemories, line 166: This is an additional measure which would have made this bug more obvious in the first place.

.github/workflows/build.yml
```diff
@@ -55,6 +55,7 @@ jobs:
             for (const pr of items) {
               const prNumber = pr.number;
               const { data: commits } = await github.rest.pulls.listCommits({ owner: 'ariel-os', repo: 'ariel-os', pull_number: prNumber });
+              // TODO: handle pagination
```
ROMemories, line 58: We could add support for pagination later if need be, but we'll need [to switch API](https://github.blog/changelog/2025-03-06-github-issues-projects-api-support-for-issues-advanced-search-and-more/) anyway.

ROMemories:
This was [successfully tested](https://github.com/ariel-os/ariel-os/actions/runs/17042283162/job/48308866672?pr=978) in #978, when it contained more than 30 commits:

```sh
Number of PRs containing commit fbcb9a9480ff095b8b4aa45447819a77d4f803fd: 1
Wrong PR: tip is c5c2b29aafd1abda3f21d7713d9a3cfb11472950
Did not find the PR; maybe it has more than 30 commits?
Error: Unhandled error: Did not find the PR; maybe it has more than 30 commits?
```

.github/workflows/build.yml
```diff
@@ -55,6 +55,7 @@ jobs:
             for (const pr of items) {
               const prNumber = pr.number;
               const { data: commits } = await github.rest.pulls.listCommits({ owner: 'ariel-os', repo: 'ariel-os', pull_number: prNumber });
```
bergzand, line 57: While also still a stopgap solution, have you tried increasing the number of commits per page?
```suggestion
              const { data: commits } = await github.rest.pulls.listCommits({ owner: 'ariel-os', repo: 'ariel-os', pull_number: prNumber, per_page: 100 });
```

ROMemories: I didn't try it, but that's orthogonal and I would rather do this in a follow-up, which would be cheap thanks to `ci-build:skip`.

bergzand: I agree, just wanted to know if you already looked at it and found a reason not to increase the page size

bergzand (approved):
Thanks for fixing!

PR #1211, opened 2025-08-19
chore(deps): remove our now unneeded `embassy-net` patch

ROMemories:
# Description

Following #1188, our `embassy-net` patch can now be removed.

This PR also bumps `embassy-net` from v0.6.0 to v0.7.0 because it is required by `picoserve` (otherwise the `Stack` types are not the same), which is used in one of the examples. As patches [apply to transitive dependencies](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#working-with-an-unpublished-minor-version) as well, this was hidden before.

## Testing

Successfully tested the `http-server` example (which uses `picoserve`) on nrf52840dk and espressif-esp32-c6-devkitc-1.

## Issues/PRs references

Follow-up to #118.

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ embassy-embedded-hal = { version = "0.3.0", default-features = false }
 embassy-executor = { version = "0.7.0", default-features = false }
 embassy-futures = { version = "0.1.1", default-features = false }
 embassy-hal-internal = { version = "0.2.0", default-features = false }
-embassy-net = { version = "0.6", default-features = false }
+embassy-net = { version = "0.7", default-features = false }
 embassy-net-driver-channel = { version = "0.3.0", default-features = false }
 embassy-nrf = { version = "0.3.1", default-features = false }
 embassy-rp = { version = "0.4", default-features = false }
--- a/src/ariel-os-coap/src/udp_nal/util.rs
+++ b/src/ariel-os-coap/src/udp_nal/util.rs
@@ -63,7 +63,7 @@ impl embedded_io_async::Error for Error {
             }
             Self::AddressFamilyUnavailable => embedded_io_async::ErrorKind::AddrNotAvailable,
             // These should not happen b/c our sockets are typestated.
-            Self::SendError(udp::SendError::SocketNotBound) |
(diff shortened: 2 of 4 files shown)


kaspar030:
This also drops printing of dhcp addresses, right?

ROMemories:
> This also drops printing of dhcp addresses, right?

Right, we'll wait before merging this; I set to draft for now.

Cargo.toml
```diff
@@ -70,7 +70,7 @@ embassy-embedded-hal = { version = "0.3.0", default-features = false }
 embassy-executor = { version = "0.7.0", default-features = false }
 embassy-futures = { version = "0.1.1", default-features = false }
 embassy-hal-internal = { version = "0.2.0", default-features = false }
-embassy-net = { version = "0.6", default-features = false }
+embassy-net = { version = "0.7", default-features = false }
```
ROMemories, line 73: Despite the version bump, it didn't [seem to have any (relevant) breaking changes](https://github.com/embassy-rs/embassy/blob/main/embassy-net/CHANGELOG.md#07---2025-05-06).
Edit: actually the very Clippy error is because of a (documented) breaking in the UDP API

src/ariel-os-coap/src/udp_nal/util.rs
```diff
@@ -63,7 +63,7 @@ impl embedded_io_async::Error for Error {
             }
             Self::AddressFamilyUnavailable => embedded_io_async::ErrorKind::AddrNotAvailable,
             // These should not happen b/c our sockets are typestated.
-            Self::SendError(udp::SendError::SocketNotBound) |
+            Self::SendError(udp::SendError::SocketNotBound | udp::SendError::PacketTooLarge) |
```
ROMemories, line 66: @chrysn Can you check that this semantically makes sense? I didn't find a better suited [`ErrorKind` variant](https://docs.rs/embedded-io-async/latest/embedded_io_async/enum.ErrorKind.html).

chrysn: Other is fine, we already have Truncated in here as well.

ROMemories:
Superseded by #2125, and the current state of this PR is outdated anyway.

PR #1212, opened 2025-08-19
 feat(support-matrix): collect all errors instead of one 

bergzand:
# Description

This pr changes the validation and functionality mapping by collecting all errors into one if there are any errors during the process. This has the advantage that all errors are printed during a single execution instead of only one at a time.

So far it is still split into validation errors and board mapping errors, printing only validation errors if there are some, and omitting any board mapping errors.

The goal is to make the generator a bit more user friendly and prevent CI whack-a-mole when adding a new feature

The generated output should not have changed

## Issues/PRs references

None

## Change checklist

### DIFF
--- a/doc/gen_support_matrix_html.rs
+++ b/doc/gen_support_matrix_html.rs
@@ -214,7 +214,9 @@ fn main() -> miette::Result<()> {
 
     validate_input(&matrix)?;
 
-    let html = render_html(&matrix, args.tier())?;
+    let board_info = gen_functionalities(&matrix)?;
+
+    let html = render_html(&matrix, board_info, args.tier())?;
 
     match args.command {
         SubCommand::Generate(_) => {
@@ -280,21 +282,27 @@ fn validate_input(matrix: &schema::Matrix) -> Result<(), Error> {
     }
 }
 
-fn render_html(matrix: &schema::Matrix, board_support_tier: &SupportTier) -> Result<String, Error> {
-    use minijinja::{Environment, context};
+fn gen_functionalities(matrix: &schema::Matrix) -> Result<Vec<BoardSupport>, Error> {
+    let boards = matrix
+        .boards
+        .iter()
+        .map(|(board_technical_name, board_info)| {
+            let board_name = &board_info.name;
+            let chip = &board_info.chip;
 
+            // Implement chip info inheritance
+            let chip_info = matrix.chips.get(chip).ok_or(vec![Error::InvalidChipName {
+                found: chip.to_owned(),
+                board: board_name.to_owned(),
+            }])?;
 
-    let mut boards = matrix.boards.iter().map(|(board_technical_name, board_info)| {
-        let board_name = &board_info.name;
-
-        let functionalities = matrix.functionalities
-            .iter()
-            .map(|functionality_info| {
+            let functionalities = matrix.functionalities.iter().map(|functionality_info| {
                 let name = &functionality_info.name;
 
                 let support_key = if let Some(support_info) = board_info.support.get(name) {
                     let status = support_info.status();
-                    matrix.support_keys
+                    matrix
+                        .support_keys
                         .iter()
                         .find(|s| s.name == status)
                         .ok_or(Error::InvalidSupportKeyNameBoard {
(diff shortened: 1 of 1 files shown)


ROMemories:
> The goal is to […] prevent CI whack-a-mole when adding a new feature

I mean you could just run it locally instead of relying on the CI.

bergzand:
> I mean you could just run it locally instead of relying on the CI.

Of course, doesn't change that getting all (or most) issues at once is a bit more friendly to the user.

doc/gen_support_matrix_html.rs
```diff
@@ -378,6 +417,11 @@ enum Error {
         err_span: miette::SourceSpan,
         source: serde_yaml::Error,
     },
+    #[error("Multiple validation issues")]
+    MultipleIssues {
+        #[related]
+        others: Vec<Error>,
+    },
```
ROMemories: I would just call it that, and not treat the case where there is a single error separately:

```suggestion
    #[error("validation issues")]
    ValidationIssues {
        #[related]
        errors: Vec<Error>,
    },
```

The error message is not capitalized so that it can be easily embedded in different (syntactic) contexts.

doc/gen_support_matrix_html.rs
```diff
@@ -321,25 +338,47 @@ fn render_html(matrix: &schema::Matrix, board_support_tier: &SupportTier) -> Res
...
+        .clone()
+        .filter_map(|f| f.err())
+        .flatten()
+        .collect::<Vec<_>>();
+    if errors.is_empty() {
+        Ok(boards.map(|f| f.unwrap()).collect())
+    } else if errors.len() == 1 {
+        Err(errors.pop().unwrap())
+    } else {
+        Err(Error::MultipleIssues { others: errors })
```
ROMemories: (See comment below)

```suggestion
    } else {
        Err(Error::ValidationIssues { errors })
```

doc/gen_support_matrix_html.rs
```diff
@@ -175,6 +175,25 @@ struct SubCommandCheck {
...
+    functionalities: Vec<FunctionalitySupport>,
+}
+
+#[derive(Debug, Serialize)]
+struct FunctionalitySupport {
+    icon: String,
+    description: String,
+    // TODO: add comments
+    // TODO: add the PR link
+}
```
ROMemories, line 195: Would be easier to review if moving these structs around was its own commit.

jkr-ariel (approved):
Verified that the output is unchanged by comparing it against `gen_support_matrix_html.rs` on `main`

doc/gen_support_matrix_html.rs
bergzand: > I would just call it that, and not treat the case where there is a single error separately:

Sure, refactored in 75288307a33eb2c15cca8c3d894a78f314536f7b

> The error message is not capitalized so that it can be easily embedded in different (syntactic) contexts.

:+1:

doc/gen_support_matrix_html.rs
bergzand: Fixed in 5ae787883946ab5f08e2ae13859b43926dd82b94

doc/gen_support_matrix_html.rs
bergzand: Inserted 168f9a73fffd5c27f0b2d814a9064bd8c8bc82f0 in between the existing commits

doc/gen_support_matrix_html.rs
```diff
@@ -321,25 +336,45 @@ fn render_html(matrix: &schema::Matrix, board_support_tier: &SupportTier) -> Res
...
+
+    let errors = boards
+        .clone()
+        .filter_map(|f| f.err())
+        .flatten()
+        .collect::<Vec<_>>();
+    if errors.is_empty() {
+        Ok(boards.map(|f| f.unwrap()).collect())
+    } else {
+        Err(Error::ValidationIssues { errors: errors })
```
ROMemories: optional: shorter and more idiomatic, otherwise it looks like there could be a typo: 
```suggestion
        Err(Error::ValidationIssues { errors })
```

doc/gen_support_matrix_html.rs
```diff
@@ -223,68 +244,65 @@ fn main() -> miette::Result<()> {
...
-        if let Some(f) = invalid_functionality_name {
-            return Err(Error::InvalidFunctionalityNameChip {
-                found: f.to_owned(),
-                chip: chip_info.name.to_owned(),
-            });
-        }
+    if errors.is_empty() {
+        Ok(())
+    } else {
+        Err(Error::ValidationIssues { errors: errors })
```
ROMemories: optional: shorter and more idiomatic, otherwise it looks like there could be a typo: 
```suggestion
        Err(Error::ValidationIssues { errors })
```

ROMemories (approved):
LGTM, after possibly getting the two suggestions in, and autosquashing

doc/gen_support_matrix_html.rs
bergzand: Appreciated, should have spotted that one myself…

doc/gen_support_matrix_html.rs
bergzand: Same as above

PR #1214, opened 2025-08-19
feat(ci): add self hosted runners

nponsard:
# Description

Enables self-hosted runners managed with GitHub ARC on Kubernetes.

## Change checklist


.github/workflows/build.yml
```diff
@@ -98,6 +98,7 @@ jobs:
 
     steps:
       - name: Free up space
+        if: runner.name != 'arc-runner-set'
```
kaspar030: can we factor out this condition somehow? e.g., have one step that defines a true/false condition, and use that in the following steps?
Right now "if arc-runner-set" is correct, but the same skipping of prerequisites installation is also needed for #1055 but slightly different.

kaspar030: the `should-skip` step provides a template for passing on variables between steps.

.github/workflows/build.yml
```diff
@@ -110,6 +123,13 @@ jobs:
             /usr/share/dotnet \
             /usr/share/swift \
             &
+      - name: Set env for self-hosted runners
```
kaspar030: these should be pre-set in the runner environment. in the incus runners, I add worker environment to `.env` in the runner folder - is that possible with the k3s workers?

.github/workflows/build.yml
```diff
@@ -133,6 +153,28 @@ jobs:
           echo "${result}"
           echo "${result}" >> $GITHUB_OUTPUT
 
+
+      - name: Rust cache
```
kaspar030: ... and can we make use of `https://github.com/falcondev-oss/github-actions-cache-server` already? that allows using local disk cache or local s3 as well, just transparent to the workflow. meaning, less changes needed.

.github/workflows/build.yml
```diff
@@ -257,6 +299,8 @@ jobs:
 
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --select nightly --disable xtensa
 
+      - name: show sccache stats
```
kaspar030, line 359: this maybe only do on self hosted, as the sccache action does that already if used.

.github/workflows/build.yml
```diff
@@ -97,7 +98,19 @@ jobs:
         partition: ["1/20", "2/20", "3/20", "4/20", "5/20", "6/20", "7/20", "8/20", "9/20", "10/20", "11/20", "12/20", "13/20", "14/20", "15/20", "16/20", "17/20", "18/20", "19/20", "20/20"]
 
     steps:
+
+      - name: check runner environment
+        id: check-runner
+        run: |
+          echo "name=${{ runner.name }}" >> $GITHUB_OUTPUT
+          if [[ "${{ contains( runner.name, 'arc-runners-inria' ) }}" ]]; then
+            echo "self-hosted=true" >> $GITHUB_OUTPUT
```
kaspar030: "self-hosted" is kinda already available as `runner.environment == "self-hosted"`, see https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context.

.github/workflows/build.yml
nponsard: I can set environment variables yes, though the `SCCACHE_GHA_ENABLED` env variable is set in the workflow definition so I think I will have to keep this step to override that one.

kaspar030: > the `SCCACHE_GHA_ENABLED` env variable is set in the workflow

maybe make that conditional on `self-hosted==false` instead?

.github/workflows/build.yml
```diff
@@ -261,6 +281,15 @@ jobs:
         run: |
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --select nightly --disable xtensa
 
+
+      - name: save nightly build cache
+        run: |
+          mkdir -p /cache/raw/${{ matrix.partition }}
```
kaspar030: `matrix.partition` contains a slash. works here anyways, but with some subfolders. :)

nponsard: Yeah that's not final, this is just a quick test

ROMemories: Could also have spaces, that should be quoted.

nponsard:
That's weird, the cache seems to work for some partitions but not for others. With the same cache on the same node, 3 partitions will finish under 10 minutes and the two others end up taking more than 25 minutes.

nponsard:
> The local storage only supports a single sccache server at a time. Multiple concurrent servers will race and cause spurious build failures.
In the [sccache documentation](https://github.com/mozilla/sccache/blob/main/docs/Local.md) 

I wonder if its affecting us.

ROMemories:
How does it work when we'll need to add new build-time dependencies? https://github.com/ariel-os/garm-worker-image/pull/7 leads me to think that they need to be added to the cache *first*, right? What happens otherwise? Do we also need to wait for the cache to be re-generated?

nponsard:
> How does it work when we'll need to add new build-time dependencies? [ariel-os/garm-worker-image#7](https://github.com/ariel-os/garm-worker-image/pull/7) leads me to think that they need to be added to the cache _first_, right? What happens otherwise? Do we also need to wait for the cache to be re-generated?

With the current setup, the worker image contains the build tools, that's what you need to update if you want to add new tools, this image takes around 10 minutes on each machine to build (can be parallelized), no need to refresh the cache.

If you need a new tool it could be installed inside the CI workflow temporarily and then added to the worker image.

The aim here is to reduce the execution time of a CI run by removing the tools installation steps.

.github/workflows/build.yml
```diff
@@ -97,14 +101,35 @@ jobs:
...
+            "9/10",
+            "10/10",
+          ]
 
     steps:
+      - name: reset git-cache update barrier
+        run: pkill -f -SIGUSR1 /usr/local/bin/git-cache-daemon | true
+
       - name: Free up space
+        if: runner.environment != 'self-hosted'
```
ROMemories, line 137: Where is this environment variable coming from? Is it from the worker image or is it set from GitHub web UI?

kaspar030: It is set by [github](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context).

ROMemories:
Currently if we wanted/needed to switch back to GitHub-hosted runners, we'd need to edit the workflow file again, right? Could this be based on an environment variable instead (which we'd set in GitHub's web UI) so that we could switch it quickly and easily? The workflow should be using GH-hosted runners by default (so that it works transparently on forks), and use our self-hosted runners when the environment variable is present.

nponsard:
> Currently if we wanted/needed to switch back to GitHub-hosted runners, we'd need to edit the workflow file again, right?

Yes


> Could this be based on an environment variable instead (which we'd set in GitHub's web UI) so that we could switch it quickly and easily? 

With current setup where we select a runner group, for free GH-hosted runners to work we need to remove that selector. That's something we can do, we just need to move back the runners to the `Default` group and select them using labels instead. The `runs-on` label selector should then be able to be modified using GH variables.

kaspar030:
@nponsard I'll be looking at this again now, maybe rebase & check if it is up-to-date tooling wise?

.github/workflows/build-cache.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
         with:
-          # we need to access the Main branch for the cache
+          # we need to access all the commits to know when the file was last modified
           fetch-depth: 0
```
kaspar030, line 62: ~~so depth needs to be unlimited here? (drop this line?)~~ that's what `fetch-depth:0` does actually.

ROMemories:
*That* was a fast merge queue run! (4 min 11 s for a full build.)

PR #1216, opened 2025-08-19
fix(clippy): allow the `collapsible_if` lint

jkr-ariel:
# Description

This PR allows the `collapsible_if` Clippy lint which triggered new lint errors due to the new `if let` chains introduced in 1.89. We allow this lint since we do not want to bump the MSRV to 1.89 yet.

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,6 +168,7 @@ unexpected_cfgs = "allow"
 
 [workspace.lints.clippy]
 # Clippy lints are documented here: https://rust-lang.github.io/rust-clippy/master/index.html
+collapsible_if = "allow"                     # We can remove this when we bump the MSRV to 1.89 or higher.
 indexing_slicing = "deny"
 missing_safety_doc = { level = "warn" }
 pedantic = { level = "warn", priority = -1 }


ROMemories:
So [`collapsible_if` isn't new](https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if), it's been there since 1.29. What's new is the [`if let` chains](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/let-chains.html), which allow for more opportunities of collapsing.

src/lib/coapcore/src/seccontext.rs
```diff
@@ -634,6 +634,8 @@ impl<
                 return Err(CoAPError::bad_request());
             };
 
+            // We can remove this when the MSRV is 1.89 or higher
+            #[allow(clippy::collapsible_if)]
```
ROMemories: You could use `reason` instead of a comment, and maybe allow this at module-level for once, this is purely a cosmetic lint.

jkr-ariel: Would it make sense to directly allow it at the workspace level?

ROMemories: I suppose so; the issue is that we'll miss normal `if`s that could be collapsed, but that's not a big deal.

ROMemories:
Switched to `ci-build:skip`, as this change is only about Clippy.

Edit: you can re-run the workflow when it is eventually canceled properly.

PR #1220, opened 2025-08-20
refactor(sensors): make appropriate iterators `ExactSizeIterator`s

ROMemories:
# Description

These iterators are constructed from arrays, so it makes sense to guarantee that they impl `ExactSizeIterator` so that consumers can rely on it.

This is not a breaking change as these items are not yet documented.

## How to review this PR

With #1219 in, the `ariel-os-sensors` crate's documentation can be generated with:

```sh
cargo +nightly doc -p ariel-os-sensors
```

---

`ci-build:skip` is enough because there is currently no implementors and Clippy and rustdoc make sure the rest is fine. 

## Issues/PRs references

Mostly extracted from #1191.

## Change checklist

### DIFF
--- a/src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
+++ b/src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
@@ -106,7 +106,7 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             /// [`Samples`].
             /// [`Iterator::zip()`] can be useful to zip the returned iterator with the one
             /// obtained with [`Reading::samples()`].
-            pub fn iter(&self) -> impl Iterator<Item = ReadingChannel> + '_ {
+            pub fn iter(&self) -> impl ExactSizeIterator<Item = ReadingChannel> + '_ {
                 match self {
                     #(#samples_iter),*,
                 }
--- a/src/ariel-os-sensors/src/registry.rs
+++ b/src/ariel-os-sensors/src/registry.rs
@@ -33,7 +33,8 @@ impl Registry {
     }
 
     /// Returns an iterator over registered sensor driver instances.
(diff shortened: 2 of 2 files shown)


src/ariel-os-sensors/src/registry.rs
```diff
@@ -33,7 +33,8 @@ impl Registry {
     }
 
     /// Returns an iterator over registered sensor driver instances.
-    pub fn sensors(&self) -> impl Iterator<Item = &'static dyn Sensor> {
+    #[must_use]
```
ROMemories, line 36: `must_use` seems to be required now, it seems it is neither on `ExactSizeIterator` or inherited from its super-trait `Iterator` (Clippy yells about it).

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -106,7 +106,7 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             /// [`Samples`].
             /// [`Iterator::zip()`] can be useful to zip the returned iterator with the one
             /// obtained with [`Reading::samples()`].
-            pub fn iter(&self) -> impl Iterator<Item = ReadingChannel> + '_ {
+            pub fn iter(&self) -> impl ExactSizeIterator<Item = ReadingChannel> + '_ {
```
ROMemories, line 109: It would likely make sense to guarantee `FusedIterator` as well, but this is out of scope of this PR.

ROMemories: See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=361f80a9ad883e731c11290b004405a0) for an example.

ROMemories: Proposed in #1226.

kaspar030 (approved):

PR #1223, opened 2025-08-20
fix(sensors): provide a custom `defmt::Format` impl for `MeasurementUnit`

ROMemories:
# Description

Without this the units do not get pretty printed when using `defmt`.

---

`ci-build:skip` is enough because there is currently no implementors and Clippy and rustdoc make sure the rest is fine.

## Issues/PRs references

Extracted from #1191.
Depends on #1224.

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/measurement_unit.rs
+++ b/src/ariel-os-sensors/src/measurement_unit.rs
@@ -8,7 +8,6 @@
 // and https://bthome.io/format/#sensor-data
 // and https://www.iana.org/assignments/senml/senml.xhtml
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum MeasurementUnit {
     /// [Acceleration *g*](https://en.wikipedia.org/wiki/G-force#Unit_and_measurement).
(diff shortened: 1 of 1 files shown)


chrysn:
I don't think that's the right approach, it stands in the way of not linking any standard library formatting.

Maybe a (reusable) macro that impls Display and Format in one go using the shared subset of the two write!, as we do for logging?

src/ariel-os-sensors/src/measurement_unit.rs
```diff
@@ -82,44 +81,58 @@ pub enum MeasurementUnit {
...
 
+macro_rules! provide_unit_fmt {
+    ($unit:expr, $f:expr) => {
+        match $unit {
+            Self::AccelG => write!($f, "g"),
+            Self::Ampere => write!($f, "A"),
+            Self::Becquerel => write!($f, "Bq"),
+            Self::Bool => write!($f, ""),
+            Self::Candela => write!($f, "cd"),
+            Self::Celsius => write!($f, "°C"), // The Unicode Standard v15 recommends using U+00B0 + U+0043.
```
chrysn: Feel free to ignore as out-of-scope, but this could be:

```suggestion
            Self::Celsius => write!($f, "°C"), // As recommended by the Unicode Standard v15 (U+00B0 + U+0043)
```

because the old comment sounds like we're willfully deviating, which AIU we do not do (and then it'd need a reason)

ROMemories:
> this is the kind of change where build tests do their job

So actually not yet, at least because we need #1224, so I made this PR depend on it.

src/ariel-os-sensors/src/measurement_unit.rs
ROMemories: I'll get it in, that's definitely clearer phrasing. What I had in mind was why this wasn't using "℃" (U+2103), which actually only exists for compatibility with CJK languages.

ROMemories: I'll also update to the Unicode Standard v16 (which is out now!), while I'm at it.

ROMemories:
Tested again by merging with #1191.

PR #1225, opened 2025-08-20
fix(clippy): fix the last Clippy lints

jkr-ariel:
# Description

This PR fixes the last Clippy lints that were left in the `coapcore` crate.

## Open Questions

After fixing those lints a new one appeared in the `ariel-os` crate:

```
error: item in documentation is missing backticks
 --> src/ariel-os/src/lib.rs:1:92
  |
1 | //! Ariel OS is an operating system for secure, memory-safe, low-power Internet of Things (IoT).
  |                                                                                            ^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown
  = note: `-D clippy::doc-markdown` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(clippy::doc_markdown)]`
help: try
  |
1 - //! Ariel OS is an operating system for secure, memory-safe, low-power Internet of Things (IoT).
1 + //! Ariel OS is an operating system for secure, memory-safe, low-power Internet of Things (`IoT`).
  |
```

It does not look like "IoT" should be enclosed in backticks. Should we allow the lint at the module level, or is there a way to allow it for this specific doc-comment ?

## Change checklist


src/lib/coapcore/src/lib.rs
```diff
@@ -36,7 +36,7 @@
 //! facility, representing a compromise between development and runtime complexity. Should
 //! benchmarks show this to be a significant factor in code size in applications that need error
 //! handling, more fine grained control can be implemented (eg. offering an option to make
-//! Debug2Format merely print the type name or even make it empty).
+//! `Debug2Format` merely print the type name or even make it empty).
```
ROMemories: Might need to also wrap the line so it's not too long:
```suggestion
//! [`Debug2Format`](defmt_or_log::Debug2Format) merely print the type name or even make it empty).
```

src/lib/coapcore/src/ace.rs
```diff
@@ -369,6 +369,10 @@ impl AceCborAuthzInfoResponse {
     ///
     /// The implementation may fail like [any CoAP response
     /// rendering][coap_handler::Handler::extract_request_data()].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the encoding of the message fails due to an insufficient size.
```
ROMemories: The size of what?

ROMemories: Adding @chrysn as a reviewer as author of this module.

ROMemories:
You'll need to re-run the workflow so `Build` runs.

ROMemories:
> It does not look like "IoT" should be enclosed in backticks. Should we allow the lint at the module level, or is there a way to allow it for this specific doc-comment ?

Clearly we don't want backticks around it; it's a bit weird it's only happening now though. The best move I think would be add [an exception for that word](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown) and then to contribute it to Clippy as we've done before.

src/lib/coapcore/src/ace.rs
chrysn: This function will not panic for any user input; that's what the const asserts and calculations are for.

I don't think that functions that don't panic on any input need a "panics" section, so this is a false positive.

(Sure there can be a bug in the calculation, and the Rust compiler may or may not be able to eliminate the code for the panic, but neither is a reason for a panic documentation.)

chrysn: As there are no precise docs on the lint, at least [some comment by a clippy contributor](https://github.com/rust-lang/rust-clippy/issues/13229#issuecomment-2273783854) says that if a panic only happens if the function is buggy, that shouldn't require a Panics doc.

ROMemories: Agreed

src/lib/coapcore/src/ace.rs
```diff
@@ -369,6 +369,10 @@ impl AceCborAuthzInfoResponse {
     ///
     /// The implementation may fail like [any CoAP response
     /// rendering][coap_handler::Handler::extract_request_data()].
+    #[allow(
```
ROMemories, line 372: We should always use `expect` when possible, to avoid forgetting to remvoe the attribute if it becomes unneeded in the future. The only cases where we still add new `allow`s is in case of conditional compilation, where `cfg_attr`ing the `expect` is just too much trouble. There may be remaining `allow`s in the codebase, but they predate the stabilization of `expect`.

```suggestion
    #[expect(
```

src/lib/coapcore/src/ace.rs
```diff
@@ -369,6 +369,10 @@ impl AceCborAuthzInfoResponse {
...
+        clippy::missing_panics_doc,
+        reason = "Will never panic for any user input"
```
ROMemories: [Lint `reason`s](https://doc.rust-lang.org/reference/attributes/diagnostics.html#lint-reasons) should not be capitalized so they can be easily embedded in other (syntactic) contexts:

```suggestion
        reason = "will never panic for any user input"
```

src/lib/coapcore/src/ace.rs
jkr-ariel: For some reason, when using `expect`, Clippy on the CI does not trigger the lint, which causes an [unfulfilled_lint_expectations](https://github.com/ariel-os/ariel-os/actions/runs/17103177960/job/48505037278#step:13:862) error.

ROMemories: Right, because Clippy's version in CI is nightly-2025-02-25, which is why it's not yelling in the first place. Fine with `allow` for now then; ideally we'll switch to `expect` when we bump the MSRV.

ROMemories (approved):

PR #1227, opened 2025-08-20
docs(examples): add (currently BLE) clock example

chrysn:
# Description

This example, in its first and current phase, implements a BLE clock that prints time to stdout.

Before it should be merged, it should reach a state where it prints time per second (or per minute) and when it is updated by some source.

Over time, more displays and time backends can come in.

As with the CoAP example in its early stages, the idea here is that the code is verbose right now, but as we enhance our abstractions and OS facilities, this becomes as few lines as it should be.

## Issues/PRs references

See discussion at https://github.com/ariel-os/ariel-os/discussions/642#discussioncomment-14153832

## Open questions

- [ ] Do we have an area where maturing examples can sit? (Some CoAP stuff is in tests for that reason).
- [ ] Should we differentiate between examples (that show how something is done didactically) and (demo) applications, that aim to be ready-to-use?

## Change checklist

### DIFF
--- a/examples/clock/Cargo.toml
+++ b/examples/clock/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "clock-app"
+license.workspace = true
+edition.workspace = true
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["time"] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+defmt.workspace = true
+
+embassy-futures = { version = "0.1.1" }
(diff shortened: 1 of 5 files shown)


chrysn:
The current README has a few lines on how to use it. Sadly, "hook it up to gadgetbridge" is not [yet](https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/5281) part of this.

stappersg:
(2026-02-04 was Ariel-OS version 0.3 released, it has **BLE** support.)

What is blocking this merge request?

chrysn:
Thanks for the reminder: Mainly, I need to come back to the example and utilize that now we do have boards with actual digit output.

PR #1230, opened 2025-08-22
docs(book): document how to compile every application for a builder

ROMemories:
# Description

This mentions `laze build --global` to make it easier to compile applications when adding new boards.

## Change checklist

### DIFF
--- a/book/src/adding-board-support.md
+++ b/book/src/adding-board-support.md
@@ -47,6 +47,12 @@ builders:
         - CONFIG_SWI=USART2
 ```
 
+> To build every example and test for a board the following command can be used (as this is only for compilation the credentials do not need to be valid):
+>
+> ```sh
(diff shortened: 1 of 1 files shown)


nponsard:
It may be useful to say that if this command is used only to test if everything builds for a board, the WIFI credentials don't have to be valid.

book/src/adding-board-support.md
```diff
@@ -47,6 +47,12 @@ builders:
         - CONFIG_SWI=USART2
 ```
 
+> To build every example and test for a board the following command can be used (as this is only for compilation the credentials do not need to be valid):
+>
+> ```sh
+> CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build --global -b <builder>
```
ROMemories, line 53: @nponsard I copied the credentials we use in CI and explained why they don't matter; that's one fewer decision to make for the user.

nponsard: This seems even better :+1:

PR #1232, opened 2025-08-22
ci(build): remove `sccache` as it does not get any hits

ROMemories:
# Description

This removes the sccache setup, as it was not getting any hits; hopefully that makes the build a bit faster.

Using `ci-build:nrf` to measure whether that makes a difference.

## Change checklist


ROMemories:
Based on one run, compared to a single other data point, this seems to be making the build about 15% faster.

Issue #1217, opened 2025-08-19
ci: introducing `cargo-manifest-check` piece by piece


ROMemories:
Now that [`cargo-manifest-check`](https://codeberg.org/AudaciousAxiom/cargo-manifest-check) accepts paths of directories to check, we can start integrating it into our CI piece by piece.

The first step would be to add it as a CI lint job (through `taiki-e/install-action`), and make it check a crate that does not change often (to avoid conflicts) and whose manifest requires few changes: e.g., ariel-os-identity.

Then, we'll be able to slowly make every manifest compliant with it, while avoiding conflicts with open PRs.

---

This is not high priority however, and OS features should be prioritized over this.

PR #1235, opened 2025-08-25
feat(ci): introduce new `cargo-manifest-check` lint

jkr-ariel:
# Description

This PR introduces a new lint that makes use of [`cargo-manifest-check`](https://codeberg.org/AudaciousAxiom/cargo-manifest-check).  
Currently, the only crate checked by this lint is `ariel-os-identity` to avoid conflict with open PRs.

## Issues/PRs references

Linked to #1217.

## Change checklist


.github/workflows/main.yml
```diff
@@ -86,6 +86,11 @@ jobs:
         with:
           tool: cargo-deny
 
+      - name: Install cargo-manifest-check
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-manifest-check@0.2.2
```
ROMemories: No need to pin the patch version:

```suggestion
          tool: cargo-manifest-check@0.2
```

ROMemories:
Thinking about this again, I think the check should be moved out of the `lint` job, so it would not fail before getting to Clippy, and moved into either a dedicated job or merged into the one checking TOML files.

PR #1236, opened 2025-08-26
ci(lint): bump `typos-cli` to `1.35`

ROMemories:
# Description

This bumps `typos` from v1.34.0 to v1.35. As usual this brings new corrections and some fixes for false positives. The new corrections should include the ones for #1121 which I had reported upstream for inclusion as part as the non-critical corrections for July 2025.
The complete [`CHANGELOG` can be found here](https://github.com/crate-ci/typos/blob/master/CHANGELOG.md).

This also removes the patch version from the version pin, so that new compatible versions get automatically used, since they only seem to include fixes for false positives.

## Change checklist


.github/workflows/main.yml
```diff
@@ -406,7 +406,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: typos
-        uses: crate-ci/typos@v1.34.0
+        uses: crate-ci/typos@v1.35.5
```
kaspar030, line 409: this does not remove the patch version?

kaspar030: ah, updated dPR description. :slightly_smiling_face:

jkr-ariel, line 409: I think it is fine to keep using `crate-ci/typos` since we would still need to update the minor version occasionally even if we were using `taiki-e/install-action@v2`.

PR #1238, opened 2025-08-26
chore: update lockfile to reduce the number of PR initially raised by renovate

anlavandier:
# Description

Ran `cargo +nightly update` to update the lockfile.

## Issues/PRs references

Aiming to reduce PR's that renovate would open, see #1237.

## Change checklist

### DIFF
--- a/src/ariel-os-macros/tests/ui/spawner/missing_peripherals_param.stderr
+++ b/src/ariel-os-macros/tests/ui/spawner/missing_peripherals_param.stderr
@@ -10,4 +10,4 @@ error: ... because this function has a second parameter
  --> tests/ui/spawner/missing_peripherals_param.rs:9:9
   |
 9 | fn main(_spawner: Spawner, _peripherals: Peripherals) {}
-  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  |         ^^^^^^^^
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,9 +94,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.18"
+version = "0.6.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
+checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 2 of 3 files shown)


anlavandier:
There are already some issues with rust's MSRV and with `cargo test` in `ariel-os-macros`

kaspar030:
```
error: rustc 1.85.1 is not supported by the following package:
  sequential-storage@4.0.3 requires rustc 1.87
```

maybe we should just bump that as well?

kaspar030:
> There are already some issues with rust's MSRV and with `cargo test` in `ariel-os-macros`

Looks like the span changed. Rerunning locally with `TRYBUILD=overwrite` would update the expected results.

anlavandier:
> > There are already some issues with rust's MSRV and with `cargo test` in `ariel-os-macros`
> 
> Looks like the span changed. Rerunning locally with `TRYBUILD=overwrite` would update the expected results.

I tried to run `laze build -DCARGO_ARGS+='--locked' --builders host --multiple-tasks --global  --keep-going=0 test` as is done in the [ci](https://github.com/ariel-os/ariel-os/blob/18f6cbf9866d574aa637d406d967239a8a7a1e3e/.github/workflows/main.yml#L44) but it doesn't work because of our patches. I tried to also add `-s nightly` but it doesn't seem to add anything. How can I run these tests locally ?

kaspar030:
> How can I run these tests locally ?

Try overriding the default toolchain, run `rustup override set nightly` in the ariel checkout root, then again the laze command.

*edit* that will make cargo itself be nightly, not the whole build.

anlavandier:
Bumping the MSRV requires also to changed the pinned nightly toolchain of `laze-project.yml` which is currently [`2025-02-25`](https://github.com/ariel-os/ariel-os/blob/9514daa229d405936c6e33c2d46c1347ddf783e7/laze-project.yml#L864)

ROMemories:
Could we first try to set the workspace's [`resolver`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) to `"3"` so that `rust-version` gets taken into account when selecting dependency versions?

anlavandier:
Changed the resolver and also (and that's probably more important) I used `cargo +nightly-2025-02-25` when updating the lockfile instead of `cargo +nightly`

src/ariel-os-macros/tests/ui/spawner/missing_peripherals_param.stderr
```diff
@@ -10,4 +10,4 @@ error: ... because this function has a second parameter
  --> tests/ui/spawner/missing_peripherals_param.rs:9:9
   |
 9 | fn main(_spawner: Spawner, _peripherals: Peripherals) {}
-  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  |         ^^^^^^^^
```
ROMemories, line 13: Can you move this to a separate PR? This likely comes from bumping `trybuild`.

anlavandier: You mean move both the bump of `trybuild` and this span change ?

anlavandier:
The errors we have here show why going one package at a time even for a "simple" lockfile update is good

src/ariel-os-macros/tests/ui/spawner/missing_peripherals_param.stderr
ROMemories: If this does come from `trybuild`, yes; hopefully this PR can only be about bumping the resolver version and updating the lockfile.

kaspar030:
> The errors we have here show why going one package at a time even for a "simple" lockfile update is good

Yeah you're right. So maybe let's just have renovate do its job.

src/ariel-os-macros/tests/ui/spawner/missing_peripherals_param.stderr
anlavandier: Running only `cargo update+nightly-2025-02-25` and then `laze test` doesn't show this issue so it seems like it's more complicated than just `trybuild` going from `1.0.105` to `1.0.110`

ROMemories: Yeah it's a bit weird; fine to keep it in that PR then I guess.

ROMemories:
Seems like the current CI's error is a bug in `sequential-storage` v4.0.2, which was [fixed in v4.0.3](https://github.com/tweedegolf/sequential-storage/blob/master/CHANGELOG.md#403-18-06-25). Not sure why that latter version is not being selected though.

anlavandier:
It's not selected because of the [MSRV bump](https://github.com/tweedegolf/sequential-storage/commit/cfcf45377cf9d594754dc63027cddf3ebde19524) that landed one commit before v4.0.3

ROMemories:
> It's not selected because of the [MSRV bump](https://github.com/tweedegolf/sequential-storage/commit/cfcf45377cf9d594754dc63027cddf3ebde19524) that landed one commit before v4.0.3

Indeed, they "bumped" it to 1.87 in https://github.com/tweedegolf/sequential-storage/pull/83 because they used `is_multiple_of()`.
I think we should pin it to v4.0.1 for now then.

anlavandier:
> > It's not selected because of the [MSRV bump](https://github.com/tweedegolf/sequential-storage/commit/cfcf45377cf9d594754dc63027cddf3ebde19524) that landed one commit before v4.0.3
> 
> Indeed, they "bumped" it to 1.87 in [tweedegolf/sequential-storage#83](https://github.com/tweedegolf/sequential-storage/pull/83) because they used `is_multiple_of()`. I think we should pin it to v4.0.1 for now then.

Done in #1240

anlavandier:
IIUC the plan is to close this PR and deals with updates as they come when Renovate opens PRs. This hopefully will allow us to understand for example why it's not actually changing `trybuild`'s patch that changes the expected output of the tests of `ariel-os-macros`

PR #1239, opened 2025-08-26
fix(sensors): be more specific about relative humidity

ROMemories:
# Description

This clarifies that the existing "humidity" variants are about relative humidity (we can add absolute humidity later if needed).
This is not a breaking change as this is not yet rendered in the docs.

---

`ci-build:skip` is enough because there is no users of these types for now anyway.

## How to review this PR

Docs can be rendered with:

```sh
cargo doc -p ariel-os-sensors
```

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/category.rs
+++ b/src/ariel-os-sensors/src/category.rs
@@ -31,10 +31,10 @@ pub enum Category {
     Color,
     /// Gyroscope.
     Gyroscope,
-    /// Humidity sensor.
-    Humidity,
-    /// Humidity & temperature sensor.
-    HumidityTemperature,
+    /// Relative humidity sensor.
+    RelativeHumidity,
+    /// Relative humidity & temperature sensor.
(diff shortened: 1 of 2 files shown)


bergzand:
Why do we want to distinguish between absolute and relative humidity at the category level? The difference in provided units is not sufficient to distinguish between the two types (and possibly convert between them based on supplied temperature and pressure)?

ROMemories:
> Why do we want to distinguish between absolute and relative humidity at the category level? The difference in provided units is not sufficient to distinguish between the two types (and possibly convert between them based on supplied temperature and pressure)?

My understanding is that relative humidity and absolute humidity sensors [work very differently](https://www.rfwireless-world.com/terminology/humidity-sensors-types-working-principles) (they're different technologies of sensors), and therefore I think it makes to have different categories for them.

bergzand (approved):
Thanks for the explanation, no other remarks, lgtm

PR #1240, opened 2025-08-26
fix: pin sequential-storage to v4.0.1 to avoid v4.0.2 regression and v4.0.3 MSRV

anlavandier:
# Description

## Change checklist


ROMemories:
Switched to `ci-build:rp` as the change is not HAL-specific, and the RPs do support storage so `sequential-storage` will get used in the build.

Issue #1243, opened 2025-08-27
Bumping MSRV to 1.89+


ROMemories:
This issue is intended to keep track of what we could do after bumping the MSRV to 1.89+:

- [ ] #1242
- [ ] Simplify/remove some of our linker scripts, thanks to https://github.com/rust-lang/rust/pull/140872
- [ ] Remove [`collapsible_if`](https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if) from the list of workspace-`allow`ed lints, see #1216

---

MSRV bump checklist:

The MSRV in the README and the [`rust-version` manifest key](https://doc.rust-lang.org/cargo/reference/rust-version.html) must be updated atomically (in the same PR).

chrysn:
I suggest we do the actual bumping first to get stuff unblocked (https://github.com/ariel-os/ariel-os/pull/1593), and then reap more benefits.

We're not in a state yet that we should need to discuss bumping the MSRV -- our collective time is way better spent on actually getting things done. If and when people are starting to depend on the MSRV and provide maintainer time to care for this, then we can start discussing what needs to be done to not bump this at the very moment the act of bumping stable is less work than any workaround.

PR #1267, opened 2025-08-28
fix(deps): update proc-macro2 to v1.0.101 and related tests

anlavandier:
# Description

Supesedes #1262 by also changing the expected test outputs

## Change checklist


ROMemories (commented):
Now that I see this in isolation and I know what crate this comes from this change looks a bit weird and not like an improvement. Could you track down why there was a change regarding spans upstream (at least to make sure it's not just a regression)?

anlavandier:
> Now that I see this in isolation and I know what crate this comes from this change looks a bit weird and not like an improvement. Could you track down why there was a change regarding spans upstream (at least to make sure it's not just a regression)?

I think that [this PR](https://github.com/dtolnay/proc-macro2/pull/517/files) is where this was changed. 
From what I understand, this change in spans was added with the intent of being available on rust 1.88+ stable but was also made available on older rust if they use nightly (which is our case when running tests)

anlavandier:
So all in all I think that this is fine to merge as-is and not a regression

ROMemories:
> I think that [this PR](https://github.com/dtolnay/proc-macro2/pull/517/files) is where this was changed.
> From what I understand, this change in spans was added with the intent of being available on rust 1.88+ stable but was also made available on older rust if they use nightly (which is our case when running tests)

> So all in all I think that this is fine to merge as-is and not a regression

Actually I do think it is (a temporary) one: I was able to isolate that the behavior changed with v1.0.100 (with the PR you linked, or the other one from that release), *when testing with nightly-2025-02-25*:

```sh
cargo +nightly-2025-02-25 test -p ariel-os-macros
```

However, when testing with today's nightly (`cargo 1.91.0-nightly (a6c58d430 2025-08-26)`), there is no change in behavior, and that adjustment is not needed.

I suppose there's been a change in the implementation on nightly between 2025-02-25 and today, and we'd have to change it back when bumping our toolchain. Also looking again at the test output, clearly that's not what the span should look like.

So I think we should pin `proc-macro2` to v1.0.99 (the latest version that works), and add that bump to #1243.

PR #1271, opened 2025-08-28
refactor(env-utils): remove the dependency on `konst`

ROMemories:
# Description

[`konst` v0.4 has removed some items](https://github.com/rodrimati1992/konst/releases/tag/0.4.0) we were relying upon, as they are now `const` in the std. This PR refactors a couple of env-related macros in `ariel-os-utils` so we can entirely remove the dependency.

## How to review this PR

## Testing

This requires careful testing, because many uses of these macros are feature-gated; they can be found with the following:

```sh
grep -rE '(usize|u8|bool)_from_env_or' src
```

Testing networking and USB on an MCU with a Synopsis IP should cover all occurrences.

Testing with invalid values to see it fail at compile time can also help:

```
CONFIG_NETWORK_MAX_CONCURRENT_SOCKETS=<not-a-usize> CONFIG_VBUS_DETECTION=not-a-bool> laze -C examples/udp-echo/ build -d network-config-dhcp -s usb-ethernet -b st-nucleo-h755zi-q
```

I have also tested the value of `CONFIG_VBUS_DETECTION` in hardware by adding a log line, and it was successfully obtained from the environment variable.

## Change checklist

### DIFF
--- a/src/ariel-os-utils/src/env.rs
+++ b/src/ariel-os-utils/src/env.rs
@@ -37,7 +37,7 @@ define_env_with_default_macro!(u8_from_env_or, u8, "a u8");
 #[macro_export]
 macro_rules! bool_from_env_or {
     // $doc is currently unused
-    ($env_var:literal, $default:expr, $doc:literal) => {
+    ($env_var:literal, $default:expr, $doc:literal $(,)?) => {
         const {
             if let Some(str_value) = option_env!($env_var) {
                 // Manual implementation of `bool::from_str()` because it is not `const` yet.


anlavandier:
More of a remark regarding the upstream policy but it feels weird that there isn't some kind of minimum rust version that has `const`-ed all the things that were removed. From what I can tell `from_str_radix` is only const since 1.82 for example.

ROMemories:
> Wondering if it wouldn't make sense to also manually expand the two remaining macros to also add support for optional trailing commas.

We're only using these macros internally for now, so that's not worth the duplication. Hopefully we'd get [`macro_metavar_expr`](https://github.com/rust-lang/rust/issues/83527) before that if we do end up exposing them in the future.

ROMemories:
> More of a remark regarding the upstream policy but it feels weird that there isn't some kind of minimum rust version that has `const`-ed all the things that were removed. From what I can tell `from_str_radix` is only const since 1.82 for example.

I think they did bump the MSRV, but in a [separate PR to prepare for the release](https://github.com/rodrimati1992/konst/pull/35).

Issue #1231, opened 2025-08-22
feat: structured board descriptions


kaspar030:
This issue tracks the design of structured board descriptions.

chrysn:
There's the aspect of modelling external connectors, so that a composite board with some device plugged in can be modelled. To take a hard case (not necessarily "if you get that right you can do everything", but also harder than the typical feather), look at [Omnimo nRF52840](https://www.crowdsupply.com/eafaq/omnimo-nrf52840#products): It has feather and mikrobus boards overlapping, switchable voltage for the daughterboards, a Qwiic connector (that can be turned into a Grove through an adapter) and a PMOD (typical FPGA connector, so daughterboards will go wild with pinout there).

PR #1273, opened 2025-08-28
refactor(hal): move peripheral code from a-o-embassy into a-o-hal

kaspar030:
This PR moves ~~the~~ some peripheral library code from `ariel-os-embassy` into `ariel-os-hal`

My main motivation was to separate the initialization code from the peripheral code (dependency wise), to make it possible for `ariel-os-embassy::init()` to call out to a potential board init function, which in turn should be able to use the peripheral API. Trying that I realized that this API is exported from this crate, so there was a circular dependency.

This is a prerequisite of #1231.

The `ariel-os-embassy` crate became quite overloaded over time - this might slim it a bit.

So far I moved:

- [x] gpio
- [x] i2c
- [ ] spi
- [ ] usb?
- [ ] eth?
- [ ] ble?

~~A lot needs fixing up still,~~ I basically just tested `examples/blinky` and `test/i2c_controller` on nrf52840dk.

### DIFF
--- a/src/ariel-os-storage/src/lib.rs
+++ b/src/ariel-os-storage/src/lib.rs
@@ -13,7 +13,7 @@ mod storage;
 
 use core::ops::Range;
 
-use ariel_os_hal::{
+use ariel_os_hal::hal::{
     OptionalPeripherals,
     storage::{Flash, FlashError, init as flash_init},
 };


src/ariel-os-hal/src/hal.rs
```diff
@@ -0,0 +1,53 @@
...
+//!
+//! # Portability
+//!
+//! To ensure portability of your application, it is recommended to use the generic,
+//! MCU-family-agnostic items provided in other modules.
+//! Items from this module should only be used when MCU-specific settings are *necessary* for your
+//! application.
+
+#[macro_use]
+pub mod define_peripherals;
```
ROMemories, line 27: This module now shows up in the docs while it shouldn't.

src/ariel-os-hal/src/hal/define_peripherals.rs
ROMemories, line 1: Do we really need to have this duplicated with the one directly in `src`?

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -0,0 +1,94 @@
...
+/// This macro allows to group peripheral structs defined with
+/// [`define_peripherals!`](crate::define_peripherals!) into a single peripheral struct.
+#[macro_export]
+macro_rules! group_peripherals {
+    (
+        $(#[$outer:meta])*
+        $group:ident {
+            $(
+                $(#[$inner:meta])*
+                $peripheral_name:ident : $peripherals:ident
```
ROMemories: This was changed to `path` in #1280.

src/ariel-os-hal/Cargo.toml
```diff
@@ -9,7 +9,16 @@ rust-version.workspace = true
 workspace = true
 
 [dependencies]
+ariel-os-embassy-common = { workspace = true }
+
 cfg-if.workspace = true
+const_panic.workspace = true
+
+embassy-embedded-hal = { workspace = true, optional = true }
+embassy-sync = { workspace = true }
```
ROMemories: Based on grepping, it seems that one could be optional, and only enabled for `i2c`.

src/ariel-os-hal/src/hal/define_peripherals.rs
kaspar030: no, one is enough :-) removed the other.

src/ariel-os-hal/src/hal.rs
kaspar030: fixed.

src/ariel-os-hal/Cargo.toml
```diff
@@ -51,6 +58,9 @@ external-interrupts = [
 ]
```
ROMemories, line 59: This crate also needs to propagate to `ariel-os-embassy-common` with `"ariel-os-embassy-common/external-interrupts"` (GH doesn't let me add a suggestion on the correct line).

Found this by trying to generate the docs for `ariel-os-hal` with the following:

```sh
cargo +nightly doc -p ariel-os-hal --features external-interrupts
```

src/ariel-os-hal/src/lib.rs
```diff
@@ -1,51 +1,24 @@
...
+#[cfg(feature = "i2c")]
+pub mod i2c;
 
-#[doc(hidden)]
-pub mod define_peripherals;
+pub mod hal;
 
-pub use define_peripherals::*;
+// All items of this module are re-exported at the root of `ariel_os`.
+pub mod api {
```
ROMemories, line 11: Rendering the docs with the above command also hints at hiding that module:
```suggestion
#[doc(hidden)]
pub mod api {
```

ROMemories (commented):
Successfully tested the `gpio` and `i2c-controller` example and test on the stm32u083c-dk. After this last round, this should be good to go.

ROMemories (approved):

PR #1274, opened 2025-08-29
fix(sensors): avoid missing docs error on macro-generated enum variants

ROMemories:
# Description

The `#[doc(hidden)]` attribute must be added on each enum variant, not just the first one, otherwise we get `missing_docs` errors when using `ariel-os-sensors` Cargo features `max-sample-min-count-$c` with `$c` > 1.

I'm also working on improving the documentation of these enums separately.

## How to review this PR

This can be tested using the following:

```sh
cargo +nightly doc -p ariel-os-sensors --features max-sample-min-count-2
```

This should fail on `main` and should not anymore with this PR.

## Change checklist

### DIFF
--- a/src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
+++ b/src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
@@ -70,8 +70,10 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
         /// This type is automatically generated, the number of variants is automatically adjusted.
         #[derive(Debug, Copy, Clone)]
         pub enum Samples {
-            #[doc(hidden)]
-            #(#samples_variants),*
+            #(
+                #[doc(hidden)]
+                #samples_variants
+            ),*
         }
 
         impl Reading for Samples {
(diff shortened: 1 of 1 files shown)


src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -70,8 +70,10 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
         /// This type is automatically generated, the number of variants is automatically adjusted.
         #[derive(Debug, Copy, Clone)]
         pub enum Samples {
-            #[doc(hidden)]
-            #(#samples_variants),*
+            #(
+                #[doc(hidden)]
+                #samples_variants
```
ROMemories, line 75: [Relevant interpolation documentation of `quote!`](https://docs.rs/quote/latest/quote/macro.quote.html#interpolation)

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -95,8 +97,10 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
         /// This type is automatically generated, the number of variants is automatically adjusted.
         #[derive(Debug, Copy, Clone)]
         pub enum ReadingChannels {
-            #[doc(hidden)]
-            #(#reading_channels_variants),*,
+            #(
+                #[doc(hidden)]
+                #reading_channels_variants
+            ),*,
```
ROMemories, line 103: That trailing comma is really not needed, but that's out of scope of this PR.

PR #1280, opened 2025-08-29
fix(macros): group_peripherals ident -> path

kaspar030:
# Description

trying to use `group_peripherals` with a path include (e.g., `pins::LedPeripherals` vs `LedPeripherals` makes it choke:

```
error: no rules expected `::`                                                                                                                                  
  --> examples/gpio/src/main.rs:7:15                                                                                                                           
   |                                                                                                                                                           
 7 |     leds: pins::LedPeripherals,                                                                                                                           
   |               ^^ no rules expected this token in macro call               
   |                                                                                                                                                           
note: while trying to match `}`                                                                                                                                
```

IIUC the macro does expect a type, so this PR changes the macro to using `ty` instead of `ident`.

I did test the changed macro in my use case.

## Change checklist

### DIFF
--- a/src/ariel-os-hal/src/define_peripherals.rs
+++ b/src/ariel-os-hal/src/define_peripherals.rs
@@ -61,7 +61,7 @@ macro_rules! group_peripherals {
         $group:ident {
             $(
                 $(#[$inner:meta])*
-                $peripheral_name:ident : $peripherals:ident
+                $peripheral_name:ident : $peripherals:path
             ),*
             $(,)?
         }


src/ariel-os-hal/src/define_peripherals.rs
```diff
@@ -61,7 +61,7 @@ macro_rules! group_peripherals {
         $group:ident {
             $(
                 $(#[$inner:meta])*
-                $peripheral_name:ident : $peripherals:ident
+                $peripheral_name:ident : $peripherals:ty
```
ROMemories: Looking at [the definition](https://doc.rust-lang.org/reference/types.html#grammar-Type), `ty` seems maybe a bit too generic, maybe that should be `path` (a type path) instead:

```suggestion
                $peripheral_name:ident : $peripherals:path
```

ROMemories (commented):
Fine to merge after changing to `path` if that works, after testing it.

src/ariel-os-hal/src/define_peripherals.rs
kaspar030: works too, but a path is actually more generic, as in, it resolves to Type at some point: https://doc.rust-lang.org/reference/paths.html#r-paths.type

kaspar030: huh, there's a circle. `Type -> ... -> TypePath`, `TypePath -> ... -> Type`

ROMemories: > a path is actually more generic

I don't think it is, `Type` only appears *within* `TypePathFn`, while `TypePath` is one possible `Type`.

kaspar030:
> Fine to merge after changing to `path` if that works, after testing it.

Works with `path`, too.

ROMemories (approved):

PR #1282, opened 2025-08-29
feat: generate board support from structured board descriptions

kaspar030:
# Description

This is an initial PoC for auto-generated board support.

I've modeled all current boards, but they're mostly untested. The board descriptions sit in `boards/`.

`src/ariel-os-boards` is now completely auto-generated from those files.

Currently, this generates/handles:

- the boards crate
- leds, buttons structs using `define_peripherals`, sets `provides: has_leds/has_buttons` for example build filtering
- laze file with builders
- CONFIG_SWI, CONFIG_VBUS_DETECTION

~~The actual sbd file format will probably *heavily* change, so maybe don't build on it for now.~~
First round of bike-shedding on the sbd file format is done.

## How to play with it

The generator can be found at https://github.com/ariel-os/sbd, or installed using `cargo install sbd-gen`.

After changing any of the sbd yamls, or adding one, `ariel-os-boards` needs to be re-generated:

`sbd-gen generate-ariel --mode update boards -o src/ariel-os-boards`

## Issues/PRs references

Depends on and currently includes #1273.
Implements #1231.

## Open Questions

TODO:

- [x] ~~currently, one needs to `rm -rf src/ariel-os-boards` before re-generating the crate after changes. add some `regenerate` and/or `check` flags to sbd tool.~~ => fixed, "--mode update"

As follow-up:

- [ ] (stm32) clock config
- [ ] support OOT board crate(s)
- [ ] i2c/spi/... buses
- [ ] debugger
- [ ] many, *many* more :slightly_smiling_face: 

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -357,7 +357,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
 
       - name: rustfmt
-        run: cargo fmt --check --all
+        run: cargo fmt --check --all -- --config format_generated_files=false
 
   lint-yamllint:
     runs-on: ubuntu-latest


.github/workflows/main.yml
```diff
@@ -462,3 +462,19 @@ jobs:
         with:
```
kaspar030, line 457: Blocking comment preventing accidental merge.

src/ariel-os-boards/build.rs
```diff
@@ -0,0 +1,36 @@
+// @generated
+pub fn main() {
+    println!("cargo::rustc-check-cfg=cfg(context, values(\"ai-c3\"))");
```
ROMemories, line 4: Wouldn't it be possible to simply use the following?
```suggestion
    println!("cargo::rustc-check-cfg=cfg(context, values(any()))");
```

src/ariel-os-boards/.sbd-gen
```diff
@@ -0,0 +1 @@
+The contents of this directory were generated by sbd-gen.
```
ROMemories: optional: I know this is generated, but this could have a trailing new line.

src/ariel-os-boards/src/bbc-microbit-v1.rs
```diff
@@ -0,0 +1,23 @@
...
+ariel_os_hal::define_peripherals!(ButtonPeripherals {
+button0: P0_17,
+button1: P0_26,
+});
+}
+
+#[allow(unused_variables)]
+pub fn init(peripherals: &mut ariel_os_hal::hal::OptionalPeripherals) {
+{
+    // Set the LED matrix column for led0 to low
```
ROMemories: I understand that this is generated from the quirks section, but that looks really weird to me: this is really just used in a couple of Ariel OS examples (i.e., applications) and has nothing to do with the board itself, *especially* as AFAIU other applications have no way of opting out of this right now.

boards/st-nucleo-f042k6.yaml
```diff
@@ -0,0 +1,12 @@
...
+    chip: stm32f042k6
+    ariel:
+      swi: USART2
+    leds:
+      led0:
+        pin: PA5
+        # color: green
+        # active: high
+#   buttons:
+# Note: the board only has one physical button which is hardwired to NRST
```
ROMemories: I don't think this comment is needed, but I guess it calls for naming `buttons` differently: `user_buttons`/`user-buttons`. Same I suppose for the LEDs, which are specifically user LEDs.

boards/nrf52840dk.yaml
```diff
@@ -0,0 +1,40 @@
+description: test sbd file
```
ROMemories: Is it?

boards/ai-c3.yaml
```diff
@@ -0,0 +1,3 @@
+boards:
```
ROMemories, line 1: I'm not sure why each individual sbd file has to have a `boards` (plural) key, considering I suppose they are intended to only contain info about one single board? If so it would seem better to me to simply remove that nesting level.

boards/dfrobot-firebeetle2-esp32-c6.yaml
```diff
@@ -0,0 +1,15 @@
+boards:
+  dfrobot-firebeetle2-esp32-c6:
+    chip: esp32c6
+    leds:
+      led0:
+        pin: GPIO15
+        # color: green
+        # active: high
```
ROMemories: I'm not sure I understand why some of these are commented out for some boards and not for others.

boards/heltec-wifi-lora-32-v3.yaml
```diff
@@ -0,0 +1,12 @@
+boards:
+  heltec-wifi-lora-32-v3:
+    chip: esp32s3
+    leds:
+      led0:
+        pin: GPIO35
+        color: white
```
ROMemories, line 7: Can we differentiate between cold white and warm white?

boards/rpi-pico2.yaml
```diff
@@ -0,0 +1,14 @@
...
+    chip: rp235xa
+    flags:
+      - has_usb_device_port
+    leds:
+      led0:
+        pin: PIN_25
+        color: green
+        # active: high
+#    buttons:
+#      button0:
```
ROMemories: Why is this commented out? There is no user button on that board.

src/ariel-os-boards/src/ai-c3.rs
ROMemories, line 1: Have you considered using [`prettyplease`](https://docs.rs/prettyplease/latest/prettyplease/) for formatting the Rust files? I understand that this would make (part of) sbd more biased towards Rust, and we may prefer keeping it more language-agnostic.

boards/bbc-microbit-v2.yaml
```diff
@@ -0,0 +1,23 @@
...
+          - --cfg=board_init
+    quirks:
+      - type: set_pin
+        description: "Set the LED matrix column for led0 to low"
+        pin: P0_28
+        level: low
+
+    leds:
+      led0:
+        pin: P0_21
```
ROMemories: I don't think LEDs from the matrix should be treated as individual user LEDs.

src/ariel-os-boards/Cargo.toml
```diff
@@ -1,15 +1,23 @@
...
 
-[lints]
+[package.edition]
 workspace = true
 
-[dependencies]
+[package.license]
+workspace = true
+
+[package.rust-version]
```
ROMemories, line 12: Can you move this key before `license`, so that `cargo manifest-check src/ariel-os-boards` passes?

src/ariel-os-boards/laze.yml
ROMemories, line 1: Is this file generated?

src/ariel-os-embassy/src/lib.rs
```diff
@@ -205,6 +205,9 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
 
     debug!("ariel-os-embassy::init_task()");
 
+    #[cfg(board_init)]
+    ariel_os_boards::init(&mut peripherals);
```
ROMemories, line 209: I think this should be moved just after the `EXTINT_REGISTRY` init.

boards/nordic-thingy-91-x-nrf9151.yaml
```diff
@@ -0,0 +1,12 @@
+boards:
+  nordic-thingy-91-x-nrf9151:
+    chip: nrf9151
+    leds:
+      led0:
+        pin: P0_29
+        # color: green
```
ROMemories: How should bicolor and RGB LEDs (in a single package) be handled here? The Thingy:91 X for instance has two such RGB LEDs.

boards/ai-c3.yaml
kaspar030: > they are intended to only contain info about one single board

Not necessarily. I wanted to keep the path open for variants within the same file.
(Internally sbd-gen merges all `*.yaml` into one large object.)

boards/bbc-microbit-v2.yaml
kaspar030: Agreed. This would need some "led_matrix: cols: rows: mapping: ...".

boards/dfrobot-firebeetle2-esp32-c6.yaml
kaspar030: I kept the ones that are not verified commented out.

boards/heltec-wifi-lora-32-v3.yaml
kaspar030: I'm not sure that makes sense, aren't there too many variants?
We can add the temperature in kelvin.

src/ariel-os-boards/src/ai-c3.rs
kaspar030: Had considered and thought it only works on ASTs. Another look shows that parsing is just one extra line, will add.

src/ariel-os-boards/src/bbc-microbit-v1.rs
kaspar030: yeah, so, I basically hacked this into the Ariel generator so we don't lose blinky on the microbits, and keep blinky board-agnostic. I consider this temporary, until we have proper led types, and led matrix support.

src/ariel-os-boards/build.rs
kaspar030: It would. The intention is that like this the values are checked.

src/ariel-os-boards/Cargo.toml
kaspar030: Not easily, this is a `BTreeMap`, so iterates alphabetically sorted, IIUC.

ROMemories: Can you mark it as [`@generated`](https://generated.at/) as well then?

src/ariel-os-boards/laze.yml
kaspar030: Yes. Does taplo understand this?

src/ariel-os-boards/Cargo.toml
ROMemories: Assuming that you intended to reply here with [the following](https://github.com/ariel-os/ariel-os/pull/1282/files#r2394967015):

> Will do. Does taplo understand this?

I don't think so, but `cargo-manifest-check` does.

boards/heltec-wifi-lora-32-v3.yaml
ROMemories: > I'm not sure that makes sense, aren't there too many variants?

No more than greenish LEDs I think, and RGBCW/RGBWW/RGB+CCT is well established.
The exact temperature might be overkill at first, but I suppose having `cold-white` and `warm-white` should be pretty simple.
BTW is the color an enumeration or just an arbitrary string?

src/ariel-os-boards/src/bbc-microbit-v1.rs
ROMemories: Maybe I'm misunderstanding: doesn't this grab `P0_28` in every application (and pull it low) with no way for users to get it back?

kaspar030: Yes, it needs to put the pin back in OptionalPeripherals.

ROMemories: No the issue is not about the type, it's that the board shouldn't be doing that, this needs to be moved back into the application.

kaspar030: > the board shouldn't be doing that, this needs to be moved back into the application

The application shouldn't have to do this either. We need to be pragmatic. Until we properly support the matrix, we can set this one pin high so blinky works.

ROMemories: > The application shouldn't have to do this either.

Why? Pulling that GPIO low (COL1) is just something I wrote on a whim for that board so that we could quickly use D2 on that matrix for the blinky example. To use the matrix you'd then need to set that pin high again before switching to COL2.

src/ariel-os-boards/.sbd-gen
kaspar030: Will be in next sbd-gen release: https://github.com/ariel-os/sbd/pull/42

src/ariel-os-boards/src/bbc-microbit-v1.rs
kaspar030: dropped matrix leds for microbits.

boards/dfrobot-firebeetle2-esp32-c6.yaml
kaspar030: removed all comments

boards/heltec-wifi-lora-32-v3.yaml
kaspar030: > BTW is the color an enumeration or just an arbitrary string?

Currently, arbitrary string.

> I suppose having cold-white and warm-white should be pretty simple.

Maybe. IMO having "white" is better than no color specification. As it is arbitrary ATM, let's postpone being more specific.

src/ariel-os-boards/src/ai-c3.rs
kaspar030: Latest `sbd-gen` is using prettyplease. It's not perfect (e.g., cfg-if looks horrible, and all non-doc comments are always stripped).

src/ariel-os-boards/Cargo.toml
kaspar030: marked

boards/nordic-thingy-91-x-nrf9151.yaml
kaspar030: For now I added the individual channel as individual led.

src/ariel-os-embassy/src/lib.rs
kaspar030: No, `board::init()` gets to handle this first. Why would extint be special?

boards/nrf52840dk.yaml
kaspar030: dropped from the files. also, no file level description field anymore in the sbd schema.

kaspar030:
~~ah, prettyprint actually breaks cfg-if. :(~~ No, I did. :facepalm:

examples/gpio/src/pins.rs
```diff
@@ -1,130 +0,0 @@
...
-    led1: PB0
-});
-
-#[cfg(context = "st-nucleo-wb55")]
-ariel_os::hal::define_peripherals!(Peripherals {
-    led1: PB5,
-    btn1: PC4
-});
-
-#[cfg(any(context = "stm32u073kc", context = "stm32u083mc"))]
```
kaspar030, line 114: these are not boards, right?

kaspar030: (stm32u073kc got added recently and is now dropped. the other has a board now.)

src/ariel-os-embassy/src/lib.rs
ROMemories: Because the MCU has to be initialized before initializing the board. `EXTINT_REGISTRY` manages the MCU's external interrupts. The boad init *has* to be moved after that.

.github/workflows/main.yml
```diff
@@ -457,3 +457,19 @@ jobs:
...
           args: --offline --require-https --no-progress --include-fragments --include-verbatim -v '*.md' book/book
           fail: true
+
+  sbd-gen-check:
```
ROMemories, line 461: *After this lands*, we'll need to add this to the set of required CI checks.

boards/ariel-chips.yaml
```diff
@@ -0,0 +1,36 @@
...
+# generating an Ariel OS BSP crate.
+ariel:
+  # This list contains all currently supported chips. The names must correspond
+  # to an Ariel OS laze context.
+  chips:
+    - esp-c3-01m
+    - esp-wroom-32
+    - esp32-c3-mini-1
+    - esp32-c6-wroom-1
+    - esp32-s3-wroom-1
```
ROMemories, line 11: None of these are chips, they're boards. I suppose this is a current limitation of sbd? We'll need to be able to use boards on boards.

examples/gpio/src/main.rs
```diff
@@ -1,36 +1,37 @@
 #![no_main]
 #![no_std]
 
-mod pins;
+use ariel_os_boards::pins;
+
+ariel_os::hal::group_peripherals!(Peripherals {
```
ROMemories: Could you move this below the `use` block below? I don't see a reason why those would be split.

boards/nrf52840dk.yaml
```diff
@@ -0,0 +1,39 @@
+boards:
+  nrf52840dk:
+    chip: nrf52840
+    flags:
+      - has_usb_device_port
+    leds:
+      led0:
+        color: green
+        pin: P0_13
+        active: low
```
ROMemories, line 10: As we discussed before, I think the `color` and `active` fields should just be removed for now, as it is simply not used. Also when it is, the `active` fields should be a required field.

boards/skeleton.yaml
ROMemories, line 1: Do we really need this file? It results in a warning when using sbd and it seems to me that it would get out of date quickly.

src/ariel-os-boards/Cargo.toml
```diff
@@ -1,15 +1,25 @@
+# @generated
+
 [package]
 name = "ariel-os-boards"
-version = "0.2.0"
```
ROMemories, line 3: Do we intend to eventually publish this crate? If so how are we going to control the version number?

examples/gpio/src/pins.rs
ROMemories: No they're not boards. Dropping `stm32u073kc` from this example is fine.

boards/ariel-chips.yaml
kaspar030: > None of these are chips, they're boards.

They're "modules", right? IMO it is fine to not be too pedantic on terminology.

> I suppose this is a current limitation of sbd?

Well, there is a line where the sbd concept of "board" starts.

> We'll need to be able to use boards on boards.

Maybe. Can we get some use-cases? But, I think that's future work.

src/ariel-os-boards/Cargo.toml
kaspar030: Good point.https://github.com/ariel-os/sbd/issues/45

boards/skeleton.yaml
kaspar030: I'd like to have one skeleton file that shows all supported data points. Maybe having it in `sbd-gen` repo, or just rename it, is better. I'll drop it here for now.

boards/nrf52840dk.yaml
kaspar030: No, we didn't discuss dropping unused but accurate information. And we discussed that incomplete information is also fine.

src/ariel-os-embassy/src/lib.rs
kaspar030: I disagree. The board needs a hook to do something before Ariel takes over.
Maybe both is needed, `early init` and `late init`.

laze-project.yml
```diff
@@ -450,6 +450,30 @@ contexts:
   - name: esp32-s3-wroom-1
     parent: esp32s3
 
+  - name: native_chip
```
ROMemories: Is there a reason to go with snake-case for this one? 
```suggestion
  - name: native-chip
```

src/ariel-os-embassy/src/lib.rs
ROMemories: > The board needs a hook to do something before Ariel takes over.

Maybe; I would have hoped the boards could be less privileged than that by construction.

src/ariel-os-boards/Cargo.toml
ROMemories: Without a `version` field, the version defaults to `0.0.0` (look at the lockfile), so this PR downgrades this crate.

boards/heltec-wifi-lora-32-v3.yaml
ROMemories: > let's postpone being more specific

That would be a breaking change however.

boards/ariel-chips.yaml
ROMemories: > They're "modules", right?

They're hardware modules yes, so definitely not chips. However even if this is semantically wrong, that's structurally ok for now, so fine as is for now.

> Can we get some use-cases?

Any actual PCB that uses a hardware module is a use case, we'll need to be able to use the pin names of the hardware modules when defining the custom board for instance.

boards/nrf52840dk.yaml
ROMemories: So I'll have to go through every schematic to review the rest of this PR now.

laze-project.yml
kaspar030: no, will fix

boards/heltec-wifi-lora-32-v3.yaml
kaspar030: so what

boards/nrf52840dk.yaml
kaspar030: No you don't. We will have to do that (or rely on testing) when we bump `sbd-gen` to a version that makes use of them.

src/ariel-os-boards/Cargo.toml
kaspar030: It is not released, so for now, is that an issue?

boards/dfrobot-firebeetle2-esp32-c6.yaml
```diff
@@ -0,0 +1,12 @@
+boards:
+  dfrobot-firebeetle2-esp32-c6:
+    chip: esp32c6
+    leds:
+      led0:
+        pin: GPIO15
+    buttons:
+      button0:
+        pin: GPIO9
```
ROMemories, line 9: Tested this, the `gpio` example does work 👍

boards/dwm1001.yaml
```diff
@@ -0,0 +1,30 @@
+boards:
+  dwm1001:
+    chip: nrf52832
+    leds:
+      led0:
+        pin: P0_14
+        aliases:
+          - D12
+          - GPIO14
```
ROMemories, line 9: Should we keep the underscore in these names?

boards/heltec-wifi-lora-32-v3.yaml
ROMemories: Breaking changes are bad if we can avoid them?

src/ariel-os-boards/Cargo.toml
ROMemories: Not really I think.

boards/dwm1001.yaml
kaspar030: That was the silkscreen name IIUC

kaspar030:
Thanks for the review!

PR #1289, opened 2025-09-01
docs(sensors): define the concept of sensor channels

ROMemories:
# Description

This clarifies the meaning of "sensor channels" early on, which was not included in #474.

## How to review this PR

The `ariel-os-sensors` crate's documentation can be generated with:

```sh
cargo doc -p ariel-os-sensors
```

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/lib.rs
+++ b/src/ariel-os-sensors/src/lib.rs
@@ -6,6 +6,9 @@
 //!
 //! - A *sensor device* is a device measuring one or multiple physical quantities and reporting
 //!   them as one or more digital values—we call these values *samples*.
+//! - The number of samples a sensor device exposes is its number of *channels*.
+//!   For instance, a relative humidity and temperature sensor device has two channels, and a
+//!   3-axis accelerometer has three.
 //! - Sensor devices measuring the same physical quantity are said to be part of the same *sensor
 //!   category*.
 //!   A sensor device may be part of multiple sensor categories.


src/ariel-os-sensors/src/lib.rs
```diff
@@ -6,6 +6,9 @@
 //!
 //! - A *sensor device* is a device measuring one or multiple physical quantities and reporting
 //!   them as one or more digital values—we call these values *samples*.
+//! - The number of samples a sensor device exposes is its number of *channels*.
```
bergzand, line 9: I think this can be clarified a bit better, for example if I have a sensor that supports multiple consecutive measurement of a fixed arbitrary number (let's say 5) values of the same underlying sensing mechanism with a single trigger. With the current phrasing it is not completely clear to me whether that would be 5 channels or one channel providing 5 samples.

I think it should be very clear here that a channel whether a channel is an individual sensing mechanism (for a lack of better description right now).

ROMemories: In your example, how are the multiple consecutive values reported? With a FIFO?

bergzand: I was thinking of a vec of values, or a continuous set of registers in hardware.

bergzand: The best example I can come up with so far is an ADC + DMA that writes N measurement values into a memory area after a single trigger

ROMemories: So these N values would be consecutive measurements of the same physical quantity? This is either unsupported by this sensor API, or we can redefine "samples" and "channels" to refer to values exposed by the sensor *driver*, instead of the sensor *device*.

bergzand: I'm fine with this use case not being supported, I consider it an edge case. However, I think it should be very clear from the documentation that the number of samples a sensor device exposes is from different sensing mechanisms (for lack of better wording) and *not* consecutive measurements.

ROMemories: I understand your point, I think I need to think about some more and try some things out.

Issue #856, opened 2025-02-25
Dependency tree audit chain


chrysn:
We generally do look at crates we introduce at dependencies, but we don't have a formal way of acknowledging them (and we don't check anything on updates, and we rack up a huge transitive dependency tree as things happen to be).

I think we should use some acknowledgement process for crates in our dependency tree. Potential tools are cargo-crev and cargo-audit; I've only used the former, while the latter appears to have a larger user base.

I'd like to emphasize that this doesn't need to be a code review of those crates. A simple "I've looked at this crate superficially, it doesn't look fishy" can suffice as long as it is documented to not be a thorough review (cargo-crev has levels roughly described by "spending an hour by file", "spending a minute by file" and "spending a few seconds by file"), but it will track that we *did* do that check.

When selecting the tool, we may want to look at whether it has the metadata (or support for custom metadata) to support https://github.com/ariel-os/ariel-os/issues/845.

chrysn:
Related concerns:

* https://github.com/ariel-os/ariel-os/issues/845
* We might want to extend this also to crates we don't build but ask the user to install (eg. `cargo generate` in our guides), ideally in a way that allows users to do the cargo-install with a bit more faith. (Thanks @jkarinkl for pointing this out in the context of tutorials)

chrysn:
In light of @anlavandier's work on integrating an updater, we may want to also look into assigning trust to authors of crates. That might help with the tension between "we don't auto-update so we don't get the latest security fixes" vs "we auto-update and get the latest supply chain attacks". IIUC, cargo-vet supports this to some extent.

PR #1296, opened 2025-09-02
ci: add `cargo-vet` to the CI process

anlavandier:
# Description

This PR initializes [`cargo-vet`](https://mozilla.github.io/cargo-vet/index.html) for our repository and adds a CI workflow to check the our dependencies are properly vetted.

Currently, running `cargo (+nightly) vet` against our repo tells us that we have 237 fully audited dependencies (either because they were audited by trusted third parties, _e.g._ google and or because their publisher is trusted by trusted third parties _e.g_ `libm` is considered fully audited because it's published by `rust-lang-owner`) and 308 exempted ones. running `cargo (+nightly) vet suggest` shows which dependencies are yet to be audited and how many lines it amounts to. 

The CI file is adapted from [the  `cargo vet` docs](https://mozilla.github.io/cargo-vet/configuring-ci.html). 

## Issues/PRs references
 Linked to #856 

## Change checklist

### DIFF
--- a/.github/workflows/vetting.yml
+++ b/.github/workflows/vetting.yml
@@ -1,0 +1,29 @@
+name: Dependency Vetting
+on:
+  pull_request:
+    branches:
+      - main
+  merge_group:
+
+jobs:
+  cargo-vet:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+
+      - id: get_toolchain
+        run: echo "toolchain=$(scripts/rust-toolchain.sh)" >> $GITHUB_OUTPUT
+
+      - name: Install toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ steps.get_toolchain.outputs.toolchain }}
+
+      - name: Install cargo-vet
+        uses: taiki-e/install-action@c0dee14250395ae6b1754a99f67c1d693138b102 # v2
+        with:
+          tool: cargo-vet@0.10.0
+
+      - name: Invoke cargo-vet
+        run: cargo vet --locked
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,7 @@ jobs:
       # a correct use for it (specifically referring to the GUI concept), it is
       # totally fine to remove this lint.
       - name: Check for occurrences of the word fold_er (without the underscore)
-        run: set +e; grep -ri "fold""er" --line-number --exclude-dir=.git; test $? -eq 1
+        run: set +e; grep -ri "fold""er" --line-number --exclude-dir=.git --exclude-dir=supply-chain; test $? -eq 1
 
       # This should be replaced with a Clippy lint when one exists.
       - name: Check for leftover occurrences of the `authors` key in manifests
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -1,0 +1,571 @@
+
+# cargo-vet audits file
+
+[[audits.log]]
+who = "chrysn <chrysn@fsfe.org>"
+criteria = "safe-to-deploy"
+delta = "0.4.27 -> 0.4.28"
+
+[[trusted.aho-corasick]]
+criteria = "safe-to-deploy"
+user-id = 189 # Andrew Gallant (BurntSushi)
+start = "2019-03-28"
+end = "2026-09-01"
+
+[[trusted.anstream]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2023-03-16"
+end = "2026-09-01"
+
+[[trusted.anstyle]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2022-05-18"
+end = "2026-09-01"
+
+[[trusted.anstyle-parse]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2023-03-08"
+end = "2026-09-01"
+
+[[trusted.anstyle-query]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2023-04-13"
+end = "2026-09-01"
+
+[[trusted.anstyle-wincon]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2023-03-08"
+end = "2026-09-01"
+
+[[trusted.anyhow]]
+criteria = "safe-to-deploy"
+user-id = 3618 # David Tolnay (dtolnay)
+start = "2019-10-05"
+end = "2026-09-01"
+
+[[trusted.basic-toml]]
+criteria = "safe-to-deploy"
+user-id = 3618 # David Tolnay (dtolnay)
+start = "2023-01-25"
+end = "2026-09-01"
+
+[[trusted.cc]]
+criteria = "safe-to-deploy"
+user-id = 1 # Alex Crichton (alexcrichton)
+start = "2019-03-01"
+end = "2026-09-02"
+
+[[trusted.cc]]
+criteria = "safe-to-deploy"
+user-id = 2915 # Amanieu d'Antras (Amanieu)
+start = "2024-02-20"
+end = "2026-09-02"
+
+[[trusted.clap]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2021-12-08"
+end = "2026-09-02"
+
+[[trusted.clap_builder]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2023-03-28"
+end = "2026-09-01"
+
+[[trusted.clap_derive]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2021-12-08"
+end = "2026-09-02"
+
+[[trusted.clap_lex]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2022-04-15"
+end = "2026-09-01"
+
+[[trusted.colorchoice]]
+criteria = "safe-to-deploy"
+user-id = 6743 # Ed Page (epage)
+start = "2023-04-13"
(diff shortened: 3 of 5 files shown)


.github/workflows/vetting.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          path: ${{ runner.tool_cache }}/cargo-vet
+          key: cargo-vet-bin-${{ env.CARGO_VET_VERSION }}
+
+      - name: Add the tool cache directory to the search path
+        run: echo "${{ runner.tool_cache }}/cargo-vet/bin" >> $GITHUB_PATH
+
+      - name: Ensure that the tool cache is populated with the cargo-vet binary
+        run: |
+          cargo install --root ${{ runner.tool_cache }}/cargo-vet --version ${{ env.CARGO_VET_VERSION }} cargo-vet
```
ROMemories: This installation process seems pretty convoluted to me; I think the upstream docs just [have not been updated in three years](https://github.com/mozilla/cargo-vet/blob/5c601456787f3d07db23bfa47ab316ca1306c721/book/src/configuring-ci.md). Could you try using [`taiki-e/install-action`](https://github.com/taiki-e/install-action/) instead? `cargo-vet` is surprisingly not part of its [directly supported tools](https://github.com/taiki-e/install-action/blob/8cbaef1f655c5ac1460784a8d60f453df16beb34/TOOLS.md), but it'll just fallback to using `cargo-binstall` to install the binary from `cargo-vet`'s GitHub releases. However it seems `cargo-vet` doesn't have a GH release for the latest patch release, which is a bit surprising, but does not seem like a big deal for now.

ROMemories (commented):
I've only reviewed the CI setup for now, I'll review the config in another round.

When we eventually merge this, we need to remember to add it as a required check for PRs, in the repo settings.

ROMemories:
Regarding the "folder" CI check that is currently failing, you can simply exclude the whole `supply-chain` directory from that check.

anlavandier:
`taplo` doesn't seem to allow users to set [`exclude`](https://taplo.tamasfe.dev/configuration/file.html#exclude) through the command line so I left the lint failing for now. We would need to use a config file or maybe to call it on specific files and directories  instead of simply using [`taplo fmt --check`](https://github.com/ariel-os/ariel-os/blob/main/.github/workflows/main.yml#L387).

.github/workflows/vetting.yml
anlavandier: done in [56c6587](https://github.com/ariel-os/ariel-os/pull/1296/commits/56c65876f38ff4892921717e41037abc5f835bac)

anlavandier:
@ROMemories Unrelated but I don't know where else to put it. I can't seem to connect to Matrix today so if you have something to tell me, emails are the way

ROMemories:
> @ROMemories Unrelated but I don't know where else to put it. I can't seem to connect to Matrix today so if you have something to tell me, emails are the way

[Matrix.org is currently having a major outage.](https://status.matrix.org/)

supply-chain/trusted-authors.yaml
```diff
@@ -0,0 +1,49 @@
+# This file is intended as a list of list of trusted publishers and why
+# This file is NOT used by cargo-vet and is just meant for our convenience
+# as maintainers.
```
ROMemories: I'm not sure where that file comes from; is it something you wrote? I suppose this is related to the issue of tracking trust in authors/publishers you've mentioned? If so should it be called trusted publishers instead?

anlavandier: Yes it is a file I wrote. The trusted publishers can be seen in the [`supply-chain/config.toml`](https://github.com/ariel-os/ariel-os/blob/a59f6ebfe4e76454f7384317c5f2c0292c148e6a/supply-chain/config.toml) file as imports (because we import their audits). This file is meant to track trusted authors and the first line and its many typos needs to be changed to reflect that.

ROMemories: I think what matters for supply chain security is that they are trusted *publishers*, not just authors. So this file is intended to track *why* we trust these publishers, as that trust originates from third-party sources? I guess there should be the `user-id` in there as well then, because it's the only thing that's really guaranteed to be unique (with possibly the username as well currently, but that's only because GH is the only supported auth provider on crates.io right now).

Also could you make that file TOML instead, for consistency with other supply chain config files (should be a matter of a simple macro)?

ROMemories:
> `taplo` doesn't seem to allow users to set [`exclude`](https://taplo.tamasfe.dev/configuration/file.html#exclude) through the command line so I left the lint failing for now. We would need to use a config file or maybe to call it on specific files and directories instead of simply using [`taplo fmt --check`](https://github.com/ariel-os/ariel-os/blob/main/.github/workflows/main.yml#L387).

After having a look at what makes the format check fail, it turns out it's only about having comments aligned. Instead of excluding these files, I think we could disable [`align_comments`](https://taplo.tamasfe.dev/configuration/formatter-options.html) entirely? That still requires a taplo config file (which to me is fine), but that setting was already a source of friction anyway, with little value imho.

If we do this, I would prefer this be done in a separate PR.

supply-chain/trusted-authors.yaml
ROMemories: When running `cargo vet` it does show messages like the following:

```
NOTE: mozilla and bytecode-alliance trust David Tolnay (dtolnay) - consider cargo vet trust proc-macro2 or cargo vet trust --all dtolnay
```

so I don't think the information is lost, it's just not encoded in the TOML because that's not needed. I'm not sure we really need this file then.

anlavandier: It does but after you do `cargo vet trust dtolnay --all` or `cargo vet trust proc-macro2`, the only things that shows up in `audits.toml` is something to the likes of 
```toml
[[trusted.proc-macro2]]
criteria = "safe-to-deploy"
user-id = 3618 # David Tolnay (dtolnay)
start = "2019-04-23"
end = "2026-09-01"
```
This doesn't show why we trust this library which is what I wanted to track with this file. I don't have strong opinions though so if you feel like this is not necessary, this can go. Either way I only saw your comment after being done modifying the [file](https://github.com/ariel-os/ariel-os/blob/4a933eaf00b6dbd2e8045b960ff8567f6909fffa/supply-chain/trusted-publishers.toml).

anlavandier:
> > `taplo` doesn't seem to allow users to set [`exclude`](https://taplo.tamasfe.dev/configuration/file.html#exclude) through the command line so I left the lint failing for now. We would need to use a config file or maybe to call it on specific files and directories instead of simply using [`taplo fmt --check`](https://github.com/ariel-os/ariel-os/blob/main/.github/workflows/main.yml#L387).
> 
> After having a look at what makes the format check fail, it turns out it's only about having comments aligned. Instead of excluding these files, I think we could disable [`align_comments`](https://taplo.tamasfe.dev/configuration/formatter-options.html) entirely? That still requires a taplo config file (which to me is fine), but that setting was already a source of friction anyway, with little value imho.
> 
> If we do this, I would prefer this be done in a separate PR.

I added the config file to that branch to test it's effects in CI. I will remove it when squashing commits and open a PR with just the config file once I'm confident that this works as it should

anlavandier:
> > > `taplo` doesn't seem to allow users to set [`exclude`](https://taplo.tamasfe.dev/configuration/file.html#exclude) through the command line so I left the lint failing for now. We would need to use a config file or maybe to call it on specific files and directories instead of simply using [`taplo fmt --check`](https://github.com/ariel-os/ariel-os/blob/main/.github/workflows/main.yml#L387).
> > 
> > 
> > After having a look at what makes the format check fail, it turns out it's only about having comments aligned. Instead of excluding these files, I think we could disable [`align_comments`](https://taplo.tamasfe.dev/configuration/formatter-options.html) entirely? That still requires a taplo config file (which to me is fine), but that setting was already a source of friction anyway, with little value imho.
> > If we do this, I would prefer this be done in a separate PR.
> 
> I added the config file to that branch to test it's effects in CI. I will remove it when squashing commits and open a PR with just the config file once I'm confident that this works as it should

After pushing this commit, I realized that should have checked locally first for the impact of `align_comment = false`. In any case, not only this does solve the issue with `supply-chain/audits.toml` files but it also now finds errors in `ariel-os/Cargo.toml`, `ariel-os/_typos.toml` and `ariel-os/ariel-os-cargo.toml`. In both cases, the `trusted-publishers.toml` file is causing issues because it lacks a newline at the end...

supply-chain/trusted-authors.yaml
anlavandier: Removed the file from the index (I'm keeping it on local  for now) in the [latest commit](https://github.com/ariel-os/ariel-os/pull/1296/commits/08105c93acc93d85893dca119a3623ecb2d86652)

anlavandier:
@ROMemories The only thing left to decide is what to do about the TOML lints. The comment by default conform to `align_comments = false` but that is not the case of our others TOML files that conform to `align_comments = true`. Should I open a PR to add `align_comments = false` to the taplo config and modify the TOML files that need to be ?

ROMemories:
> Should I open a PR to add `align_comments = false` to the taplo config and modify the TOML files that need to be ?

I tried it locally and to me this looks better (and avoids issues with the `supply-chain` files), so yes please!

anlavandier:
A few questions remain I think :
- how do we document the effects of this PR ? 
- This PR once merged will lead to failing CI pretty much everywhere (in fact this fails already in this branch because of patches of some dependencies e.g. `build-rs` that have been merged through PR raised by renovate leading to unvetted versions of existing dependencies), what do we do about this ? 
- In relation to the previous point, what is our policy regarding the vetting process ? 
- In its current state, this might lead to friction for external contributions if they happen to bring new dependencies or patches of existing dependencies through lock file updates. We can't rely on external contributors to vet our dependencie

ROMemories:
This is currently waiting on me documenting the vetting process in the book.

Edit: this is now in #1331.

ROMemories:
Now that #1331 has been reviewed, @anlavandier you can get this PR in a state that would make it ready-to-merge I think, and then we'll merge it.

anlavandier:
This now contains the audit from @chrysn 
https://github.com/anlavandier/ariel-os/pull/21

ROMemories:
Seems like this PR's second commit got a bit busy when rebasing.

anlavandier:
@ROMemories Thanks for catching that. I had rebased the PR on my fork's main (which wasn't synced at the time) instead of ariel os's main. This should be good now

PR #1313, opened 2025-09-05
docs(sensors): render the documentation of the sensor API

### LINKED ISSUE #334: "portable" sensor reading

ROMemories:
# Description

This introduces an `unstable` Cargo feature and renders the docs for the sensor API.

## How to review this PR

The sensor API docs can be read using the docs preview.

Examples of usage of the API can be found in #1363 and #1412 (for sensor implementations), and in #1530 and in the `sensors-debug` example (for accessing sensor readings).

## Issues/PRs references

- Closes #334
- Depends on #1191
- Depends on #1522

## Changelog entry

A custom sensor abstraction has been introduced: sensor drivers can be written against it, and sensor driver instances can be registered in a sensor registry inside an application. The registry then allows to query sensor driver instances and fetch their readings asynchronously. The `sensors-debug` example is available for testing. See the documentation of `ariel_os::sensors` for details.

## Change checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -84,6 +84,7 @@ jobs:
                   no-boards,
                   random,
                   ariel-os-coap/doc,
+                  sensors,
                   spi,
                   storage,
                   tcp,
--- a/src/ariel-os/Cargo.toml
+++ b/src/ariel-os/Cargo.toml
@@ -66,8 +66,7 @@ random = ["dep:ariel-os-random", "ariel-os-embassy/random"]
 csprng = ["dep:ariel-os-random", "ariel-os-random?/csprng"]
 # Enables seeding the random number generator from hardware.
 hwrng = ["ariel-os-embassy/hwrng"]
-# Enables support for sensors.
-# *Currently experimental and undocumented.*
+## Enables unified support for sensors.
 sensors = ["dep:ariel-os-sensors"]
 
 #! ## Network protocols


src/ariel-os/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub use ariel_os_power as power;
 pub use ariel_os_random as random;
 #[doc(hidden)]
 pub use ariel_os_rt as rt;
-#[cfg(feature = "sensors")]
+#[cfg(all(feature = "unstable", feature = "sensors"))]
```
ROMemories: The `ariel_os::sensors` module being gated behind `unstable` *only* means that bumping the `ariel_os_sensors` dependency *of* `ariel_os` to an incompatible version is not a breaking change for the `ariel_os` crate. Breaking changes to the `ariel_os_sensors` crate (that sensor driver crates directly depend on) still *require* bumping its version number to an incompatible version (therefore fragmenting the sensor driver ecosystem).

kaspar030:
IMO we should mark stuff "unstable" once we hit 1.0 ...

ROMemories:
> IMO we should mark stuff "unstable" once we hit 1.0 ...

Do you want me to remove it? I'm fine dropping it.

kaspar030:
> Do you want me to remove it? I'm fine dropping it.

yes, drop for now. Thanks!

src/ariel-os/src/lib.rs
ROMemories: Removed the `unstable` Cargo feature.

PR #1314, opened 2025-09-08
chore: specify the `rust-version` of each `ariel-os*` crate

ROMemories:
# Description

This specifies the `rust-version` (i.e., the MSRV) of each `ariel-os*` crate. Before that, only the one of `ariel-os` was specified for tools to use.
This also clarifies the versioning policy of the `ariel-os-sensors` crate.

---

Using `ci-build:skip` as manifests are checked by Clippy and CI builds with Rust 1.85, which is the specified MSRV.

## How to review this PR

All `ariel-os*` manifests must now have a `rust-version` key.

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,9 @@ resolver = "3"
 [workspace.package]
 license = "MIT OR Apache-2.0"
 edition = "2024"
+# This should be the Ariel OS MSRV.
+# Some crates may have a different MSRV.
+rust-version = "1.85"
 repository = "https://github.com/ariel-os/ariel-os"
 
 [workspace.dependencies]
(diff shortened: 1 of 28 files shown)


ROMemories:
> What do you think about adding (perhaps in a separate PR) a CI check for the presence of `rust-version` or `rust-version.workspace = true` even in every relevant `Cargo.toml`.

I think it's a good idea; couldn't find an appropriate [Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html?groups=cargo) however, so I'll have to come up with a grep check similar to the ones we already have. I'll do that in a follow-up.

Edit: #1315 is the follow-up.

PR #793, opened 2025-02-11
 feat(async): expose `yield_now()`

ROMemories:
# Description

Exposes an async function which can be useful but is sometimes a bit difficult to find in the ecosystem.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1320, opened 2025-09-08
chore(ariel-os-coap): remove the unused `embassy-futures` dependency

ROMemories:
# Description

This removes the `embassy-futures` dependency from `ariel-os-coap`, which seems unused (grepping for `embassy_futures` returns no matches). It seems I did not bother checking whether it was used in #793. 

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,7 +224,6 @@ dependencies = [
  "coap-numbers",
  "coapcore",
  "critical-section",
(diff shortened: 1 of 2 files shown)


chrysn (approved):
Thanks for spotting this. Code moved between ariel-os-coap and coapcore, and I must have missed removing it when the last usage went away.

Issue #1229, opened 2025-08-20
doc: Type provenance unclear


chrysn:
As a user of our documentation, I find it very unclear where types come from. Example: [Delay](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/time/struct.Delay.html)

This *should* allow me to learn that this is a pub re-export of embassy_time in some way.

There might be better ways in rustdoc even (I think I saw it show things as "pub reexport", but I didn't find an example), but the fallback that is usually available is to click on the source link and see where it's from.

As we are building with `--no-deps`, that road is blocked.

This is not only an issue for external types (where the question is more "is this really and publicly the type I'm getting from that other crate so I can mix them"), but also for internal types: for components such as [i2c](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/i2c/), I have to resort to filesystem grep/find to see where it comes from, as the wildcard re-exports of ariel-os/src/lib.rs can't be traced easily.

@ROMemories, do you have ideas about whether we could get more useful information out of rustdoc even with `--no-deps`? Or could we build with dependencies included but just style down those we don't want to show?

ROMemories:
> This should allow me to learn that this is a pub re-export of embassy_time in some way.

I think there's a couple of things at play here, the main one being that "type identity" (is there a better word for it?) is leaky.

For instance, if a function expects a specific type, the arguments needs to come from the expected version of the expected crate. I think this is [nominal typing](https://en.wikipedia.org/wiki/Nominal_type_system), in opposition to [structural typing](https://en.wikipedia.org/wiki/Structural_type_system).

The thing is that items from the `ariel_os::time` module can de facto be used for two different use cases: in a self-contained way, and as arguments to functions that specifically expect a type from `embassy_time`. In fact, only the first use case was intended I think, the goal was to enable something as simple as the following:

```rust
ariel_os::time::Timer::after_secs(1).await;
```

In that case, the type identity/provenance is irrelevant, it just works in isolation.

For the second use case—i.e., when an `embassy_time` type is specifically needed—you should use the `ariel_os::reexports::embassy_time` re-export instead. I agree that there is currently no hint about this in the documentation.

However, I suppose the best long-term solution for this would be to decouple the items from `ariel_os::time` from `embassy_time`, by either wrapping the types or providing our own implementations (which would also allow us to provide doc comments which are more in line with our guidelines, which we can't do for re-exports). This would however clearly be a breaking change, because "type identity" is leaky.

[...]

PR #1321, opened 2025-09-08
docs(time)!: clarify what items from the `time` module can be used for

ROMemories:
# Description

This aims to clarify what items from `ariel_os::time` can be used for, following discussions from #1229.
They should not be used *as* types from `embassy-time` and therefore their provenance (whether they are re-exports or not, and if so from which crate) should not matter. 

BREAKING CHANGE: this is arguably a breaking change as it may change how items should be used.

## How to review this PR

Having a look at the docs preview is enough.

## Open Questions

@chrysn If you think this is not the right way forward, feel free to close this.

## Changelog entry

The documentation of the `ariel_os::time` module has been clarified: its items must only be used in combination with other items from that module, and not be passed as arguments to other crates.

## Change checklist

### DIFF
--- a/src/ariel-os-embassy/src/lib.rs
+++ b/src/ariel-os-embassy/src/lib.rs
@@ -51,6 +51,24 @@ pub mod api {
     #[cfg(feature = "time")]
     pub mod time {
         //! Provides time-related facilities.
+        //!
+        //! **Note:** Even though some types from this module might be re-exports from third-party
+        //! crates (e.g., `embassy-time`), they are intended to be used "in a self-contained
+        //! way"—for instance, to sleep asynchronously:
+        //!
+        //! ```
+        //! # use ariel_os_embassy::api::time::{Duration, Timer};
+        //! # async fn example() {
+        //! Timer::after(Duration::from_secs(1)).await;
+        //! // or, equivalent but terser:
+        //! Timer::after_secs(1).await;
+        //! # }
+        //! ```
+        //!
+        //! Items from this module should *not* be passed as arguments to functions that expect
+        //! types from these third-party crates.
+        //! If such specific types are needed, please use those from `ariel_os::reexports` instead.
+
         // NOTE: we may want to re-export more items in the future, but not re-export the whole
         // crate.
         pub use embassy_time::{Delay, Duration, Instant, TICK_HZ, Timer};


src/ariel-os-embassy/src/lib.rs
```diff
@@ -51,6 +51,13 @@ pub mod api {
     #[cfg(feature = "time")]
     pub mod time {
         //! Provides time-related facilities.
+        //!
+        //! **Note:** Even though some types from this module might be re-exports from third-party
+        //! crates (e.g., `embassy-time`), they are intended to be used "in a self-contained way"
+        //! and not passed as arguments to functions that expect types from these third-party
+        //! crates.
+        //! If such specific types are needed, please use those from `ariel_os::reexports` instead.
```
ROMemories, line 70: Can't seem to be able to link to the `reexports` module unfortunately.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -51,6 +51,13 @@ pub mod api {
     #[cfg(feature = "time")]
     pub mod time {
         //! Provides time-related facilities.
+        //!
+        //! **Note:** Even though some types from this module might be re-exports from third-party
+        //! crates (e.g., `embassy-time`), they are intended to be used "in a self-contained way"
```
ROMemories: Do you think a one-line example would be useful to clarify what "in a self-contained way" means? Something like:

```rust
Timer::after_secs(1).await;
```

chrysn: Yes. Finding how to sleep is currently quite hard to find. (Maybe we can even `#[doc(alias = "sleep")]` selected functions at the re-export).

ROMemories:
> I wouldn't see this as the breaking change, because documentation IMO doesn't make semver properties go away.
>
> Is the plan to long-term make those newtypes?

I hope we can make them newtypes in the future (or even swap the implementation completely if needed). My plan was to make this a breaking change now (as, in a way, we are removing (weak, undocumented but observable) "guarantees" that these types come from `embassy-time`) to *not* make that a breaking change later.

chrysn (approved):
I don't subscribe to "it won't be breaking if we say it's not as it seems", but the change is  good one nonetheless, so let's go.

PR #1325, opened 2025-09-09
docs(sensors): explain how to use `Samples` and `ReadingChannels`

ROMemories:
# Description

This expands the documentation on the automatically adjusted types from the sensor API, to clarify how they are supposed to be used by sensor driver implementors, following #1581 and #1582.

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/Cargo.toml
+++ b/src/ariel-os-sensors/Cargo.toml
@@ -23,7 +23,7 @@ static_cell = { workspace = true }
 [features]
 defmt = ["dep:defmt"]
 
-# These features could be codegened
+# Each feature enables the previous one, in a cascading fashion.
 max-sample-min-count-2 = ["ariel-os-macros/max-sample-min-count-2"]
 max-sample-min-count-3 = [
   "ariel-os-macros/max-sample-min-count-3",


chrysn:
Following up on https://github.com/ariel-os/ariel-os/pull/1355#discussion_r2358067789 that pointed here: If the features are reworked, could the features start depending on their predecessor? Then it'd be easy for an impl that requires a 6-wide sensor reading to say `#[cfg(feature = "max-sample-min-count-6")]`, and it'd be fine if the only max-sample-min-count is 8 because 8 pulls in features 7, 6, 5, …, 2.

nponsard:
Moving [this comment](https://github.com/ariel-os/ariel-os/pull/1581#discussion_r2630413361) from #1581: 

It would be nice to add a comment to clarify what's happening with the features in `ariel-os-sensors/Cargo.toml` (where the highest feature enables the lower feature in cascade). Even if this mechanism is getting replaced later, I would prefer to have it documented.

ROMemories:
> [If the features are reworked, could the features start depending on their predecessor?](https://github.com/ariel-os/ariel-os/pull/1325#issuecomment-3306467576)

The features have been reworked in #1581 and #1582.

PR #1326, opened 2025-09-10
fix(esp-wifi): `isr_stacksize_required` is a list

kaspar030:
# Description

`isr_stacksize_required` must be a list for the `max(...)` machinery to work.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1412,7 +1412,8 @@ modules:
     env:
       global:
         # esp-wifi needs a lot of ISR stack.
-        isr_stacksize_required: "32768"
+        isr_stacksize_required:
+          - "32768"
         # esp-wifi needs 72KiB
         heapsize_required:
           - $(72*1024)


laze-project.yml
```diff
@@ -1412,7 +1412,8 @@ modules:
     env:
       global:
         # esp-wifi needs a lot of ISR stack.
-        isr_stacksize_required: "32768"
```
ROMemories, line 1415: So what was actually the behavior with just a string?

kaspar030: Depending on the order:
1. if a string gets merged with a list, it becomes the list, with the string being dropped
2. if a list gets merged with a string, the string overrides the list

In this case, other modules' `isr_stacksize_required` setting might have been shadowed, or this module's setting could have been ignored.

... this  is suboptimal. => https://github.com/kaspar030/laze/issues/778

Issue #876, opened 2025-02-27
Support matrix: Document ease-of-flashing


chrysn:
### Description

The support matrix should include some information about built-in programmers or supported bootloaders.

### Motivation

Writing a tutorial, I found the data in the support matrix insufficient to guide a newcomer: there is no indication of included flashing support.

### Requirements

Users should have all in there to make a technical choice for whether they want to use a board.

Suggested implementation:

* Column "Programming"
* Values:
  - ✅ Built-in debugger
  - ☑️ Built-in supported bootloader
  - 🚧 Bootloader or debugger present but not supported
  - ❌ Needs external debugger

### Are you planning to do it yourself in a pull request?

Maybe

PR #1327, opened 2025-09-11
docs(hardware-support): hardware support redesign

jkr-ariel:
# Description

This PR implements a first draft of what board pages should look like in our book.

We want (among other things) to generate one page per board we support, and document board specific information such as how to flash this board.

## Issues/PRs references

Linked to #1182.
Linked to #876.

## Change checklist

### DIFF
--- a/book/src/boards/rpi-pico-w.md
+++ b/book/src/boards/rpi-pico-w.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/doc/gen_book.rs
+++ b/doc/gen_book.rs
@@ -47,7 +47,7 @@ const BOARD_PAGE_TEMPLATE: &str = r#"
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -65,13 +65,6 @@ const BOARD_PAGE_TEMPLATE: &str = r#"
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/bbc-microbit-v1.md
+++ b/book/src/boards/bbc-microbit-v1.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/bbc-microbit-v2.md
+++ b/book/src/boards/bbc-microbit-v2.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/espressif-esp32-c3-lcdkit.md
+++ b/book/src/boards/espressif-esp32-c3-lcdkit.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/espressif-esp32-c6-devkitc-1.md
+++ b/book/src/boards/espressif-esp32-c6-devkitc-1.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/espressif-esp32-s3-devkitc-1.md
+++ b/book/src/boards/espressif-esp32-s3-devkitc-1.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/heltec-wifi-lora-32-v3.md
+++ b/book/src/boards/heltec-wifi-lora-32-v3.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/nordic-thingy-91-x-nrf9151.md
+++ b/book/src/boards/nordic-thingy-91-x-nrf9151.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/nrf52840dk.md
+++ b/book/src/boards/nrf52840dk.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/nrf52dk.md
+++ b/book/src/boards/nrf52dk.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
@@ -66,13 +66,6 @@
 </table>
 
 <style>
-@media (min-width: 1920px) {
-  .support-matrix {
-    position: relative;
-    left: 50%;
-    transform: translate(-50%, 0);
-  }
-}
 .support-cell {
   text-align: center;
 }
--- a/book/src/boards/nrf5340dk.md
+++ b/book/src/boards/nrf5340dk.md
@@ -14,7 +14,7 @@
 
 ## Support Matrix
 
-<table class="support-matrix">
+<table>
   <thead>
     <tr>
       <th>Functionality</th>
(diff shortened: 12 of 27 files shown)


book/src/boards/rpi-pico-w.md
bergzand, line 1: This should probably also include the tier level

book/src/boards/index.md
bergzand, line 1: This page could include the lists of boards supported, grouped by support tier level. That would be on top of the board list in the summary.

book/src/SUMMARY.md
```diff
@@ -14,6 +14,8 @@
 - [Randomness and Entropy](./randomness.md)
 - [Multithreading](./multithreading.md)
 - [Persistent Storage](./storage.md)
+- [Boards](./boards/index.md)
+  - [Raspberry Pi Pico W](./boards/rpi-pico-w.md)
```
bergzand: Is it also in scope to generate the list here?

jkr-ariel: The main reason to generate the list here is to actually render the board pages for the doc preview.

jkr-ariel:
> I'm slightly concerned with the page index when we get to >= 100 supported boards.

If what you are concerned about is the potential size of the page (or the left-side menu), I have just added an option to fold the boards section, and inside markdown we can use `<details>` to create a foldable section.

ROMemories:
Could you resolve the conflicts so the docs can be previewed?

ROMemories:
Can you update the [`.gitattributes` file](https://github.com/ariel-os/ariel-os/blob/7a7fd6128ef3e448e85ee5c64eaefcccd022e056/.gitattributes) with files that are auto-generated, so that they are folded by default in GH's view? That'll be easier to review and makes it clear which files are codegened.

ROMemories:
Can you make sure the docs job succeeds so it's easier to have a look at the current state of this PR?

.github/workflows/build-deploy-docs.yml
```diff
@@ -145,6 +145,11 @@ jobs:
           cd book
           ./cmdrun-check.sh
 
+      - name: Generate summary and board pages
+        run: |
+          doc/gen_board_pages_md.rs pages doc/support_matrix.yml book/src/boards/
+          doc/gen_board_pages_md.rs summary doc/support_matrix.yml book/src/SUMMARY.md --template book/src/SUMMARY.md.tmpl
```
ROMemories: Users shouldn't have to run commands before generating the book. Instead we should generate and commit the generated files when updating it, and have a CI check to ensure these are always up to date, as we had before #795.

doc/schema.rs
```diff
@@ -0,0 +1,95 @@
...
+impl SupportInfo {
+    pub fn status(&self) -> &str {
+        match self {
+            SupportInfo::StatusOnly(status) => status,
+            SupportInfo::Details { status, .. } => status,
+        }
+    }
+}
+
+pub enum SupportTier {
```
ROMemories: I don't think the tiers that we have really are "support tiers", but more like "testing tiers": they're not meant to represent our functionality *coverage* but instead how much these boards are tested. I think calling this `Tier` is enough.

doc/gen_board_pages_md.rs
```diff
@@ -0,0 +1,374 @@
...
+|Ariel OS Name|`{{ board_info.technical_name }}`|
+|Flashing|TODO|
+{%- for f in board_info.functionalities %}
+|{{ f.title }}|{{ f.icon}}|
+{%- endfor %}
+
+Legend:
+
+{%- for s in support_keys %}
+  {{ s.icon }} : {{ s.description }}
```
ROMemories: This should use `dl`, `dt`, and `dd` elements as for the support matrix.

book/src/SUMMARY.md.tmpl
```diff
@@ -0,0 +1,33 @@
...
+- [Building an Application](./application.md)
+- [Async Executors](./async-support.md)
+- [Debug Console](./debug-console.md)
+- [Networking](./networking.md)
+- [Randomness and Entropy](./randomness.md)
+- [Multithreading](./multithreading.md)
+- [Persistent Storage](./storage.md)
+- [Boards](./boards/index.md)
+{%- for board in boards %}
+  - [{{ board.name }}](./boards/{{ board.technical_name }}.md)
```
ROMemories: I think it would be better to have a dedicated part before the User Guide, possibly with an Overview page which could contain the current support matrix if we want to keep it.

doc/gen_board_pages_md.rs
```diff
@@ -0,0 +1,374 @@
...
+# {{ board_info.name }}
+
+## Support Matrix
+
+|Feature|Support Status|
+|---|---|
+|Support Tier|{{ board_info.tier }}|
+|Chip|{{ board_info.chip }}|
+|Ariel OS Name|`{{ board_info.technical_name }}`|
+|Flashing|TODO|
```
ROMemories: It would likely be better to move this to a separate table as these are not "Support Status"es. That would also allow us to horizontally center the status emoji.

book/book.toml
```diff
@@ -13,6 +13,11 @@ title = "Ariel OS manual"
 [output.html]
 git-repository-url = "https://github.com/ariel-os/ariel-os"
 edit-url-template = "https://github.com/ariel-os/ariel-os/edit/main/book/{path}"
+[output.html.fold]
+enable = true
+level = 0
```
ROMemories: Can you move this to a dedicated PR as it affects other existing sections?

doc/gen_board_pages_md.rs
ROMemories, line 1: Was there a reason for introducing a separate script? This seems to introduce a lot of duplication.

ROMemories (commented):
Can you revisit your commit history so that the changes are more atomic, and then move the refactor of the schema module to a separate PR?

---

The board lists should be sorted so that their generation is deterministic and that boards from the same manufacturer are more likely to be grouped together.

book/src/SUMMARY.md.tmpl
ROMemories: Also that part should be about "Hardware Support" or similar, as we'll have separate subsections for boards and MCUs in there.

doc/gen_board_pages_md.rs
jkr-ariel: The only reason was to get something running quickly; I also think we should merge some scripts together.

doc/gen_board_pages_md.rs
jkr-ariel: Fixed in 38ca0f86. Not sure how to organise the "Board Info" table though.

doc/gen_board_pages_md.rs
jkr-ariel: Fixed in 43403d07.

doc/schema.rs
jkr-ariel: Fixed in e16fa9c4.

book/src/SUMMARY.md.tmpl
jkr-ariel: Added new "Hardware Support" section in b45ec08e.

.github/workflows/build-deploy-docs.yml
jkr-ariel: Added all generated book pages in e931e111.

doc/schema.rs
```diff
@@ -66,3 +66,30 @@ impl SupportInfo {
         }
     }
 }
+
+pub enum Tier {
```
ROMemories, line 70: Can you move this refactor into a separate PR?

doc/gen_book.rs
```diff
@@ -0,0 +1,743 @@
...
+    },
+    #[error("could not write the output HTML file `{path}`")]
+    WritingOutputFile { path: PathBuf, source: io::Error },
+    #[error("could not read existing output HTML file `{path}`")]
+    ReadingExistingFile { path: PathBuf, source: io::Error },
+    #[error("existing HTML file `{path}` is not up to date")]
+    ExistingHtmlNotUpToDate { path: PathBuf },
+}
+
+mod schema {
```
ROMemories: This should reuse the shared `schema` module.

.github/workflows/build-deploy-docs.yml
```diff
@@ -12,7 +12,7 @@ on:
       - main
     paths:
       - book/**
-      - doc/gen_support_matrix_html.rs
```
ROMemories, line 15: I'm not sure what's happening here; is this file still used? It's also modified by this PR, but if it's not used anymore, it should just be deleted.

doc/gen_book.rs
```diff
@@ -0,0 +1,743 @@
...
+|||
+|---|---|
+|Tier|{{ board_info.tier }}|
+|Ariel OS Name|`{{ board_info.technical_name }}`|
+|Chip|{{ board_info.chip }}|
+|Chip Ariel OS Name|`{{ board_info.chip_technical_name }}`|
+
+## Support Matrix
+
+<table class="support-matrix">
```
ROMemories: This table needs to be horizontally centered, as is the above table.

doc/gen_book.rs
```diff
@@ -0,0 +1,743 @@
...
+{%- endfor %}
+</ul>
+"#;
+
+const BOARD_PAGE_TEMPLATE: &str = r#"
+# {{ board_info.name }}
+
+## Board Info
+
+|||
```
ROMemories: This needs either headers or to be rewritten to just not be a table.

book/src/boards/index.md
```diff
@@ -0,0 +1,21 @@
+# Board Pages
+
+## Tier 1
+
+Tier 1 hardware gets regularly tested, either automatically or manually.
+
+<!-- cmdrun ../../../doc/gen_book.rs board-list ../../../doc/support_matrix.yml /dev/stdout --tier 1 -->
```
ROMemories, line 7: Fine for keeping `mdbook-cmdrun` for now, but we'll remove it in a follow-up as we can do without it again now.

doc/gen_board_list_html.rs
ROMemories, line 1: What's with this script? It doesn't seem to be used anywhere.

doc/gen_board_pages_md.rs
ROMemories: I don't think it's been merged, has it? I don't see it being used anywhere.

doc/schema.rs
jkr-ariel: Moved in #1385.

doc/gen_book.rs
jkr-ariel: Do you mean the "Functionality" column cells should be horizontally centered ?

ROMemories: No I mean that the whole `table` element is right-aligned in the `main` element, so it's globally off-center.

jkr-ariel: That table looks globally centered on the [docs preview](https://ariel-os-docs-preview-pr1327.surge.sh/dev/docs/book/boards/bbc-microbit-v2.html). Or do you want it aligned with the headings?

ROMemories: That table is not centered inside the `main` element:

Screenshot_20250925_142745

ROMemories: This actually depends on the viewport size: it's not centered only if it's wide enough (that's likely due to the media query needing adjustment).

ROMemories (approved):
Great work on this! I had a good look at the previewed output and that looks good and that should be easy to iterate on later as we want to list other info. I had a cursory look at the implementation and checked locally that the `--check` logic worked as expected.

I have a few suggestions that I'll add later and that we'll address in a follow-up, but I'm merging now as the output already looks much better (I've checked that the existing `hardware-functionality-support.html` link is not broken, so we don't have to add an redirect).

ROMemories:
Please squash!

doc/gen_book.rs
```diff
@@ -0,0 +1,638 @@
...
+
+const LIST_TEMPLATE: &str = r#"
+<ul>
+{%- for name, board in boards %}
+  <li><a href=./{{ name }}.html>{{ board.name }}</a></li>
+{%- endfor %}
+</ul>
+"#;
+
+const BOARD_PAGE_TEMPLATE: &str = r#"
```
ROMemories, line 34: This results in a leading blank line in the file.

doc/gen_book.rs
```diff
@@ -0,0 +1,638 @@
...
+    icon: String,
+    description: String,
+    // TODO: add comments
+    // TODO: add the PR link
+}
+
+fn main() -> miette::Result<()> {
+    let args: Args = argh::from_env();
+
+    let input_path = fs::read_to_string(&args.input_path()).map_err(|source| Error::InputFile {
```
ROMemories, line 440:
```suggestion
    let input_file = fs::read_to_string(&args.input_path()).map_err(|source| Error::InputFile {
```

book/src/SUMMARY.md.tmpl
```diff
@@ -0,0 +1,36 @@
+# Summary
+
+- [Introduction](./introduction.md)
+
+# Hardware Support
+
+- [Overview](./hardware-functionality-support.md)
```
ROMemories, line 7: Can you [set a more specific title](https://rust-lang.github.io/mdBook/format/mdbook.html#controlling-page-title) for this page (without changing its name in the summary)? Otherwise we just get "Overview" in the tab bar.

.github/workflows/main.yml
```diff
@@ -424,11 +424,29 @@ jobs:
...
       - name: Check the support matrix generation exit code
         run: |
-          ./doc/gen_support_matrix_html.rs generate doc/support_matrix.yml /dev/null --tier 1
-          ./doc/gen_support_matrix_html.rs generate doc/support_matrix.yml /dev/null --tier 2
-          ./doc/gen_support_matrix_html.rs generate doc/support_matrix.yml /dev/null --tier 3
+          ./doc/gen_book.rs matrix doc/support_matrix.yml /dev/null --tier 1
+          ./doc/gen_book.rs matrix doc/support_matrix.yml /dev/null --tier 2
+          ./doc/gen_book.rs matrix doc/support_matrix.yml /dev/null --tier 3
+
+      - name: Check the book summary generation exit code
```
ROMemories, line 439: Isn't this redundant with the first check above?

.github/workflows/main.yml
```diff
@@ -424,11 +424,29 @@ jobs:
...
+        run: ./doc/gen_book.rs summary doc/support_matrix.yml /dev/null --template-path book/src/SUMMARY.md.tmpl
+
+      - name: Check the board list markdown generation exit code
+        run: |
+          ./doc/gen_book.rs board-list doc/support_matrix.yml /dev/null --tier 1
+          ./doc/gen_book.rs board-list doc/support_matrix.yml /dev/null --tier 2
+          ./doc/gen_book.rs board-list doc/support_matrix.yml /dev/null --tier 3
+
+      - name: Check the board pages generation exit code
```
ROMemories, line 448: Isn't this redundant with the second check above?

ROMemories (commented):
Here are the comments I mentioned above.

PR #1328, opened 2025-09-11
feat(deps)!: bump embassy & esp-hal deps

kaspar030:
# Description

This is my WIP branch for bumping our main embassy deps, plus the rats tail of transitive dependencies.

~~I've concentrated on getting nrf52840dk to build again, the rest is TODO.~~

## Issues/PRs references

split out issues:

- [ ] https://github.com/ariel-os/ariel-os/issues/1673
- [ ] https://github.com/ariel-os/ariel-os/issues/1679
- [ ] https://github.com/ariel-os/ariel-os/issues/1680
- [x] https://github.com/ariel-os/ariel-os/issues/1681

## Open Questions / TODOs

~~- [ ] find solution to `spawner_for_current_executor()` unsoundness and wording~~ => postponed into https://github.com/ariel-os/ariel-os/issues/1673
- [ ] some esp issues still: => split into #1678
  - [ ] Wifi connect flakey
  - [ ] Xiao esp32-s3,c6 Wifi connects only if channel is 1 (!)
- [x] vetting of the new dependencies => added exemptions as discussed
- [x] nrf5340-app examples/sensors-debug pins wrong => ~~enabled `lfxo-as-gpio-pins`~~ using other pins now
- [x] nucleo-f767zi single-bank/dual-bank => made configurable through laze
- [x] switch from `kaspar030` urls
- [x] plain esp32 wifi not working correctly => works OK since backporting https://github.com/esp-rs/esp-hal/pull/4706
- [x] merge @nponsard's riscv threading code
  - [x] merged, basic threading works
  - [x] timer isr doesn't trigger
- [ ] ~~bump to now-current state of deps~~ (postponed)
- [ ] after merge: close https://github.com/probe-rs/embedded-test/issues/77

## Changelog entry

The Embassy and `esp-hal`-related crates have been updated to the following versions:

- `embassy-executor@0.9.1`
- `embassy-net@0.7.1`
- `embassy-nrf@0.8.0`
- `embassy-rp@0.8.0`
- `embassy-stm32@0.4.0`
- `embassy-sync@0.7.2`
- `embassy-time@0.5.0`
- `embassy-usb@0.5.1`
- `embassy-usb-driver@0.2.0`
- `esp-hal@1.0.0`

This comes with some potentially breaking changes:

- The GPIO, I2C, and SPI drivers have been adjusted to take instances of `IntoPeripheral`s following the upstream removal of the unsound `Peripheral` type, which has been replaced with different solutions in Embassy and `esp-hal`. The `IntoPeripheral` trait unifies this in Ariel OS and should most often require no changes to applications.
- Some of these crates were re-exported in `ariel_os::reexports`, upstream breaking changes are therefore exposed this way.
- The `executor-single-thread` async executor flavor has been removed.
- The `Mutex` type required by `ariel_os::i2c::controller::I2cDevice::new()` and `ariel_os::spi::main::SpiDevice::new()` now comes from `embassy-sync@0.7.2`.
- The network stack returned by `ariel_os::net::network_stack()` now comes from `embassy-net@0.7.1`.
- The module-level types from `ariel_os::async` now come from `embassy-executor@0.9.1`.
- The `ariel_os::usb::UsbBuilder` type now comes from `embassy-usb@0.5.1` and the `ariel_os::usb::UsbDriver` type from `embassy-usb-driver@0.2.0`.
- (ESP32-S3) Multicore support on ESP32-S3 had to be disabled as a temporary workaround following changes in `esp-hal` requiring that peripheral instances be pinned to the core they have been initialized on.
- (ESP32) The GPIO peripheral types have been revisited: there is now a dedicated type per GPIO, instead of type aliases.
- (STM32F767ZI) The single-bank flash setup is now enabled on this MCU.

## Change checklist

### DIFF
--- a/examples/threading/laze.yml
+++ b/examples/threading/laze.yml
@@ -1,4 +1,6 @@
 apps:
   - name: example-threading
+    conflicts:
+      - context::st-nucleo-f042k6
     selects:
       - sw/threading


ariel-os-cargo.toml
```diff
@@ -12,38 +12,30 @@ debug = true # Required for defmt
...
-# branch = "embassy-net-v0.6.0+ariel-os"
-embassy-net = { git = "https://github.com/ariel-os/embassy", rev = "3d9f77ee05cdb18f1cd676233bcda9b6f7be3369" }
-# branch = "embassy-rp-v0.4.0+ariel-os+trng-panic-fix"
-embassy-rp = { git = "https://github.com/ariel-os/embassy", rev = "1dd58227a73efa8da2aef96aaccf17b8749b9f7b" }
-# branch = "embassy-stm32-v0.2.0+ariel-os"
-embassy-stm32 = { git = "https://github.com/ariel-os/embassy", rev = "90c8da09646d5fd97b7d416720688dcd3da1ecfb" }
-# branch = "embassy-time-v0.4.0+ariel-os"
-embassy-time = { git = "https://github.com/ariel-os/embassy", rev = "f9b4c8132b07ef696668005846ee2240a8f57573" }
+# branch = "embassy-hal-internal-v0.3.0+ariel-os"
+embassy-hal-internal = { git = "https://github.com/kaspar030/embassy", rev = "0feffeb3c90748112a64c6570a39e1251311accb" }
```
kaspar030: note: these are pointing to my fork now, have to change back before merging

ariel-os-cargo.toml
```diff
@@ -12,38 +12,30 @@ debug = true # Required for defmt
...
+embassy-stm32 = { git = "https://github.com/kaspar030/embassy", rev = "1b75c3d6e61a99a312e2d12f039fa6e76e9c3a9c" }
 
-# ariel-os esp-hal fork, using branch = "v0.23.1+ariel-os-threads"
-esp-alloc = { git = "https://github.com/ariel-os/esp-hal", rev = "aa8d1995fc06eb5e6d04a2022ecf3679b4929e63" }
-esp-hal = { git = "https://github.com/ariel-os/esp-hal", rev = "aa8d1995fc06eb5e6d04a2022ecf3679b4929e63" }
-esp-hal-embassy = { git = "https://github.com/ariel-os/esp-hal", rev = "aa8d1995fc06eb5e6d04a2022ecf3679b4929e63" }
-esp-println = { git = "https://github.com/ariel-os/esp-hal", rev = "aa8d1995fc06eb5e6d04a2022ecf3679b4929e63" }
-esp-wifi = { git = "https://github.com/ariel-os/esp-hal", rev = "aa8d1995fc06eb5e6d04a2022ecf3679b4929e63" }
-xtensa-lx-rt = { git = "https://github.com/ariel-os/esp-hal", rev = "aa8d1995fc06eb5e6d04a2022ecf3679b4929e63" }
+# ariel-os esp-hal fork, using branch = "esp-hal-20250911+ariel-os"
```
kaspar030: note: esp-hal has no released version supporting e-executor 0.9, so pointing at main + ariel patches

Cargo.toml
```diff
@@ -146,16 +146,16 @@ once_cell = { version = "=1.19.0", default-features = false, features = [
...
+rand_core = { version = "0.9.3", default-features = false }
 rtt-target = { version = "0.6.0" }
 
 rp-pac = { version = "7.0", default-features = false }
 serde = { version = "1.0.197", default-features = false }
 static_cell = { version = "2.1.1", default-features = false }
-trouble-host = "0.1"
-bt-hci = { version = "0.2.0" }
-nrf-sdc = { git = "https://github.com/ariel-os/nrf-sdc.git", rev = "8ce4780b691ec6850b05a0382cef395846b67e02" }
+trouble-host = { git = "https://github.com/embassy-rs/trouble", rev = "26ec6794d0a976dd91cd694b631c72b16ddf651b" }
```
kaspar030: needed for embassy updates. 0.3.0 around the corner.

src/ariel-os-embassy-common/src/cell.rs
```diff
@@ -52,15 +52,17 @@ impl<T> SameExecutorCell<T> {
     ///
     /// Despite being async, this function never blocks/yields, it returns instantly.
     pub async fn new_async(inner: T) -> Self {
-        let spawner = Spawner::for_current_executor().await;
+        // SAFETY: safe unless intentionally exploited.
```
kaspar030: I don't really know how to phrase this better.

src/ariel-os-hal/src/gpio.rs
```diff
@@ -343,13 +343,13 @@ pub mod output {
...
-    use crate::hal::{self, gpio::output::OutputPin as HalOutputPin, peripheral::Peripheral};
+    use crate::hal::{self, gpio::output::OutputPin as HalOutputPin, peripheral::Peri};
 
     use super::{HalDriveStrength, HalSpeed, Output};
 
     /// Builder type for [`Output`], can be obtained with [`Output::builder()`].
-    pub struct OutputBuilder<P: Peripheral<P: HalOutputPin>> {
-        pub(crate) pin: P,
+    pub struct OutputBuilder<P: HalOutputPin> {
+        pub(crate) pin: Peri<'static, P>,
```
kaspar030: This did not previously had any mention of `'static`, but the builder had, so I think it is not more restrictive than before.

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -30,7 +30,7 @@ macro_rules! define_peripherals {
         pub struct $peripherals {
             $(
                 $(#[$inner])*
-                pub $peripheral_name: peripherals::$peripheral_field
+                pub $peripheral_name: $crate::peripheral::Peri<'static, peripherals::$peripheral_field>
```
kaspar030: This is required as upstream `Peripherals` now contains `Peri<'static, <type>`.

src/ariel-os-nrf/src/ble.rs
```diff
@@ -6,14 +6,16 @@ use nrf_sdc::{
...
+use trouble_host::{Stack, prelude::DefaultPacketPool};
 
 use ariel_os_debug::log::debug;
 use ariel_os_embassy_common::ble::MTU;
 
-use crate::irqs::Irqs;
+use crate::{irqs::Irqs, peripheral::Peri};
 
-static STACK: OnceLock<Stack<'static, SoftdeviceController<'static>>> = OnceLock::new();
+pub type BleStack = Stack<'static, SoftdeviceController<'static>, DefaultPacketPool>;
```
kaspar030, line 19: I pulled out the type as it was used twice.

src/ariel-os-nrf/src/hwrng.rs
```diff
@@ -10,16 +10,15 @@ pub fn construct_rng(peripherals: &mut crate::OptionalPeripherals) {
...
-            let rng = embassy_nrf::rng::Rng::new(
+            let p =
                 peripherals
                     .RNG
-                    // We don't even have to take it out, just use it to seed the RNG
-                    .as_mut()
-                    .expect("RNG has not been previously used"),
-                Irqs,
-            );
+                    .take()
```
kaspar030, line 16: Here I didn't manage to make the `Rng` constructor take a `Peri` *reference`. The hwrng is now taken forever, even though it is only used for initialization. Suggestions welcome.

src/ariel-os-nrf/src/i2c/controller/mod.rs
```diff
@@ -168,9 +170,17 @@ macro_rules! define_i2c_drivers {
                     // peripheral multiple times.
                     let twim_peripheral = unsafe { peripherals::$peripheral::steal() };
 
+                    // NOTE(hal): the I2C implementation for nrf needs a buffer, define it here.
```
kaspar030, line 175: yeah ... We need to figure out if this size defines the maximum allowed transfer in all cases, or only when using non-ram sources. And what is a reasonable default. ...

src/ariel-os-nrf/src/i2c/controller/mod.rs
```diff
@@ -196,15 +206,15 @@ macro_rules! define_i2c_drivers {
 // We cannot impl From because both types are external to this crate.
 fn from_error(err: embassy_nrf::twim::Error) -> ariel_os_embassy_common::i2c::controller::Error {
     use embassy_nrf::twim::Error::{
-        AddressNack, BufferNotInRAM, DataNack, Overrun, Receive, RxBufferTooLong, Timeout,
+        AddressNack, DataNack, Overrun, RAMBufferTooSmall, Receive, RxBufferTooLong, Timeout,
```
kaspar030, line 211: The old error doesn't exist anymore, but there's a new one.

src/ariel-os-random/src/lib.rs
```diff
@@ -84,8 +84,21 @@ impl RngCore for FastRng {
     fn fill_bytes(&mut self, dest: &mut [u8]) {
         self.inner.fill_bytes(dest);
     }
-    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand_core::Error> {
-        self.inner.try_fill_bytes(dest)
+}
+
+impl rand_core_06::RngCore for FastRng {
```
kaspar030, line 95: After bumping (which is needed for ble), coap/lakers didn't build anymore, as it needs the old rand_core trait. So I copied what embassy-nrf is doing, and just additionally implement the rand_core v0.6 traits.

Cargo.toml
```diff
@@ -146,16 +146,16 @@ once_cell = { version = "=1.19.0", default-features = false, features = [
 ] }
 paste = { version = "1.0" }
 pin-project = "1.1.10"
-rand = { version = "0.8.5", default-features = false }
-rand_core = { version = "0.6.4", default-features = false }
+rand = { version = "0.9.2", default-features = false }
```
kaspar030, line 162: The rand bumps can be pulled out.

src/ariel-os-nrf/src/i2c/controller/mod.rs
kaspar030: upstream says this:

```
      /// `tx_ram_buffer` is required if any write operations will be performed with data that is not in RAM.                                                  
      /// Usually this is static data that the compiler locates in flash instead of RAM. The `tx_ram_buffer`                                                   
      /// needs to be at least as large as the largest write operation that will be executed with a buffer                                                     
      /// that is not in RAM. If all write operations will be performed from RAM, an empty buffer (`&[]`) may                                                  
      /// be used.                                                                                                                                             
```

chrysn:
What do you need to move this from draft to ready?

kaspar030:
> What do you need to move this from draft to ready?

time. :/

So I lost quite some time to work out the embassy `Peri` / esp-hal singleton mismatch. Turned out to not be that difficult in the end, but in the branch, I haven't fully updated all four implementations, yet. That's a day or two of grunt work.

Then there were a bunch of changes in esp-hal. E.g., the embassy timer driver that was previously usable on its own was dropped and integrated into esp-rtos, which we cannot use. So I stitched together a new timer driver for esp. That seems to be mostly working. I do have some "unhandled interrupt" panics on esp, though, that need debugging.

Next up is bumping our esp wifi scheduler integration to the new esp-wifi-preempt-driver. Haven't started with that. IIUC this now needs semaphores with some new semantics, that could take a bit of time.

chrysn:
If it is time as in "I personally have no time to work": All fine, let me know how we can all help.

If it is time as in "during my work time anyone thinks that something else is more urgent": I'd be glad to discuss with that someone.

(Some but not all of that is "We could already have awesome BLE examples, but with the trouble version we have due to all the other dependencies, effort goes into making examples on platform-specific Embassy-based code rather than working on Ariel")

kaspar030:
> If it is time as in "I personally have no time to work": All fine, let me know how we can all help.

It's this, or "so much else also important and also all the meetings".

Some help would be great. An easy to delegate task comes to mind: this branch contains a commit bumping rand to a newer version, but also keeping the old 0.6 trait impls. Would be great if someone(tm) could cut that out and PR seperately. There's a somewhat related [issue](https://github.com/ariel-os/ariel-os/issues/1396).

nponsard:
Seems like RISCV esp targets will need more work for the scheduler: https://github.com/esp-rs/esp-hal/pull/3903

chrysn:
Awesome work, now running my BLE experiments on this! (With little changes, see https://github.com/kaspar030/ariel-os/pull/1)

Closes: https://github.com/ariel-os/ariel-os/pull/1307

src/ariel-os-hal/src/gpio.rs
```diff
@@ -59,43 +60,47 @@ macro_rules! inner_impl_input_methods {
 ///
 /// If support for external interrupts is needed, use [`InputBuilder::build_with_interrupt()`] to
 /// obtain an [`IntEnabledInput`].
-pub struct Input {
-    input: HalInput<'static>, // FIXME: is this ok to require a 'static pin?
+pub struct Input<'a> {
```
ROMemories: Is it absolutely necessary to introduce a lifetime parameter here? I don't see why it would be needed and that would be a serious breaking change.

chrysn: If it's needless, sure, drop it.

But this whole PR is a big breaking change and we committed to having this kind of breaking change the first time any embassy item showed up in our public APIs. Doing things in any way more complicated just because it's a breaking change is premature at this point.

ariel-os-cargo.toml
kaspar030: pointing 1t 1.0.0+ariel-os now

src/ariel-os-alloc/src/lib.rs
```diff
@@ -80,12 +80,29 @@ mod alloc {
     unsafe fn init_esp_alloc() {
         use ariel_os_debug::log::debug;
 
+        const RECLAIMED_SIZE: usize = 64000;
```
bugadani: I don't remember if I've told you this before, but this should now be available through [`memory_range!("DRAM2_UNINIT")`](https://docs.espressif.com/projects/rust/esp-metadata-generated/0.3.0/esp32s3/esp_metadata_generated/macro.memory_range.html)

kaspar030: Yes, already using this. Thanks!

src/ariel-os-esp/src/time_driver.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[cfg(all(context = "xtensa", not(context = "esp32")))]
+use esp_hal::timer::systimer::Alarm as Timer;
+
+#[cfg(all(context = "xtensa", context = "esp32"))]
+use esp_hal::timer::timg::Timer;
+
+
+#[cfg(not(context = "xtensa"))]
+use esp_hal::timer::systimer::Alarm as Timer;
```
bugadani: This can be a pair of `cfg(esp32) -> timg::Timer` and `cfg(not(esp32)) -> systimer::Alarm as Timer`.

ROMemories:
This contains breaking changes, the PR needs to be marked to reflect that, and the breaking changes listed (at least in the PR description, ideally in the Conventional Commits format).

chrysn:
At this point the proper label is probably "feat!(deps): bump embassy & esp-hal deps" (because it is more than just bumping versions, it's enabling the new use).

I wouldn't hold this for until we have a detailed list of breaking changes. A "how to migrate" can still be added later to the docs, but the title combined with knowing that we *do* pub use embassy APIs should suffice for here.

(And I see no point in annotating that commit-by-commit; the PR-as-a-whole is breaking, that should be enough indication. Not sure if you meant that.)

src/ariel-os-threads/src/sync/wait_queue.rs
ROMemories, line 1: Is there anything that prevents moving this `WaitQueue` into a separate PR?

src/ariel-os-threads/src/core_affinity.rs
ROMemories, line 1: Looks like this refactor around core affinity could be moved to a separate PR.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -204,7 +182,8 @@ fn init() {
 #[embassy_executor::task]
 #[allow(clippy::too_many_lines)]
 async fn init_task(mut peripherals: hal::OptionalPeripherals) {
-    let spawner = asynch::Spawner::for_current_executor().await;
+    // SAFETY: safe unless intentionally exploited.
```
ROMemories: (Might need reflowing)

```suggestion
    // SAFETY: this is executed in a non-public task using only Embassy executors.
```

Cargo.toml
```diff
@@ -73,19 +72,20 @@ portable-atomic = { version = "1.11.0", default-features = false, features = [
...
-embassy-stm32 = { version = "0.2", default-features = false }
-embassy-sync = { version = "0.6.1", default-features = false }
-embassy-time = { version = "0.4.0", default-features = false }
-embassy-time-queue-utils = { version = "0.1.0", default-features = false }
-embassy-usb = { version = "0.4.0", default-features = false }
+embassy-embedded-hal = { version = "0.5.0", default-features = false }
+embassy-executor = { version = "0.9.1", default-features = false }
+embassy-futures = { version = "0.1.2", default-features = false }
+embassy-hal-internal = { version = "0.3.0", default-features = false }
+embassy-net = { version = "0.7.1", default-features = false }
```
ROMemories, line 80: After some quick testing, it seems to me `embassy-net` could easily be bumped in a separate PR.

Cargo.toml
```diff
@@ -73,19 +72,20 @@ portable-atomic = { version = "1.11.0", default-features = false, features = [
...
-embassy-net-driver-channel = { version = "0.3.0", default-features = false }
-embassy-nrf = { version = "0.3.1", default-features = false }
-embassy-rp = { version = "0.4", default-features = false }
-embassy-stm32 = { version = "0.2", default-features = false }
-embassy-sync = { version = "0.6.1", default-features = false }
-embassy-time = { version = "0.4.0", default-features = false }
-embassy-time-queue-utils = { version = "0.1.0", default-features = false }
-embassy-usb = { version = "0.4.0", default-features = false }
+embassy-embedded-hal = { version = "0.5.0", default-features = false }
+embassy-executor = { version = "0.9.1", default-features = false }
```
ROMemories, line 77: After some testing, I think we could bump `embassy-executor` in a separate PR stacked on top of dropping `executor-single-thread`. We would end up with two versions of it for ESP but the overhead seems very minimal as a temporary measure (< 100 bytes of RAM and < 100 bytes of flash).

chrysn: Would that benefit the outcome and speediness? Sure this *can* be done in steps, but this effort has been going on for almost half a year. Sure, if there are tangible benefits to splitting, why not, but we shouldn't pile any more work on this if it's just for merge history cleanliness.

src/ariel-os-hal/src/hal/dummy/peripheral.rs
```diff
@@ -1,11 +1,23 @@
-pub use embassy_hal_internal::Peripheral;
-
 /// Dummy type.
 ///
 /// See the `OptionalPeripherals` type of your Embassy HAL crate instead.
 #[doc(hidden)]
 pub struct OptionalPeripherals;
 
+#[doc(hidden)]
+pub trait IntoPeripheral<'a, P> {
```
ROMemories, line 8: Why is this `doc(hidden)`? We can't have doc-hidden traits as input types in the docs.

ariel-os-cargo.toml
kaspar030: done

src/ariel-os-hal/src/hal/dummy/peripheral.rs
kaspar030: fixed

src/ariel-os-hal/src/gpio.rs
kaspar030: I removed the lifetimes again, let's revisit after this is in.

ROMemories: There are still new public lifetime generics in the dummy HAL module, in the `usb-keyboard` example, on `SpiDevice`, and in the arguably private-ish ESP crates.

src/ariel-os-nrf/Cargo.toml
```diff
@@ -59,6 +59,7 @@ nrf-sdc = { workspace = true, features = ["nrf52840"], optional = true }
 [target.'cfg(context = "nrf5340")'.dependencies]
 # Disable NFC support for now, as we do not support it yet.
 embassy-nrf = { workspace = true, features = [
+  "lfxo-pins-as-gpio",
```
ROMemories: Do we really want to do this? I think we should instead change the few tests/examples that use the `P0_00` and `P0_01` pins on the nRF5340.

kaspar030: done

ROMemories: Any chance you'd move that to a separate PR so I can approve that fix right away?

src/ariel-os-random/src/lib.rs
ROMemories, line 1: I think this fix should actually be on `main`, we don't even have to update anything for it (even though `embassy-sync` v0.7 *requires* it): can you open a PR for it?

src/ariel-os-embassy-common/src/cell.rs
kaspar030: marked them as TODO, will open issue

examples/threading-waitqueue/README.md
```diff
@@ -0,0 +1,43 @@
+# Threading Events
```
kaspar030: TODO (readme is from "events" still)

src/ariel-os-coap/src/udp_nal/util.rs
```diff
@@ -69,6 +69,7 @@ impl embedded_io_async::Error for Error {
             // error.
             // FIXME we're not there in this impl yet.
             Self::RecvError(udp::RecvError::Truncated) => embedded_io_async::ErrorKind::Other,
+            Self::SendError(udp::SendError::PacketTooLarge) => todo!(),
```
kaspar030, line 72: `todo!()` addressed in #1653.

src/ariel-os-embassy-common/src/cell.rs
kaspar030: https://github.com/ariel-os/ariel-os/issues/1673

src/ariel-os-esp/src/wifi/esp_wifi/wait_queue.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    // Safety: caller must ensure not deleting an in-use wait queue
+    unsafe fn delete(queue: WaitQueuePtr) {
+        let q = unsafe { Box::from_raw(queue.cast::<ArielWaitQueue>().as_ptr()) };
+        core::mem::drop(q);
+    }
+
+    unsafe fn wait_until(queue: WaitQueuePtr, deadline_instant: Option<u64>) {
+        #[expect(unused_unsafe)]
+        let queue = unsafe { ArielWaitQueue::from_ptr(queue) };
```
ROMemories: Why is there an `unsafe` block on these?

src/ariel-os-esp/src/lib.rs
```diff
@@ -42,130 +44,45 @@ pub mod usb;
...
 }
 
+#[cfg(feature = "time")]
+mod time_driver;
+
 #[doc(hidden)]
 pub use esp_hal::peripherals::OptionalPeripherals;
 
-#[cfg(feature = "executor-single-thread")]
+// TODO(bump):
```
ROMemories: What's the current state of this TODO?

src/ariel-os-esp/src/lib.rs
```diff
@@ -180,7 +97,10 @@ pub fn init() -> OptionalPeripherals {
         }
     };
 
-    esp_hal_embassy::init(embassy_timer);
+    crate::time_driver::init(embassy_timer);
```
ROMemories, line 98: Looks like this should be gated on the `time` Cargo feature, and then `ariel-os-hal` wouldn't have to enable the `time` Cargo feature on `ariel-os-esp` (which shouldn't be done by `ariel-os-hal` anyway if it's now a hard requirement of `ariel-os-esp`).

src/ariel-os-macros/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license.workspace = true
 proc-macro = true
 
 [dependencies]
-proc-macro-crate = "3.1.0"
+proc-macro-crate = "3.4.0"
```
ROMemories, line 14: This seems entirely unrelated?

src/ariel-os-stm32/src/lib.rs
```diff
@@ -67,6 +67,21 @@ static SHARED_DATA: MaybeUninit<SharedData> = MaybeUninit::uninit();
 #[doc(hidden)]
 pub static EXECUTOR: Executor = Executor::new();
 
+// TODO(bump):
```
ROMemories: What's the current state of this TODO?

laze-project.yml
```diff
@@ -576,6 +576,7 @@ contexts:
       PROBE_RS_CHIP: STM32F767ZITx
       RUSTFLAGS:
         - --cfg capability=\"hw/stm32-rng\"
+        - --cfg use_stm32_single_bank
```
ROMemories, line 579: Shouldn't this use the `capability` `cfg` as done above?

src/ariel-os-threads/src/arch/riscv.rs
```diff
@@ -90,80 +138,313 @@ const fn default_trap_frame() -> TrapFrame {
...
-            .cpu_intr_from_cpu_0()
-            .modify(|_, w| w.cpu_intr_from_cpu_0().clear_bit());
-
-        sched(trap_frame)
-    }
-}
+global_asm!(
+    r#"
+
+    .section .trap.rust, "ax"          // FIXME: is this right ?
```
ROMemories, line 157: Should this FIXME be checked?

src/ariel-os-threads/src/arch/riscv.rs
```diff
@@ -90,80 +138,313 @@ const fn default_trap_frame() -> TrapFrame {
...
             }
-            *scheduler.current_tid_mut() = Some(next_tid);
+            let next = scheduler.get_unchecked_mut(next_tid);
+            next.data.mstatus = mstatus;
 
-            copy_registers(&scheduler.get_unchecked(next_tid).data, trap_frame);
-            true
+            let next_high_regs = &raw mut next.data;
+            trace!("next cleanup: {}", next.data.ra);
+            Some((current_high_regs as u32, next_high_regs as u32))
```
ROMemories, line 418: `current_high_regs` is zero at this point if `scheduler.current_tid_mut()` returned `None` above, is this fine?

laze-project.yml
```diff
@@ -470,9 +469,10 @@ contexts:
     selects:
       - xtensa
     provides:
-      - has_multi_core_support
+      # TODO: re-enable as soon as upstream peripheral drivers become `Send`
+      # in combination with multi-core.
+      # - has_multi_core_support
```
ROMemories, line 474: So this completely disables multicore support on the ESP32-S3 for now? That's unfortunate.

src/ariel-os-embassy/Cargo.toml
```diff
@@ -61,6 +61,8 @@ embassy-executor = { workspace = true, default-features = false, features = [
 embassy-nrf = { workspace = true }
 embedded-io-async = { workspace = true }
 
+[target.'cfg(context = "esp")'.dependencies]
+esp-hal = { workspace = true, default-features = false }
```
ROMemories: Is this needed? Grepping for `esp_hal` in `ariel-os-embassy` doesn't find anything.

ROMemories (commented):
Definitely not a complete review; I'd also like to see SAFETY comments on every new `unsafe` block.

src/ariel-os-threads/src/arch/riscv.rs
nponsard: I put this comment here, I don't know enough about how those statements work, I got it from https://github.com/esp-rs/esp-hal/blob/main/hil-test/src/bin/interrupt.rs .

src/ariel-os-rp/src/uart.rs
```diff
@@ -182,10 +186,10 @@ macro_rules! define_uart_drivers {
                 /// This never returns an error.
                 #[expect(clippy::new_ret_no_self)]
                 pub fn new(
-                    rx_pin: impl Peripheral<P: RxPin<peripherals::$peripheral>> + 'd,
-                    tx_pin: impl Peripheral<P: TxPin<peripherals::$peripheral>> + 'd,
-                    rx_buf: &'d mut [u8],
-                    tx_buf: &'d mut [u8],
+                    rx_pin: Peri<'static, impl RxPin<peripherals::$peripheral>>,
+                    tx_pin: Peri<'static, impl TxPin<peripherals::$peripheral>>,
```
bergzand:
```suggestion
                    rx_pin: Peri<'d, impl RxPin<peripherals::$peripheral>>,
                    tx_pin: Peri<'d, impl TxPin<peripherals::$peripheral>>,
```

src/ariel-os-nrf/src/uart.rs
```diff
@@ -236,9 +236,9 @@ macro_rules! define_uart_drivers {
                 ///
                 /// This never returns an error.
                 #[expect(clippy::new_ret_no_self)]
-                pub fn new(
-                    rx_pin: impl Peripheral<P: GpioPin> + 'd,
-                    tx_pin: impl Peripheral<P: GpioPin> + 'd,
+                pub fn new<T: GpioPin, U: GpioPin>(
+                    rx_pin: Peri<'d, T>,
+                    tx_pin: Peri<'d, U>,
```
bergzand: For consistency with the other drivers
```suggestion
                pub fn new(
                    rx_pin: Peri<'d, impl GpioPin>,
                    tx_pin: Peri<'d, impl GpioPin>,
```

src/ariel-os-threads/src/arch/riscv.rs
nponsard: This is handled in the assembly code, line 195 of the file

src/ariel-os-esp/src/wifi/esp_wifi/scheduler.rs
```diff
@@ -0,0 +1,188 @@
...
+        _name: &str,
+        task: extern "C" fn(*mut c_void),
+        param: *mut c_void,
+        priority: u32,
+        pin_to_core: Option<u32>,
+        task_stack_size: usize,
+    ) -> ThreadPtr {
+        trace!("task_create()");
+
+        let stack_slice = alloc_aligned_leaked_buffer(task_stack_size, 8);
```
bugadani: I believe the minimum required stack alignment is 16 bytes on both architectures so this may undersize the stack just a bit

kaspar030: fixed!

src/ariel-os-embassy/Cargo.toml
kaspar030: dropped

laze-project.yml
kaspar030: Yes it is. Not much we can do about it. Upstream is aware and promised a fix down the line.

src/ariel-os-macros/Cargo.toml
kaspar030: maybe, 3.4.0 is aligned with the other crates though

src/ariel-os-esp/src/lib.rs
kaspar030: actually done.

src/ariel-os-esp/src/wifi/esp_wifi/wait_queue.rs
kaspar030: not needed, I dropped them

examples/threading/laze.yml
```diff
@@ -1,4 +1,6 @@
 apps:
   - name: example-threading
+    conflicts:
+      - context::st-nucleo-f042k6
```
kaspar030, line 4: Adding the extra `deadline` field to `Thread` shoves this example 68 bytes over the line, not enough memory.
There are multiple options to explore. Let's do that as follow-up.

kaspar030: https://github.com/ariel-os/ariel-os/issues/1679

src/ariel-os-nrf/src/i2c/controller/mod.rs
kaspar030: https://github.com/ariel-os/ariel-os/issues/1680

src/ariel-os-random/src/lib.rs
kaspar030: not splitting out

src/ariel-os-stm32/src/lib.rs
kaspar030: done actually

src/ariel-os-threads/src/core_affinity.rs
kaspar030: not moving out

src/ariel-os-threads/src/sync/wait_queue.rs
kaspar030: not moving out

Cargo.toml
kaspar030: it is already bumped here. let's merge and then rebase #1653 on top of this. I've cherry-picked the vetting commit from #1653.

Cargo.toml
kaspar030: not splitting out.

examples/threading-waitqueue/README.md
kaspar030: removed the example

laze-project.yml
kaspar030: Didn't we agree on `has_FOO` and `use_BAR`?

src/ariel-os-threads/src/arch/riscv.rs
kaspar030: the section seems fine to me, so do the attributes. Let's leave the comment so the next one looking at this has a reason to be distracted.

src/ariel-os-esp/src/lib.rs
kaspar030: https://github.com/ariel-os/ariel-os/issues/1681

laze-project.yml
ROMemories: I'm not sure we really discussed this in the case of MCUs (only for SBDs, for boards), but after looking into what this does a bit more, `use_` seems fine, especially as that's an actual configuration choice.

PR #1330, opened 2025-09-12
feat(sensors): add a sensor driver for the built-in nRF91 GNSS

nponsard:
# Description

This PR adds the GNSS support for the nrf91 SiP, exposing it via the sensor abstraction.

## Testing 

Doc can be open using:

```sh
cargo doc -p ariel-os-sensor-nrf91-gnss --features "nrf-modem/nrf9160, ariel-os-debug-log/log" --open 
```

Can be tested on hardware using: 

```sh
laze -C examples/sensors-debug/ build -b nrf9160dk-nrf9160 run
```

## Issues/PRs references

- Depends on #1332
- Depends on #1354
- Depends on #1398
- Depends on #1351 
- Depends on #1683 
- Depends on #1981

## Open Questions

u8 for the deviation is quite limiting, for example the heading accuracy number returned by the modem is 180 when no movement is detected, this means the accuracy can only be indicated down to the degree.
Same thing for the latitude and longitude, the modem returns accuracy numbers over 25 meters on the first seconds.
Currently I clamp the deviation values to stay within the u8 limits, for example I wanted to have `-1 scaling` on the position accuracy, meaning if the reported accuracy number is greater than 25.5 meters it gets clamped down to 25.5

The main modem initialization has to happen in `ariel-os-nrf` so it can also be used to setup LTE-M networking, right now that means the app has to explicitly select the `nrf91-modem` laze feature, it would be better to have the GNSS driver express the dependency. 

## Changelog Entry

The GNSS on the nRF91 SiP family can now be used as a sensor through Ariel OS's sensor API.

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ members = [
   "src/lib/ringbuffer",
   "src/sensors/ariel-os-sensor-lis2du12",
   "src/sensors/ariel-os-sensor-lps22df",
+  "src/sensors/ariel-os-sensor-nrf91-gnss",
   "src/sensors/ariel-os-sensor-stts22h",
   "tests/benchmarks/bench_sched_flags",
   "tests/benchmarks/bench_sched_yield",
@@ -152,6 +153,7 @@ ariel-os-utils = { path = "src/ariel-os-utils", default-features = false }
 # Built-in sensor drivers.
 ariel-os-sensor-lis2du12 = { path = "src/sensors/ariel-os-sensor-lis2du12" }
 ariel-os-sensor-lps22df = { path = "src/sensors/ariel-os-sensor-lps22df" }
+ariel-os-sensor-nrf91-gnss = { path = "src/sensors/ariel-os-sensor-nrf91-gnss" }
 ariel-os-sensor-stts22h = { path = "src/sensors/ariel-os-sensor-stts22h" }
 
 const-str = "1.0.0"
@@ -182,7 +184,9 @@ trouble-host = { version = "0.5.0" }
 #nrf-sdc = { version = "0.4.0" }
 nrf-sdc = { git = "https://github.com/alexmoon/nrf-sdc", rev = "e64ad081a6948b3439d1e811da36d39a5785dbda" }
 
+futures-util = { version = "0.3.32", default-features = false }
 nrf-modem = { version = "0.10.2" }
+time = { version = "0.3.47", default-features = false }
 tinyrlibc = "0.5.0"
 
 [workspace.lints.rust]
--- a/examples/sensors-debug/Cargo.toml
+++ b/examples/sensors-debug/Cargo.toml
@@ -13,6 +13,7 @@ ariel-os = { path = "../../src/ariel-os", features = [
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 ariel-os-sensor-lis2du12 = { workspace = true }
 ariel-os-sensor-lps22df = { workspace = true }
+ariel-os-sensor-nrf91-gnss = { workspace = true, optional = true }
 ariel-os-sensor-stts22h = { workspace = true }
 ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
 ariel-os-sensors-gnss-time-ext = { workspace = true, optional = true }
@@ -22,7 +23,11 @@ once_cell = { workspace = true }
 [features]
 # Enable this feature when a GNSS driver is in use to print the time.
 gnss = ["dep:ariel-os-sensors-gnss-time-ext"]
-defmt = ["ariel-os-sensors-gnss-time-ext?/defmt"]
+defmt = [
+  "ariel-os-sensor-nrf91-gnss?/defmt",
+  "ariel-os-sensors-gnss-time-ext?/defmt",
+]
+nrf-modem = ["dep:ariel-os-sensor-nrf91-gnss", "gnss"]
 
 [lints]
 workspace = true
--- a/examples/sensors-debug/laze.yml
+++ b/examples/sensors-debug/laze.yml
@@ -20,6 +20,7 @@ apps:
       - stm32u083c-dk
     selects:
       - ?defmt-app
+      - ?nrf91-modem-sensor
 
 modules:
   - name: defmt-app
@@ -29,3 +30,11 @@ modules:
       global:
         FEATURES:
           - defmt
+
+  - name: nrf91-modem-sensor
+    selects:
+      - nrf91-modem
+    env:
+      global:
+        FEATURES:
+          - nrf-modem
--- a/examples/sensors-debug/src/sensors.rs
+++ b/examples/sensors-debug/src/sensors.rs
@@ -10,6 +10,9 @@ pub async fn init() {
 
     #[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
     stts22h::init().await;
+
+    #[cfg(context = "nrf91")]
+    nrf91::init().await;
 }
 
 #[cfg(any(context = "st-steval-mkboxpro"))]
@@ -124,3 +127,24 @@ mod stts22h {
 #[allow(unused, reason = "should be directly accessible without going through the registry")]
 #[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
 pub use stts22h::STTS22H_I2C;
+
+#[cfg(context = "nrf91")]
+mod nrf91 {
+    use ariel_os_sensor_nrf91_gnss::{Nrf91Gnss, config::Config};
+
+    pub static NRF91_GNSS: Nrf91Gnss = const { Nrf91Gnss::new(Some("onboard")) };
+
+    #[ariel_os::reexports::linkme::distributed_slice(ariel_os::sensors::SENSOR_REFS)]
+    #[linkme(crate = ariel_os::reexports::linkme)]
+    static NRF91_GNSS_REF: &'static dyn ariel_os::sensors::Sensor = &NRF91_GNSS;
+    pub(super) async fn init() {
+        let mut config = Config::default();
+        config.log_nmea = true;
+        NRF91_GNSS.init(config).await;
+    }
+
+    #[ariel_os::task(autostart)]
+    pub async fn nrf91_gnss_runner() {
+        NRF91_GNSS.run().await;
+    }
+}
--- a/src/sensors/ariel-os-sensor-nrf91-gnss/Cargo.toml
+++ b/src/sensors/ariel-os-sensor-nrf91-gnss/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "ariel-os-sensor-nrf91-gnss"
+# This crate is versioned separately from Ariel OS.
+version = "0.1.0"
+edition.workspace = true
+# This crate's MSRV is decoupled from Ariel OS's.
+rust-version = "1.90"
+repository.workspace = true
+license.workspace = true
+
+[dependencies]
+ariel-os-debug-log = { workspace = true }
+ariel-os-sensors = { workspace = true, features = ["max-sample-min-count-8"] }
+ariel-os-sensors-gnss-time-ext = { workspace = true }
+ariel-os-sensors-utils = { workspace = true }
+defmt = { workspace = true, optional = true }
+embassy-futures = { workspace = true }
+embassy-sync = { workspace = true }
+futures-util = { workspace = true, default-features = false }
+libm = "0.2.15"
+nrf-modem = { workspace = true, features = ["modem-log"] }
+portable-atomic = { workspace = true }
+time = { workspace = true }
+
+[features]
+defmt = ["dep:defmt", "ariel-os-sensors/defmt", "nrf-modem/defmt"]
+
+[lints]
+workspace = true
--- a/src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
+++ b/src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
@@ -1,0 +1,54 @@
+/// Operation modes for the GNSS module.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssOperationMode {
+    /// Always keep the GNSS module active.
+    Continuous,
+    /// Update the GNSS fix periodically.
+    /// The period is defined in seconds and should be in the range 10..=65535.
+    Periodic(u16),
+    /// Try to get a GNSS fix only when requested. Timeout is defined in seconds, 300 recommended.
+    SingleShot(u16),
+}
+
+/// Configuration for the GNSS sensor.
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct Config {
+    /// The GNSS operating mode to use.
+    pub operation_mode: GnssOperationMode,
+    /// Whether NMEA messages should be logged as debug logs (adds extra processing).
+    pub log_nmea: bool,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            operation_mode: GnssOperationMode::Continuous,
+            log_nmea: false,
+        }
+    }
+}
+
+pub(crate) fn convert_gnss_config(config: &Config) -> nrf_modem::GnssConfig {
+    nrf_modem::GnssConfig {
+        // Satellites with elevation below that angle are not used.
+        // Default value in nrfxlib is 5 degrees.
+        elevation_threshold_angle: 5,
+        use_case: nrf_modem::GnssUsecase {
+            low_accuracy: false,
+            scheduled_downloads_disable: false,
+        },
+        nmea_mask: nrf_modem::NmeaMask {
+            gga: config.log_nmea,
+            gll: config.log_nmea,
+            gsa: config.log_nmea,
+            gsv: config.log_nmea,
+            rmc: config.log_nmea,
+        },
+        // Tcxo offers more precise 1PPS but uses more energy, we don't use 1PPS so Rtc makes more sense.
+        timing_source: nrf_modem::GnssTimingSource::Rtc,
+        power_mode: nrf_modem::GnssPowerSaveMode::Disabled,
+    }
+}
--- a/src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
+++ b/src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
@@ -1,0 +1,549 @@
+//! Driver for the GNSS modem of the [nRF91 SiP series] compatible with [`ariel_os_sensors::Sensor`].
+//!
+//! This driver has 3 operation modes selectable in [`config::Config`]:
+//! - `Contiunuous`: the GNSS will always stay active and send an update approximately every 3 seconds.
+//! - `Periodic`: acquires a fix every x seconds and sends an update.
+//! - `SingleShot`: acquires a fix only when requested by [`Sensor::trigger_measurement()`] with a set timeout in seconds.
+//!
+//! In `Continuous` and `Periodic` mode the driver will try to obtain a fix as soon as it's initialized.
+//! Using [`Sensor::trigger_measurement()`] then [`Sensor::wait_for_reading()`] will wait for the next update.
+//!
+//! If a fix cannot be obtained in the allowed amount of time, [`Sensor::wait_for_reading()`] will still return a reading, but some of the [`Sample`]s' values will be [`SampleError::TemporarilyUnavailable`].
+//!
+//! To access the time returned by the GNSS fix, you need to use the [`ariel_os_sensors_gnss_time_ext::GnssTimeExt`] trait.
+//!
+//! [nRF91 SiP series]: https://docs.nordicsemi.com/category/nrf-91-series
+//! [`SampleError::TemporarilyUnavailable`]: ariel_os_sensors::sensor::SampleError::TemporarilyUnavailable
+
+#![no_std]
+
+pub mod config;
+
+use core::f64::consts::PI;
+
+use embassy_futures::select::{Either, select};
+use embassy_sync::{
+    blocking_mutex::raw::CriticalSectionRawMutex, channel::Channel, once_lock::OnceLock,
+};
+use futures_util::StreamExt as _;
+use nrf_modem::{Gnss, GnssData, GnssStream};
+use time::{Date, Month, Time, UtcDateTime};
+
+use ariel_os_debug_log::{Debug2Format, debug, error, warn};
+use ariel_os_sensors::{
+    Category, Label, MeasurementUnit, Sensor,
+    sensor::{
+        Mode, ReadingChannel, ReadingChannels, ReadingError, ReadingResult, ReadingWaiter, Sample,
+        SampleMetadata, Samples, State,
+    },
+    signal::Signal,
+};
+use ariel_os_sensors_utils::AtomicState;
+
+use crate::config::{Config, GnssOperationMode, convert_gnss_config};
+
+// From WGS 84, Mean Radius of the Three Semi-axes in meters.
+// Source: table 3.5 in https://nsgreg.nga.mil/doc/view?i=4085
+const EARTH_RADIUS: f64 = 6_371_008.771_4;
+// The fraction of degrees representing a meter for the latitude (and the longitude at the equator).
+// Computed at build time to improve performance.
+const DEGREES_PER_METER_BASE: f64 = 360.0 / (EARTH_RADIUS * 2.0 * PI);
+
+#[derive(Debug)]
+enum Command {
+    Start,
+    Trigger,
+    Stop,
+}
+
+// Clamp to allowed u8 values and convert it to u8
+#[expect(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+fn clamp_to_u8(value: f32) -> u8 {
+    value.clamp(u8::MIN.into(), u8::MAX.into()) as u8
+}
+
+pub struct Nrf91Gnss {
+    config: OnceLock<config::Config>,
+    label: Option<&'static str>,
+    state: AtomicState,
+    command_channel: Channel<CriticalSectionRawMutex, Command, 1>,
+    result_signal: Signal<ReadingResult<Samples>>,
+}
+
+impl Nrf91Gnss {
+    /// Creates a new uninitialized GNSS driver with the corresponding label.
+    #[must_use]
+    pub const fn new(label: Option<&'static str>) -> Self {
+        Self {
+            config: OnceLock::new(),
+            label,
+            state: AtomicState::new(State::Uninitialized),
+            command_channel: Channel::new(),
+            result_signal: Signal::new(),
+        }
+    }
+
+    /// Initializes the driver with a configuration. Needs to be run before triggering measurements.
+    #[expect(
+        clippy::unused_async,
+        reason = "uniformity with other drivers so it can be codegened"
+    )]
+    pub async fn init(&self, config: config::Config) {
+        self.state.set(State::Enabled);
+
+        let _ = self.config.init(config);
+
+        debug!("nRF91 GNSS driver initialized");
+    }
+
+    /// At this point the sensor assume the modem is already initialized with the GNSS feature enabled.
+    /// In single shot mode, taking a measurement will return until a fix is obtained or the timeout is reached.
+    /// In continuous or periodic mode, taking a measurement will return the current status of the GNSS module, even if a fix has not been obtained yet.
+    ///
+    /// # Panics
+    ///
+    /// When the modem library refuses the config or fails to initialize the GNSS system.
+    pub async fn run(&'static self) {
+        let configuration = self.config.get().await;
+        loop {
+            // Wait until the state is Enabled.
+            if self.state.get() != State::Enabled
+                && !matches!(self.command_channel.receive().await, Command::Start)
+            {
+                continue;
+            }
+
+            let gnss_stream = match configuration.operation_mode {
+                GnssOperationMode::Continuous => Gnss::new()
+                    .await
+                    .unwrap()
+                    .start_continuous_fix(convert_gnss_config(configuration))
+                    .expect("Continuous fix initialization"),
+                GnssOperationMode::Periodic(period) => Gnss::new()
+                    .await
+                    .unwrap()
+                    .start_periodic_fix(convert_gnss_config(configuration), period)
+                    .expect("Periodic fix initialization"),
+
+                GnssOperationMode::SingleShot(timeout) => {
+                    match self.command_channel.receive().await {
+                        Command::Trigger => Gnss::new()
+                            .await
+                            .unwrap()
+                            .start_single_fix(convert_gnss_config(configuration), timeout)
+                            .expect("Single shot fix initialization"),
+
+                        Command::Stop => {
+                            warn!("Trying to stop the GNSS module when it is already stopped");
+                            continue;
+                        }
+                        Command::Start => {
+                            debug!("Ignoring Start command in SingleShot mode");
+                            continue;
+                        }
+                    }
+                }
+            };
+
+            self.handle_gnss_stream(gnss_stream, configuration).await;
+        }
+    }
+
+    async fn handle_gnss_stream(
+        &'static self,
+        mut gnss_stream: GnssStream,
+        configuration: &Config,
+    ) {
+        let mut latest_data = None;
+        let mut should_send_update = false;
+
+        loop {
+            if !matches!(self.state.get(), State::Enabled | State::Measuring) {
+                error!("Invalid state found");
+                break;
+            }
+
+            match select(self.command_channel.receive(), gnss_stream.next()).await {
+                Either::First(Command::Start) => {
+                    warn!("GNSS sensor already started");
+                }
+                Either::First(Command::Stop) => {
+                    break;
+                }
+                Either::Second(None) => {
+                    // In single shot mode, the stream ending means it has found a fix.
+                    if matches!(
+                        configuration.operation_mode,
+                        GnssOperationMode::SingleShot(_)
+                    ) {
+                        self.result_signal.clear();
+                        if let Some(data) = latest_data {
+                            let samples = self.convert_to_samples(&data);
+                            self.result_signal.signal(Ok(samples));
+                        } else {
+                            self.result_signal.signal(Err(ReadingError::SensorAccess));
+                        }
+                    }
+                    break;
+                }
+                Either::First(Command::Trigger) => {
+                    // Ignore, already running
+                    if should_send_update
+                        || matches!(
+                            configuration.operation_mode,
+                            GnssOperationMode::SingleShot(_)
+                        )
+                    {
+                        warn!("Received Trigger command while already processing one");
+                    } else {
+                        should_send_update = true;
+                    }
+                }
+                Either::Second(Some(Ok(message))) => match message {
+                    GnssData::PositionVelocityTime(pos) => {
+                        if should_send_update {
+                            let samples = self.convert_to_samples(&pos);
+                            self.result_signal.signal(Ok(samples));
+                            should_send_update = false;
+                        }
+
+                        // Only matters if we're in SingleShot mode.
+                        latest_data = Some(pos);
+                    }
+                    GnssData::Nmea(nmea_message) => {
+                        debug!("NMEA: {}", nmea_message.as_str());
+                    }
+                    GnssData::Agps(_) => {
+                        // Ignore AGPS data
+                    }
+                },
+                Either::Second(Some(Err(e))) => {
+                    warn!("GNSS error: {}", e);
+                }
+            }
+        }
+        let _ = gnss_stream.deactivate().await;
+    }
+
+    /// Convert time from `nrf_modem` to parts that can be put in samples.
+    ///
+    /// # Panics
+    ///
+    /// When the date is too far in the future / past.
+    fn convert_to_time_parts(
+        data: &nrf_modem::nrfxlib_sys::nrf_modem_gnss_pvt_data_frame,
+    ) -> Option<(i32, i32)> {
+        let parsed_date = Date::from_calendar_date(
+            data.datetime.year.into(),
+            Month::try_from(data.datetime.month).ok()?,
+            data.datetime.day,
+        )
+        .ok()?;
+
+        let time = Time::from_hms_milli(
+            data.datetime.hour,
+            data.datetime.minute,
+            data.datetime.seconds,
+            data.datetime.ms,
+        )
+        .ok()?;
+
+        // Default year when no GNSS fix.
+        if data.datetime.year == 1980 {
+            return None;
+        }
+
+        let timestamp = UtcDateTime::new(parsed_date, time).unix_timestamp_nanos();
+
+        Some(ariel_os_sensors_gnss_time_ext::convert_datetime_to_parts(timestamp).unwrap())
+    }
+
+    #[expect(clippy::cast_possible_truncation)]
+    fn convert_to_samples(
+        &'static self,
+        data: &nrf_modem::nrfxlib_sys::nrf_modem_gnss_pvt_data_frame,
+    ) -> Samples {
+        let time_parts = Self::convert_to_time_parts(data);
+
+        let (time_seconds_part, time_nanos_part) = if let Some(time_parts) = time_parts {
+            (
+                Sample::new(time_parts.0, SampleMetadata::UnknownAccuracy),
+                Sample::new(time_parts.1, SampleMetadata::UnknownAccuracy),
+            )
+        } else {
+            (
+                Sample::new(0, SampleMetadata::ChannelTemporarilyUnavailable),
+                Sample::new(0, SampleMetadata::ChannelTemporarilyUnavailable),
+            )
+        };
+
+        let latitude_accuracy = (f64::from(data.accuracy) * DEGREES_PER_METER_BASE) as f32;
+
+        // For longitude, the distance represented by a degree changes depending on the latitude
+        //
+        // The perimeter of the circle formed by the latitude is `cos(latitude_radians) * EARTH_RADIUS * 2 * PI`
+        // Full formula here is `longitude_accuracy = accuracy * 360 / (cos(latitude_radians) * EARTH_RADIUS * 2 * PI)`. We have 360 / (EARTH_RADIUS * 2 * PI) already pre-computed.
+        let longitude_accuracy = (f64::from(data.accuracy) * DEGREES_PER_METER_BASE
+            / libm::cos(data.latitude.to_radians())) as f32;
+
+        let latitude_value = (data.latitude * 10_000_000f64) as i32;
+        let longitude_value = (data.longitude * 10_000_000f64) as i32;
+        let altitude_value = (data.altitude * 100f32) as i32;
+
+        let fix_valid = (u32::from(data.flags)
+            & nrf_modem::nrfxlib_sys::NRF_MODEM_GNSS_PVT_FLAG_FIX_VALID)
+            != 0;
+
+        let (latitude, longitude, altitude) = if fix_valid {
+            let latitude = Sample::new(
+                latitude_value,
+                SampleMetadata::SymmetricalError {
+                    // One meter is approximately 0.000009 degrees. Accuracy value usually between 1 and 50 meters.
+                    deviation: clamp_to_u8(latitude_accuracy * 100_000f32),
+                    bias: 0,
+                    scaling: -5,
+                },
+            );
+            let longitude = Sample::new(
+                longitude_value,
+                SampleMetadata::SymmetricalError {
+                    deviation: clamp_to_u8(longitude_accuracy * 100_000f32),
+                    bias: 0,
+                    scaling: -5,
+                },
+            );
+            let altitude = Sample::new(
+                altitude_value,
+                SampleMetadata::SymmetricalError {
+                    deviation: clamp_to_u8(data.altitude_accuracy * 10f32),
+                    bias: 0,
+                    scaling: -1,
+                },
+            );
+            (latitude, longitude, altitude)
+        } else {
+            (
+                Sample::new(
+                    latitude_value,
+                    SampleMetadata::ChannelTemporarilyUnavailable,
+                ),
+                Sample::new(
+                    longitude_value,
+                    SampleMetadata::ChannelTemporarilyUnavailable,
+                ),
+                Sample::new(
+                    altitude_value,
+                    SampleMetadata::ChannelTemporarilyUnavailable,
+                ),
+            )
+        };
+
+        let horizontal_speed_value = (data.speed * 1_000_000f32) as i32;
+        let vertical_speed_value = (data.vertical_speed * 1_000_000f32) as i32;
+        let heading_value = (data.heading * 1_000_000f32) as i32;
+
+        let velocity_valid = (u32::from(data.flags)
+            & nrf_modem::nrfxlib_sys::NRF_MODEM_GNSS_PVT_FLAG_VELOCITY_VALID)
+            != 0;
+
+        let (horizontal_speed, vertical_speed, heading) = if velocity_valid {
+            let horizontal_speed = Sample::new(
+                horizontal_speed_value,
+                SampleMetadata::SymmetricalError {
+                    deviation: clamp_to_u8(data.speed_accuracy * 10f32),
+                    bias: 0,
+                    scaling: -1,
+                },
+            );
+
+            let vertical_speed = Sample::new(
+                vertical_speed_value,
+                SampleMetadata::SymmetricalError {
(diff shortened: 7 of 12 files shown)


src/ariel-os-sensors/src/category.rs
ROMemories, line 1: We'll want to move the changes to `ariel-os-sensors` to a separate PR, but that's not really urgent.

src/ariel-os-nrf91-gnss/src/state_atomic.rs
ROMemories, line 1: I'm going to introduce a crate (likely `ariel-os-sensor-utils`) to share these between sensor driver implementations, without them being part of the `ariel-os-sensor` crate.

nponsard: Ping me when you have a PR open so I can depend on it !

ROMemories: That would be #1351 (haven't added the `Depends on` myself because these items are technically only implementation details of this driver, so we could simply swap them later).

ROMemories:
In #1156, there were `Option`s in `GnssData` in case not all data were available; how does that work with this design? I know the sensor API doesn't currently provide such facility. I mentioned before introducing a `Label` for that but that doesn't make sense because it needs to be dynamic, so I suppose we could introduce a special variant of `Accuracy` (e.g., `TemporaryUnavailable`) instead if needed.

nponsard:
> In #1156, there were `Option`s in `GnssData` in case not all data were available; how does that work with this design? I know the sensor API doesn't currently provide such facility. I mentioned before introducing a `Label` for that but that doesn't make sense because it needs to be dynamic, so I suppose we could introduce a special variant of `Accuracy` (e.g., `TemporaryUnavailable`) instead if needed.

The modem API returns a default value (almost everything is zeroes) when no fix has been acquired.
My goal in #1156 was to have driver implementers return None instead of a default value when no fix is available. In the implementation in this PR I return the values returned by the modem API without checking if it's the default values.

ROMemories:
> My goal in #1156 was to have driver implementers return None instead of a default value when no fix is available. In the implementation in this PR I return the values returned by the modem API without checking if it's the default values.

I do agree it's way better to have a proper way of signaling that than having sensor-specific sentinel values. Could you open a separate PR to add such `TemporaryUnavailable` variant to `Accuracy`?

On a separate but related note, I think it would also make sense to have two or three variations of `SymmetricalError` for 2δ and 3δ errors (but that's for another PR).

src/ariel-os-nrf91-gnss/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+edition.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os-debug = { path = "../ariel-os-debug" }
+ariel-os-embassy = { path = "../ariel-os-embassy" }
```
ROMemories: I don't think this is needed, is it? If not can you remove it, that should make building docs for this driver a bit easier:
```suggestion
```

src/ariel-os-nrf91-gnss/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+] }
+embassy-sync = { workspace = true }
+portable-atomic = { workspace = true }
+futures = { workspace = true, default-features = false }
+defmt = { workspace = true, optional = true }
+time = { workspace = true, features = ["macros"] }
+
+# nrf-modem = { workspace = true }
+[target.'cfg(context = "nrf9151")'.dependencies]
+nrf-modem = { workspace = true, features = ["nrf9151", "modem-log"] }
```
ROMemories: Usually what we do in this case is also have `nrf-modem` as a non-platform-depedent dependency, with the `modem-log` Cargo feature factored out, and only select the platform-dependent feature in these (the set of Cargo features will be unified).

src/ariel-os-sensors/src/label.rs
```diff
@@ -53,10 +57,34 @@ impl core::fmt::Display for Label {
...
             Self::Y => write!(f, "Y"),
             Self::Z => write!(f, "Z"),
         }
     }
 }
+
+impl Label {
+    /// Returns `true` if the value tied to this label is opaque.
+    #[must_use]
+    pub fn is_opaque(&self) -> bool {
```
ROMemories: Another possibility would be to have a dedicated `Opaque` `Label` (which could be used multiple times), and the `Samples` and `ReadingChannels` iterators could simply skip over them. The driver would still know which one is which in `Nrf91GnssExt` based on their index.

nponsard: Oh that sounds like a good Idea, but I'm encountering an issue where if I filter in `Samples::samples()` and `ReadingChannels::iter()` the return type cannot be `std::iter::ExactSizeIterator`

ROMemories: Right, because [`Filter`](https://doc.rust-lang.org/stable/core/iter/struct.Filter.html) (returned by `Iterator::filter()`) doesn't implement `ExactSizeIterator`. There's two possible ways forward I think: not using `.filter()` but instead defining a custom iterator types (similar to `Filter`) which does implement `ExactSizeIterator` (if this is possible and not too cumbersome, that's likely the best option), otherwise we'll drop the `ExactSizeIterator` guarantee.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -43,34 +45,61 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+    let reading_channels_iter = (1..=count).map(|i| {
+        let variant = variant_name(i);
+        quote! { InnerReadingChannels::#variant(ref channels) => channels.iter().copied() }
+    });
+    let reading_channels_iter_cloned = reading_channels_iter.clone();
+    let reading_channels_iter_cloned_cloned = reading_channels_iter.clone();
 
     let expanded = quote! {
+        #[derive(Copy, Clone, Debug)]
+        struct SizedIterator<I: Iterator + core::iter::FusedIterator> {
```
ROMemories: I don't think the underlying iterator has to be a `FusedIterator`, does it?

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -94,9 +149,20 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             }
 
             fn samples(&self) -> impl ExactSizeIterator<Item = Sample> + core::iter::FusedIterator {
-                match self.samples {
-                    #(#samples_iter),*
-                }
+                let size = self.sensor().reading_channels().iter().size_hint().0;
```
ROMemories: That iterator is an `ExactSizedIterator` I think, so you should be able to use `ExactSizedIterator::len()` on it:
```suggestion
                let size = self.sensor().reading_channels().iter().len();
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -126,9 +192,25 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+                    #(#reading_channels_iter_cloned),*
+                }
+                .filter(|channel| channel.label() != Label::Opaque);
+                let size = cloned.count();
+
+                SizedIterator::new(iter, size)
+            }
+
+            /// FIXME
+            pub fn iter_raw(&self) -> impl ExactSizeIterator<Item = ReadingChannel> + core::iter::FusedIterator + '_ {
```
ROMemories: This fortunately doesn't have to be public I think:
```suggestion
            pub(crate) fn iter_raw(&self) -> impl ExactSizeIterator<Item = ReadingChannel> + core::iter::FusedIterator + '_ {
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -126,9 +192,25 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             /// [`Iterator::zip()`] can be useful to zip the returned iterator with the one
             /// obtained with [`Reading::samples()`].
             pub fn iter(&self) -> impl ExactSizeIterator<Item = ReadingChannel> + core::iter::FusedIterator + '_ {
-                match self.channels {
+                let iter = match self.channels {
```
ROMemories: This could reuse `iter_raw()`

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -43,34 +45,61 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+        quote! { InnerSamples::#variant(ref samples) => samples.iter().copied() }
+    });
+    let samples_iter_filtered = samples_iter.clone();
+
+    let reading_channels_iter = (1..=count).map(|i| {
+        let variant = variant_name(i);
+        quote! { InnerReadingChannels::#variant(ref channels) => channels.iter().copied() }
+    });
+    let reading_channels_iter_cloned = reading_channels_iter.clone();
+    let reading_channels_iter_cloned_cloned = reading_channels_iter.clone();
```
ROMemories: I think you could `collect()` the initial iterator into a `Vec<_>` instead, which would be easier to read (not saying this would be more *performant*).

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
nponsard: The iterators we return in the functions need to implement `+ core::iter::FusedIterator`, I prefer to guarantee that the inner iterator also has this guarantee.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
nponsard: It does indeed build as `pub(crate)`

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
ROMemories: AFAIK it's more idiomatic to only have that bound on the `impl` block that impls `FusedIterator` on the type.

nponsard: Ahh I didn't think about that :+1:

src/ariel-os-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssValueError {
+    /// Error reading value from sample, see inner error.
+    Reading(ValueError),
+    /// Sensor does not provide the correct channels
+    InvalidSensor,
+}
+
+pub trait Nrf91GnssExt {
```
ROMemories: We'll need to rename this to make it more generic, the goal being that it should be possible to swap the GNSS device for another as transparently as possible:
```suggestion
pub trait GnssExt {
```

We'll later move this to a dedicated crate (but separate from `ariel-os-sensors`), but no need to do this for now.

src/ariel-os-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+        }
+    }
+    fn convert_to_samples(
+        &'static self,
+        data: &nrf_modem::nrfxlib_sys::nrf_modem_gnss_pvt_data_frame,
+    ) -> Samples {
+        let date = Date::from_calendar_date(
+            data.datetime.year.into(),
+            Month::try_from(data.datetime.month).unwrap(),
+            data.datetime.day,
```
nponsard: Found during testing that a value of 0 can be returned by the nrf API when the modem is busy with LTE-M. Currently this panics because of the `unwrap()`. I will have to handle this case.

nponsard:
I found out that in the PVT struct returned by the modem library there is a `flags` field where I can read if a fix is valid or not.

nponsard:
@ROMemories I feel like the `GnssExt` should not be in this specific driver since it would be re-used. Should this be moved in `ariel-os-sensors` and maybe a different PR ? Also would it make sense in the name that this trait is about exposing time ? Something like `GnssTimeExt`.

src/ariel-os-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        Samples::from_8(
+            self,
+            [
+                Sample::new(
+                    (data.latitude * 10_000_000f64) as i32,
+                    if fix_valid {
+                        SampleMetadata::SymmetricalError {
+                            // Accuracy in meters, max value is 25,5. Need to watch out for overflows.
+                            deviation: clamp_to_u8(data.accuracy * 10f32),
```
nponsard: This needs to be converted to degrees

src/ariel-os-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+        };
+
+        let latitude_accuracy = data.accuracy * DEGREES_PER_METER_BASE;
+
+        // For longitude, the distance represented by a degree changes depending on the latitude
+        //
+        // The perimeter of the circle formed by the latitude is `cos(latitude_radians) * EARTH_RADIUS * 2 * PI`
+        // Full formula here is `longitude_accuracy = accuracy * 360 / (cos(latitude_radians) * EARTH_RADIUS * 2 * PI)`. We have 360 / (EARTH_RADIUS * 2 * PI) already pre-computed.
+        let longitude_accuracy =
+            data.accuarcy * DEGREES_PER_METER_BASE / data.latitude.to_radians().cos();
```
nponsard: Could there be a division by 0 here ?

src/ariel-os-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,458 @@
...
+        // FIXME: return an error if the channel is full
+        let _ = self.command_channel.try_send(Command::Trigger);
+
+        Ok(())
+    }
+
+    fn wait_for_reading(&'static self) -> ariel_os_sensors::sensor::ReadingWaiter {
+        if self.state.get() != State::Enabled {
+            return ReadingWaiter::Err(ReadingError::NonEnabled);
+        }
```
nponsard: Todo: handle the `Measuring` state, error out if not in this state.

kaspar030:
holy shit, just a minute! :rocket:

nponsard:
Yeah CI seems to work well

src/sensors/ariel-os-sensor-nrf91-gnss/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[dependencies]
+ariel-os-debug-log = { workspace = true }
+ariel-os-sensors = { workspace = true, features = ["max-sample-min-count-8"] }
+ariel-os-sensors-gnss-time-ext = { workspace = true }
+ariel-os-sensors-utils = { workspace = true }
+defmt = { workspace = true, optional = true }
+embassy-futures = { workspace = true }
+embassy-sync = { workspace = true }
+futures = { workspace = true, default-features = false }
```
ROMemories: I think this is only used for `StreamExt`, right? Can you use [`futures-util`](https://docs.rs/futures-util/latest/futures_util/stream/trait.StreamExt.html) direcly if so?

examples/sensors-debug/src/sensors.rs
```diff
@@ -124,3 +127,27 @@ mod stts22h {
...
 pub use stts22h::STTS22H_I2C;
+
+#[cfg(context = "nrf91")]
+mod nrf91 {
+    use ariel_os_sensor_nrf91_gnss::{
+        Nrf91Gnss,
+        config::{Config, GnssOperationMode},
+    };
+
+    pub static NRF91_GNSS: Nrf91Gnss = const { Nrf91Gnss::new(Some("Gnss")) };
```
ROMemories: There *might* be multiple GNSS sensors/modules, but this the *onboard* one:
```suggestion
    pub static NRF91_GNSS: Nrf91Gnss = const { Nrf91Gnss::new(Some("onboard")) };
```

examples/sensors-debug/laze.yml
```diff
@@ -29,3 +30,10 @@ modules:
       global:
         FEATURES:
           - defmt
+  - name: nrf91-modem-sensor
```
ROMemories, line 34:
```suggestion

  - name: nrf91-modem-sensor
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
+/// Operation modes for the GNSS module.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssOperationMode {
+    /// Always keep the GNSS module active.
+    Continuous,
+    /// Update the GNSS fix periodically. Period is defined in seconds and should be in range 10...65535.
```
ROMemories: I suppose 65535 should be in range?
```suggestion
    /// Update the GNSS fix periodically.
    /// The period is defined in seconds and should be in the range 10..=65535.
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+    Periodic(u16),
+    /// Try to get a GNSS fix only when requested. Timeout is defined in seconds, 300 recommended.
+    SingleShot(u16),
+}
+
+impl core::fmt::Display for GnssOperationMode {
```
ROMemories: Do we really a `Display` impl for this? It's not an error type.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+/// Configuration for the GNSS sensor.
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config {
+    /// The GNSS operating mode to use.
+    pub operation_mode: GnssOperationMode,
+    /// If NMEA messages should be logged as debug (adds extra processing).
```
ROMemories:
```suggestion
    /// Whether NMEA messages should be logged as debug logs (adds extra processing).
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+impl core::fmt::Display for GnssOperationMode {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Debug::fmt(self, f)
+    }
+}
+
+/// Configuration for the GNSS sensor.
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Config {
```
ROMemories: This should be `non_exhaustive` so we can expand the type later, similarly to the other sensor drivers. Does it also *have* to be `Copy` and `Clone`? Config types generally don't need to be, and it might make adding other settings later more difficult.
```suggestion
#[derive(Debug, Default)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[non_exhaustive]
pub struct Config {
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Config::const_default()
+    }
+}
+
+impl core::fmt::Display for Config {
```
ROMemories: Do we really need a `Display` impl?

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    /// Creates a new [`Config`] with default options.
+    #[must_use]
+    pub const fn const_default() -> Self {
+        Config::new(GnssOperationMode::Continuous, false)
+    }
+}
+
```
ROMemories: I don't think we should have these: the intent is that this eventually gets codegened from declarative configuration. In any case the config types should be designed to be instantiated in the following way:

1. Obtain the default `Config` through `Config::default()` (`non_exhaustive` forbids instantiation by specifying every field).
1. Possibly tune the settings individually.

```suggestion
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+impl core::fmt::Display for Config {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Debug::fmt(self, f)
+    }
+}
+
+pub(crate) fn convert_gnss_config(config: &Config) -> nrf_modem::GnssConfig {
+    nrf_modem::GnssConfig {
+        elevation_threshold_angle: 5,
```
ROMemories, line 38: This could use a comment, I don't know where that number comes from.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+            scheduled_downloads_disable: false,
+        },
+        nmea_mask: nrf_modem::NmeaMask {
+            gga: config.log_nmea,
+            gll: config.log_nmea,
+            gsa: config.log_nmea,
+            gsv: config.log_nmea,
+            rmc: config.log_nmea,
+        },
+        timing_source: nrf_modem::GnssTimingSource::Tcxo,
```
ROMemories: Can the TCXO be omitted, or is there always one for the nRF91? It it can be omitted, this setting value would make the config board-dependent. In that case I suppose we can use `Rtc` instead, and make the timing source a config setting of the driver later.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+//!
+//! This driver has 3 operation modes selectable in [`config::Config`]:
+//! - Contiunuous: the GNSS will always stay active and send an update approximately every 3 seconds.
+//! - Periodic: acquires a fix every x seconds and sends an update.
+//! - SingleShot: acquires a fix only when requested by [`Sensor::trigger_measurement()`] with a set timeout in seconds.
+//!
+//! In `Continuous` and `Periodic` mode the driver will try to obtain a fix as soon as it's initialized.
+//! Using [`Sensor::trigger_measurement()`] then [`Sensor::wait_for_reading()`] will wait for the next update.
+//!
+//! If a fix cannot be obtained in the allowed amount of time, [`Sensor::wait_for_reading()`] will return a reading and some of the [`Sample`]s values will be [`SampleError::TemporarilyUnavailable`].
```
ROMemories:
```suggestion
//! If a fix cannot be obtained in the allowed amount of time, [`Sensor::wait_for_reading()`] will still return a reading, but some of the [`Sample`]s' values will be [`SampleError::TemporarilyUnavailable`].
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+    label: Option<&'static str>,
+    state: AtomicState,
+
+    command_channel: Channel<CriticalSectionRawMutex, Command, 1>,
+    result_signal: Signal<ReadingResult<Samples>>,
+    started: Watch<CriticalSectionRawMutex, bool, 2>,
+}
+
+impl Nrf91Gnss {
+    /// Create a new GNSS driver with the corresponding label.
```
ROMemories:
```suggestion
    /// Creates a new uninitialized GNSS driver with the corresponding label.
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+    fn categories(&self) -> &'static [ariel_os_sensors::Category] {
+        &[Category::Gnss]
+    }
+
+    fn label(&self) -> Option<&'static str> {
+        self.label
+    }
+
+    fn display_name(&self) -> Option<&'static str> {
+        Some("NRF91 GNSS")
```
ROMemories:
```suggestion
        Some("nRF91 GNSS")
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+                }
+            }
+            Mode::Disabled | Mode::Sleeping => {
+                if let Err(e) = self.command_channel.try_send(Command::Stop) {
+                    error!("Couldn't send the command: {:?}", Debug2Format(&e));
+                }
+            }
+        }
+
+        return Ok(old);
```
ROMemories: The crate should be added to Clippy in CI.

```suggestion
        Ok(old)
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+                if let Err(e) = self.command_channel.try_send(Command::Start) {
+                    error!("Couldn't send the command: {:?}", Debug2Format(&e));
+                }
+            }
+            Mode::Disabled | Mode::Sleeping => {
+                if let Err(e) = self.command_channel.try_send(Command::Stop) {
+                    error!("Couldn't send the command: {:?}", Debug2Format(&e));
+                }
+            }
+        }
```
ROMemories: This can be factored out, which apparently saves 56 bytes of flash on nordic-thingy-91-x-nrf9151.

ROMemories, line 521: I think this should be made more robust: the state atomic is properly updated anyway, so the run loop should I think check the state at the end of the loop, and act on it. This way, either "waking up" the loop succeeds, or sending the command over the channel fails, but the state change is still properly taken into account, in which case there is no need to log anything (a comment would be useful however).

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+    fn label(&self) -> Option<&'static str> {
+        self.label
+    }
+
+    fn display_name(&self) -> Option<&'static str> {
+        Some("NRF91 GNSS")
+    }
+
+    fn part_number(&self) -> Option<&'static str> {
+        None
```
ROMemories:
```suggestion
        Some("nRF91xx")
```

ROMemories (commented):
I'm still to go through the core logic of the driver.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+    Stop,
+}
+
+// Clamp to allowed u8 values and convert it to u8
+fn clamp_to_u8(value: f32) -> u8 {
+    value.clamp(u8::MIN.into(), u8::MAX.into()) as u8
+}
+
+pub struct Nrf91Gnss {
+    config: Mutex<CriticalSectionRawMutex, config::Config>,
```
ROMemories: It seems it might be feasible to refactor this to not use a mutex, but no need to do this now, it can refactored later.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+            result_signal: Signal::new(),
+
+            started: Watch::new(),
+        }
+    }
+
+    /// Initializes the driver with a configuration. Needs to be run before triggering measurements.
+    ///
+    /// `run()` needs to be spawned and awaited in a concurrent task, otherwise `init` will block indefinitely.
+    pub async fn init(&self, config: config::Config) {
```
ROMemories, line 91: Because this async function is expected to be used in a task that only runs once (on startup), it should do as little async work as possible, so its `Future` is as small as possible to not waste RAM. In particular, this should really just initialize the sensor *driver* and only reset the sensor *device* (if such resets exists). This should also simplify the relationship between `init()` and `run()`.

ROMemories (commented):
I'll go through `run()` and what it calls then.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
nponsard: Upstream documentation is unclear about that: https://github.com/nrfconnect/sdk-nrfxlib/blob/e4af79800ff41ea59fb0007a84e1e8dedda98ac9/nrf_modem/include/nrf_modem_gnss.h#L1044

But my guess would be yes.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: I have an issue here with my message passing, if the task awaiting `run()` doesn't get time to run between `init()` and the first `trigger_measurement()`, the line `while !receiver.changed().await {}` is allowing that.

I'll see what I can do to reduce the impact.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+            None
+        } else if let Ok(date) = date
+            && let Ok(time) = time
+        {
+            let datetime = UtcDateTime::new(date, time).unix_timestamp_nanos();
+
+            Some(ariel_os_sensors_gnss_time_ext::convert_datetime_to_parts(datetime).unwrap())
+        } else {
+            None
+        };
```
ROMemories: I think this could be split out into a separate function, and that would make it easy to simply return `None` in case of a parsing error.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+        // For longitude, the distance represented by a degree changes depending on the latitude
+        //
+        // The perimeter of the circle formed by the latitude is `cos(latitude_radians) * EARTH_RADIUS * 2 * PI`
+        // Full formula here is `longitude_accuracy = accuracy * 360 / (cos(latitude_radians) * EARTH_RADIUS * 2 * PI)`. We have 360 / (EARTH_RADIUS * 2 * PI) already pre-computed.
+        let longitude_accuracy = f64::from(data.accuracy) * DEGREES_PER_METER_BASE
+            / libm::cos(data.latitude.to_radians());
+
+        Samples::from_8(
+            self,
+            [
```
ROMemories: Could you refactor this by first building the eight samples individually, before returning the `Samples` from them? That would allow easily naming each them, because it's a bit hard for me to follow which is which at the moment.

ROMemories (commented):
Still need to go through the physics part.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: Alright I'll re-request on you when it's refactored.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
nponsard: > Using TCXO instead of RTC during GNSS sleep periods might be beneficial when used with 1PPS.

We're not using 1PPS, switching to `Rtc`.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
nponsard: Removed the `Display` impl.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
nponsard: Removed the `Display` impl, I don't remember why I did it.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: Refactored to use a Watch instead, also solves the synchronization with init().

src/sensors/ariel-os-sensor-nrf91-gnss/Cargo.toml
nponsard: switched to `futures-util`

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
ROMemories: The sensor crate needs to be added to the Clippy list:

https://github.com/ariel-os/ariel-os/blob/099070b27d9091884da9d8abb1223850a2cbcb4f/.github/workflows/main.yml#L291-L301

Cargo.toml
```diff
@@ -182,7 +184,9 @@ trouble-host = { version = "0.5.0" }
 #nrf-sdc = { version = "0.4.0" }
 nrf-sdc = { git = "https://github.com/alexmoon/nrf-sdc", rev = "e64ad081a6948b3439d1e811da36d39a5785dbda" }
 
+futures-util = { version = "0.3", default-features = false }
```
ROMemories: Otherwise this means that `0.3.0` is ok, but there's been a soundness fix in 0.3.31 (and the lockfile having 0.3.32 means we don't test for `0.3.0`):
```suggestion
futures-util = { version = "0.3.32", default-features = false }
```

examples/sensors-debug/Cargo.toml
```diff
@@ -13,6 +13,9 @@ ariel-os = { path = "../../src/ariel-os", features = [
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 ariel-os-sensor-lis2du12 = { workspace = true }
 ariel-os-sensor-lps22df = { workspace = true }
+ariel-os-sensor-nrf91-gnss = { workspace = true, features = [
+  "defmt",
```
ROMemories: Couldn't this be enabled only when the app's `defmt` feature is enabled?

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -0,0 +1,46 @@
+/// Operation modes for the GNSS module.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssOperationMode {
+    /// Always keep the GNSS module active.
+    #[default]
```
ROMemories: I don't think there should be a `Default` impl on this type actually (`Continuous` isn't that an obvious default choice for this setting in isolation), but there should be one on `Config`, which then needs to a manual impl (and in that case using `Continuous` for that is fine).

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
ROMemories: Yes, they definitely seem to be using the ellipsis for inclusive ranges in other parts of the docs.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
ROMemories: Does it even need to the `Clone`?

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
ROMemories: Even simpler then :slightly_smiling_face:

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+            config: Watch::new(),
+            label,
+            state: AtomicState::new(State::Uninitialized),
+            command_channel: Channel::new(),
+            result_signal: Signal::new(),
+        }
+    }
+
+    /// Initializes the driver with a configuration. Needs to be run before triggering measurements.
+    pub fn init(&self, config: config::Config) {
```
ROMemories: Even if there is no `.await` inside, I think this method still needs to be `async` for consistency with other sensor drivers, so we can codegen this later:
```suggestion
    pub async fn init(&self, config: config::Config) {
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// Clamp to allowed u8 values and convert it to u8
+#[expect(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+fn clamp_to_u8(value: f32) -> u8 {
+    value.clamp(u8::MIN.into(), u8::MAX.into()) as u8
+}
+
+pub struct Nrf91Gnss {
+    config: Watch<CriticalSectionRawMutex, config::Config, 1>,
```
ROMemories: I think this can be done without relying on a `Watch`, which is relatively costly RAM-wise, but this refactor is better done in a follow-up.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+    ///
+    /// # Panics
+    ///
+    /// When the date is too far in the future / past.
+    fn convert_to_time_parts(
+        data: &nrf_modem::nrfxlib_sys::nrf_modem_gnss_pvt_data_frame,
+    ) -> Option<(i32, i32)> {
+        let Ok(parsed_date) = Date::from_calendar_date(
+            data.datetime.year.into(),
+            Month::try_from(data.datetime.month).unwrap_or(Month::January),
```
ROMemories: That's one of the benefit of having split that into a function; we can just bail out if we happen to get a 13th month:
```suggestion
            Month::try_from(data.datetime.month).ok()?,
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    /// Convert time from `nrf_modem` to parts that can be put in samples.
+    ///
+    /// # Panics
+    ///
+    /// When the date is too far in the future / past.
+    fn convert_to_time_parts(
+        data: &nrf_modem::nrfxlib_sys::nrf_modem_gnss_pvt_data_frame,
+    ) -> Option<(i32, i32)> {
+        let Ok(parsed_date) = Date::from_calendar_date(
```
ROMemories: This can also simply use `.ok()?` now:
```suggestion
        let parsed_date = Date::from_calendar_date(
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+            data.datetime.year.into(),
+            Month::try_from(data.datetime.month).unwrap_or(Month::January),
+            data.datetime.day,
+        ) else {
+            return None;
+        };
+
+        let Ok(time) = Time::from_hms_milli(
```
ROMemories: Same:

```suggestion
        let time = Time::from_hms_milli(
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+        Mode, ReadingChannel, ReadingChannels, ReadingError, ReadingResult, ReadingWaiter, Sample,
+        SampleMetadata, Samples, State,
+    },
+    signal::Signal,
+};
+use ariel_os_sensors_utils::AtomicState;
+
+use crate::config::{Config, GnssOperationMode, convert_gnss_config};
+
+// From WGS 84, Mean Radius of the Three Semi-axes in meters.
```
ROMemories, line 45: Can you add a link or a more precise reference of where that value comes from? I see it is pretty much [the mean radius from that figure](https://en.wikipedia.org/wiki/File:WGS84_mean_Earth_radius.svg), but I can't seem to find that exact value.

ROMemories, line 45: It might also be useful to link to the relevant [Wikipedia pages](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS_84) (or other references) somewhere.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
nponsard: `Clone` is needed here to store it in the `Watch`. 

The use of a `Watch` was to allow for changes later. But I can use a `OnceLock` if we don't need the configuration to change after init.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+    /// Initializes the driver with a configuration. Needs to be run before triggering measurements.
+    pub fn init(&self, config: config::Config) {
+        self.state.set(State::Enabled);
+
+        self.config.sender().send(config);
+
+        debug!("nRF91 GNSS driver initialized");
+    }
+
+    /// At this point the sensor assume the modem is already initialized with the GNSS feature enabled.
```
ROMemories, line 99: I don't think this is the case anymore, is it?

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+    /// In single shot mode, taking a measurement will return until a fix is obtained or the timeout is reached.
+    /// In continuous or periodic mode, taking a measurement will return the current status of the GNSS module, even if a fix has not been obtained yet.
+    ///
+    /// # Panics
+    ///
+    /// Panics if configuration cannot be retrieved from the internal [`Watch`].
```
ROMemories: Which would be a bug, there's no need to document these (an `expect` with a `reason` should be used instead).

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+    pub async fn run(&'static self) {
+        let mut configuration_receiver = self.config.receiver().unwrap();
+        loop {
+            let configuration = configuration_receiver.get().await;
+            // Wait until the state is Enabled.
+            if self.state.get() != State::Enabled
+                && !matches!(self.command_channel.receive().await, Command::Start)
```
ROMemories, line 111: You can impl `PartialEq` (and `Eq`) on `Command`, which would make this a bit simpler, but this is better done a follow-up.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+                ReadingWaiter::new_err(ReadingError::NonEnabled)
+            }
+        }
+    }
+
+    fn reading_channels(&self) -> ariel_os_sensors::sensor::ReadingChannels {
+        ReadingChannels::from([
+            // Putting these first so `GnssExt` doesn't spend more time searching for them.
+            ReadingChannel::new(
+                // Seconds since Ariel epoch (2024-01-01)
```
ROMemories:
```suggestion
                // Seconds since Ariel epoch (2025-01-01)
```

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+            }
+            State::Enabled => ReadingWaiter::new_err(ReadingError::NotMeasuring),
+            State::Disabled | State::Uninitialized | State::Sleeping => {
+                ReadingWaiter::new_err(ReadingError::NonEnabled)
+            }
+        }
+    }
+
+    fn reading_channels(&self) -> ariel_os_sensors::sensor::ReadingChannels {
+        ReadingChannels::from([
```
ROMemories, line 446: I went through these, the necessary explanations are there; the phrasing could be improved a bit, but this we can be done in a follow-up.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+                        }
+                    }
+                }
+            };
+
+            self.handle_gnss_stream(gnss_stream, configuration).await;
+        }
+    }
+
+    async fn handle_gnss_stream(
```
ROMemories, line 152: I went through this loop, it might be possible to refactor it a bit later, but I didn't find any obvious flaws.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+                    deviation: clamp_to_u8(data.heading_accuracy),
+                    bias: 0,
+                    scaling: 0,
+                },
+            );
+            (horizontal_speed, vertical_speed, heading)
+        } else {
+            (
+                Sample::new(
+                    horizontal_speed_value,
```
ROMemories, line 380: By using `0` instead of a known-invalid value, the computations right before the `if` could be moved inside the other branch, but this is better done in a follow-up.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+            return None;
+        }
+
+        let timestamp = UtcDateTime::new(parsed_date, time).unix_timestamp_nanos();
+
+        Some(ariel_os_sensors_gnss_time_ext::convert_datetime_to_parts(timestamp).unwrap())
+    }
+
+    #[expect(clippy::cast_possible_truncation)]
+    fn convert_to_samples(
```
ROMemories, line 262: I went through this and didn't find any obvious issues.

ROMemories (commented):
I've been through the rest of the implementation (including the math), I didn't find any obvious issues. After rebasing and addressing the couple of code comments/doc-related comments, I think we'll land this and do a couple of refactoring follow-ups (no need to tackle any of my comments referring to follow-ups in *this* PR).

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: Execution of the `run()` command still needs the correct initialization of the `nrf-modem` crate by `ariel-os-embassy`.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: would https://nsgreg.nga.mil/doc/view?i=4085 work ? that's the source of the constant too.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: This was intentional, we can discuss the utility in a follow-up.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
ROMemories: That's the actual standard? Sure!

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: This message is actually out of date, there is another panic source that comes from `nrf-modem` that needs to be documented and shouldn't be ignored.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
nponsard: From what I see this seems to be the standard yeah.

ROMemories (approved):
LGTM, we'll do a couple of refactors and clarify some source comments in follow-ups.

src/sensors/ariel-os-sensor-nrf91-gnss/src/lib.rs
```diff
@@ -0,0 +1,549 @@
...
+        }
+    }
+
+    async fn handle_gnss_stream(
+        &'static self,
+        mut gnss_stream: GnssStream,
+        configuration: &Config,
+    ) {
+        let mut latest_data = None;
+        let mut should_send_update = false;
```
ROMemories, line 158: Could you add a comment on this to clarify what it represents, especially across two consecutive iterations of the loop?

PR #1331, opened 2025-09-12
docs(book): document our current dependency vetting process

ROMemories:
# Description

As discussed in #1296, this documents our expected process for vetting Rust dependencies through `cargo-vet`.

## Issues/PRs references

Depends on #1296.

## Open Questions

I expect this to require quite some discussion and refinements before merging.

## Change checklist

### DIFF
--- a/book/src/development-processes.md
+++ b/book/src/development-processes.md
@@ -52,6 +52,22 @@ This comment contains links that allow accessing the deployed documentation.
 
 > When making a PR with changes that affect the documentation, the preview should be checked to make sure the docs are rendered as expected.
 
+## Dependency Vetting
+
+We are currently experimenting with the process of vetting our Rust dependencies through [`cargo-vet`][cargo-vet-repo], with the aim of defending against supply-chain attacks.
+Please see [`cargo-vet`'s book][cargo-vet-book] for what it can do and how it works.
+
+Our `cargo-vet` configuration [imports the audits][cargo-vet-importing-audits] of the main organizations that make their audits available (and are in the official `cargo-vet` registry) and trusts either well-known publishers or publishers that are trusted by these organizations.
+
+We invite all contributors to document audits they performed on crates that their PRs introduce.
+As merging them is a statement by the project, the PR adding the audit needs to be reviewed as carefully as if performing the audit.
+The team members are prepared to vet dependencies if external contributors introduce reasonable dependencies for functionality they add.
+Currently, `cargo-vet` is run in CI for each PR; if this proves to block the PR process too often, we may later revisit this and only vet dependencies for releases.
+
+Vetting a dependency involves either [performing an audit][cargo-vet-performing-audits], [trusting (all of) its publishers][cargo-vet-trusting-publishers], or adding an exemption (in which case it is not actually “vetted”).
+As we are still figuring out our vetting workflow, we are currently fine with adding new exemptions for both new crates and dependency updates.
+We may also introduce a [custom criteria][cargo-vet-custom-criteria] to reflect the actual meaning of our auditing process.
+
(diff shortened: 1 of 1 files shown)


book/src/development-processes.md
```diff
@@ -52,6 +52,20 @@ This comment contains links that allow accessing the deployed documentation.
...
 > When making a PR with changes that affect the documentation, the preview should be checked to make sure the docs are rendered as expected.
 
+## Dependency Vetting
+
+We are currently experimenting with the process of vetting our Rust dependencies through [`cargo-vet`][cargo-vet-repo], with the aim of defending against supply-chain attacks.
+Please see [`cargo-vet`'s book][cargo-vet-book] for what it can do and how it works.
+
+Our `cargo-vet` configuration [imports the audits][cargo-vet-importing-audits] of the main organizations that make their audits available (and are in the official `cargo-vet` registry) and trusts either well-known publishers or publishers that are trusted by these organizations.
+
+Only team members are allowed to vet dependencies, which is required when either adding new dependencies or updating existing ones.
```
ROMemories: "Team members" is currently meant as [defined in the governance document](https://github.com/ariel-os/ariel-os/blob/main/GOVERNANCE.md#team-members); I'm not exactly sure there is a one-to-one mapping though.

kaspar030: I think the wording is fine.

kaspar030:
This looks fine from my side. Passing review token to @anlavandier.

ROMemories:
@chrysn has mentioned introducing a custom criteria multiple times before so passing the review token to him as well.

book/src/development-processes.md
chrysn: I wouldn't even go this far: If a contributor has vetted a repo, that's fine -- it'll just be up to the reviewer to confirm that in more detail than is usually warranted for a single-line change. Maybe:

```suggestion
We invite all contributors to document audits they performed on crates which their PRs introduce.
As merging them is a statement by the project, the PR adding the audit needs to be performed as carefully as the audit itself.
The team members are prepared to audit dependencies (or add temporary exceptions for them) if external contributors introduce reasonable dependencies for functionality they add.
```

(Also, the actions are "audit", "trust" or "add exceptions", so "allowed to vet" is not a thing a person does).

book/src/development-processes.md
```diff
@@ -52,6 +52,20 @@ This comment contains links that allow accessing the deployed documentation.
...
+Currently, `cargo-vet` is run in CI for each PR; if this proves too time-consuming, we may later revisit this and only vet dependencies for releases.
```
chrysn:
```suggestion
Currently, `cargo-vet` is run in CI for each PR; if this proves to block the PR process too often, we may later revisit this and only vet dependencies for releases.
```

This was previously ambiguous: time could be CI time or person time.

book/src/development-processes.md
ROMemories: I rephrased this a bit, and used "are prepared to vet dependencies" instead, as trusting the publishers would also be valid, and "vetting a dependency" is defined in the following paragraph.

I'm not sure what you mean by the following however:

> "allowed to vet" is not a thing a person does

chrysn: Thanks, LGTM.

The "vet" was mainly me misreading the paragraph after it, which (as I now probably more correctly read it) defines the action of "to vet" as doing either of the three "audit", "trust" or "add exemption". The text you rephrased things into captured this correctly, I think: External contributors either do nothing or add an audit; team members then either accept that audit or perform either of the three vetting steps.

ROMemories:
Merging as we agreed on the process and as #1296 is now in.

PR #1332, opened 2025-09-12
feat(sensors): add extra Cargo features for adjustable sensor data enums

ROMemories:
# Description

Adds a couple of variants to the sensor data enum that were left out of #474 because it wasn't obvious they were useful. #1330 shows that `max-sample-min-count-8` is, so this also adds `max-sample-min-count-5` to complete the sequence.

Using `ci-build:skip` is there is users in the tree yet.

## Change checklist


ROMemories:
@nponsard This should be useful for #1330 if I read that right. It it is, please make #1330 depend on this PR.

PR #1339, opened 2025-09-14
feat(random): Add support for getrandom

SimonIT:
# Description

This adds support for using Ariel OS's random as a backend for [getrandom](https://github.com/rust-random/getrandom) as described [here](https://github.com/rust-random/getrandom?tab=readme-ov-file#custom-backend).

Other crates using `rand` can then automatically use Ariel OS's random backend. The application just needs to be set up to use the custom backend, as described [here](https://github.com/rust-random/getrandom?tab=readme-ov-file#opt-in-backends).

## Issues/PRs references

rust-random/getrandom#716
Depends on #1414.

## Open Questions

I'm not sure about the extra feature `getrandom`. What's your opinion?

I just had the idea of making the type of random generator which is used feature dependent if csprng is enabled or not. What do you think?

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1307,6 +1307,8 @@ modules:
       global:
         FEATURES:
           - ariel-os/hwrng
+        RUSTFLAGS:
+          - --cfg getrandom_backend=\"custom\"
 
   - name: has_hwrng
     selects:
@@ -1435,17 +1437,6 @@ modules:
       global:
         FEATURES:
           - ariel-os/random
-
-  - name: random-crypto
-    help: A system-wide cryptographically secure RNG is available.
-    selects:
-      - random
-    env:
-      global:
-        FEATURES:
-          - ariel-os/csprng
-        RUSTFLAGS:
-          - --cfg getrandom_backend=\"custom\"
 
   - name: sw/benchmark
     help: provided if a target supports `benchmark()`
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1436,6 +1436,17 @@ modules:
         FEATURES:
           - ariel-os/random
 
+  - name: random-crypto
+    help: A system-wide cryptographically secure RNG is available.
+    selects:
+      - random
+    env:
+      global:
+        FEATURES:
+          - ariel-os/csprng
+        RUSTFLAGS:
+          - --cfg getrandom_backend=\"custom\"
+
   - name: sw/benchmark
     help: provided if a target supports `benchmark()`
     selects:
--- a/src/ariel-os-random/Cargo.toml
+++ b/src/ariel-os-random/Cargo.toml
@@ -17,7 +17,9 @@ embassy-sync.workspace = true
 rand_pcg = "0.3.1"
 rand_chacha = { version = "0.3.1", default-features = false, optional = true }
 
+getrandom = { version = "0.3.3", optional = true }
+
 [features]
 ## If set, the one global RNG is also a cryptographically secure pseudo
 ## random number generator (CSPRNG), and thus, a `CryptoRng` can be produced.
-csprng = ["dep:rand_chacha"]
+csprng = ["dep:rand_chacha", "dep:getrandom"]
--- a/src/ariel-os-random/src/lib.rs
+++ b/src/ariel-os-random/src/lib.rs
@@ -27,6 +27,8 @@
 use core::{cell::RefCell, marker::PhantomData};
 
 use embassy_sync::once_lock::OnceLock;
+#[cfg(feature = "csprng")]
+use getrandom::Error;
 use rand_core::{RngCore, SeedableRng as _};
 
 /// A global RNG.
@@ -245,4 +247,32 @@ pub fn crypto_rng_send() -> CryptoRngSend {
             rand_chacha::ChaCha20Rng::from_rng(i).expect("Global RNG is infallible")
         }),
     }
+}
+
+/// Implements RNG access for the `getrandom` crate
+///
+/// # Safety
+///
+/// See the `getrandom` documentation on custom backends.
+///
+/// # Errors
+///
+/// See the `getrandom` documentation on custom backends.
+///
+/// At the time of writing, all Ariel OS RNGs are infallible.
+///
+/// # Panics
+///
+/// The function panics if error conversion fails.
+#[cfg(feature = "csprng")]
+#[unsafe(no_mangle)]
+unsafe extern "Rust" fn __getrandom_v03_custom(dest: *mut u8, len: usize) -> Result<(), Error> {
+    let buf = unsafe {
+        core::ptr::write_bytes(dest, 0, len);
+        core::slice::from_raw_parts_mut(dest, len)
+    };
+    #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+    crypto_rng()
+        .try_fill_bytes(buf)
+        .map_err(|e| Error::new_custom(e.raw_os_error().unwrap() as u16))
 }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -157,7 +157,7 @@ jobs:
             --
             --deny warnings
 
-      - run: echo 'RUSTFLAGS=--cfg context="esp32c6"' >> $GITHUB_ENV
+      - run: echo 'RUSTFLAGS=--cfg context="esp32c6" --cfg getrandom_backend="custom"' >> $GITHUB_ENV
       - name: clippy for ESP32
         uses: clechasseur/rs-clippy-check@69da786488ddcceb16285b0219841750089369ba # v5
         with:
@@ -176,7 +176,7 @@ jobs:
             --
             --deny warnings
 
-      - run: echo 'RUSTFLAGS=--cfg context="cortex-m" --cfg context="rp" --cfg context="rp2040"' >> $GITHUB_ENV
+      - run: echo 'RUSTFLAGS=--cfg context="cortex-m" --cfg context="rp" --cfg context="rp2040" --cfg getrandom_backend="custom"' >> $GITHUB_ENV
       - name: clippy for RP
         uses: clechasseur/rs-clippy-check@69da786488ddcceb16285b0219841750089369ba # v5
         with:
@@ -195,7 +195,7 @@ jobs:
             --
             --deny warnings
 
-      - run: echo 'RUSTFLAGS=--cfg context="nrf52840" --cfg context="nrf52"' >> $GITHUB_ENV
+      - run: echo 'RUSTFLAGS=--cfg context="nrf52840" --cfg context="nrf52" --cfg getrandom_backend="custom"' >> $GITHUB_ENV
       - name: clippy for nRF
         uses: clechasseur/rs-clippy-check@69da786488ddcceb16285b0219841750089369ba # v5
         with:
@@ -215,7 +215,7 @@ jobs:
             --
             --deny warnings
 
-      - run: echo 'RUSTFLAGS=--cfg context="stm32wb55rg"' >> $GITHUB_ENV
+      - run: echo 'RUSTFLAGS=--cfg context="stm32wb55rg" --cfg getrandom_backend="custom"' >> $GITHUB_ENV
       - name: clippy for STM32
         uses: clechasseur/rs-clippy-check@69da786488ddcceb16285b0219841750089369ba # v5
         with:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -454,6 +454,7 @@ name = "ariel-os-random"
 version = "0.2.0"
 dependencies = [
  "embassy-sync 0.6.2",
+ "getrandom 0.3.3",
  "rand_chacha",
  "rand_core",
  "rand_pcg",


src/ariel-os-random/src/getrandom.rs
```diff
@@ -0,0 +1,17 @@
...
+//! Enable it by using the custom backend from getrandom. See [Opt-in Backends](https://github.com/rust-random/getrandom?tab=readme-ov-file#opt-in-backends)
+
+use crate::crypto_rng;
+use getrandom::Error;
+use rand_core::RngCore;
+
+#[unsafe(no_mangle)]
+unsafe extern "Rust" fn __getrandom_v03_custom(dest: *mut u8, len: usize) -> Result<(), Error> {
+    let buf = unsafe {
+        core::ptr::write_bytes(dest, 0, len);
```
chrysn: I doubt that this is really necessary (for every u8 value is inhabited, and if there are poison values in there, our try_fill_bytes will not read them), but as it's the upstream guidance, we better just follow it. (Also I'm not 100% sure that my understanding of poison values and &mut is right there).

chrysn (commented):
Looks good code-wise.

On the feature, I think we can just export that symbol without needing a feature; this is an operating system, and getrandom has getting it from the operating system in its description. It'll be a dead symbol anyway if unused.

getrandom has strong randomness in its description/requirements ("assumed that the system always provides high-quality, cryptographically secure random data"), so there's no need (or option) to switch over to a non-CS source.

src/ariel-os-random/laze.yml
```diff
@@ -0,0 +1,6 @@
+apps:
+  - name: crates/ariel-os-random
+    env:
+      global:
+        RUSTFLAGS:
+          - '--cfg getrandom_backend="custom"'
```
ROMemories: Would it make sense to enable this in the manifest instead? Otherwise that wouldn't be usable without laze I think.

SimonIT: I'm not sure what manifest you mean... I did it like that because it's done the same way here: https://github.com/ariel-os/ariel-os/blob/main/src/ariel-os-rp/laze.yml

ROMemories: I meant the `Cargo.toml`; it's unfortunate that `getrandom` removed the `custom` Cargo feature in v0.3. I guess that's indeed the best way to do it then.

chrysn: I think it's fine that this is enabled through laze (any baked form would just set this on the baked command line); should be in laze-project.toml though (and then the getrandom feature of ariel-os-random would also be enabled through that laze module – if it needs to be optional at all, see https://github.com/ariel-os/ariel-os/pull/1339#discussion_r2348556122).

SimonIT: >  should be in laze-project.toml though

The laze-project.yml in the root of the ariel-os repo? Or that users have to specify it in their own projects?

> if it needs to be optional at all

Because `crypto_rng` is behind the csprng feature I added the getrandom to it too as we're using it. Or do I understand your comment wrong? So I add a module `getrandom` to the root laze-project.yml of the repo which sets the environment variable and enables the csprng feature?

chrysn: The one at the root of the Ariel OS project (given that that is the project where it's available from to all others).

As to module dependencies: Is there any good reason why this should be a module at all? I think it can just be present unconditionally, provided there is a CSPRNG available. (There's no runtime cost to it if the linker decides that nobody looks at the generated symbol and GC's it). Looking at the current modules however, I see that we don't have the crypto-safety of the RNG as a feature in Laze. Do you want to try adding this here too, or rather let one of the maintainers fix the infrastructure separately?

SimonIT: All right, I hope it's now how you wanted it

chrysn: Roughly yes; what I still don't understand (from first comment): Why would you make your getrandom implementation optional? (We don't have concrete criteria, but the main driving forces are optionality-induced complexity aka ifdef hell on one side, and runtime overhead on the other side -- and this does not seem to induce the latter)

SimonIT: Perhaps there has been a misunderstanding on my part. At first, I made it an additional feature because I thought that perhaps someone might not want the getrandom dependency. Then you said it didn't need to be optional, so I removed the additional feature and added RUSTFLAGS to laze.yml. I should then have put the flag in the root laze-project.yml file. Where exactly should I put the flag? In the csprng module of the root laze-project.yml file? Or is there another place? Sorry for the misunderstanding

laze-project.yml
```diff
@@ -1418,6 +1418,21 @@ modules:
         FEATURES:
           - ariel-os/random
 
+  - name: csprng
```
chrysn:
```suggestion
  - name: random-crypto
    help: A system-wide cryptographically secure RNG is available.
```

Suggested name: random-crypto (that's usually backed by a CSPRNG, but that's not what we promise, which )

src/ariel-os-random/laze.yml
chrysn: I see no issue with having getrandom as a dependency. The flag can go into the feature that enables it, which is the csprng (or random-crypto, see suggestion) right now. See code suggestion.

laze-project.yml
```diff
@@ -1418,6 +1418,21 @@ modules:
...
+      global:
+        FEATURES:
+          - ariel-os/csprng
+
+  - name: getrandom
+    help: Enables the use of the custom backend for the getrandom crate.
+    selects:
+      - csprng
+    env:
+      global:
```
chrysn:
```suggestion
          - ariel-os/csprng
```

I don't think this needs to be optional.

src/ariel-os-random/Cargo.toml
```diff
@@ -17,7 +17,10 @@ embassy-sync.workspace = true
 rand_pcg = "0.3.1"
 rand_chacha = { version = "0.3.1", default-features = false, optional = true }
 
+getrandom = { version = "0.3.3", optional = true }
```
chrysn, line 20:
```suggestion
getrandom = "0.3.3"
```

(and leaving the features below as-is – for some reason the GitHub GUI won't let me suggest that directly)

src/ariel-os-random/laze.yml
SimonIT: Alright, thank you!

laze-project.yml
```diff
@@ -1418,6 +1418,15 @@ modules:
...
           - ariel-os/random
 
+  - name: random-crypto
+    help: A system-wide cryptographically secure RNG is available.
+    env:
+      global:
+        FEATURES:
+          - ariel-os/csprng
+        RUSTFLAGS:
+          - '--cfg getrandom_backend="custom"'
```
chrysn:
```suggestion
          - '--cfg getrandom_backend="\"custom\""'
```

With this change, it works for me and passes a manual test on native and particle-xenon.

chrysn (approved):

laze-project.yml
SimonIT: That's strange. Why does it need double quotation marks?

chrysn: Because the computer said so ;-)

More seriously: The RUSTFLAGS are expanded into a shell invocation, which would strip away the unescaped quotation marks – whereas some quotation marks need to arrive at the rustc invocation which otherwise (and we're back to computer-said-so) complains that

```
  error: invalid `--cfg` argument: `getrandom_backend=custom` (expected `key` or `key="value"`, ensure escaping is appropriate for your shell, try 'key="value"' or key=\"value\")
```

(Actually we wouldn't need the outer quotes, but they also don't do harm, so I'd merge things as they are.)

chrysn:
For the larger review context: I'm merging this without related documentation changes. We don't have regular per-laze-feature documentation, and we don't have a comprehensive list of third party crates whom we set things up for.

Adding documentation would still be nice, but it's something we can evolve as the whole docs evolve.

SimonIT:
Do I need to do something about the CI failures? I'm confused what _check-labels_ and _Lint Commits_ is about. Do I need to update the lock file for _cargo-vet_?

src/ariel-os-random/src/lib.rs
```diff
@@ -246,3 +248,15 @@ pub fn crypto_rng_send() -> CryptoRngSend {
         }),
     }
 }
+
+#[cfg(feature = "csprng")]
```
chrysn, line 267:
```suggestion
/// Implements RNG access for the `getrandom` crate
///
/// # Safety
///
/// See the `getrandom` documentation on custom backends.
///
/// # Errors
///
/// See the `getrandom` documentation on custom backends.
///
/// At the time of writing, all Ariel OS RNGs are infallible.
///
/// # Panics
///
/// The function panics if error conversion fails.
#[cfg(feature = "csprng")]
```

chrysn:
Good point – frankly, I didn't check (for those would refuse on trouble anyway), but were probably not visible for lack of tags.

Yes, the CI complaints would need to be addressed:
* For the lints I have a proposal inline, not sure if that's all of them
* For the Cargo.lock, just check that in along with the rest.
* For the cargo vet, this doesn't need any new vetting (getrandom was already in our chain, it's just complaining about the Cargo.lock), If you've spent some time with the getrandom source code and care to provide an actual statement that you have looked at 0.3.3 and consider it safe to deploy, we're a bit wobbly around policies so far but generally happy to take in a review you might provide through cargo-vet, but there's no pressing need for action.

SimonIT:
How do I generate the Cargo.lock?

Running `cargo generate-lockfile` gives me this error:
```
error: failed to select a version for `embassy-rp`.
    ... required by package `ariel-os-threads v0.2.0 (/mnt/data/Software/Entwicklung/ariel-os/src/ariel-os-threads)`
    ... which satisfies path dependency `ariel-os-threads` of package `ariel-os v0.2.0 (/mnt/data/Software/Entwicklung/ariel-os/src/ariel-os)`
    ... which satisfies path dependency `ariel-os` of package `example-alloc v0.0.0 (/mnt/data/Software/Entwicklung/ariel-os/examples/alloc)`
versions that meet the requirements `^0.4` are: 0.4.0

package `ariel-os-threads` depends on `embassy-rp` with feature `fifo-handler` but `embassy-rp` does not have that feature.


failed to select a version for `embassy-rp` which could resolve this conflict
```

chrysn:
There should not need to be any manual process involved: If you have a clean (no modified files) checkout of this branch and run any Cargo activity that involves the lockfile (typically, running a test), this will leave the lockfile changed, and that change you can add and commit.

SimonIT:
grafik
I don't know why, but I can't do anything because of that error

chrysn:
It may help to go `cargo +nightly` here -- while Ariel OS can be built on stable, that can require some handholding which in typical builds laze does offer.

laze-project.yml
```diff
@@ -1418,6 +1418,15 @@ modules:
         FEATURES:
           - ariel-os/random
 
+  - name: random-crypto
+    help: A system-wide cryptographically secure RNG is available.
```
ROMemories: It's very important that this selects `random`, otherwise the CSPRNG would not get seeded from the HWRNG at startup and would panic on first use:
```suggestion
    help: A system-wide cryptographically secure RNG is available.
    selects:
      - random
```

laze-project.yml
```diff
@@ -1418,6 +1418,15 @@ modules:
...
+    env:
+      global:
+        FEATURES:
+          - ariel-os/csprng
+        RUSTFLAGS:
+          - '--cfg getrandom_backend="\"custom\""'
```
ROMemories: Equivalent but simpler, and consistent with how we do this for other `--cfg`s:
```suggestion
          - --cfg getrandom_backend=\"custom\"
```

ROMemories (changes requested):

laze-project.yml
SimonIT: Are you sure this works? Some tests complain that they can't find the custom backend

ROMemories:
> Are you sure this works? Some tests complain that they can't find the custom backend

Yes, but that's not related to my above suggestion. The issue is that `getrandom` needs to be optional and only enabled when the `csprng` Cargo feature is enabled (which I know is exactly what you had before), otherwise it's expecting a backend which is not provided.

Edit: To recap: `getrandom` does not need a dedicated laze module, separate from the CSPRNG, however it still needs to be an optional dependency, which can be enabled *as soon as* `csprng` is.

SimonIT:
However, it is now optional, and the error remains

ROMemories:
The current CI failures are due the `csprng` feature being enabled by two different things in ways that don't go through the `random-crypto` laze module and therefore don't pass `--cfg getrandom_backend="custom"`:
- Our Clippy checks that enable Cargo features that themselves enable `csprng` (currently that's the (unsupported) `ble-*` features)
- Applications that transitively enable the `csprng` feature

I need to think for a bit about how to solve these (possibly selecting the `random-crypto` laze module in the `hwrng` module, and passing `--cfg getrandom_backend="custom"` manually to Clippy).

ROMemories:
@SimonIT After all I went with passing `--cfg getrandom_backend="custom"` manually to our Clippy CI checks for now, and also passing the `--cfg` flag from the `hwrng` module. The new `random-crypto` module is not needed anymore for this PR (we *could* have a laze module that enables the `csprng` Cargo feature, but this is in itself now out of scope of this PR). I took the liberty of pushing this to your branch.

@chrysn There seems to be an issue with liboscore's build script, which seems to also be expecting the getrandom symbol but I'm not sure why. Could you have a look?

ROMemories:
@SimonIT @chrysn Found the issue with the `lint` CI check failing: the failing step was running Clippy on `coapcore`, which depends on `liboscore`, which itself relies on `cbindgen`. `cbindgen` has a transitive dependency on `getrandom`, through `tempfile` (this I've found using `cargo tree`).

It so happens that we weren't resetting our `RUSTFLAGS` environment variable before running that Clippy check and that I had added the `--cfg getrandom_backend="custom"` in it in the previous Clippy checks. What was happening is therefore that `getrandom`, which runs on the host as a dependency of `tempfile`, was still expecting a custom backend, as it was still getting that `--cfg` flag.

I've for now pushed a commit that resets `RUSTFLAGS` before the `coapcore` Clippy check.
I'm going to move this fix to a separate PR, and when this is in, we'll rebase this PR, which should then pass CI.

ROMemories (approved):
Tested successfully on a bbc-microbit-v2 in the `random` example by generating the random value with the following:

```rust
let value = getrandom::u32().unwrap();
```
after enabling the `csprng` Cargo feature on the example.

ROMemories:
@SimonIT Thanks a lot for this!

SimonIT:
Awesome! Thank you guys a lot!

ROMemories:
This new feature should be documented in #1416.

PR #1345, opened 2025-09-15
feat: add time unit/label for GNSS

nponsard:
# Description

Adds units, labels and category needed to represent a GNSS receiver with the sensor abstraction

## Issues/PRs references

- Depends on ##1373

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/label.rs
+++ b/src/ariel-os-sensors/src/label.rs
@@ -32,10 +32,11 @@ pub enum Label {
     Heading,
     /// Temperature.
     Temperature,
-    /// Time elapsed since Ariel OS epoch (2024-01-01).
-    Time,
-    /// Time elapsed since last full second.
-    SecondFraction,
+    /// First 4 bytes of the i64 timestamp (milliseconds since UNIX epoch).
+    /// Needs to be combied using [`datetime::combine_timestamp`].
+    DateTimeUpper,
+    /// Last 4 bytes of the i64 timestamp.
+    DateTimeLower,
     /// Vertical speed.
     VerticalSpeed,
     /// X axis.
@@ -57,8 +58,8 @@ impl core::fmt::Display for Label {
             Self::RelativeHumidity => write!(f, "Relative humidity"),
             Self::Heading => write!(f, "Heading"),
             Self::Temperature => write!(f, "Temperature"),
-            Self::Time => write!(f, "Time"),
-            Self::SecondFraction => write!(f, "Second fraction"),
+            Self::DateTimeUpper => write!(f, "Time (upper part)"),
+            Self::DateTimeLower => write!(f, "Time (lower part)"),
             Self::VerticalSpeed => write!(f, "Vertical speed"),
             Self::X => write!(f, "X"),
             Self::Y => write!(f, "Y"),
--- a/src/ariel-os-sensors/src/datetime.rs
+++ b/src/ariel-os-sensors/src/datetime.rs
@@ -1,0 +1,64 @@
+//! Utilities to handle datetime samples.
+//!
+//! Samples are limited to i32 values, a timestamp expressing microseconds since UNIX epoch needs to be stored in an i64.
+//! To work around the samples limitation, we have to split the timestamp into two i32 samples.
+//! You need to use [`combine_timestamp`] to reconstruct the i64 timestamp from the two samples labeled with [`Label::DateTimeUpper`] and [`Label::DateTimeLower`].
+//!
+//! For implementors you need to use [`split_timestamp`] to split an i64 timestamp into two samples.
+
+use crate::sample::Sample;
+
+/// Contains two samples representing a split i64 timestamp.
+pub struct DateTimeSplit {
+    /// Sample containing the upper 4 bytes
+    pub upper: Sample,
+    /// Sample containing the lower 4 bytes
+    pub lower: Sample,
+}
+
+/// Splits a timestamp into two samples. Has to be used by sensor driver implementors to return a timestamp.
+#[must_use]
+pub fn split_timestamp(timestamp: i64) -> DateTimeSplit {
+    let lower_bytes = (timestamp & 0xFFFF_FFFF) as i32;
+    let upper_bytes = ((timestamp >> 32) & 0xFFFF_FFFF) as i32;
+    let upper = Sample::new(upper_bytes, crate::sample::Accuracy::Unknown);
+    let lower = Sample::new(lower_bytes, crate::sample::Accuracy::Unknown);
+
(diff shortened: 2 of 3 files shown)


ROMemories:
Switched to `ci-build:skip` as there is no users of these items in the tree for now anyway.

src/ariel-os-sensors/src/measurement_unit.rs
```diff
@@ -49,6 +53,8 @@ pub enum MeasurementUnit {
     Lux,
     /// Meter (m)
     Meter,
+    /// Meters per second (m/s).
+    MetersPerSecond,
```
ROMemories: All of these should be [singular](https://www.iana.org/assignments/senml/senml.xhtml):
```suggestion
    /// Meter per second (m/s).
    MeterPerSecond,
```

src/ariel-os-sensors/src/measurement_unit.rs
```diff
@@ -104,6 +112,7 @@ macro_rules! provide_unit_fmt {
             Self::Lumen => write!($f, "lm"),
             Self::Lux => write!($f, "lx"),
             Self::Meter => write!($f, "m"),
+            Self::MetersPerSecond => write!($f, "m/s"),
```
ROMemories:
```suggestion
            Self::MeterPerSecond => write!($f, "m/s"),
```

src/ariel-os-sensors/src/measurement_unit.rs
```diff
@@ -26,6 +26,10 @@ pub enum MeasurementUnit {
     Coulomb,
     /// Decibel (dB).
     Decibel,
+    /// Decimal Degrees (°).
```
ROMemories: Should not be title-capitalized (Celsius is only because that's a proper noun):

```suggestion
    /// Decimal degrees (°).
```

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[non_exhaustive]
 pub enum Label {
+    /// Altitude
```
ROMemories:
```suggestion
    /// Altitude.
```

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
...
+    Altitude,
+    /// Ground Speed
```
ROMemories:
```suggestion
    /// Ground speed.
```

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
...
+    GroundSpeed,
+    /// Latitude
```
ROMemories:
```suggestion
    /// Latitude.
```

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
...
+    Latitude,
+    /// Longitude
```
ROMemories:
```suggestion
    /// Longitude.
```

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
...
+    Longitude,
     /// Used for sensor drivers returning a single [`Sample`](crate::sensor::Sample).
     Main,
     /// Relative humidity.
     RelativeHumidity,
+    /// Heading
+    Heading,
     /// Temperature.
     Temperature,
+    /// Time elapsed since Ariel epoch (2024-01-01).
```
ROMemories: We need to agree on a specific epoch, I think other candidates include 2000-01-01 and 2025-01-01. This also needs to include the exact time (in ISO format), not just the date.

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
...
+    Time,
+    /// Time elapsed since last full second.
+    SecondFraction,
```
ROMemories: I think this needs another name and maybe definition; I need to think about this some more, in the context of the GNSS API.

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,12 +16,28 @@
...
+    GroundSpeed,
+    /// Latitude
+    Latitude,
+    /// Longitude
+    Longitude,
     /// Used for sensor drivers returning a single [`Sample`](crate::sensor::Sample).
     Main,
     /// Relative humidity.
     RelativeHumidity,
+    /// Heading
```
ROMemories:
```suggestion
    /// Heading.
```

src/ariel-os-sensors/src/label.rs
ROMemories:
```suggestion
    /// Time elapsed since Ariel OS epoch (2024-01-01).
```

src/ariel-os-sensors/src/label.rs
nponsard: I couldn't find a better name, but I'm open to suggestions.

src/ariel-os-sensors/src/label.rs
nponsard: Should this be a subject for tomorrow's meeting ?

src/ariel-os-sensors/src/label.rs
ROMemories: I need to keep thinking about this, but I don't think the label should include a unit; maybe `TimeDelta`? And the documentation would be something like "Additional time resolution." (No need to include this for now, this needs some more pondering.)

src/ariel-os-sensors/src/label.rs
ROMemories: After some offline discussion we agreed on not using an epoch, but simply making that value opaque and to provide a helper function/type to take the associated sample and the one associated to `SecondFraction`/`TimeDelta` to return a date-time type.

After some more thinking, I think we could instead simply return an `i64` which would be a Unix timestamp (possibly in microseconds), which could then easily be used to instantiate such data-time types (`time`'s or `jiff`'s), while keeping these *out* of the sensor API crate.

nponsard: Would it make sense then to use the two i32 to store the bottom and top half of the i64 ?
The values wouldn't make much sense on their own, but converting them should be easy.

nponsard:
@ROMemories I've pushed a commit that adds helper functions to split an i64 timestamp into two samples, I'm not sure how to properly express it in the documentation.

ROMemories:
Can you split this PR so we can land what's not related to date+time now?

ROMemories:
Closing this as we opted for doing things differently. We can always re-open later if needed.

PR #1346, opened 2025-09-15
refactor: introduce the `ram-small` module for MCUs unable to do HTTP

ROMemories:
# Description

This introduces a `ram-small` laze module, similar to `ram-tiny` (#1053), for MCUs having more RAM than `ram-tiny` but still not enough to do HTTP or CoAP.

## How to review this PR

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -550,6 +550,7 @@ contexts:
     parent: stm32
     selects:
       - cortex-m0-plus
+      - ram-small
     disables:
       # TODO: not enough RAM for HTTP examples
       - network
@@ -856,6 +857,9 @@ modules:
       global:
         isr_stacksize_required_default: "512"
         executor_stacksize_required_default: "512"
+
+  - name: ram-small
+    help: This module marks MCUs with little RAM (currently, the ones which cannot run HTTP or CoAP)
 
   - name: stable
     help: build with stable Rust
@@ -1290,6 +1294,8 @@ modules:
       # unless provided by the context.
       - c-function-assert
       - c-function-abort
+    conflicts:
+      - ram-small
     env:
       global:
         FEATURES:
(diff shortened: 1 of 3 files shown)


laze-project.yml
```diff
@@ -843,6 +841,8 @@ modules:
 
   - name: ram-tiny
     help: This module marks boards with very little RAM (currently, <=16KiB)
+    selects:
+      - ram-small
```
ROMemories: Even if not strictly required (for now), I think it makes that `ram-tiny` selects `ram-small`.

laze-project.yml
```diff
@@ -550,9 +550,7 @@ contexts:
     parent: stm32
     selects:
       - cortex-m0-plus
-    disables:
-      # TODO: not enough RAM for HTTP examples
-      - network
```
ROMemories, line 555: Documenting this newly unlocked support is addressed by #1177.

laze-project.yml
kaspar030: Hm. This would make it possible for applications to "conflict: ram-small", implying "conflicts: ram-small".
But at the same time, it means that all ram-tiny boards have ram-small in their module list, which completely defeats the "this module marks boards", and we cannot use the modules to do any runtime/buildtime configuration.

We can't do it the other way around (ram-small selects ram-tiny), as then we cannot conflict ram-tiny without conflicting ram-small.

The solution is probably to split the markers and the selectors, as in, have something like "select: ram-at-least-small".

For this PR I suggest dropping this selects here, go with "conflicts: [`ram-small`, `ram-tiny`]" where applicable, and solve as next step.

ROMemories: > For this PR I suggest dropping this selects here

Done.

> go with "conflicts: [`ram-small`, `ram-tiny`]" where applicable

Didn't do anything in this direction, as `ram-tiny` already marks the conflict with `network`.

PR #1348, opened 2025-09-16
refactor(network): clarify the names of IPv4-only laze modules

ROMemories:
# Description

This renames laze modules that are actually IPv4-only to make room for the IPv6 modules that will be introduced by #915. This only renames laze modules for now and leaves Cargo features untouched.

This is not a breaking change as old modules are kept and only deprecated.

## How to review this PR

Check for forgotten occurrences in the codebase.

Test the `udp-echo` example (or any other network example):

```sh
laze -C examples/udp-echo/ build -s network-config-dhcp -b nrf52840dk
```

and pass `--disable network-config-dhcpv4` and `--disable network-config-dhcp` to it, possibly testing in hardware and/or checking the size of the binary to verify these are actually equivalent. The same can be done for `--select` to verify these indeed behave as aliases.

## Issues/PRs references

Extracted from #915.

## Change checklist

### DIFF
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -31,7 +31,7 @@ CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze build ...
 ### Network Configuration
 
 DHCPv4 is used by default for network configuration, including for IP address allocation.
-This is enabled by the `network-config-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
+This is enabled by the `network-config-ipv4-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
 
 In order to provide a static configuration, select the `network-config-static` [laze module](./build-system.md#laze-modules), which will take precedence.
 The configuration can be customized with the following environment variables:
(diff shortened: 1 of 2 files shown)


book/src/networking.md
```diff
@@ -31,9 +31,9 @@ CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze build ...
 ### Network Configuration
 
 DHCPv4 is used by default for network configuration, including for IP address allocation.
-This is enabled by the `network-config-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
+This is enabled by the `network-config-dhcpv4` [laze module](./build-system.md#laze-modules), selected by default.
 
-In order to provide a static configuration, select the `network-config-static` [laze module](./build-system.md#laze-modules), which will take precedence.
+In order to provide a static configuration, select the `network-config-static-ipv4` [laze module](./build-system.md#laze-modules), which will take precedence.
```
ROMemories: Should this be `network-config-ipv4-static` instead?

bergzand: For consistency, yes. I would even go as far as `network-config-ipv4-dhcp` so that it is clear those are all `network-config-ipv4` configuration mechanisms, and so they all show up when looking for `network-config-ipv4` configuration mechanisms.

ROMemories: I went even farther and made it `network-config-ipv4-dhcpv4` as DHCPv4 and DHCPv6 are different protocols (not just DHCP over IPv4/IPv6).

kaspar030: IMO that v4 is redundant, there'll not be `ipv4-dhcpv6` or `ipv6-dhcpv4`.

bergzand: Thinking about this a bit more, they should be named after the configuration mechanism (dhcp), and not the exact protocol implementation (dhcpv4 and dhcpv6).

PR #1354, opened 2025-09-17
feat: add TemporarilyUnavailable variant to sensor Accuracy

nponsard:
# Description

This adds a way to indicate the current sample should be ignored because this channel of the sensor isn't sending valid data yet. 

## Open Questions

Is this the right place to put the indication that the data should be ignored ?
The alternative would be to split one sensor into multiple sensors and have the ones that need more time to initialize/produce a value make the user wait on `Sensor::wait_for_reading()`  

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/sample.rs
+++ b/src/ariel-os-sensors/src/sample.rs
@@ -96,6 +96,9 @@ pub enum Accuracy {
         /// [`bias`](Accuracy::SymmetricalError::bias).
         scaling: i8,
     },
+    /// Indicate that the value in this channel is temporarily unavailable and the value in the sample should be ignored.
+    /// Happens when parts of a sensor aren't ready yet, but data is already available for other channels.
+    TemporarilyUnavailable,
 }
 
 /// Implemented on [`Samples`](crate::sensor::Samples), returned by


ROMemories:
Can you use this in #1330 so it's easier for me to see what it looks like in there?

nponsard:
> Can you use this in #1330 so it's easier for me to see what it looks like in there?

I made a commit showing how it would be used https://github.com/ariel-os/ariel-os/pull/1330/commits/b207c1bb94745a7b04e875846c39709b4a184d85

nponsard:
To add more context, the Accuracy is always a positive value and is 0 when no the modem has no fix. Time information can be obtained earlier than a full fix, that's why I have a different condition.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -96,6 +96,9 @@ pub enum Accuracy {
         /// [`bias`](Accuracy::SymmetricalError::bias).
         scaling: i8,
     },
+    /// Indicate that the value in this channel is temporarily unavailable and the value in the sample should be ignored.
+    /// Happens when parts of a sensor aren't ready yet, but data is already available for other channels.
```
ROMemories:
```suggestion
    /// Indicates that the value from this channel is temporarily unavailable and that the [`Sample`]'s value is considered opaque and should be ignored.
    /// Happens when parts of a sensor are not ready yet, but data is already available for other channels.
```

ROMemories (commented):
Looking at #1330, this looks as I expected; we'll wait until we actually want to land #1330 to land this PR, so we have some more time to ponder this.

nponsard:
> Looking at #1330, this looks as I expected; we'll wait until we actually want to land #1330 to land this PR, so we have some more time to ponder this.

Okay, I added it to the dependencies of #1330

bergzand:
This is only to fulfill the edge case where a sensor has multiple channels where some of those channels might not produce valid data while others already do?

I see some issues with hiding that info in the accuracy info and leave the actual measurement value as seemingly valid. To me this looks like a possible footgun as the sample value can be used without checking the accuracy for validity. For example a user that does not care about and does not display the gnss accuracy might use invalid samples

ROMemories:
> I see some issues with hiding that info in the accuracy info and leave the actual measurement value as seemingly valid. To me this looks like a possible footgun as the sample value can be used without checking the accuracy for validity. For example a user that does not care about and does not display the gnss accuracy might use invalid samples

We'll definitely need to give this more visibility in the docs. Unfortunately I don't see another way that wouldn't significantly increase the memory footprint.

bergzand:
> We'll definitely need to give this more visibility in the docs. Unfortunately I don't see another way that wouldn't significantly increase the memory footprint.

May I suggest adding an getter for the `value` that includes a check for a valid accuracy, and returns a `Result<i32, SomethingSomethingError>`. That's the best I can come up with without sacrificing the plain `i32` for the `value`

nponsard:
> May I suggest adding an getter for the `value` that includes a check for a valid accuracy, and returns a `Result<i32, SomethingSomethingError>`. That's the best I can come up with without sacrificing the plain `i32` for the `value`

That sounds like it could work, we could change the current `Sample::value()` getter or add something like `Sample::checked_value()`

bergzand:
I agree with keeping the current `Sample::value()` and add the `Sample::checked_value()`. Changing the current `Sample::value()` to a `Sample::unchecked_value()` is going to discourage users (not necessarily a bad thing), but for many sensors the extra checks will not be required and the simple (unchecked) case is fine.

ROMemories:
Definitely sounds like a good idea! I'm not too sure about the name though: `checked` is usually reserved for things that could otherwise panic (like [`i32::checked_add()`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.checked_add)) or that work in pair with an `unchecked` variant that is `unsafe` ([`C-GETTER`](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)). Maybe `value()` and `raw_value()`? Not a huge fan of that either though, the latter sounds maybe a bit *too* raw (however *renaming* the `value()` method is still ok at this point if need be)?

nponsard:
I've added a commit to make `Sample::value()` return an error when the value is temporarily unavailable. The other getter is `Sample::value_unchecked()`. Naming still needs to be discussed.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -96,6 +115,9 @@ pub enum Accuracy {
         /// [`bias`](Accuracy::SymmetricalError::bias).
         scaling: i8,
     },
+    /// Indicates that the value from this channel is temporarily unavailable and that the [`Sample`]'s value is considered opaque and should be ignored.
+    /// Happens when parts of a sensor are not ready yet, but data is already available for other channels.
+    TemporarilyUnavailable,
```
ROMemories: Actually maybe that should be called `ValueTemporarilyUnavailable`, otherwise that reads as if *the accuracy* was temporarily unavailable:

```suggestion
    ValueTemporarilyUnavailable,
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -43,8 +50,20 @@ impl Sample {
     }
 
     /// Returns the sample value.
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if this value is temporarily unavailable (this part of the sensor is not ready).
+    pub fn value(&self) -> Result<i32, ValueError> {
+        if self.accuracy == Accuracy::TemporarilyUnavailable {
```
ROMemories: I would rather this be written more defensively with an exhaustive `match` (without a wildcard pattern), so that it wouldn't be possible to forget to update this if the `Accuracy` enum were to change later.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -1,3 +1,10 @@
+/// Errors returned when trying to interpret a sensor value.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
ROMemories, line 4:
```suggestion
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -43,8 +50,20 @@ impl Sample {
...
+        if self.accuracy == Accuracy::TemporarilyUnavailable {
+            return Err(ValueError::TemporarilyUnavailable);
+        }
+        Ok(self.value)
+    }
+
+    /// Returns the sample value without checking if it is valid.
     #[must_use]
-    pub fn value(&self) -> i32 {
+    pub fn value_unchecked(&self) -> i32 {
```
ROMemories: I think this is the good way forward, we just need to find a better name for this that doesn't suggest this could cause UB.

nponsard: I renamed it to `raw_value()`. I don't know if there's a better way to express it.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -43,8 +51,26 @@ impl Sample {
...
+            } => Ok(self.value),
+        }
+    }
+
+    /// Returns the sample value without checking if it is valid.
     #[must_use]
-    pub fn value(&self) -> i32 {
+    pub fn raw_value(&self) -> i32 {
         self.value
     }
```
ROMemories: As discussed offline, I don't think we actually need to provide this method. If we really we later realized there were a use case for it, we'd be able to add it back it and that wouldn't be a breaking change.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -96,6 +122,9 @@ pub enum Accuracy {
         /// [`bias`](Accuracy::SymmetricalError::bias).
         scaling: i8,
     },
+    /// Indicates that the value from this channel is temporarily unavailable and that the [`Sample`]'s value is considered opaque and should be ignored.
```
ROMemories: I think we can remove the end of that sentence as we're removing `raw_value()` and the raw value is therefore not exposed anymore in this case:
```suggestion
    /// Indicates that the value from this channel is temporarily unavailable.
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -1,3 +1,11 @@
+/// Errors returned when trying to interpret a sensor value.
```
ROMemories: "Sample" is better defined in the sensor abstraction definitions (in the crate root):
```suggestion
/// Errors returned when trying to interpret a sample.
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -1,3 +1,11 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ValueError {
+    /// The value is not available at the moment (i.e., part of the sensor is not ready yet).
```
ROMemories: This should be an example, not an equivalence:
```suggestion
    /// The value is not available at the moment (e.g., part of the sensor is not ready yet).
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -43,8 +51,26 @@ impl Sample {
...
+    pub fn value(&self) -> Result<i32, ValueError> {
+        match self.accuracy {
+            Accuracy::ValueTemporarilyUnavailable => Err(ValueError::TemporarilyUnavailable),
+            Accuracy::NoError
+            | Accuracy::Unknown
+            | Accuracy::SymmetricalError {
+                deviation: _,
+                bias: _,
+                scaling: _,
+            } => Ok(self.value),
```
ROMemories: No need to list every field I think:
```suggestion
            | Accuracy::SymmetricalError { .. } => Ok(self.value),
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -1,3 +1,11 @@
+/// Errors returned when trying to interpret a sensor value.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ValueError {
```
ROMemories: Even though I'm usually wary of marking things `non_exhaustive`, I think it may make sense in this case as users would be able to provide a meaningful fallback (which would involve ignoring the sample):
```suggestion
#[non_exhaustive]
pub enum ValueError {
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -43,8 +51,26 @@ impl Sample {
     }
 
     /// Returns the sample value.
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if this value is temporarily unavailable (this part of the sensor is not ready).
```
ROMemories:
```suggestion
    /// An error is returned if this value is temporarily unavailable.
```

src/ariel-os-sensors/src/sample.rs
nponsard: removed `raw_value()`.

nponsard:
Squashed commits and rebased on main.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -1,3 +1,12 @@
+/// Errors returned when trying to interpret a sample.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum ValueError {
```
ROMemories: This type needs to be exported from the `sensor` module.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -43,9 +52,17 @@ impl Sample {
     }
 
     /// Returns the sample value.
-    #[must_use]
-    pub fn value(&self) -> i32 {
-        self.value
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if this value is temporarily unavailable.
```
ROMemories: This shouldn't constrain the set of errors returned by this function, otherwise there is no point in marking the error type as non-exhaustive as it's only returned by this function. This also avoids duplication.
```suggestion
    /// [`ValueError`] is returned in case of error.
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -96,6 +113,9 @@ pub enum Accuracy {
         /// [`bias`](Accuracy::SymmetricalError::bias).
         scaling: i8,
     },
+    /// Indicates that the value from this channel is temporarily unavailable.
+    /// Happens when parts of a sensor are not ready yet, but data is already available for other channels.
```
ROMemories: Somewhat similarly, this is only one possible use case of that variant:
```suggestion
    /// This may happen when parts of a sensor are not ready yet, but data is already available for other channels.
```

src/ariel-os-sensors/src/sample.rs
ROMemories: This leaves open the possibility of adding a `ChannelDisabled` error for instance, which we may need soonish (when a channel is disabled by configuration).

ROMemories (commented):
Having merged and adjusted #1412 against this for testing, I'm wondering whether we should rename this `SampleTemporarilyUnavailable`. "Value" is not part of the sensor abstraction definitions (at the crate root), while "sample" is better defined.

I'm also planning to rename `Accuracy` so it makes more sense with `*TemporarilyUnavailable` and `ChannelDisabled`. Maybe `SampleMetadata` or `SampleInfo`?

nponsard:
I renamed `ValueError` to `SampleError` and exposed it from the `sensor` module

src/ariel-os-sensors/src/sample.rs
```diff
@@ -1,5 +1,14 @@
 use crate::sensor::ReadingChannel;
 
+/// Errors returned when trying to interpret a sample.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SampleError {
+    /// The value is not available at the moment (e.g., part of the sensor is not ready yet).
```
ROMemories:
```suggestion
    /// The channel is not available at the moment (e.g., part of the sensor is not ready yet).
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -98,6 +115,9 @@ pub enum Accuracy {
         /// [`bias`](Accuracy::SymmetricalError::bias).
         scaling: i8,
     },
+    /// Indicates that the value from this channel is temporarily unavailable.
```
ROMemories:
```suggestion
    /// Indicates that the channel is temporarily unavailable.
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -98,6 +115,9 @@ pub enum Accuracy {
...
+    /// This may happen when parts of a sensor are not ready yet, but data is already available for other channels.
+    ValueTemporarilyUnavailable,
```
ROMemories:
```suggestion
    ChannelTemporarilyUnavailable,
```

src/ariel-os-sensors/src/sample.rs
```diff
@@ -45,9 +54,17 @@ impl Sample {
     }
 
     /// Returns the sample value.
-    #[must_use]
-    pub fn value(&self) -> i32 {
-        self.value
+    ///
+    /// # Errors
+    ///
+    /// [`ValueError`] is returned in case of error.
```
ROMemories:
```suggestion
    /// [`SampleError`] is returned in case of error.
```

ROMemories: That should have been caught by rustdoc 🤔

nponsard: Oh yeah weird.

PR #1355, opened 2025-09-17
feat(sensors): provide `From` constructors for sensor data enums

ROMemories:
# Description

This provides `From` constructors that are intended to be used in lieu of the enum variant constructors, making the variants an implementations details, as accessing the inner data is done through the `Reading` trait. This would allow us to later revisit how the `Sample`s and `ReadingChannel`s are internally stored if we wanted/needed to.

## Change checklist


src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -139,4 +146,31 @@ mod define_count_adjusted_enum {
     pub fn variant_name(index: usize) -> syn::Ident {
         quote::format_ident!("V{index}")
     }
+
+    pub fn get_allocation_size() -> usize {
```
ROMemories, line 150: This is somewhat unrelated, but otherwise Clippy complains that the main function is too long (and it's not completely wrong), and the refactor is trivial.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -11,33 +11,23 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+    let count = get_allocation_size();
 
     let samples_variants = (1..=count).map(|i| {
         let variant = variant_name(i);
         quote! { #variant([Sample; #i]) }
     });
+    let samples_from_impls = (1..=count)
+        .map(|i| {
+            let variant = variant_name(i);
+            quote! {
```
chrysn, line 23: Reading has a special case to access a single-element sample. Should that work both ways (`impl From<Sample> for Samples`)?

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -11,33 +11,23 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+                impl From<[Sample; #i]> for Samples {
```
chrysn, line 24: It may be preferable to emit those all the time and only `#[cfg()]` them on the relevant feature.

Then, both for documentation ("This impl is only available with the feature …") and compiler errors ("This impl is available but the feature is missing!"), the condition is accessible.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
ROMemories: `Reading::sample()` exists mainly because `Samples` is *guaranteed* to have one element or more, so users don't have to `.next().unwrap()` the iterator returned by `Reading::samples()`. This reason does not really call for a symmetrical constructor, so I'm not adding it in this PR, but we could add it later.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
ROMemories: I agree documentation can/should be improved (#1325 was doing that, it'll need to be heavily reworked). In this case this is not so easy because each `From[Sample; #i]` impl is available as soon as any `max-sample-min-count-$c` Cargo feature is enabled, with c >= i. So it would be possible to generate that `#[cfg]` attribute, but the docs would get really crowded (as `From[Sample; 2]` would be feature-gated to 11 different Cargo features).

For now I consider that out-of-scope of this PR, but definitely agree that the documentation should be clarified.

ROMemories:
Merging this now as I have other PRs stacked on this, we can add other additive constructors and improve the documentation later.

PR #1356, opened 2025-09-17
feat(probe-rs): add `attach` laze task

kaspar030:
# Description

This adds an 'attach' task to the probe-rs module, allowing to re-attach to an already flashed image for more output.

(Skipping CI build as this doesn't change any binary.)

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1027,6 +1027,12 @@ modules:
           - probe-rs download --chip ${PROBE_RS_CHIP} ${out} $@
           - probe-rs reset --chip ${PROBE_RS_CHIP}
 
+      attach:
+        help: Attach probe-rs to rtt logging
+        build: false
+        cmd:
+          - probe-rs attach --chip ${PROBE_RS_CHIP} ${out} $@
+
       debug:
         help: Starts a probe-rs gdb server
         cmd:


laze-project.yml
```diff
@@ -1027,6 +1027,12 @@ modules:
           - probe-rs download --chip ${PROBE_RS_CHIP} ${out} $@
           - probe-rs reset --chip ${PROBE_RS_CHIP}
 
+      attach:
+        help: Attach probe-rs to rtt logging
+        build: false
```
nponsard, line 1032:
```suggestion
        build: true
```

The output binary `${out}` is used in the command. I think it would be better to make sure the file exists.

kaspar030: yeah, I was struggling with that one. If the image on the device doesn't match, `attach` probably breaks. If it needs to be build, it probably doesn't match. ... I went with the fastest, but I'm not sure.

nponsard: I tried rebuilding with a different log level (trace to debug) and attaching, the log output does not make any sense and probe-rs doesn't warn about it. That's a good argument towards not regenerating the file.

kaspar030: ... we can also not just fetch the image from the device and compare with the elf, as that will "disturb" it (the device).
... the "file does not exist" will error out fast.

I suggest we stick with "build: false". :-)

nponsard: Sounds good :+1:

nponsard (approved):
After the discussion on enabling the build option or not, the changes look good to me :+1:

PR #1361, opened 2025-09-18
chore(sensor-drivers): introduce a subdirectory for sensor drivers

ROMemories:
# Description

This lays out the structure the sensor drivers will go into. This is a separate PR so we're not forced to land future sensor driver PRs in a specific order.

## Change checklist

### DIFF
--- a/src/laze.yml
+++ b/src/laze.yml
@@ -13,3 +13,4 @@ subdirs:
   - ariel-os-stm32
   - ariel-os-threads
   - lib
+  - sensors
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -24,6 +24,7 @@ members = [
   "src/ariel-os-sensors",
   "src/ariel-os-stm32",
   "src/ariel-os-storage",
+  "src/sensors/*",
```
kaspar030: I would appreciate going explicit here. With the examples already I sometimes get build errors due to leftover files.

PR #1363, opened 2025-09-18
feat(builtin-sensors): introduce an STTS22H sensor driver

ROMemories:
# Description

This adds a sensor driver for the STTS22H temperature sensor compatible with our sensor API.

## Testing

This can be tested on hardware using the following (the st-steval-mkboxpro is also supported in the example):

```sh
```

It can be useful to test on any other board to check that the example is still running as expected (which means doing nothing interesting).

## Issues/PRs references

- Depends on #1191
- Depends on #1351
- Depends on #1361
- Depends on #1589
- Depends on #1605

## Open Questions

- The sensor driver instantiation and configuration "interfaces" are pretty tied to the board descriptions.
- The commit adding the sensor to the `sensors-debug` example is marked WIP as we might not want to modify that example in this PR at all (e.g., because it is very tied to specific boards).

## Changelog entry

A sensor driver for the STTS22H, compatible with the newly-introduced sensor API, is now available.

## Change checklist


.github/workflows/main.yml
```diff
@@ -400,7 +400,7 @@ jobs:
 
       # For now, only run cargo-manifest-check on ariel-os-identity.
```
kaspar030:
```suggestion
      # For now, only run cargo-manifest-check on selected crates.
```

bergzand:
Is it possible to use (parts of) https://github.com/STMicroelectronics/stts22h-rs/ for the driver part? That driver doesn't look like it supports async, so maybe just the register definitions?

kaspar030:
> Is it possible to use (parts of) [STMicroelectronics/stts22h-rs](https://github.com/STMicroelectronics/stts22h-rs/?rgh-link-date=2025-09-18T13%3A02%3A55Z) for the driver part? That driver doesn't look like it supports async, so maybe just the register definitions?

I think this sensor is simple enough, don't think it makes much sense at this point.

src/sensors/ariel-os-sensor-stts22h/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+            MeasurementUnit::Celsius,
+        )])
+    }
+
+    fn label(&self) -> Option<&'static str> {
+        self.label
+    }
+
+    fn display_name(&self) -> Option<&'static str> {
+        Some("thermometer")
```
ROMemories:
```suggestion
        Some("temperature sensor")
```

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,257 @@
...
+);
+
+/// Driver to use an STTS22H over I2C.
+pub struct Stts22h {
+    state: AtomicState,
+    label: Option<&'static str>,
+    i2c: OnceLock<Mutex<CriticalSectionRawMutex, I2cDevice>>,
+    address: AtomicU8,
+    signaling: Signal<CriticalSectionRawMutex, ()>,
+    reading_signal: ReadingSignal<Result<Samples, ReadingError>>,
```
Ollrogge: Just saw this while checking out this PR as an example of how to implement the sensor api :P
```suggestion
    reading_signal: ReadingSignal<ReadingResult<Samples>>
```

ROMemories: Indeed, thanks!

src/sensors/ariel-os-sensor-stts22h/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+portable-atomic = { workspace = true }
+
+[dev-dependencies]
+critical-section = { workspace = true, features = ["std"] }
+embassy-executor = { workspace = true, features = [
+  "arch-std",
+  "executor-thread",
+] }
+embassy-futures = { workspace = true }
+embassy-time = { workspace = true, features = ["std"] }
```
ROMemories, line 27: I'm not exactly sure this is the best way of running these tests, but it works and we can easily revisit this later if we have more general guidelines on host tests.

src/sensors/ariel-os-sensor-stts22h/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "ariel-os-sensor-stts22h"
+# This crate is versioned separately from Ariel OS.
+version = "0.1.0"
+edition.workspace = true
+# This crate's MSRV is decoupled from Ariel OS's.
+rust-version = "1.85"
```
ROMemories: Before landing this, we might actually want to bump the MSRV of `ariel-os-sensors` (and therefore this MSRV).

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    /// Initializes the driver.
+    pub async fn init(
+        &'static self,
+        _peripherals: Peripherals,
+        mut i2c_device: I2C,
+        config: Config,
+    ) {
+        if !self.i2c.is_set() {
```
nponsard, line 81: It would be nice to emit a warning if this condition is not met (meaning init is called twice).

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+        i2c_device
+            .write(
+                address as u8,
+                &[Register::CtrlRegAddr as u8, crate::IF_ADD_INC_BITS],
+            )
+            .await
+            .map_err(|_| ())?;
+
+        i2c_device
+            .write(address as u8, &[Register::TempHLimit as u8, 0x00, 0x00])
```
nponsard, line 106: This effectively sets both `TEMP_H_LIMIT` and `TEMP_L_LIMIT` to 0, disabling the high and low interrupts 👍 .

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+            .await
+            .map_err(|_| ())?;
+
+        Ok(())
+    }
+
+    /// Listens for measurement requests generated by [`Stts22h::trigger_measurement()`], and
+    /// responds to them.
+    /// This should be called before [`Stts22h::wait_for_reading()`], as it will otherwise delay
+    /// the response.
```
nponsard: I'm having difficulties making sense of this comment. In what way does it delay the response, is it a constant delay during the whole life of the driver if you await this function too late ?

I think it should be explained more clearly that you won't obtain a response from `wait_for_reading()` until `Stts22h::run()` is awaited in a task.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+            self.signaling.wait().await;
+
+            match self.measure().await {
+                Ok(samples) => {
+                    self.reading.signal(Ok(samples));
+                }
+                Err(err) => {
+                    self.reading.signal(Err(err));
+                }
+            }
```
nponsard: Is there a reason for using a `match` statement here ? It looks like you could just pass the result of `measure()` directly to `reading.signal()`.


```suggestion
            self.reading.signal(self.measure().await);
```

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+
+impl<I2C: Send> Sensor for Stts22h<I2C> {
+    fn trigger_measurement(&self) -> Result<(), TriggerMeasurementError> {
+        match self.state.get() {
+            State::Measuring => {
+                self.reading.clear();
+            }
+            State::Enabled => {
+                self.state.set(State::Measuring);
+            }
```
nponsard, line 185: Since the state changes from `Measuring` to `Enabled` once `wait_for_reading()` is called, meaning a task is waiting for the result, we don't need to clear `self.reading` when the current state is `Enabled` 👍

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+            State::Uninitialized | State::Disabled | State::Sleeping => {
+                return Err(TriggerMeasurementError::NonEnabled);
+            }
+        }
+
+        self.signaling.signal(());
```
nponsard, line 191: There may be a problem if 3 readings are requested concurrently where one `wait_for_reading` may not get a response: 
- first signal unlocks the loop in `run()`
- second signal gets stored in `self.signaling`
- at the third call the signal `self.signaling` is already signaled so calling `signal()` on it does nothing.

I don't know how important it is to handle that edge case.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -0,0 +1,390 @@
...
+                self.state.set(State::Enabled);
+
+                ReadingWaiter::new(self.reading.wait())
+            }
+            State::Uninitialized | State::Disabled | State::Sleeping => {
+                return ReadingWaiter::new_err(ReadingError::NonEnabled);
+            }
+            State::Enabled => {
+                return ReadingWaiter::new_err(ReadingError::NotMeasuring);
+            }
```
nponsard: More of a nitpick, would be nice to have the cases handled in the same order as other functions. 
```suggestion
            State::Enabled => {
                return ReadingWaiter::new_err(ReadingError::NotMeasuring);
            }
            State::Uninitialized | State::Disabled | State::Sleeping => {
                return ReadingWaiter::new_err(ReadingError::NonEnabled);
            }
```

src/sensors/ariel-os-sensor-stts22h/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+// CTRL register bits.
+const ONE_SHOT_BITS: u8 = 1 << 0;
+const IF_ADD_INC_BITS: u8 = 1 << 3;
+const BDU_BITS: u8 = 1 << 6;
+
+// STATUS register bits.
+const BUSY_BITS: u8 = 1 << 0;
+
+#[expect(dead_code)]
+const DEVICE_ID: u8 = 0xa0;
```
nponsard, line 37: Is this needed somewhere ?

nponsard (changes requested):

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: Exactly; I've added a comment to make that extra clear.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: That's a leftover from when `SensorSignaling` was a thing and had different methods for these branches. Thanks for spotting that!

src/sensors/ariel-os-sensor-stts22h/src/lib.rs
ROMemories: It's not currently used; in a way it's a remnant of the development process (which usually starts with trying to read the WHOAMI register). I'd rather keep it as it doesn't hurt and can always come in handy.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: This initialization is intended to be completely abstracted and codegened; I'd rather not add a log statement for this as that would slightly increase the binary size and add an extra dependency, for something that should eventually be guaranteed not to happen, by construction (otherwise indeed that would be useful).

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: Indeed, I'll rephrase.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: I need to think about this a bit.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: I've rephrased a bit.

src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
ROMemories: Thanks for spotting this behavior and bringing this up. After giving it quite some thought, I think we can consider it an edge case in the sensor driver that we don't need to specifically handle (unless there is some specific case I haven't thought of). The underlying issue is that the internal signaling is only a 1-bit flag and therefore multi-triggers are forgotten which can indeed lead to multiple, concurrent waiters never resolving. 

However this should never happen in a well-formed application (and I can't think of any legitimate use cases for trying to do that). There *would* be ways to fix things, but they all come at a cost, which seems unwarranted:

- We could introduce a `TriggerMeasurementError::MeasurementInProgress` error in the sensor API, which `trigger_measurement()` would return if a measurement has already been triggered. That would prevent the behavior from happening, but that turns an application mis-design into a runtime error that the application would have no meaningful way to respond to.
- The sensor driver could use another construct than a 1-bit signal, but that would come at a runtime cost for something that should just never happen.
- The sensor driver could simply panic in case of concurrent triggers (possibly only when `debug-assertions` are enabled); this would be cheap and allow for easily spotting the misuse, but it's unclear whether that would be compliant with the `Sensor` trait definition.

All in all I think this doesn't need to be handled in the sensor driver and we can simply keep the current behavior (if we later discovered there are indeed use cases for that it would be possible to fix the sensor driver, without requiring changes to the sensor API).

nponsard: Sounds good to me, It may be useful to document this somewhere but this is outside the scope of this PR.

examples/sensors-debug/src/sensors.rs
ROMemories, line 1: One of the open questions was whether we wanted this in this PR, and now I think we do. I've reworked this slightly so it can easily accommodate the sensors from other boards (and removed the change that restricted the example to the two supported boards with an STTS22H: the example would still compile for other contexts, but do nothing interesting). Do keep in mind that this is all temporary until we are able to codegen this from SBD files.

PR #1365, opened 2025-09-18
 feat: add initial uart abstraction

bergzand:
# Description

Split out from #978

This PR provides an abstraction layer for UART peripherals.
The focus is on the shared UART TX/RX and provide the buffered mode of the peripherals.
The drivers implements the [embedded-io-async::Read](https://docs.rs/embedded-io-async/0.6.1/embedded_io_async/trait.Read.html), [embedded-io-async::ReadReady](https://docs.rs/embedded-io-async/0.6.1/embedded_io_async/trait.ReadReady.html), and [embedded-io-async::Write](https://docs.rs/embedded-io-async/0.6.1/embedded_io_async/trait.Write.html) traits. The [BufRead](https://docs.rs/embedded-io-async/0.6.1/embedded_io_async/trait.BufRead.html) can be added for peripherals that support it.

## Issues/PRs references

Split out from #978 (and rebased), now only contains the common uart parts with the test application 

## Changelog entry

A UART abstraction has been introduced, similar to the I2C and SPI abstractions. Drivers are provided for each currently supported HAL: ESP32, nRF, RP, and STM32.

## Change checklist

### DIFF
--- a/tests/uart-loopback/Cargo.toml
+++ b/tests/uart-loopback/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "uart-loopback"
+license.workspace = true
+edition.workspace = true
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["time", "uart"] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+embedded-io-async = { workspace = true }
--- a/tests/uart-loopback/README.md
+++ b/tests/uart-loopback/README.md
@@ -1,0 +1,19 @@
+# UART loopback test
+
+## About
+
+This application tests the UART peripheral via an external loopback wire.
+
+## How to run
+
+1. Ensure a wire is present between the TX and RX pins.
+2. In this directory, run:
+
+    laze build -b nrf52840dk run
+
+The test attempts to do a transfer and compares if what was sent has been read back.
+
+## Potential issues
+
+- An existing serial connection on the board can interfere with the test.
+  The TX pin of the other serial endpoint can conflict with the loopback operation.
--- a/tests/uart-loopback/src/main.rs
+++ b/tests/uart-loopback/src/main.rs
@@ -1,0 +1,53 @@
+//! This is a test for UART loopback operation
+
+#![no_main]
+#![no_std]
+
+mod pins;
+
+use ariel_os::{
+    debug::{
+        ExitCode, exit,
+        log::{Hex, info},
+    },
+    hal,
+    time::{Duration, with_timeout},
+    uart::Baudrate,
+};
+
+use embedded_io_async::{Read as _, Write as _};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    info!("Starting UART test");
+
+    let mut config = hal::uart::Config::default();
+    config.baudrate = Baudrate::_115200;
+    info!("Selected configuration: {:?}", config);
+
+    let mut rx_buf = [0u8; 32];
+    let mut tx_buf = [0u8; 32];
+
+    let mut uart = pins::TestUart::new(
+        peripherals.uart_rx,
+        peripherals.uart_tx,
+        &mut rx_buf,
+        &mut tx_buf,
+        config,
+    )
+    .expect("Invalid UART configuration");
+
+    const OUT: &str = "Test Message";
+    let mut input = [0u8; OUT.len()];
+
+    uart.write_all(OUT.as_bytes()).await.unwrap();
+    uart.flush().await.unwrap();
+    info!("Wrote bytes");
+    let _ = with_timeout(Duration::from_secs(5), uart.read_exact(&mut input)).await;
+
+    info!("Got: {}", Hex(input));
+    assert_eq!(OUT.as_bytes(), input);
+    info!("Test passed!");
+
+    exit(ExitCode::SUCCESS);
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5639,6 +5639,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb77c29baba9e4d3a6182d51fa75e3215c7fd1dab8f4ea9d107c716878e55fc0"
 
 [[package]]
+name = "uart-loopback"
+version = "0.0.0"
+dependencies = [
+ "ariel-os",
+ "ariel-os-boards",
+ "embedded-io-async",
+]
+
+[[package]]
 name = "udp-echo"
 version = "0.0.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ members = [
   "tests/threading-lock",
   "tests/threading-mutex",
   "tests/threading-fpu",
+  "tests/uart-loopback",
 ]
 
 exclude = ["src/lib", "doc"]
--- a/tests/laze.yml
+++ b/tests/laze.yml
@@ -12,3 +12,4 @@ subdirs:
   - threading-fpu
   - threading-lock
   - threading-mutex
+  - uart-loopback
--- a/tests/uart-loopback/laze.yml
+++ b/tests/uart-loopback/laze.yml
@@ -1,0 +1,3 @@
+apps:
+  - name: uart-loopback
+    context: []
--- a/tests/uart-loopback/src/pins.rs
+++ b/tests/uart-loopback/src/pins.rs
@@ -1,0 +1,1 @@
+use ariel_os::hal::{peripherals, uart};


src/ariel-os-hal/src/hal/dummy/uart.rs
```diff
@@ -0,0 +1,21 @@
...
+/// driver provided by this module.
+// NOTE: we keep this type public because it may still required in user-written type signatures.
+pub enum Uart {
+    // Make the docs show that this enum has variants, but do not show any because they are
+    // MCU-specific.
+    #[doc(hidden)]
+    Hidden,
+}
+
+pub fn init(_peripherals: &mut crate::OptionalPeripherals) {
```
kaspar030:
```suggestion
pub fn init(_peripherals: &mut crate::hal::OptionalPeripherals) {
```

src/ariel-os-hal/Cargo.toml
```diff
@@ -75,6 +75,8 @@ spi = [
   "ariel-os-stm32/spi",
 ]
 
+uart = []
```
ROMemories:
```suggestion
uart = ["ariel-os-embassy-common/uart"]
```

src/ariel-os-embassy/src/uart.rs
ROMemories, line 1: Following #1273, should this be moved into `ariel-os-hal` instead?

tests/uart-loopback/README.md
```diff
@@ -0,0 +1,19 @@
+# UART loopback test
+
+## About
+
+This application tests the UART peripheral via an external loopback wire
```
ROMemories:
```suggestion
This application tests the UART peripheral via an external loopback wire.
```

tests/uart-loopback/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## How to run
+
+1. Ensure a wire is present between the TX and RX pins
```
ROMemories:
```suggestion
1. Ensure a wire is present between the TX and RX pins.
```

tests/uart-loopback/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "uart-loopback"
+license.workspace = true
+edition.workspace = true
+
```
ROMemories, line 6: Other tests and example have that key, to prevent us from mistakenly publish these on crates.io; even if not strictly necessary because there's isn't a `version` key anyway, I would still prefer we have for consistency:

```suggestion
publish = false
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _4800,
+    /// 9600 bauds
+    _9600,
+    /// 19200 bauds
+    _19200,
+    /// 38400 bauds
+    _34800,
+    /// 57600 bauds
+    _57600,
+    /// 57600 bauds
```
nponsard:
```suggestion
    /// 115200 bauds
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
+//! Provides HAL-agnostic UART-related types.
+
+/// Common UART baud rates.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Baud {
+    /// Custom baud rate
```
ROMemories:
```suggestion
    /// Custom baud rate.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    Baud(u32),
+    /// 2400 bauds,
```
ROMemories:
```suggestion
    /// 2400 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _2400,
+    /// 4800 bauds
```
ROMemories:
```suggestion
    /// 4800 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _4800,
+    /// 9600 bauds
```
ROMemories:
```suggestion
    /// 9600 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _9600,
+    /// 19200 bauds
```
ROMemories:
```suggestion
    /// 19200 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _19200,
+    /// 38400 bauds
```
ROMemories:
```suggestion
    /// 38400 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _34800,
+    /// 57600 bauds
```
ROMemories:
```suggestion
    /// 57600 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    _57600,
+    /// 115200 bauds
```
ROMemories:
```suggestion
    /// 115200 bauds.
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
+//! Provides HAL-agnostic UART-related types.
+
+/// Common UART baud rates.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Baud {
+    /// Custom baud rate
+    Baud(u32),
```
ROMemories: I'm not sure we should have this variant in this shared type; all values of this type should work on every MCU we support ever (115,200 bauds seems fine as that's even supported on AVR MCUs). Have a look at how it's done for SPI: each HAL has a dedicated `Frequency` type with the max frequency being checked at const time. There's also a helper `highest_freq_in()` const function in `ariel-os-embassy` to make this HAL-agnostic based on a range of frequencies.

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for Baud {
+    fn format(&self, f: defmt::Formatter<'_>) {
+        use defmt::write;
+        write!(f, "{=u32}", Into::<u32>::into(*self));
+    }
+}
+
+/// UART parity.
```
ROMemories:
```suggestion
/// Parity bit.
```

src/ariel-os-embassy/src/uart.rs
ROMemories, line 1: Following what was started with #1273, shouldn't all this be moved into `ariel-os-hal`?

tests/uart-loopback/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+mod pins;
+
+use ariel_os::{
+    debug::{
+        ExitCode, exit,
+        log::{Hex, info},
+    },
+    hal,
+    time::Duration,
+    time::with_timeout,
```
ROMemories:
```suggestion
    time::{Duration, with_timeout},
```

tests/uart-loopback/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+};
+
+use embedded_io_async::{Read as _, Write as _};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    info!("Starting UART test");
+
+    let mut config = hal::uart::Config::default();
+    config.baudrate = Baud::_9600;
```
ROMemories: This is debatable, but maybe we could default to 115,200 bauds nowadays (same for the defaults in every HAL)?

tests/uart-loopback/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+    let mut uart = pins::TestUart::new(
+        peripherals.uart_rx,
+        peripherals.uart_tx,
+        &mut rx_buf,
+        &mut tx_buf,
+        config,
+    );
+
+    const OUT: &str = "Test Message";
+    let mut in_ = [0u8; OUT.len()];
```
ROMemories: There's other ways to name that variables to not have to resort to using an underscore:

```suggestion
    let mut input = [0u8; OUT.len()];
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+    };
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! impl_async_uart_bufread_for_driver_enum {
+    ($driver_enum:ident, $( $peripheral:ident ),*) => {
+        impl embedded_io_async::BufRead for $driver_enum<'_> {
```
ROMemories, line 91: Do we really need to keep these lifetimes on the drivers? I don't see a strong use case for switching the UART peripheral at runtime, and we already erase them for SPI, I2C and GPIOs.

src/ariel-os-hal/Cargo.toml
bergzand: Is this the change that's needed because #1364 landed in the mean time?

ROMemories: Yes, but more importantly #1273, we're moving everything to `ariel-os-hal` now.

bergzand:
> I've done a first review of this, but haven't tested any of it yet. I'll move to reviewing one of the HAL and may get back to this if I find other things when looking at how this is used. Hopefully we can limit this to two other rounds.

Makes sense to me. I'd suggest to look at least at the esp hal, in addition to one of the others. the esp implementation is a bit of an outlier compared to the others as it has hardware buffers.

src/ariel-os-embassy-common/src/uart.rs
bergzand: I don't think the same structure with `highest_freq_in()` as done with SPI is going to work with UART. With SPI the exact frequency usually doesn't matter as long as it doesn't exceed the max frequency of the peripheral. That doesn't work with UART, you need a matching baud rate (within bounds) otherwise communication won't work.
 I'm fine with restricting this type to a common set of uart baud rates, but an escape to custom rates is still useful to have as not all uart hardware uses the 'standard' baud rates

bergzand: I'm fine with a different method of supplying a custom baud rate, but this looked the most convenient to me for users of the api

tests/uart-loopback/src/main.rs
bergzand: No strong opinion on this. I went with the conservative option here. It's not significant for the duration of the test. I also expect both baud rates to work fine over a jumper wire, with 9600 baud having a *very slightly* better chance of working.

src/ariel-os-embassy-common/src/uart.rs
ROMemories: Ah yes that's right, I had in mind the case where two identical MCUs would communicate over UART, in which case that *would* work. I need to think for a bit whether we should still move that arbitrary baudrate config into the HAL crates. How does that work if you were to provide an unsupported baudrate (say 1 Mbit/s to an MCU that doesn't support that)? We don't want this to fail at runtime.

src/ariel-os-embassy-common/src/uart.rs
bergzand: The compiler complains if I erase them here. As far as I can tell these are caused by how the buffers are currently passed into the drivers, see also https://github.com/ariel-os/ariel-os/pull/978#discussion_r2050661153

ROMemories: Right, because that lifetime is tied to that of the buffers as well, not just the underlying peripheral.

src/ariel-os-embassy-common/src/uart.rs
bergzand: > How does that work if you were to provide an unsupported baudrate (say 1 Mbit/s to an MCU that doesn't support that)? We don't want this to fail at runtime.

I have to think whether this is possible.

As an example that requires arbitrary baud rates, [SWO](https://kb.segger.com/SWO) has uart framing and works at baud rates divided from the cpu frequency (which *might* result in common uart baud rates).

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// 2400 bauds,
+    _2400,
+    /// 4800 bauds
+    _4800,
+    /// 9600 bauds
+    _9600,
+    /// 19200 bauds
+    _19200,
+    /// 38400 bauds
+    _34800,
```
ROMemories:
```suggestion
    _38400,
```

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl From<Baud> for u32 {
+    fn from(b: Baud) -> u32 {
+        match b {
+            Baud::Baud(b) => b,
+            Baud::_2400 => 2400,
+            Baud::_4800 => 4800,
+            Baud::_9600 => 9600,
+            Baud::_19200 => 19200,
+            Baud::_34800 => 34800,
```
ROMemories:
```suggestion
            Baud::_38400 => 38400,
```

src/ariel-os-embassy-common/src/uart.rs
bergzand: Shit, good catch!

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+
+/// UART parity.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Parity {
+    /// No parity bit.
+    None,
+    /// Even parity bit.
+    Even,
+    /// Odd parity bit.
+    Odd,
```
ROMemories: Considering odd parity is not supported on nRF, I'm not sure it should be able on this shared type.

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+/// UART number of stop bits.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum StopBits {
+    /// One stop bit.
+    Stop1,
+    /// Two stop bits.
+    Stop2,
```
ROMemories: Same for this actually, nRF doesn't support it and it seems to me extremely rare anyway. I think it should be moved into HAL types only.

bergzand: I agree and fine with leaving this out.

tests/uart-loopback/README.md
bergzand: Fixed in a26d2854bea58b9a8907e2cafa1a77cd98ba63df

tests/uart-loopback/README.md
bergzand: fixed in a26d2854bea58b9a8907e2cafa1a77cd98ba63df

tests/uart-loopback/Cargo.toml
bergzand: Added in 69cf45220918b92e26cfc78f527230275287cfdb and added in the test I copied from in #1374

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in 17a3b3822c54bae8f750ac0dfcc8eda6b34a503a

tests/uart-loopback/src/main.rs
bergzand: Applied in c94fcfd58b1f7aa0b1b7b880e48c7e4452c8f169

tests/uart-loopback/src/main.rs
bergzand: Copied from the SPI loopback test :innocent:, fixed in c94fcfd58b1f7aa0b1b7b880e48c7e4452c8f169. Let me know if you find it worth the effort to also rename the variable in the spi loopback test :smile:

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in ca1db85f9e2991703f383f9d3ca7e2c71a7399ed

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fixed in ca1db85f9e2991703f383f9d3ca7e2c71a7399ed

src/ariel-os-embassy-common/src/uart.rs
bergzand: Is there a good way to still support odd parity on the other platforms? Is it acceptable to use conditional compilation for odd parity?

I'd prefer to keep the odd parity available for the other platforms as faults that result in sending all zeroes or all ones are detected via the parity.

src/ariel-os-embassy/src/uart.rs
bergzand: Moved in 1cfeb679e5750150bbe2eaaefe63c2a61d59af7e, I think I moved the necessary bits, but this is not a part I'm very confident in.

src/ariel-os-embassy-common/src/uart.rs
ROMemories: After having another look at this I think `Baud`, `Parity`, and `StopBits` should have a `Hal` variant as we do for `gpio::{DriveStrength, Speed}`, so that the common values are shared and easy to use while still allowing to provide HAL-specific values when needed.

src/ariel-os-embassy-common/src/uart.rs
ROMemories: See my comment on `Baud`

src/ariel-os-embassy-common/src/uart.rs
ROMemories: Same as `Baud`

src/ariel-os-hal/Cargo.toml
ROMemories: I think that one still needs to be added

src/ariel-os-embassy-common/src/uart.rs
bergzand: > > How does that work if you were to provide an unsupported baudrate (say 1 Mbit/s to an MCU that doesn't support that)? We don't want this to fail at runtime.
> 
> I have to think whether this is possible.

Looking at the embassy-stm32 implementation for USART, this is already hard to guarantee. Different from SPI, the embassy [USART constructors](https://docs.embassy.dev/embassy-stm32/git/stm32f446re/usart/struct.BufferedUart.html#method.new) allow for returning an error on incorrect configuration. Guaranteeing we don't trigger any cases is going to be hard, for example, during the [configuration of the device](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L1679-L1813), the baud rate divider is [determined](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/usart/mod.rs#L1568) based on the clock supplied to the peripheral and the possible dividers in the peripheral.

Digging through, the embassy-nrf and embassy-rp do not have this issue and always return an Uart instance from a config, the esp-hal is similar to embassy-stm32 in that in can return a `ConfigError`

src/ariel-os-hal/Cargo.toml
bergzand: Added!

src/ariel-os-embassy-common/src/uart.rs
bergzand: > After having another look at this I think `Baud`, `Parity`, and `StopBits` should have a `Hal` variant as we do for `gpio::{DriveStrength, Speed}`, so that the common values are shared and easy to use while still allowing to provide HAL-specific values when needed.

I've added HAL-specific variants for these types. I've updated the stm32 PR with these changes, to have an implementation example ready.

src/ariel-os-embassy-common/src/uart.rs
bergzand: Added a HAL-specific variant for the parity

src/ariel-os-embassy-common/src/uart.rs
bergzand: Added a HAL-specific variant for the stop bits

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,263 @@
...
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConfigError {
+    /// The requested baud rate is not supported.
+    BaudrateNotSupported,
+    /// Data bits and parity combination not supported.
+    DataParityNotSupported,
+    /// Peripheral-specific error.
+    ConfigurationNotSupported,
+}
```
bergzand, line 13: This is the set of configuration-related errors I came up with, when merging the stm32 and the esp configuration errors into one generic error struct. The generic `ConfigurationNotSupported` is used for configuration errors that occur on configuration very unlikely to change during the device lifetime (fifo sizes, driver enable timings, RX and TX both disabled, etc.). 
Both have a set of different baud rate related errors that get squashed into one. Only the stm32 have an error for data and parity combinations, which I expect that might happen during device reconfiguration, so it's added here.

ROMemories: That seems like a reasonable set of errors 👍

bergzand:
I've removed the Hal traits again and the display implementations as discussed offline. The `defmt::Format` implementations are now all in the derive macros.

tests/uart-loopback/src/main.rs
bergzand: Current consensus is that this should be 115200

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Common UART baud rates.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Baud<A> {
```
ROMemories: I should have seen this before, but I think this should be `Baudrate`: we call these based on the physical quantity, not the unit (e.g., `DriveStrength`, `Speed`).

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,148 @@
...
+/// UART number of data bits.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum DataBits<A> {
+    /// HAL-specific number of data bits per character.
+    Hal(A),
+    /// 7 bits per character.
+    Data7,
+    /// 8 bits per character.
+    Data8,
```
ROMemories: I don't think nRF allows configuring this, if not only the value common to all HALs should be kept.

src/ariel-os-embassy-common/src/uart.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/// UART number of data bits.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum DataBits<A> {
+    /// HAL-specific number of data bits per character.
+    Hal(A),
+    /// 7 bits per character.
+    Data7,
```
bergzand:
```suggestion
```

The Nordic chips don't support this case, so it has to go.

src/ariel-os-embassy-common/src/uart.rs
bergzand: Fair, renamed

src/ariel-os-embassy-common/src/uart.rs
bergzand: Yup, removed

doc/support_matrix.yml
ROMemories, line 1: I think this is good to go; can you rebase onto main, and then either drop the changes to this file or, alternatively, run the following command to update the board pages?

```sh
./doc/gen_book.rs board-pages doc/support_matrix.yml book/src/boards/ --template-path book/templates/board-page.md.tmpl
```

Edit: also, please autosquash!

bergzand: rebased, squashed and dropped the support matrix changes for a later PR

PR #1368, opened 2025-09-18
feat(rp): Add UART HAL support

bergzand:
# Description

This adds support for the UART HAL on the Raspberry Pi Pico series devices

## Issues/PRs references

depends on #1365

## Change checklist

### DIFF
--- a/tests/uart-loopback/laze.yml
+++ b/tests/uart-loopback/laze.yml
@@ -1,3 +1,5 @@
 apps:
   - name: uart-loopback
-    context: []
+    context:
+      - rp2040
+      - rp235xa
--- a/tests/uart-loopback/src/pins.rs
+++ b/tests/uart-loopback/src/pins.rs
@@ -1,1 +1,9 @@
 use ariel_os::hal::{peripherals, uart};
+
+#[cfg(context = "rp")]
+pub type TestUart<'a> = uart::UART0<'a>;
+#[cfg(context = "rp")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    uart_rx: PIN_17,
+    uart_tx: PIN_16,
+});


src/ariel-os-rp/Cargo.toml
```diff
@@ -22,6 +22,7 @@ embassy-rp = { workspace = true, default-features = false, features = [
   # "unstable-traits",
 ] }
 embedded-hal-async = { workspace = true }
+embedded-io-async = { workspace = true }
```
ROMemories: Should be made optional if only used for UART

```suggestion
embedded-io-async = { workspace = true, optional = true }
```

src/ariel-os-rp/Cargo.toml
```diff
@@ -67,6 +68,9 @@ i2c = ["ariel-os-embassy-common/i2c"]
 ## Enables SPI support.
 spi = ["dep:embassy-embedded-hal", "ariel-os-embassy-common/spi"]
 
+## Enables UART support.
+uart = ["ariel-os-embassy-common/uart"]
```
ROMemories:
```suggestion
uart = ["dep:embedded-io-async", "ariel-os-embassy-common/uart"]
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+    uart::{BufferedInterruptHandler, BufferedUart, RxPin, TxPin},
+};
+
+/// UART interface configuration.
+#[derive(Clone)]
+#[non_exhaustive]
+pub struct Config {
+    /// The baud rate at which the bus should operate.
+    pub baudrate: Baud,
+    /// Number of data bits
```
ROMemories:
```suggestion
    /// Number of data bits.
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub data_bits: DataBits,
+    /// Number of stop bits
```
ROMemories:
```suggestion
    /// Number of stop bits.
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+use embassy_rp::{
+    Peripheral, bind_interrupts, peripherals,
+    uart::{BufferedInterruptHandler, BufferedUart, RxPin, TxPin},
+};
+
+/// UART interface configuration.
+#[derive(Clone)]
+#[non_exhaustive]
+pub struct Config {
+    /// The baud rate at which the bus should operate.
```
ROMemories: UART is not a bus:
```suggestion
    /// The baud rate at which UART should operate.
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Number of stop bits
+    pub stop_bits: StopBits,
+    /// Parity mode used for the interface.
+    pub parity: Parity,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            baudrate: Baud::_9600,
```
ROMemories: Maybe it makes sense to go with 115,200 bauds these days?

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+                    rx_buf: &'d mut [u8],
+                    tx_buf: &'d mut [u8],
+                    config: Config,
+                ) -> Uart<'d> {
+                    // Make this struct a compile-time-enforced singleton: having multiple statics
+                    // defined with the same name would result in a compile-time error.
+                    paste::paste! {
+                        #[allow(dead_code)]
+                        static [<PREVENT_MULTIPLE_ $peripheral>]: () = ();
+                    }
```
ROMemories: As was discussed in the public chat the other day, these statics should actually be moved outside of this function (between the struct definition and the impl block), because statics are scoped to the function and would actually not conflict as intended (but they're likely redundant anyway).

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+define_uart_drivers!(
+   UART0_IRQ => UART0,
+   UART1_IRQ => UART1,
+);
+
+#[doc(hidden)]
+pub fn init(peripherals: &mut crate::OptionalPeripherals) {
+    // Take all SPI peripherals and do nothing with them.
+    cfg_if::cfg_if! {
+        if #[cfg(context = "rp")] {
```
ROMemories: We *want* this to error out if a future `rp` MCU had more than two UART peripherals, so we don't forget to take them here:

```suggestion
        if #[cfg(any(context = "rp2040", context = "235xa"))] {
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+impl_defmt_display_for_config!();
+
+macro_rules! define_uart_drivers {
+    ($( $interrupt:ident => $peripheral:ident ),* $(,)?) => {
+        $(
+            /// Peripheral-specific UART driver.
+            pub struct $peripheral<'d> {
```
ROMemories, line 165: Do we really need the lifetime here (see my [other comment](https://github.com/ariel-os/ariel-os/pull/1365/files#r2362218426))?

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+                uart: BufferedUart<'d, peripherals::$peripheral>,
+            }
+
+            impl<'d> $peripheral<'d> {
+                #[expect(clippy::new_ret_no_self)]
+                #[must_use]
+                /// Returns a driver implementing [`embedded-io`] for this Uart
+                /// peripheral.
```
ROMemories:
```suggestion
                /// Returns a driver implementing [`embedded-io`] for this Uart
                /// peripheral.
                #[expect(clippy::new_ret_no_self)]
                #[must_use]
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+                    // peripheral multiple times.
+                    let uart_peripheral = unsafe { peripherals::$peripheral::steal() };
+
+                    let uart = BufferedUart::new(
+                        uart_peripheral,
+                        Irqs,
+                        tx_pin,
+                        rx_pin,
+                        tx_buf,
+                        rx_buf,
```
ROMemories, line 212: Could you add a comment saying that the RX/TX order is indeed different (so that's not a typo)? Unfortunately these cannot be typechecked

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+        $(
+            /// Peripheral-specific UART driver.
+            pub struct $peripheral<'d> {
+                uart: BufferedUart<'d, peripherals::$peripheral>,
+            }
+
+            impl<'d> $peripheral<'d> {
+                #[expect(clippy::new_ret_no_self)]
+                #[must_use]
+                /// Returns a driver implementing [`embedded-io`] for this Uart
```
ROMemories: Doesn't it also/only implement `embedded-io-async` traits? Also `embedded-io` is a crate, this should either list the actual traits or just say "embedded-io traits".

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+define_uart_drivers!(
+   UART0_IRQ => UART0,
+   UART1_IRQ => UART1,
+);
+
+#[doc(hidden)]
+pub fn init(peripherals: &mut crate::OptionalPeripherals) {
+    // Take all SPI peripherals and do nothing with them.
```
ROMemories:
```suggestion
    // Take all UART peripherals and do nothing with them.
```

src/ariel-os-rp/Cargo.toml
bergzand: Added in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/Cargo.toml
bergzand: Added in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: Fixed in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: Fixed in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: Fixed in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: As mentioned in https://github.com/ariel-os/ariel-os/pull/1365#discussion_r2362392972, a very slight preference to the conservative option from me.

src/ariel-os-rp/src/uart.rs
bergzand: removed in f8a417a07db3ff4c581d95f585b748ad8743b1e4 and added in the suggested place in a65a81f2

src/ariel-os-rp/src/uart.rs
bergzand: Fixed in https://github.com/ariel-os/ariel-os/commit/f8a417a07db3ff4c581d95f585b748ad8743b1e4 and a65a81f2

src/ariel-os-rp/src/uart.rs
bergzand: For now, [yes](https://github.com/ariel-os/ariel-os/pull/1365#discussion_r2362571376)

src/ariel-os-rp/src/uart.rs
bergzand: Fixed in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: :+1: Added in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: rephrased to say "embedded-io traits" in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
bergzand: Fixed in f8a417a07db3ff4c581d95f585b748ad8743b1e4

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,150 @@
+//! UART bus configuration.
```
ROMemories: Missed that one before:

```suggestion
//! UART configuration.
```

bergzand:
This PR is now also up to date with the latest in #1365

doc/support_matrix.yml
```diff
@@ -176,6 +181,7 @@ chips:
       hwrng: supported
       i2c_controller: supported
       spi_main: supported
+      uart: supported
```
ROMemories:
```suggestion
      uart: not_currently_supported
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,248 @@
...
+            stop_bits: ariel_os_embassy_common::uart::StopBits::Stop1,
+            parity: ariel_os_embassy_common::uart::Parity::None,
+        }
+    }
+}
+
+/// UART baud rate.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct Baud {
```
ROMemories:
```suggestion
pub struct Baudrate {
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,248 @@
...
+    uart::{BufferedInterruptHandler, BufferedUart, RxPin, TxPin},
+};
+
+/// UART interface configuration.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub struct Config {
+    /// The baud rate at which UART should operate.
+    pub baudrate: ariel_os_embassy_common::uart::Baud<Baud>,
```
ROMemories: Would it be an issue to use the HAL-specific types directly? That would make the `Config` type slightly smaller, and look simpler in the docs:  
```suggestion
    pub baudrate: Baud,
```

bergzand: This is the config that is supplied to the constructor by the user, such as in the test [here](https://github.com/ariel-os/ariel-os/pull/1368/files#diff-ae10a110391622353d83a1c4fdd8920045b0f90206afa91317a4983d3bf8843bR24). Changing to the HAL-specific types would remove the use of the common abstraction types there right?

ROMemories: I mean currently we need this to set the parity for instance:

```rust
config.parity = ariel_os::uart::Parity::Hal(hal::uart::Parity::Even);
```

If we used the HAL-specific types directly, it would look like this:

```rust
config.parity = hal::uart::Parity::Even;
```

bergzand: >     config.parity = ariel_os::uart::Parity::Hal(hal::uart::Parity::Even);

This was the way I understood it, and the hoops a user has to go through to get a custom mode.

>     config.parity = hal::uart::Parity::Even;

I agree this looks better, but how would this work for the `Baudrate`, that's an u32 on most targets, except on the nRF, where it is an enum with distinct values.

ROMemories: Actually isn't the HAL-specific `Baud`/`Baudrate` type missing a public constructor?
If we added a `From<u32>` on it, we could simply do:

```rust
config.baudrate = 1_000_000.into();
```

bergzand: As discussed offline. Changed the types in the config struct to the HAL-specific types. Except for the baud rate because nRF makes everything more complicated with the enum there

ROMemories: Now we can do both of these:

```rust
config.stop_bits = ariel_os::hal::uart::StopBits::Stop2;
config.stop_bits = ariel_os::uart::StopBits::Stop1.into();
```

ROMemories: > Actually isn't the HAL-specific `Baud`/`Baudrate` type missing a public constructor?

Can you show me how to set a 1 Mbaud baud rate?

bergzand: With the latest fixup:

```Rust
    config.baudrate = Baudrate::Hal(1_000_000.into());
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,259 @@
...
+
+use ariel_os_embassy_common::{impl_async_uart_for_driver_enum, uart::ConfigError};
+
+use embassy_rp::{
+    Peripheral, bind_interrupts, peripherals,
+    uart::{BufferedInterruptHandler, BufferedUart, RxPin, TxPin},
+};
+
+/// UART interface configuration.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
ROMemories: I think this [should be `Copy` as well](https://doc.rust-lang.org/stable/std/marker/trait.Copy.html#when-should-my-type-be-copy):

```suggestion
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,259 @@
...
+
+impl From<Parity> for embassy_rp::uart::Parity {
+    fn from(parity: Parity) -> embassy_rp::uart::Parity {
+        match parity {
+            Parity::None => embassy_rp::uart::Parity::ParityNone,
+            Parity::Even => embassy_rp::uart::Parity::ParityEven,
+            Parity::Odd => embassy_rp::uart::Parity::ParityOdd,
+        }
+    }
+}
```
ROMemories: I think it would be better not to do this with a `From` impl, because it otherwise ends up in the docs, while `embassy_rp::uart__Parity` is a private type. Using a free function should be fine.

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,259 @@
...
+
+/// UART number of stop bits.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum StopBits {
+    /// One stop bit.
+    Stop1,
+    /// Two stop bit.
```
ROMemories:
```suggestion
    /// Two stop bits.
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -0,0 +1,259 @@
...
+                static [<PREVENT_MULTIPLE_ $peripheral>]: () = ();
+            }
+
+            impl<'d> $peripheral<'d> {
+                /// Returns a driver implementing embedded-io traits for this Uart
+                /// peripheral.
+                ///
+                /// # Errors
+                ///
+                /// This never returns an error.
```
ROMemories, line 182: It would also be possible to use [`Infallible`](https://doc.rust-lang.org/stable/std/convert/enum.Infallible.html) as an error type, but that would make constructors inconsistent across HALs, so maybe it's best not to do it.

bergzand:
> However on an rpi-pico2-w, I'm consistently receiving an extra null byte before getting the data, so the test fails.

Trying to reproduce this on my rpi-pico2. Did not yet have a successful test. Oscilloscope measurements show it's not in the transmit path:
image

**edit:** Clarifying, the uart-loopback test only succeed once per power cycle per uart peripheral on the rpi-pico2 for me. 

To reproduce:
1. Flash an unrelated example on the board (e.g. `examples/hello-world`).
2. Power cycle the board.
3. Flash and run the uart-loopback test on the board:
    ```
    [INFO ] Starting UART test (uart_loopback uart-loopback/src/main.rs:22)
    [INFO ] Selected configuration: Config { baudrate: _115200, data_bits: Data8, stop_bits: Stop1, parity: None } (uart_loopback uart-loopback/src/main.rs:26)
    [INFO ] Wrote bytes (uart_loopback uart-loopback/src/main.rs:45)
    [INFO ] Got: [54, 65, 73, 74, 20, 4d, 65, 73, 73, 61, 67, 65] (uart_loopback uart-loopback/src/main.rs:48)
    [INFO ] Test passed! (uart_loopback uart-loopback/src/main.rs:50)
    Firmware exited successfully
    ```
    test succeeds.
4. Run the test again:
     ```
    [INFO ] Starting UART test (uart_loopback uart-loopback/src/main.rs:22)
    [INFO ] Selected configuration: Config { baudrate: _115200, data_bits: Data8, stop_bits: Stop1, parity: None } (uart_loopback uart-loopback/src/main.rs:26)
    [WARN ] Framing error (embassy_rp src/uart/buffered.rs:534)
    [WARN ] Break error (embassy_rp src/uart/buffered.rs:540)
    [INFO ] Wrote bytes (uart_loopback uart-loopback/src/main.rs:45)
    [INFO ] Got: [00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00] (uart_loopback uart-loopback/src/main.rs:48)
    panicked at tests/uart-loopback/src/main.rs:49:5:
    assertion `left == right` failed
      left: [84, 101, 115, 116, 32, 77, 101, 115, 115, 97, 103, 101]
      right: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
    ```
    Test fails now
5. Change the UART peripheral to `UART1` and `PIN_7` and `PIN_6`.
6. Flash and run the test again:
    ```
    [INFO ] Starting UART test (uart_loopback uart-loopback/src/main.rs:22)
    [INFO ] Selected configuration: Config { baudrate: _115200, data_bits: Data8, stop_bits: Stop1, parity: None } (uart_loopback uart-loopback/src/main.rs:26)
    [INFO ] Wrote bytes (uart_loopback uart-loopback/src/main.rs:45)
    [INFO ] Got: [54, 65, 73, 74, 20, 4d, 65, 73, 73, 61, 67, 65] (uart_loopback uart-loopback/src/main.rs:48)
    [INFO ] Test passed! (uart_loopback uart-loopback/src/main.rs:50)
    Firmware exited successfully
    ```
7. Rerunning the test, even when changing it back to UART0, it keeps failing.

**edit2:** A reset of the peripheral during initialization of the peripheral fixes the test. The whole issue is likely caused by that a debugger reset only resets the CPU core and none of the peripherals. The following very quick and dirty patch on Embassy resolves the issue with the uart loopback test here:
```diff
diff --git i/embassy-rp/src/uart/mod.rs w/embassy-rp/src/uart/mod.rs
index 8d12aeef6..dec306bfb 100644
--- i/embassy-rp/src/uart/mod.rs
+++ w/embassy-rp/src/uart/mod.rs
@@ -844,6 +844,11 @@ impl<'d, T: Instance + 'd, M: Mode> Uart<'d, T, M> {
         cts: Option<PeripheralRef<'_, AnyPin>>,
         config: Config,
     ) {
+        let mut resets = rp_pac::resets::regs::Peripherals::default();
+        resets.set_uart0(true);
+        resets.set_uart1(true);
+        crate::reset::reset(resets);
+        crate::reset::unreset_wait(resets);
         let r = T::regs();
         if let Some(pin) = &tx {
             let funcsel = {
```

src/ariel-os-rp/src/uart.rs
bergzand: That's exactly why I kept the `ConfigError` there and documented it this way.

src/ariel-os-rp/src/uart.rs
bergzand: I assume this also applies to the `DataBits` and `StopBits`?

ROMemories: Yes

src/ariel-os-rp/src/uart.rs
bergzand: Yup, missed that one somehow. Added in e55744ca

bergzand:
Thanks again for all the reviewing and testing @ROMemories :heart:

PR #1315, opened 2025-09-08
ci(lint): check for missing `rust-version` keys in `ariel-os*` crates

ROMemories:
# Description

Follow-up to #1314, makes sure that each and every `ariel-os*` crate has a `rust-version` key from now on.

## Issues/PRs references

Depends on #1314.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1374, opened 2025-09-19
chore(tests): Add missing publish = false

bergzand:
# Description

Other tests and example have the key, to prevent mistakenly publish these tests on crates.io. For consistency this adds the key to the tests where it is still missing.

## Change checklist


ROMemories (approved):
Haven't checked that these were all of them. We could also add a CI check similar to #1315 (that would also check examples).

bergzand:
> Haven't checked that these were all of them. We could also add a CI check similar to #1315 (that would also check examples).

I checked only the tests here directory, this should be all of them

PR #1377, opened 2025-09-22
docs(net): document support for static IPv6

ROMemories:
# Description

Document support for static IPv6 newly added by #915.

## Issues/PRs references

- Depends on #915
- Depends on #1571
- Depends on #1651

## Changelog entry

Support for IPv6 has been added, which can be used alongside IPv4 or in place of it. Only static configuration is currently supported. See [the networking documentation](https://ariel-os.github.io/ariel-os/dev/docs/book/networking.html) for details.

## Change checklist

### DIFF
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -52,7 +52,7 @@ The configuration can be customized with the following environment variables:
 Support for IPv6 is not currently enabled by default, but can be enabled by selecting the `ipv6` [laze module](./build-system.md#laze-modules).
 IPv4 and IPv6 can both be enabled at the same time.
 
-IPv6 currently only supports static configuration, which can be selected with the `network-config-ipv6-static` [laze module](./build-system.md#laze-modules).
+IPv6 currently only supports static configuration, which is therefore enabled by default; it can also be explicitly selected with the `network-config-ipv6-static` [laze module](./build-system.md#laze-modules).
 
 The configuration must be customized with the following environment variables:
 


book/src/networking.md
```diff
@@ -30,10 +30,15 @@ CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze build ...
...
 
-DHCPv4 is used by default for network configuration, including for IP address allocation.
-This is enabled by the `network-config-ipv4-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
+#### IPv4
 
-In order to provide a static configuration, select the `network-config-ipv4-static` [laze module](./build-system.md#laze-modules), which will take precedence.
+Support for IPv4 is enabled by default.
+When unneeded, it can be disabled by disabling the `ipv4` [laze module](./build-system.md#laze-modules).
+
+DHCPv4 is used by default for IPv4 network configuration, including for address allocation.
```
bergzand:
```suggestion
DHCP is used by default for IPv4 network configuration, including for address allocation.
```

Nitpicking here, but I think calling it plain DHCP is less confusing to the reader.

book/src/networking.md
```diff
@@ -30,10 +30,15 @@ CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze build ...
 
 ### Network Configuration
 
-DHCPv4 is used by default for network configuration, including for IP address allocation.
-This is enabled by the `network-config-ipv4-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
+#### IPv4
 
-In order to provide a static configuration, select the `network-config-ipv4-static` [laze module](./build-system.md#laze-modules), which will take precedence.
+Support for IPv4 is enabled by default.
+When unneeded, it can be disabled by disabling the `ipv4` [laze module](./build-system.md#laze-modules).
```
bergzand, line 36:
```suggestion
When unneeded, it can be disabled by [disabling](./build-system.md#forbidding-laze-modules-for-an-application) the `ipv4` [laze module](./build-system.md#laze-modules):

   apps:
     - name: udp-echo
       conflicts:
          - ipv4
```

Keeps it easier on the reader. (please check if my markdown is correct)

book/src/networking.md
```diff
@@ -42,6 +47,21 @@ The configuration can be customized with the following environment variables:
 | `CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN` | `24`         |
 | `CONFIG_NET_IPV4_STATIC_GATEWAY_ADDRESS` | `10.42.0.1`  |
 
+#### IPv6
+
+Support for IPv6 is not currently enabled by default, but can be enabled by selecting the `ipv6` [laze module](./build-system.md#laze-modules).
```
bergzand, line 52:
```suggestion
Support for IPv6 is not currently enabled by default, but can be enabled by [selecting](./build-system.md#enabling-laze-modules-for-an-application) the `ipv6` [laze module](./build-system.md#laze-modules).
```

bergzand (changes requested):
Looks good to me, just some clarifications.

book/src/networking.md
bergzand: Main reason for me to include this snippet in the docs (and not one below) is that this pattern is less common.

ROMemories: The reason why I hadn't added that link is that this tends to hide the ability of using the CLI as well (the CLI section is linked to the config file section, but not the other way around, and you'd need to scroll *up* to find it), and users will have to have digested the Build System page anyway.

Also I'd rather not have that 4-line snippet here, which duplicates what already is in the Build System page and, to me, gives too much emphasis to this (and again tends to hide the possibility of also using the CLI).

Do you really want this here? I'd rather slightly rework the Build System page if it's not clear enough.

book/src/networking.md
ROMemories: Same as above

book/src/networking.md
ROMemories: Attempt at slightly clarifying the Build System page in https://github.com/ariel-os/ariel-os/pull/1570.

book/src/networking.md
```diff
@@ -42,6 +47,21 @@ The configuration can be customized with the following environment variables:
...
+IPv4 and IPv6 can both be enabled at the same time.
+
+IPv6 currently only supports static configuration, which can be selected with the `network-config-ipv6-static` [laze module](./build-system.md#laze-modules).
+
+The configuration can be customized with the following environment variables:
```
chrysn:
```suggestion
Support for IPv6 is not enabled by default until address auto-configuration becomes available.
IPv4 and IPv6 can both be enabled at the same time.

Currently, IPv6 can be enabled by selecting the `network-config-ipv6-static` [laze module](./build-system.md#laze-modules)
and providing the following environment variables:
```

Proposing this for two reasons:

* Set the expectation that IPv6 *will* be enabled by default as soon as autoconfig works.
* Phrase static IPv6 configuration as *the* way to start it, no point in enabling it without static config set.

book/src/networking.md
```diff
@@ -42,6 +47,21 @@ The configuration can be customized with the following environment variables:
...
+
+| Variable                                 | Default                     |
+| --                                       | --                          |
+| `CONFIG_NET_IPV6_STATIC_ADDRESS`         | `2001:db8::61`              |
```
chrysn: This is not a suitable default address: Expect that this address is filtered (as per [Section 3 of RFC3849](https://www.rfc-editor.org/rfc/rfc3849#section-3)).

If there is really a need to have an equivalent default to the 10.42.x.x addresses, roll up a ULA -- but really, I think it's better to just err on lack of configured addresses. Yes, that makes build-testing more of a hassle, but in build tests, setting those from the environment is fine.

(I realize this is not just affecting the docs but also the code, but at this point here is where that choice becomes visible and would be documented as intentional.)

ROMemories: > This is not a suitable default address: Expect that this address is filtered (as per [Section 3 of RFC3849](https://www.rfc-editor.org/rfc/rfc3849#section-3)).
> […]
> I think it's better to just err on lack of configured addresses. Yes, that makes build-testing more of a hassle, but in build tests, setting those from the environment is fine.

I totally agree, [that's exactly what I said](https://github.com/ariel-os/ariel-os/pull/915#discussion_r2563932054). I think we should just go back to not having defaults for these.

chrysn: Sorry, I had planned to do a review pass on that today, saw it merged, and went for the docs instead :-)

If this is the only remaining comment, I have no objections to merging it with the defaults in, as that simplifies a follow-up step before the release when the defaults are removed.

ROMemories: > If this is the only remaining comment

It's not :slightly_smiling_face:
And I'd rather do this now, so I opened https://github.com/ariel-os/ariel-os/pull/1571 for that.

bergzand: My intention there is that there's no difference in the expectation between static IPv4 and static IPv6. Given that IPv4 already has a default enabled for the address and gateway, I argued for at least some semblance on the IPv6 side, with the documentation prefix being the safest option in my opinion.

So from my side I'm fine with removing the IPv6 default, if we also remove the default on the IPv4 side.

book/src/networking.md
ROMemories: > Set the expectation that IPv6 _will_ be enabled by default as soon as autoconfig works.

I understand the intent, but the docs state that IPv6 is only "currently" not enabled by default, and there is also a note about the clear intent ("will be supported") to support auto-configuration so I think that makes it clear that we don't see IPv4-only-as-default as the endgame configuration.

My concern with this suggestion is that it is less clear (to me at least) whether us enabling IPv6 would then be a breaking change in the future. And, for once, I think we *should* make it one: enabling IPv6 (while keeping IPv4 enabled) impacts the binary size in a non-negligible way (a couple dozens of kilobytes) so we might want to completely switch from IPv4-only to IPv6-only (maybe that was even your suggestion?).

book/src/networking.md
ROMemories: I mean I wouldn't be the one against removing the IPv4 defaults, but tbh I don't think there is really a symmetry in usage between IPv4 and IPv6 so I'm not convinced that would go against expectations, and anyway that's just properly documented. No users could use the IPv6 defaults for anything, so the docs are a mandatory stop.

book/src/networking.md
bergzand: In short, I'm fine with leaving this out, and thanks for the clarification in #1570.

What I do think we need for the book is an easy way to template these kind of snippets for users to copy and expand (both with cli and `laze.yaml`), in a way that they're not too intrusive in the text itself. Whether that means a collapsible bit in these subsections, or a larger example at the bottom I don't know yet (and might not matter too much). 

My issue here is that we starting to stack multiple levels of indirection which increases the barrier for new users. Here the names of the module and that it must be disabled, a link to the build system with instructions on how to disable *a module* via the cli and the example snippet on how to disable multi-threading. That's multiple steps a (new) user has to go through where the user might not follow or incorrectly replace the example with the specific case. Sprinkling small snippets that can be copied and adapted by users helps them on their way with that, where extra info can be found in the build info section.

What I proposed here doesn't cut it for this so I'm fine with the current state.

book/src/networking.md
chrysn: I'm fine with removing the default from IPv4 as well. (Likely needs a bit of CI massaging to ensure no builds are lacking something set). But I don't think there is an expectation of parity because we do not have parity -- after all, we have the default tools implemented for v4 and for v6; and static configuration *is* more niche in v6 than it is in v4.

bergzand: > I mean I wouldn't be the one against removing the IPv4 defaults, but tbh I don't think there is really a symmetry in usage between IPv4 and IPv6 so I'm not convinced that would go against expectations, and anyway that's just properly documented. No users could use the IPv6 defaults for anything, so the docs are a mandatory stop.

Maybe I'm missing something here, but to me what you describe here for IPv6 is also the case for IPv4. The documentation range must not be routed, but works fine for a local subnet (e.g. the nrf52840dk via USB ethernet to a computer). To me that matches the IPv4 experience, our 'random' `10.42.0.61` address works between my computer and the device, but I don't want to use it beyond that and will have to hit the documentation to figure out how to change it.

book/src/networking.md
chrysn: (Yes, I do recommend going single-stack v6 once we have it, but that's beside the point, just getting it out of the way.)

These things *will* change, and I do not think that it any of those are breaking changes. If a user cares particularly about some setting to be the default, especially in an area where we provide a what-we-think-is-best default, they can make it an explicit choice and select one option and/or deselect the other. Unless they do that, selections are best-effort. (If we do *very* well at some point, that best-effort may even involve looking at the available ROM and expected binary size, but even when not, nothing in the tools we use guarantees firmwares don't get bigger).

We should be transparent about changes to our defaults (like releasnotes:highlight), but let's not go through breaking-change discussions just to enable something that was planned all along.

book/src/networking.md
chrysn: Quoting the RFC:

> This is not a local-use address prefix, and the filters may be used in both local and public contexts.

I read that in that any system may filter out 2001:db8:: messages anywhere, even when no routing is involved. By comparison, an IPv4 user can justifiably expect to add that 10.42.x.x prefix to their home router and things will work.

book/src/networking.md
chrysn: And on the editorial clarity: One thing unclear about the text in the PR is what it means to enable IPv6 -- that is shown as an option in the text, but without any config enabled, what happens?

book/src/networking.md
ROMemories: > I read that in that any system may filter out 2001:db8:: messages anywhere, even when no routing is involved.

That was my reading of this as well and the understanding I had; in insight I should have expanded on this in #915.

book/src/networking.md
ROMemories: > What I do think we need for the book is an easy way to template these kind of snippets for users to copy and expand

Maybe but I don't think it's worth focusing on that right now.

> My issue here is that we starting to stack multiple levels of indirection which increases the barrier for new users.

Honestly there's two very small sections to read in the Build System page, that users *need* to understand anyway, and which are in the natural reading order (the Build System page is right after the Getting Started page, and is linked from it as the next step).

book/src/networking.md
bergzand: Yeah, I also missed that bit in the RFC. I've never had the issue or heard of it being filtered before. RIOT used it as [default subnet](https://github.com/RIOT-OS/RIOT/blob/master/examples/networking/gnrc/border_router/Makefile#L108) for the network examples for as long as I can remember.

ROMemories: Dropped the commit that was documenting the defaults as they were removed in #1571. We're back to no default addresses for IPv6.

book/src/networking.md
bergzand: Another bit of nitpicking

```suggestion
The configuration must be configured with the following environment variables:
```

ROMemories: Got the "must" in but kept "customized" because "configuration […] configured" doesn't read too well imo.

kaspar030 (approved):
lgtm

book/src/networking.md
kaspar030: > until address auto-configuration becomes available.

Reading this, IMO docs should document current state, at least if the time frame is unclear.

kaspar030: > And on the editorial clarity: One thing unclear about the text in the PR is what it means to enable IPv6 -- that is shown as an option in the text, but without any config enabled, what happens?

Right. Enabling `ipv6` defaults to the static configuration. `network-config-ipv6-static` does not select `ipv6` currently.

kaspar030: looks fine now.

PR #1378, opened 2025-09-22
refactor(dhcpv4): enable the `dhcpv4` feature through laze

ROMemories:
# Description

Until now, DHCPv4 was hard-coded as always enabled on `embassy-net` even when when unused. This PR enables DHCPv4 only when the `network-config-ipv4-dhcp` laze module is selected, which is necessary for supporting IPv6 and makes the binary smaller when DHCP is not used.

## How to review this PR

Hiding whitespace helps a bit.

## Testing

Successfully tested with DHCP on espressif-esp32-c6-devkitc-1 and without it on nrf52840dk.

## Issues/PRs references

Extracted from #915.

## Changelog entry

DHCPv4 is now only enabled when the `network-config-ipv4-dhcp` laze module (formerly `network-config-dhcp`) is enabled, instead of always being enabled. This may reduce the size of applications not using DHCP.

## Change checklist

### DIFF
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -22,8 +22,8 @@ embassy-executor = { workspace = true }
 embassy-futures = { workspace = true }
 embassy-hal-internal = { workspace = true }
 embassy-net = { workspace = true, optional = true, features = [
-  "dhcpv4",
   "medium-ethernet",
+  "proto-ipv4",
 ] }
 embassy-sync = { workspace = true }
 embassy-time = { workspace = true, optional = true }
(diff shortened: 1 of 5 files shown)


src/ariel-os-embassy/Cargo.toml
```diff
@@ -22,8 +22,8 @@ embassy-executor = { workspace = true }
 embassy-futures = { workspace = true }
 embassy-hal-internal = { workspace = true }
 embassy-net = { workspace = true, optional = true, features = [
-  "dhcpv4",
   "medium-ethernet",
+  "proto-ipv4",
```
ROMemories, line 26: This will be removed when adding support for IPv6 in #915.

src/ariel-os/Cargo.toml
```diff
@@ -70,6 +70,8 @@ hwrng = ["ariel-os-embassy/hwrng"]
 sensors = ["dep:ariel-os-sensors"]
 
 #! ## Network protocols
+# Enables support for DHCPv4.
+dhcpv4 = ["ariel-os-embassy/dhcpv4"]
```
ROMemories, line 74: That Cargo feature is kept out of the docs on purpose, as it should only be enabled through laze.

kaspar030 (approved):
Tested the default on nucleo-wb55.

PR #1382, opened 2025-09-23
feat(sensors): add variants required for IMUs

ROMemories:
# Description

This adds `Label`, `MeasurementUnit`, and `Category` variants need for adding IMU drivers.

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/label.rs
+++ b/src/ariel-os-sensors/src/label.rs
@@ -24,6 +24,12 @@ pub enum Label {
     AccelerationZ,
     /// Altitude.
     Altitude,
+    /// Angular velocity about the X axis.
+    AngularVelocityX,
+    /// Angular velocity about the Y axis.
+    AngularVelocityY,
+    /// Angular velocity about the Z axis.
+    AngularVelocityZ,
     /// Ground speed.
     GroundSpeed,
     /// Latitude.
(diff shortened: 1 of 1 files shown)


nponsard:
For acceleration you may need the [meter per second squared](https://en.wikipedia.org/wiki/Metre_per_second_squared) unit (m/s^2)

ROMemories:
> For acceleration you may need the [meter per second squared](https://en.wikipedia.org/wiki/Metre_per_second_squared) unit (m/s^2)

No need, we already have `AccelG` (which has a slightly different definition), and that is the unit used by the ST sensors I currently have in mind.

nponsard:
Oh I forgot about this one, that works :+1:

src/ariel-os-sensors/src/label.rs
```diff
@@ -16,8 +16,20 @@
...
 pub enum Label {
+    /// Acceleration along the X axis.
+    AccelerationX,
+    /// Acceleration along the Y axis.
+    AccelerationY,
+    /// Acceleration along the Z axis.
+    AccelerationZ,
     /// Altitude.
     Altitude,
+    /// Angular velocity about the X axis.
```
nponsard, line 27: "about" doesn't sound right to me, is this wording used often ?

ROMemories: Yes, "about" clarifies that the thing is [spinning in place about the axis, not orbiting around it](https://english.stackexchange.com/questions/45566/rotate-about-vs-rotate-around). See for instance [this Wikipedia page about rotation](https://en.wikipedia.org/wiki/Rotation_formalisms_in_three_dimensions).

bergzand (approved):

PR #1388, opened 2025-09-25
refactor(stm32): move the RCC config to a dedicated pure function

ROMemories:
# Description

This is just a small refactor to make future changes to the way the RCC configuration is provided easier, and in particular so that these changes do not constantly result in conflicts as we land support for other STM32 MCUs. Now is a good time to do this as we do not have any in-flight PRs that add support for STM32 MCUs (except for #1183, which I will handle).

## Testing

Successfully tested the `random` example (which is known to require proper clock configuration) on the stm32u083c-dk.

## How to review this PR

This PR is better reviewed commit by commit.

## Issues/PRs references

The aforementioned future changes are in #1389.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1389, opened 2025-09-25
feat(stm32): allow to provide RCC config from OoT applications

ROMemories:
# Description

This builds on #1388 and adds the ability to provide custom RCC configuration *out-of-tree*, through the existing `#[ariel_os::config]` attribute macro. This is particularly important as the clock configuration is not just board-specific but actually application-specific.

## How to review this PR

## Testing

1. Pick an example and a board that require a non-default clock configuration to work: e.g., `udp-echo` on an st-nucleo-wb55.
1. Remove the [hard-coded RCC configuration provided in `rcc_config()`](https://github.com/ariel-os/ariel-os/blob/77da309aff272826353bbde5e96fcbd7bc09ead6/src/ariel-os-stm32/src/lib.rs#L104) for the board.
1. Check that the example doesn't work anymore (as expected).
1. Add `embassy-stm32` as a dependency of the example.
1. Enable the `rcc-config-override` Cargo feature on `ariel-os`.
1. Provide the RCC configuration in the example:
    ```rust
    #[ariel_os::config(rcc)]
    const RCC_CONFIG: embassy_stm32::rcc::Config = {
        use embassy_stm32::rcc::*;

        let mut rcc = embassy_stm32::rcc::Config::new();

        // Copy-paste from `rcc_config()`.

        rcc
    };
    ```
    It is necessary to use `embassy_stm32::rcc::Config::new()` instead of `default()`, because the newly added `new()` is const (this is what required waiting for `embassy-stm32` v0.3.0).
1. Check that the example works again.

## Issues/PRs references

- Depends on #1388
- Depends on #1328
- Depends on #1612 (because of conflicts in the docs)

~~This is currently blocked on us updating to `embassy-stm32` v0.3.0 or newer, as this requires the ability to instantiate the RCC configuration in a `const` context.~~

## Changelog Entry

(STM32) It is now possible to provide an RCC configuration from out-of-tree applications, allowing in particular to adjust the clock configuration to the board and the application.

## Change checklist

### DIFF
--- a/src/ariel-os-macros/src/config.rs
+++ b/src/ariel-os-macros/src/config.rs
@@ -9,10 +9,11 @@
 ///
 /// - The name of the driver the constant provides configuration for.
 ///
-/// | Driver    | Expected type                      | How to enable                                         |
-/// | --------- | ---------------------------------- | ----------------------------------------------------- |
-/// | `network` | `embassy_net::Config`              | See the [book section][network-config-override-book]. |
-/// | `usb`     | `embassy_usb::Config`              | Enable the `override-usb-config` Cargo feature.       |
+/// | Functionality | Expected type                      | How to enable                                         |
+/// | ------------- | ---------------------------------- | ----------------------------------------------------- |
+/// | `network`     | `embassy_net::Config`              | See the [book section][network-config-override-book]. |
+/// | `rcc`         | `embassy_stm32::rcc::Config`       | Enable the `rcc-config-override` Cargo feature.       |
+/// | `usb`         | `embassy_usb::Config`              | Enable the `override-usb-config` Cargo feature.       |
 ///
 /// # Note
 ///
--- a/src/ariel-os/Cargo.toml
+++ b/src/ariel-os/Cargo.toml
@@ -126,6 +126,7 @@ usb-hid = ["ariel-os-embassy/usb-hid"]
 #! configuration is taken into account.
 # Enables custom network configuration.
 network-config-override = ["ariel-os-embassy/network-config-override"]
+## Enables custom RCC configuration (currently only available on STM32).
 rcc-config-override = ["ariel-os-embassy/rcc-config-override"]
 ## Enables custom USB configuration.
 override-usb-config = ["ariel-os-embassy/override-usb-config"]


src/ariel-os/Cargo.toml
```diff
@@ -119,6 +119,7 @@ usb-hid = ["ariel-os-embassy/usb-hid"]
 ble-config-override = ["ariel-os-embassy/ble-config-override"]
 ## Enables custom network configuration.
 network-config-override = ["ariel-os-embassy/network-config-override"]
+rcc-config-override = ["ariel-os-embassy/rcc-config-override"]
```
ROMemories, line 130: This will need to be documented.
Edit: we might want to wait to decide whether to document this.
Edit: if we don't document it that way, it should also *not* appear in the macro suggestions in case of typo.

src/ariel-os-macros/src/config.rs
```diff
@@ -9,10 +9,11 @@
...
 /// - The name of the driver the constant provides configuration for.
 ///
-/// | Driver    | Expected type                      | How to enable                                         |
-/// | --------- | ---------------------------------- | ----------------------------------------------------- |
-/// | `network` | `embassy_net::Config`              | See the [book section][network-config-override-book]. |
-/// | `usb`     | `embassy_usb::Config`              | Enable the `override-usb-config` Cargo feature.       |
+/// | Functionality | Expected type                      | How to enable                                         |
+/// | ------------- | ---------------------------------- | ----------------------------------------------------- |
+/// | `network`     | `embassy_net::Config`              | See the [book section][network-config-override-book]. |
+/// | `rcc`         | `embassy_stm32::rcc::Config`       | Enable the `rcc-config-override` Cargo feature.       |
```
kaspar030, line 15: Im wondering. This needs a very specific format. Would it make sense to clarify that at least a bit in its name? (calll it `stm32-rcc?` / `stm32-rcc-config-override` )?

ROMemories: I'm not sure (what do you mean by specific format? Do you mean a type from a specific HAL?), I think we could reuse that config setting for other HALs (e.g., RP) if needed, by making the expected type depend on the current HAL, and because they are of course mutually exclusive, this makes sure only one is enabled at a time.

kaspar030: > what do you mean by specific format?

I actually meant "type", but yeah, it is not that specific - it depends on the HAL.

ok, let's keep as is.

PR #1392, opened 2025-09-26
docs(book): fix up book

jkr-ariel:
# Description

This PR addresses the last few comments left on #1327 after being merged.

## Issues/PRs references

Follow up to #1327.

## Change checklist

### DIFF
--- a/book/src/hardware-functionality-support.md
+++ b/book/src/hardware-functionality-support.md
@@ -1,3 +1,5 @@
+{{#title Hardware & Functionality Support Overview - Ariel OS manual}}
+
 # Hardware & Functionality Support
 
 > The tables below indicate whether we support using the piece of functionality in a portable manner, through an abstraction layer and platform-aware configuration.


book/src/hardware-functionality-support.md
```diff
@@ -1,3 +1,5 @@
+{{#title Hardware & Functionality Support Overview}}
```
ROMemories: Generated titles have the book's title in them; can you check whether mdbook exposes it in some way so we don't have to hard-code it as below?
```suggestion
{{#title Hardware & Functionality Support Overview - Ariel OS manual}}
```

jkr-ariel: It does not seem like the book title is exposed directly, but it is possible to write a custom preprocessor that replaces `{{book_title}}` (for example) with the book title. Though I don't think this is really needed in this case.

ROMemories: Yeah I wasn't aware of any way to do it, but it might have changed. Agree it's not worth the trouble, so you can just the suggestion in (as I think this still needs to be consistent) and this should be good to go after that.

book/src/hardware-functionality-support.md
```diff
@@ -1,5 +1,9 @@
+{{#title Hardware & Functionality Support Overview - Ariel OS manual}}
+
 # Hardware & Functionality Support
 
+{{book_title}}
+
```
ROMemories:
```suggestion
```

ROMemories (approved):

PR #1393, opened 2025-09-26
docs(book): remove `mdbook-cmdrun`

jkr-ariel:
# Description

This PR removes the dependency on `mdbook-cmdrun` for the ArielOS book. This basically means going back to the setup that we had before #795 where auto-generated files were commited on the repo and checked by CI to make sure they are up to date.

This choice was motivated by the fact that #1327 already introduced commands that generate files to be commited on the repo. Also this enables us to get rid of the complexity of the `cmdrun-check.sh` script used by CI.

## Issues/PRs references

Linked to #1327.
Linked to #795.

## Change checklist


book/src/boards/index.md.tmpl
```diff
@@ -0,0 +1,25 @@
+# Board Pages
+
+## Tier 1
+
+Tier 1 hardware gets regularly tested, either automatically or manually.
+
+{%- for board_technical_name, board in tier1_boards %}
```
ROMemories: I would prefer to have a blank line before the vertical list, so that this is valid "Markdown Classic". The following should avoid whitespace to be trimmed before the list:

```suggestion
{% for board_technical_name, board in tier1_boards %}
```

ROMemories: Same for the other lists below

book/src/support_matrix_tier1.html
ROMemories, line 1: Can you declare these files as generated files, as you've done in #1327?

ROMemories (commented):
Otherwise the output looks identical to me (which is what we want), so after these this should be good to go!

ROMemories:
I suppose it would also be possible to re-enable the printing feature? If so, that can be done in a follow-up, no need to do this in this PR.

book/src/boards/index.md.tmpl
jkr-ariel: There is already a blank line after the paragraph which should not get trimmed.

.github/workflows/main.yml
```diff
@@ -436,21 +436,17 @@ jobs:
...
+      - name: Check the board index page markdown generation exit code
         run: |
-          ./doc/gen_book.rs board-list doc/support_matrix.yml /dev/null --tier 1
-          ./doc/gen_book.rs board-list doc/support_matrix.yml /dev/null --tier 2
-          ./doc/gen_book.rs board-list doc/support_matrix.yml /dev/null --tier 3
+          ./doc/gen_book.rs board-index doc/support_matrix.yml /dev/null --template-path book/src/boards/index.md.tmpl
 
-      - name: Install mdbook and mdbook utils
+      - name: Check the board pages generation exit code
+        run: ./doc/gen_book.rs board-pages doc/support_matrix.yml /tmp/
```
ROMemories: Isn't this redundant with the `board-pages` up-to-date check above?

jkr-ariel: It is. Fixed now

PR #1394, opened 2025-09-26
feat(sensors-api): add opaque sample values

nponsard:
# Description

This PR adds opaque sample values, those values are meant to be read using a custom trait provided by the sensor driver or a library.

## Issues/PRs references

- Depends on #1399

## Open Questions

I just realized while preparing this PR that going this way would hurt the discoverability aspect of the sensors api. These changes would be to make opaque the storage of a timestamp coming from the nrf91 GNSS API.

Is this a good idea ? 

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/label.rs
+++ b/src/ariel-os-sensors/src/label.rs
@@ -38,6 +38,8 @@ pub enum Label {
     Longitude,
     /// Used for sensor drivers returning a single [`Sample`](crate::sensor::Sample).
     Main,
+    /// Opaque channel: the associated sample is intended for the sensor driver only, and no guarantees are provided.
+    Opaque,
     /// Relative humidity.
     RelativeHumidity,
     /// Heading.
@@ -68,6 +70,7 @@ impl core::fmt::Display for Label {
             Self::Latitude => write!(f, "Latitude"),
             Self::Longitude => write!(f, "Longitude"),
             Self::Main => write!(f, ""),
+            Self::Opaque => write!(f, "[opaque]"),
             Self::RelativeHumidity => write!(f, "Relative humidity"),
             Self::Heading => write!(f, "Heading"),
             Self::Temperature => write!(f, "Temperature"),


ROMemories:
Can you split this into two different PRs? Exposing the sensor driver from the `Samples` is unrelated to having opaque `Sample`s, even though that what motivates it.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -121,6 +208,18 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             /// [`Iterator::zip()`] can be useful to zip the returned iterator with the one
             /// obtained with [`Reading::samples()`].
             pub fn iter(&self) -> impl ExactSizeIterator<Item = ReadingChannel> + core::iter::FusedIterator + '_ {
+                let iter = self.iter_raw()
+                    .filter(|channel| channel.label() != Label::Opaque);
+
+                let cloned = self.iter_raw()
+                    .filter(|channel| channel.label() != Label::Opaque);
+
+                let size = cloned.count();
```
ROMemories: Untested, but it seems simpler to use [`Iterator::copied()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.copied):
```suggestion
                let size = iter.copied().count();
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -89,9 +167,18 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             }
 
             fn samples(&self) -> impl ExactSizeIterator<Item = Sample> + core::iter::FusedIterator {
-                match self.samples {
-                    #(#samples_iter),*
-                }
+                let size = self.sensor().reading_channels().iter().len();
+                let iter = self.unfiltered_samples()
+                    .enumerate()
+                    .filter(move |(i,_sample)| {
```
ROMemories: Would it be possible/simpler to use `filter_map()` here?

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -66,6 +70,54 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+
+        /// Avoid external implementations of [`SensorAccess`].
+        mod private {
+            use super::Samples;
+            pub trait Sealed {}
+
+            impl Sealed for Samples {}
+        }
+
+        #[derive(Copy, Clone, Debug)]
```
ROMemories: I'm not sure what the semantics of deriving `Copy` and `Clone` here are. Shouldn't that depend on whether the underlying Iterator is `Copy` and/or `Clone`?

src/ariel-os-sensors/src/label.rs
```diff
@@ -26,6 +26,8 @@ pub enum Label {
     Longitude,
     /// Used for sensor drivers returning a single [`Sample`](crate::sensor::Sample).
     Main,
+    /// Opaque
```
ROMemories:
```suggestion
    /// Opaque channel: the associated sample is intended for the sensor driver only, and no guarantees are provided.
```

src/ariel-os-sensors/src/label.rs
```diff
@@ -50,6 +52,7 @@ impl core::fmt::Display for Label {
             Self::Latitude => write!(f, "Latitude"),
             Self::Longitude => write!(f, "Longitude"),
             Self::Main => write!(f, ""),
+            Self::Opaque => write!(f, "Opaque"),
```
ROMemories: It's unclear how these values are to be used but I suppose adding brackets clarifies that this is a meta-value:
```suggestion
            Self::Opaque => write!(f, "[opaque]"),
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -66,6 +70,54 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+            fn next(&mut self) -> Option<Self::Item> {
+                self.iter.next()
+            }
+            fn size_hint(&self) -> (usize, Option<usize>) {
+                (self.size, Some(self.size))
+            }
+        }
+
+        impl<I: Iterator> ExactSizeIterator for SizedIterator<I> {}
+        impl<I: Iterator + core::iter::FusedIterator> core::iter::FusedIterator for SizedIterator<I> {}
```
ROMemories: `Iterator` is already a super trait of `FusedIterator`:

```suggestion
        impl<I: core::iter::FusedIterator> core::iter::FusedIterator for SizedIterator<I> {}
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -66,6 +70,54 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+
+        /// Return all samples of a sensor, without filtering out opaque channels. For driver implementors only.
+        pub trait UnfilteredSamples {
```
ROMemories: I think we could call it that: the other, filtered trait is called `Reading`, and this is in a way a private interface, in that there is no guarantee about provided to driver users.

```suggestion
        pub trait PrivateReading {
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -66,6 +70,54 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+            }
+            fn size_hint(&self) -> (usize, Option<usize>) {
+                (self.size, Some(self.size))
+            }
+        }
+
+        impl<I: Iterator> ExactSizeIterator for SizedIterator<I> {}
+        impl<I: Iterator + core::iter::FusedIterator> core::iter::FusedIterator for SizedIterator<I> {}
+
+        /// Return all samples of a sensor, without filtering out opaque channels. For driver implementors only.
```
ROMemories: I think this documentation could be a bit more aligned with that of `Reading`, and make it very clear that the trait is only intended to be both implemented *and used* only by driver implementors.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
ROMemories:
```suggestion
                    .filter(move |(i, _sample)| {
```

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
nponsard: According to the documentation, `.copied()` is an equivalent of 

```rust 
.map(|&x| x) 
```

So this doesn't copy the whole iterator, only its elements as the iterator advances. Meaning if I call `.count()` on it, all the elements of the iterator will be consumed and the iterator returned by the function will only be an iterator with no more values.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
nponsard: On structs that are generic it will implement `Clone` and/or `Copy` only if the generic type has `Clone` and/or `Copy` respectively.
https://doc.rust-lang.org/std/clone/trait.Clone.html#derivable

> For a generic struct, #[derive] implements Clone conditionally by adding bound Clone on generic parameters.

So this sounds like it already does what we want.

nponsard:
Applied suggestions and rebased on main

src/ariel-os-sensors/src/label.rs
ROMemories, line 1: After some offline discussion, we've concluded that it would be simpler to keep only the changes to this file; i.e., to not transparently filter samples from opaque channels. The reasoning is that, even though that puts the burden of filtering these samples on the users, sensor drivers that have opaque channels should be relatively rare (the only candidates currently are GNSS drivers) and this makes the API and its documentation simpler. In addition, #1399 has made access to the sensor driver instance from the `Samples` possible, and it's therefore much easier to obtain the information of whether a sample is from an opaque from a `Samples` than it was when this PR was started.

We'll add a `Samples::channels()` method to make this easier, but this is better done in another PR. 
We might also consider adding a `MeasurementUnit::Opaque` unit later (to avoid breaking changes for sensor drivers), but this is better done in a separate PR as well, if at all.

ROMemories: > We'll add a `Samples::channels()` method to make this easier, but this is better done in another PR.

Actually now we should instead be able make `sample()` and `samples()` return tuples of `(ReadingChannel, Sample)`. I'll take care of this.

nponsard: Returning tuples sounds good

nponsard:
This PR is now much simpler, the `Opaque` values are not hidden from the user anymore.
Rebased on main

PR #1398, opened 2025-09-29
feat(nrf): add support for nRF91 modem

nponsard:
# Description

This PR adds the initialization of the GNSS and cellular modem found in the nRF91 SiP series.

Version `0.8.0` of `nrf-modem` requires `embassy-time` version `0.5.0` that is incompatible with the current embassy version in Ariel OS 

## Open Questions

This PR includes a fix for [errata #7](https://docs.nordicsemi.com/bundle/errata_nRF9151_Rev2/page/ERR/nRF9151/Rev2/latest/anomaly_151_7.html). This workaround resets the board, and since version `v0.28` the behavior changed to catch resets as exceptions (previously `--catch-reset`) by default. This means that probe-rs will exit instantly when the workaround is triggered.
A fix for that would be to add `--no-catch-reset` to the default `probe-rs run` command, but this will make older (<v0.28) versions of `probe-rs` incompatible as it will not recognize this flag.

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,8 +82,11 @@ jobs:
         with:
           toolchain: ${{ steps.get_toolchain.outputs.toolchain }}
           # Required for checking ariel-os-esp and ariel-os-nrf
-          targets: riscv32imac-unknown-none-elf, thumbv7em-none-eabi
-          components: clippy, rustfmt
+          targets: riscv32imac-unknown-none-elf, thumbv7em-none-eabi, thumbv8m.main-none-eabihf
+          components: clippy, rustfmt, llvm-tools
+
+      - name: Install prerequisites
+        run: sudo apt-get install gcc-arm-none-eabi
 
       - name: rust cache
         uses: Swatinem/rust-cache@98c8021b550208e191a6a3145459bfc9fb29c4c0 # v2
@@ -200,7 +203,7 @@ jobs:
             --deny warnings
 
       - run: echo 'RUSTFLAGS=--cfg context="nrf52840" --cfg context="nrf52" --cfg getrandom_backend="custom"' >> $GITHUB_ENV
-      - name: clippy for nRF
+      - name: clippy for nRF52
         uses: clechasseur/rs-clippy-check@286fd9545eb133adad54b4af34bcb642e266ddd1 # v5
         with:
           args: |
@@ -217,6 +220,25 @@ jobs:
                 "
             -p ariel-os-nrf
             --target=thumbv7em-none-eabi
+            --
+            --deny warnings
+      - run: echo 'RUSTFLAGS=--cfg context="nrf9160" --cfg context="nrf91" --cfg getrandom_backend="custom"' >> $GITHUB_ENV
+      - name: clippy for nRF91
+        uses: clechasseur/rs-clippy-check@286fd9545eb133adad54b4af34bcb642e266ddd1 # v5
+        with:
+          args: |
+            --verbose
+            --locked
+            --features "
+                embassy-nrf/nrf9160-s,
+                external-interrupts,
+                i2c,
+                spi,
+                uart,
+                nrf91-modem
+                "
+            -p ariel-os-nrf
+            --target=thumbv8m.main-none-eabihf
             --
             --deny warnings
 


laze-project.yml
```diff
@@ -1247,6 +1249,26 @@ modules:
...
       - doc-only
 
+  - name: gnss
+    selects:
+      - hw/gnss
+    env:
+      global:
+        FEATURES:
+          - ariel-os/gnss
+
```
ROMemories: I think this is unrelated:
```suggestion
```

laze-project.yml
```diff
@@ -1247,6 +1249,26 @@ modules:
...
+  - name: has_nrf_modem
+    selects:
+      - doc-only
+
+  - name: nrf91-modem
+    selects:
+      - has_nrf_modem
```
ROMemories: This should either have `91` in its name, or the `nrf91-modem` shouldn't.

laze-project.yml
```diff
@@ -310,13 +310,15 @@ contexts:
     parent: nrf91
     provides:
       - has_storage_support
+      - has_nrf_modem
```
```suggestion
      - has_nrf_modem
      - has_storage_support
```

laze-project.yml
```diff
@@ -310,13 +310,15 @@ contexts:
...
     env:
       PROBE_RS_CHIP: nRF9160_xxAA
 
   - name: nrf9151
     parent: nrf91
     provides:
       - has_storage_support
+      - has_nrf_modem
```
```suggestion
      - has_nrf_modem
      - has_storage_support
```

Cargo.toml
```diff
@@ -157,6 +157,10 @@ trouble-host = "0.1"
 bt-hci = { version = "0.2.0" }
 nrf-sdc = { git = "https://github.com/ariel-os/nrf-sdc.git", rev = "8ce4780b691ec6850b05a0382cef395846b67e02" }
 
+nrf-modem = { version = "0.7.3" }
+tinyrlibc = "0.5.0"
+futures = { version = "0.3.31", default-features = false }
```
ROMemories: Do we need this? I don't see it being used.

src/ariel-os-rt/build.rs
```diff
@@ -74,6 +74,8 @@ fn write_memoryx() {
         (512, 1024)
     } else if context("nrf5340-net") {
         (64, 256)
+    } else if cfg!(feature = "nrf91-modem") {
+        (192, 1024)
```
ROMemories: Can you point me to where these values are coming from? I can't find them easily in the datasheet.

src/ariel-os/Cargo.toml
```diff
@@ -157,6 +157,9 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
 # Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
 ble-central = ["ble", "ariel-os-embassy/ble-central"]
 
+# ## Modem support for the nrf91xx SiP (nrf-modem)
```
```suggestion
# ## Enables support for the nRF91xx SiP modem.
```

src/ariel-os-nrf/Cargo.toml
```diff
@@ -102,10 +109,16 @@ ble = [
...
+
 ## Enables defmt support.
-defmt = ["dep:defmt", "embassy-nrf/defmt", "nrf-sdc?/defmt"]
+defmt = ["dep:defmt", "embassy-nrf/defmt", "nrf-sdc?/defmt", "nrf-modem?/defmt"]
 
 ## Enables the interrupt executor.
-executor-interrupt = ["embassy-executor/executor-interrupt"]
+executor-interrupt = [
+  "embassy-executor/executor-interrupt",
+  "nrf-modem?/os-irq",
```
ROMemories, line 124: Would it make sense to add a comment explaining why `os-irq` is needed? I couldn't easily find documentation about it.

src/ariel-os-nrf/Cargo.toml
```diff
@@ -32,7 +32,12 @@ ariel-os-random = { workspace = true, optional = true }
 ariel-os-rt = { workspace = true, features = ["memory-x"] }
 static_cell = { workspace = true }
 embassy-sync = { workspace = true }
+embassy-time = { workspace = true, optional = true }
```
ROMemories: This doesn't seem to be used?
```suggestion
```

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+        NVIC::unmask(Interrupt::IPC);
+    }
+
+    let system_mode = SystemMode {
+        lte_support: true,
+        lte_psm_support: true,
+        nbiot_support: true,
+        gnss_support: true,
+        preference: ConnectionPreference::None,
+    };
```
ROMemories, line 125: We'll need to be able to configure this, but this doesn't have to be part of this PR. Would disabling some of these decrease the power consumption if unused?

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+        spu.periphid(4).perm().write(|w| w.set_secattr(false));
+        // Set clock and power access registers to nonsecure
+        spu.periphid(5).perm().write(|w| w.set_secattr(false));
+        // Set IPC access register to nonsecure
+        spu.periphid(42).perm().write(|w| w.set_secattr(false));
+        ipc_start
+    }
+    let ipc_start = configure_modem_non_secure();
+
+    unsafe {
```
ROMemories: This needs a SAFETY comment.

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    uicr_hfxo_workaround();
+
+    // from https://github.com/diondokter/nrf-modem/issues/31
+    // allows us to use the modem in S mode
+    fn configure_modem_non_secure() -> u32 {
+        // The RAM memory space is divided into 32 regions of 8 KiB.
+        // Set IPC RAM to nonsecure
+        const SPU_REGION_SIZE: u32 = 0x2000; // 8kb
+        const RAM_START: u32 = 0x2000_0000; // 256kb
```
ROMemories:
```suggestion
        const SPU_REGION_SIZE: u32 = 0x2000; // 8 KiB
        const RAM_START: u32 = 0x2000_0000; // 256 KiB
```

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+        let ipc_start: u32 = unsafe { &_MODEM_start as *const u8 } as u32;
+        let ipc_reg_offset = (ipc_start - RAM_START) / SPU_REGION_SIZE;
+        let ipc_reg_count = (unsafe { &_MODEM_length as *const u8 } as u32) / SPU_REGION_SIZE;
```
ROMemories: These need `SAFETY` comments. It's insta UB to create invalid references.

src/ariel-os-nrf/src/modem.rs
ROMemories: I think that can be read from linker symbols.

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+        let spu = embassy_nrf::pac::SPU;
+        let range = ipc_reg_offset..(ipc_reg_offset + ipc_reg_count);
+        debug!(
+            "marking region as non secure: {}..{}",
+            range.start, range.end
+        );
+        for i in range {
+            spu.ramregion(i as usize).perm().write(|w| {
+                w.set_execute(true);
```
ROMemories: I don't think the IPC RAM has to be executable:
```suggestion
                w.set_execute(false);
```

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+                w.set_write(true);
+                w.set_read(true);
+                w.set_secattr(false);
+                w.set_lock(false);
+            })
+        }
+
+        // Set regulator access registers to nonsecure
+        spu.periphid(4).perm().write(|w| w.set_secattr(false));
```
ROMemories: Can you define constants for these peripheral IDs? I think these are from Table 4 from the nRF9151 datasheet.

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+
+impl typelevel::Handler<typelevel::IPC> for InterruptHandler {
+    unsafe fn on_interrupt() {
+        nrf_modem::ipc_irq_handler();
+    }
+}
+
+bind_interrupts!(struct Irqs{
+    IPC => InterruptHandler;
+});
```
ROMemories: I think this we simplified by using the `#[interrupt]` macro attribute (from `embassy-nrf`):
```rust
#[interrupt]
#[expect(non_snake_case)]
fn IPC() {
    nrf_modem::ipc_irq_handler();
}
```
as is [done here](https://github.com/diondokter/nrf-modem/issues/31#issuecomment-2637010536).

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+    },
+};
+use nrf_modem::{ConnectionPreference, MemoryLayout, SystemMode};
+// need to make the symbols available so nrf_modem can link against them
+extern crate tinyrlibc as _;
+
+use ariel_os_debug::log::debug;
+
+#[cfg(feature = "executor-interrupt")]
+use cortex_m::interrupt::InterruptNumber;
```
ROMemories: That was otherwise confusing to me as it's not obvious that's a trait:
```suggestion
use cortex_m::interrupt::InterruptNumber as _;
```

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+});
+
+unsafe extern "C" {
+    static _MODEM_start: u8;
+    static _MODEM_length: u8;
+}
+
+// Workaround used in the nrf mdk: file system_nrf91.c , function SystemInit(), after `#if !defined(NRF_SKIP_UICR_HFXO_WORKAROUND)`
+// Use this until commit `3e2b23d2f453d10324896484f9d045d2821bd567` is included in the embassy-nrf version we use.
+fn uicr_hfxo_workaround() {
```
ROMemories, line 31: I think this needs to be updated with the upstream version; I agree with the upstream recommendation to use `cortex_m::interrupt::free()`.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -153,6 +153,11 @@ pub(crate) fn init() {
 
     #[cfg(any(context = "nrf", context = "rp", context = "stm32"))]
     {
+        #[cfg(feature = "nrf91-modem")]
+        {
+            use crate::hal::interrupt::{InterruptExt, Priority};
```
ROMemories:
```suggestion
            use crate::hal::interrupt::{InterruptExt as _, Priority};
```

src/ariel-os-nrf/src/modem.rs
```diff
@@ -0,0 +1,157 @@
...
+        nbiot_support: true,
+        gnss_support: true,
+        preference: ConnectionPreference::None,
+    };
+
+    let memory_layout = MemoryLayout {
+        base_address: ipc_start,
+        tx_area_size: 0x2000,
+        rx_area_size: 0x2000,
+        trace_area_size: 0x1000,
```
ROMemories: Where do these numbers come from?

src/ariel-os-rt/build.rs
nponsard: This is to reserve 64k of ram to communicate between the modem and the processor.
The [nrf-modem readme](https://github.com/diondokter/nrf-modem/tree/master?tab=readme-ov-file#memory) shows the memory setup. I've set 64K when troubleshooting, this reminds me that I need to try with 32K, see if it still works.

src/ariel-os-nrf/src/modem.rs
nponsard: I cannot find documentation on how these settings impact performance and power consumption.

Cargo.toml
nponsard: I removed the `futures` dependency

src/ariel-os-nrf/src/modem.rs
nponsard: Tried using the GNSS feature of the modem with execute disabled in the memory region and didn't see any changes. Seems like it is indeed unnecessary to mark the region as execute.

src/ariel-os-nrf/src/modem.rs
nponsard: Here RAM_START represents the start of the physical RAM where the `_RAM_start` defined in the linker symbols represent the start of the usable RAM for Ariel OS. When the modem is used, the IPC is reserved at the start of the physical RAM, making the usable RAM start with an offset.

nponsard: If we make sure that the IPC memory is always reserved at the start we can remove this constant altogether and make `ipc_reg_offset` a constant with the value 0.

src/ariel-os-nrf/src/modem.rs
nponsard: Those are the default sizes, I changed the struct creation to use `Default::default()`

src/ariel-os-rt/build.rs
nponsard: Set it back to 32KiB and it's working, I've also moved this to a constant.

src/ariel-os-nrf/Cargo.toml
nponsard: Added a comment trying to quickly explain why this feature is needed.

src/ariel-os-nrf/src/modem.rs
nponsard: Seems like the interrupt gets enabled by the `#[interrupt]` macro so this is not needed anymore (removed it)

src/ariel-os-nrf/src/modem.rs
chrysn: Which we enable will also depend on the firmware that is loaded, so I expect to do some of the follow-up work with many `false`s in there.

src/ariel-os-nrf/src/modem.rs
ROMemories: Actually, I don't think these need to go through references at all, see how the same thing is done [in the storage API](https://github.com/ariel-os/ariel-os/blob/d415dad08931ae5154037ff5a0073aa0cc2b2782/src/ariel-os-storage/src/lib.rs#L53-L59).

nponsard:
> Can this be tested (even minimally) without relying on #1330 or #1148?

You can test the initialization by enabling the `nrf91-modem` module in any example: 

```sh
laze -C examples/blinky build -DLOG=trace -b nrf9160dk-nrf9160 -s nrf91-modem run 
```

nponsard:
> Actually, I don't think these need to go through references at all, see how the same thing is done [in the storage API](https://github.com/ariel-os/ariel-os/blob/d415dad08931ae5154037ff5a0073aa0cc2b2782/src/ariel-os-storage/src/lib.rs#L53-L59).

Changed to the way it's done in the storage API and it seems like it didn't even change the output binary :+1:

ROMemories:
I think this requires a new compile-time dependency that needs to be documented. On Debian I otherwise get the following compilation errors:

```
warning: nrf-modem@0.7.3: wrapper/nrf_modem_os_log.c:1:10: fatal error: stdio.h: No such file or directory
warning: nrf-modem@0.7.3:     1 | #include <stdio.h>
warning: nrf-modem@0.7.3:       |          ^~~~~~~~~
warning: nrf-modem@0.7.3: compilation terminated.
```

```
warning: tinyrlibc@0.5.0: ./src/snprintf.c:18:10: fatal error: string.h: No such file or directory
warning: tinyrlibc@0.5.0:    18 | #include <string.h>
warning: tinyrlibc@0.5.0:       |          ^~~~~~~~~~
warning: tinyrlibc@0.5.0: compilation terminated.
```

Installing `libnewlib-arm-none-eabi` was sufficient to make it compile and make the following command work in hardware:

```sh
laze -C examples/blinky build -DLOG=trace -b nrf9160dk-nrf9160 -s nrf91-modem run 
```

nponsard:
> I think this requires a new compile-time dependency that needs to be documented. On Debian I otherwise get the following compilation errors:

I tried on debian 13 (docker container), I didn't get the same error, instead my error was 
```
Could not find `llvm-objcopy`. Is it installed? Use `rustup component add llvm-tools` to install it or select the `arm-none-eabi-objcopy` feature if you have that tool installed.: NotFound
```
Running the suggested command (`rustup component add llvm-tools`) allowed the build command to succeed.

So there is indeed a new dependency that needs to be installed, possibly by `laze build install-toolchain`.

nponsard:
I could not reproduce the `fatal error: stdio.h: No such file or directory` on debian 12 or debian 13 containers when following the Getting Started section of the book to install dependencies.
I added the installation of the `llvm-tools` component in hte `install-toolchain` script.

src/ariel-os-rt/Cargo.toml
```diff
@@ -56,6 +56,7 @@ _panic-handler = []
 single-core = ["cortex-m/critical-section-single-core"]
 multi-core = ["embassy-rp/critical-section-impl"]
 memory-x = ["dep:ld-memory"]
+nrf91-modem = []
```
ROMemories, line 59: Unfortunate that we need something so specific in this runtime crate, but I don't see a better option for now.

src/ariel-os-nrf/Cargo.toml
```diff
@@ -106,10 +111,17 @@ ble = [
 ble-peripheral = ["ble", "nrf-sdc/peripheral"]
 ble-central = ["ble", "nrf-sdc/central"]
 
+## Enables modem support.
+nrf91-modem = ["dep:cortex-m", "dep:nrf-modem", "dep:tinyrlibc"]
```
ROMemories, line 115: Actually we should try to have this feature enabled for Clippy in CI.

ROMemories (approved):
The new Clippy check seems to be adding a minute of runtime to the check, which is non-zero but still manageable.

PR #1401, opened 2025-09-30
docs(book): enable print button

jkr-ariel:
# Description

This PR re-enables the print button in the top bar of the ArielOS book, which had been disabled in #1327. In order to fix issues with broken links in the `print.html` page, bumping `mdbook` to [v0.5.0-alpha.1](https://github.com/rust-lang/mdBook/releases/tag/v0.5.0-alpha.1) was necessary.

## Issues/PRs references

Linked to #1327.

## Change checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -139,7 +139,7 @@ jobs:
       - name: Install mdbook
         uses: taiki-e/install-action@ea0eda622640ac23a17ba349cf09e2709d58f5e1 # v2
         with:
-          tool: mdbook
+          tool: mdbook@0.5.0-alpha.1
 
       - name: Build the book
         run: |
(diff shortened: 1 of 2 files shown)


chrysn:
Is this even a desirable feature?

jkr-ariel:
> Is this even a desirable feature?

Well, it is a feature enabled by default in `mdbook`. That being said, if the bump to 0.5.0-alpha.1 causes friction, I agree we should keep this feature disabled.

chrysn:
There's some saying about thinking about whether we can and whether we should…

My point more precisely is: Do we want to encourage printing this, and can we claim that this is a text that is in a good shape to be printed, with all the considerations for printing it? (For comparison, RIOT's doxygen is well capable of producing man pages and PDFs, but neither are good forms for using it). I'd answer "no" to both.

jkr-ariel:
The main point of this PR is to bring back the book in the state it was before #1327 (regarding the book printing option) without assessing if it should have been enabled in the first place. I don't have a strong opinion on this and I am fine with keeping it disabled. Adding @ROMemories to this dicussion.

PR #1403, opened 2025-09-30
docs(support-matrix): render comments in support matrix

jkr-ariel:
# Description

This PR renders comments from `doc/support_matrix.yml` in the ArielOS book.

## Issues/PRs references

Part of #1182.

## Change checklist

### DIFF
--- a/doc/gen_book.rs
+++ b/doc/gen_book.rs
@@ -382,7 +382,7 @@ struct FunctionalitySupport {
     title: String,
     icon: String,
     description: String,
-    // TODO: add comments
+    comments: Option<Vec<String>>,
     // TODO: add the PR link
 }
 
@@ -502,6 +502,7 @@ fn gen_functionalities(matrix: &schema::Matrix) -> Result<Vec<BoardSupport>, Err
                     title: functionality_info.title.to_owned(),
                     icon: support_key.icon.to_owned(),
                     description: support_key.description.to_owned(),
+                    comments: todo!(),
                 })
             });
             let errors = functionalities


jkr-ariel:
The comments in `doc/support_matrix.yml` are rendered as footnotes in each board page. It is not the case for the "Overview" page because we would need to rewrite the table in Markdown in order to use the footnote syntax which doesn't seem possible if we want to keep it as it is. Another way is to replicate the HTML generated by the footnote Markdown, but it is not trivial.

ROMemories:
Can you extract the commit that rewrites the HTML tables as Markdown tables into a separate PR? This seems easy to do as it is already atomic.

book/templates/board-page.md.tmpl
```diff
@@ -13,30 +13,13 @@
...
-      <td>{{ functionality.title }}</td>
-      <td class="support-cell" title="{{ functionality.description }}">{{ functionality.icon }}</td>
-    </tr>
-    {%- endfor %}
-  </tbody>
-</table>
+|Functionality|Support Status|
+|---|:---:|
+{%- for functionality in board_info.functionalities %}
+|{{ functionality.title }}|<span title="{{ functionality.description }}">{{ functionality.icon }}</span>{%- if functionality.comments %}{%- for comment in functionality.comments %}[^{{ comment }}]{%- endfor %}{%- endif %}|
```
ROMemories: Instead of using the raw comment as the URL fragment identifier, I think it should be slugified. It seems that minijinja doesn't provide such a template filter, but it seems [easy to provide a custom one](https://docs.rs/minijinja/latest/minijinja/filters/index.html#custom-filters). Instead of implementing slugification as in the example, it would likely be better to use the [`slug` crate](https://docs.rs/slug/latest/slug/) which makes sure only the ASCII subset of characters is used.

book/templates/board-page.md.tmpl
```diff
@@ -50,3 +33,9 @@ dt, dd {
   display: inline;
 }
 </style>
+
+{% for functionality in board_info.functionalities %}
+  {%- for comment in functionality.comments %}
+[^{{ comment }}]: {{ comment }}
```
ROMemories: Can we add a period at the end of the footnote, and capitalize the first character? The comments should be kept as-is in the data file, so that they can easily be embedded in different contexts, and we can format them as we want here. For capitalization I think we need another custom filter, that only capitalizes the first letter (minijinja provides [`capitalize`](https://docs.rs/minijinja/2.12.0/minijinja/filters/fn.capitalize.html) but it converts all other characters to lowercase, which we don't want).

jkr-ariel:
Also, I noticed that there are some URLs in the comments, so they are rendered as text, with no clickable link. Should I add an intermediate step where I scan the comment for URLs, and transform them into links?

book/templates/board-page.md.tmpl
```diff
@@ -33,3 +33,9 @@ dt, dd {
   display: inline;
 }
 </style>
+
```
ROMemories: Otherwise we get an extra blank line:
```suggestion
```

doc/gen_book.rs
```diff
@@ -324,6 +326,8 @@ impl SubCommandPages {
 
         for board in &board_info {
             let mut env = Environment::new();
+            env.add_filter("slugify", slugify::<String>);
+            env.add_filter("capitalize2", capitalize);
```
ROMemories: Sounds a bit more descriptive:
```suggestion
            env.add_filter("capitalize_first", capitalize);
```

ROMemories (commented):
After these two changes, this should be good to go! The rendering isn't perfect because the footnote's superscript numbers make the emoji not centered in the cell, and if we later add content in board pages between the support table and the end of the page, the footnotes might get a bit too far apart from the table, but we could revisit that later and maybe go with something closer that [what's done on the MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details#browser_compatibility) (no need to do that for now), but at least the comments do get rendered now!

ROMemories:
> I noticed that there are some URLs in the comments, so they are rendered as text, with no clickable link. Should I add an intermediate step where I scan the comment for URLs, and transform them into links?

No need, we can actually just write Markdown in these comments now (possibly revisiting the comments is out of scope of this PR)!

book/src/boards/bbc-microbit-v1.md
ROMemories, line 1: I suppose this added a trailing new line that was missing at the end of the file, but GitHub does not make that very clear.

ROMemories (approved):

PR #1414, opened 2025-10-03
ci: reset `RUSTFLAGS` before running Clippy on `coapcore`

ROMemories:
# Description

This resets the `RUSTFLAGS` environment variable before running the `coapcore` Clippy check, otherwise the `RUSTFLAGS` value set before still gets passed to `coapcore`, which makes it fail when passing a `--cfg` flag intended for embedded `getrandom` only, as `getrandom` is a transitive dependency of `cbindgen`, which `liboscore` uses (see https://github.com/ariel-os/ariel-os/pull/1339#issuecomment-3364623200).

## Issues/PRs references

Extracted from #1339.

## Change checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -232,6 +232,8 @@ jobs:
             --
             --deny warnings
 
+      # Reset `RUSTFLAGS`.
+      - run: echo 'RUSTFLAGS=' >> $GITHUB_ENV
       - name: clippy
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -232,6 +232,8 @@ jobs:
             --
             --deny warnings
 
+      # Reset `RUSTFLAGS`.
```
kaspar030, line 235: This could have a note on the why.

ROMemories: I mean, there's no reason *not* to reset it.

PR #1418, opened 2025-10-03
feat(builtin-sensors): introduce an LPS22DF driver

ROMemories:
# Description

This introduces a driver for the LPS22DF pressure sensor compatible with our sensor API.

## How to review this PR

This is very similar to #1363, so it might be useful to have a look at the discussions on that PR.

## Testing

Can be tested on an STEVAL-MKBOXPRO using the following:

```sh
```

Documentation can be generated with the following (but keep in mind that the instantiation and initialization is intended to be generated anyway, pretty much only the configuration is relevant):

```sh
cargo +nightly doc -p ariel-os-sensor-lps22df --open
```

## Issues/PRs references

- Depends on #1652
- Depends on #1689

## Changelog entry

A sensor driver for the LPS22DF, compatible with the newly-introduced sensor API, is now available.

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,6 +542,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ariel-os-sensor-lps22df"
+version = "0.1.0"
+dependencies = [
+ "ariel-os-hal",
+ "ariel-os-sensors",
+ "ariel-os-sensors-utils",
+ "embassy-sync 0.7.2",
+ "embassy-time",
+ "embedded-hal-async",
+ "portable-atomic",
+]
+
+[[package]]
 name = "ariel-os-sensor-stts22h"
 version = "0.1.0"
 dependencies = [
@@ -5307,6 +5320,7 @@ dependencies = [
  "ariel-os",
  "ariel-os-boards",
  "ariel-os-sensor-lis2du12",
+ "ariel-os-sensor-lps22df",
  "ariel-os-sensor-stts22h",
  "ariel-os-sensors",
  "embassy-sync 0.7.2",
--- a/examples/sensors-debug/Cargo.toml
+++ b/examples/sensors-debug/Cargo.toml
@@ -12,6 +12,7 @@ ariel-os = { path = "../../src/ariel-os", features = [
 ] }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 ariel-os-sensor-lis2du12 = { path = "../../src/sensors/ariel-os-sensor-lis2du12" }
+ariel-os-sensor-lps22df = { path = "../../src/sensors/ariel-os-sensor-lps22df" }
 ariel-os-sensor-stts22h = { path = "../../src/sensors/ariel-os-sensor-stts22h" }
 ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
 embassy-sync = { workspace = true }
--- a/examples/sensors-debug/src/sensors.rs
+++ b/examples/sensors-debug/src/sensors.rs
@@ -5,6 +5,9 @@ pub async fn init() {
     #[cfg(any(context = "st-steval-mkboxpro"))]
     lis2du12::init().await;
 
+    #[cfg(any(context = "st-steval-mkboxpro"))]
+    lps22df::init().await;
+
     #[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
     stts22h::init().await;
 }
@@ -45,6 +48,43 @@ mod lis2du12 {
 #[allow(unused, reason = "should be directly accessible without going through the registry")]
 #[cfg(any(context = "st-steval-mkboxpro"))]
 pub use lis2du12::LIS2DU12_I2C;
+
+#[cfg(any(context = "st-steval-mkboxpro"))]
+mod lps22df {
+    use ariel_os::i2c::controller::I2cDevice;
+
+    pub static LPS22DF_I2C: ariel_os_sensor_lps22df::i2c::Lps22df<I2cDevice> =
+        const { ariel_os_sensor_lps22df::i2c::Lps22df::new(Some("onboard")) };
+    #[ariel_os::reexports::linkme::distributed_slice(ariel_os::sensors::SENSOR_REFS)]
+    #[linkme(crate = ariel_os::reexports::linkme)]
+    static LPS22DF_I2C_REF: &'static dyn ariel_os::sensors::Sensor = &LPS22DF_I2C;
+
+    #[ariel_os::task(autostart)]
+    pub async fn lps22df_i2c_runner() {
+        LPS22DF_I2C.run().await
+    }
+
+    pub(super) async fn init() {
+        let mut config = ariel_os_sensor_lps22df::i2c::Config::default();
+        config.address = {
+            #[cfg(context = "st-steval-mkboxpro")]
+            let address = ariel_os_sensor_lps22df::i2c::I2cAddress::Sa0Vdd;
+            address
+        };
+
+        LPS22DF_I2C
+            .init(
+                ariel_os_sensor_lps22df::i2c::Peripherals {},
+                I2cDevice::new(crate::i2c_bus::I2C_BUS.get().unwrap()),
+                config,
+            )
+            .await;
+    }
+}
+
+#[allow(unused, reason = "should be directly accessible without going through the registry")]
+#[cfg(any(context = "st-steval-mkboxpro"))]
+pub use lps22df::LPS22DF_I2C;
 
 #[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
 mod stts22h {


src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    fn set_mode(&self, mode: SensorMode) -> Result<State, SetModeError> {
+        let new_state = self.state.set_mode(mode);
+
+        if new_state == State::Uninitialized {
+            Err(SetModeError::Uninitialized)
+        } else {
+            Ok(new_state)
+        }
+    }
```
ROMemories, line 239: The three methods above are exactly the same as for the STTS22H driver.

src/sensors/ariel-os-sensor-lps22df/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+#[expect(dead_code)]
+const DEVICE_ID: u8 = 0xb4;
+
+// Table 2 of the datasheet.
+const PRESSURE_SENSITIVITY: i32 = 4096;
+const TEMP_SENSITIVITY: i32 = 100;
+
+fn pressure_accuracy(_pressure: i32) -> SampleMetadata {
+    // Takes into account `PAccT` + `P_drift` and a rough upper bound of the temperature offset
+    // from Table 2 of the datasheet.
```
ROMemories, line 44: This could be refined by taking into account the actual temperature; I started doing that but I think this gets unnecessarily complex to just refine the accuracy by ~0.5–1 hPa.

ROMemories, line 44: Related question regarding the TCO (from Table 2): what do you think is the reference temperature to use this coefficient? Is it T = 25 °C? So the temperature-related pressure offset would be computed by taking the absolute value of the difference between the current temperature and T, and then multiplying by the TCO?

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        // Requires `IF_ADD_INC` to be set (which is the default).
+        let mut buf = [0u8; 5];
+        i2c.write_read(address, &[Register::PressOutXl as u8], &mut buf)
+            .await
+            .map_err(|_| ReadingError::SensorAccess)?;
+
+        let pressure = i32::from(self.pressure_offset.load(Ordering::Acquire))
+            + i32::from_be_bytes([0, buf[2], buf[1], buf[0]]) / crate::PRESSURE_SENSITIVITY;
```
ROMemories: Actually the 24-bit encoded pressure is signed, it needs to be sign-extended properly.

ROMemories: Added a conversion function, even though it's actually unclear whether the sensor device can actually return a negative value since the datasheet calls the device a "barometer," which is supposed to be measuring *air* pressure.

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
```diff
@@ -0,0 +1,267 @@
...
+            Ok(new_state)
+        }
+    }
+
+    fn state(&self) -> State {
+        self.state.get()
+    }
+
+    fn categories(&self) -> &'static [Category] {
+        &[Category::Pressure, Category::PressureTemperature]
```
ROMemories, line 246: Should there actually be a category dedicated to barometers (pressure sensors specifically intended to measure *air* pressure)? We could also change the display name below to "barometer."

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
```diff
@@ -0,0 +1,267 @@
...
+                .map_err(|_| ReadingError::SensorAccess)?;
+
+            // New data available.
+            let mask = crate::T_DA_BITS | crate::P_DA_BITS;
+            if buf[0] & mask == mask {
+                break;
+            }
+
+            // TODO: configuration
+            Timer::after_millis(10).await;
```
bergzand: [According to AN5699](https://www.st.com/resource/en/application_note/an5699-lps22df-lowpower-and-highprecision-mems-nano-pressure-sensor-stmicroelectronics.pdf), with the default settings of AVG[2:0] set to 000 and in oneshot mode, this should take 1.2 ms. My suggestion is setting this delay to 1500 us and moving it to the top of the loop here so that the driver first waits for the measurement and then reads the status.

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+
+    async fn reset(i2c_device: &mut I2C, address: I2cAddress) -> Result<i16, ()> {
+        let address = address as u8;
+
+        i2c_device
+            .write(address, &[Register::CtrlReg2 as u8, crate::SWRESET_BITS])
+            .await
+            .map_err(|_| ())?;
```
bergzand, line 106: From AN5699:
> The software reset procedure can take a few tens of μs; the status of the reset is signaled by the status of the
SWRESET bit of the CTRL_REG2(`@11h`) register. Once the reset is completed, this bit is automatically set low.

This needs either a comment that we can ignore the delay required because of the bus delay, or add an extra poll loop to retrieve the status.

bergzand (changes requested):
My two comments, I'll look at your comments later

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
ROMemories: Thanks a lot for looking at these timing constraints! I've applied your suggestion.

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
ROMemories: I've added a 100 µs delay to be safe: I'd rather not rely on hardware timing for this and I'd rather pay the price of a few tens of microseconds once at startup than the flash cost of adding an extra poll loop.

ROMemories:
I've added a commit that makes Clippy happy; when #1689 lands, I'll rebase and add the new sensor driver to the list, so it is checked in CI as well.

src/sensors/ariel-os-sensor-lps22df/src/lib.rs
bergzand: Agreed

bergzand: I looked briefly at both the rust and the C drivers provided by ST, both do not implement temperature compensation inside the driver, they just return the temperature to the caller. I suspect calibration by the user is required to find a reference value before this can be used.

ROMemories: > I suspect calibration by the user is required to find a reference value before this can be used.

Ah right, that would make sense.

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
bergzand: The values from the `PRESS_OUT` registers *can* be negative. The value is the difference between the measured pressure and the reference pressure stored in the `RPDS` register * 256. If we leave the `RPDS` registers at zero, the `PRESS_OUT` *should* never be negative.

See figure 6 of [AN5699](https://www.st.com/resource/en/application_note/an5699-lps22df-lowpower-and-highprecision-mems-nano-pressure-sensor-stmicroelectronics.pdf)

bergzand: I think it's fine to ignore the negative pressure case here for now though

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
bergzand: I think this is an edge case with systems measuring both atmospheric pressure and a different pressure. I think it's fine to have a single pressure category for pressure. Or do you mean distinguish between a gas pressure and other pressure (and tension) cases such as liquid pressure and mechanical pressure?

ROMemories: > I think this is an edge case with systems measuring both atmospheric pressure and a different pressure.

Yes, that's what I meant; nothing to do then :slightly_smiling_face:

src/sensors/ariel-os-sensor-lps22df/src/i2c.rs
ROMemories: > I think it's fine to ignore the negative pressure case here for now though

I mean, the driver should handle these now, through the `i32_from_i24_be_bytes()` function.

bergzand:
Tested on the `st-steval-mkboxpro`:

```
[INFO ] Will print the readings of registered sensor drivers… (sensors_debug sensors-debug/src/main.rs:22)
[INFO ] pressure sensor (onboard): 100300.0 Pa ± 150.0 Pa (Pressure) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] pressure sensor (onboard): 24.92 °C ± 1.5 °C (Temperature) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] temperature sensor (onboard): 23.74 °C ± 0.5 °C (Temperature) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] 3-axis accelerometer (onboard): 0.011712 g ± 0.011 g (AccelerationX) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] 3-axis accelerometer (onboard): 0.020496 g ± 0.011 g (AccelerationY) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] 3-axis accelerometer (onboard): -0.914512 g ± 0.011 g (AccelerationZ) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] pressure sensor (onboard): 100300.0 Pa ± 150.0 Pa (Pressure) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] pressure sensor (onboard): 24.93 °C ± 1.5 °C (Temperature) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] temperature sensor (onboard): 23.72 °C ± 0.5 °C (Temperature) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] 3-axis accelerometer (onboard): 0.008784 g ± 0.011 g (AccelerationX) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] 3-axis accelerometer (onboard): 0.022448 g ± 0.011 g (AccelerationY) (sensors_debug sensors-debug/src/main.rs:122)
[INFO ] 3-axis accelerometer (onboard): -0.914512 g ± 0.011 g (AccelerationZ) (sensors_debug sensors-debug/src/main.rs:122)
```

Works perfect. The two temperature sensors both are a bit on the high side, but are still within eachothers error margins.

PR #1419, opened 2025-10-03
feat(builtin-sensors): introduce an LIS2MDL driver

ROMemories:
# Description

This introduces a manual implementation of a driver for the LIS2MDL 3-axis magnetometer.

## Tested

Can be tested on an STEVAL-MKBOXPRO using the following:

```sh
```

## Issues/PRs references

Other sensor drivers may be of interest, especially #1418, which highlighted some of the timing requirements.

## Changelog Entry

A sensor driver for the LIS2MDL, compatible with the sensor API, is now available.

## Change checklist


examples/sensors-debug/src/sensors.rs
```diff
@@ -49,6 +52,38 @@ mod lis2du12 {
...
+    #[linkme(crate = ariel_os::reexports::linkme)]
+    static LIS2MDL_I2C_REF: &'static dyn ariel_os::sensors::Sensor = &LIS2MDL_I2C;
+
+    #[ariel_os::task(autostart)]
+    pub async fn lis2mdl_i2c_runner() {
+        LIS2MDL_I2C.run().await
+    }
+
+    pub(super) async fn init() {
+        let config = ariel_os_sensor_lis2mdl::i2c::Config::default();
```
ROMemories, line 71: Unlike other sensor drivers, this one only has a single I2C address, see Table 20 of the datasheet. (Not adding this as a source code comment here, because this module is expected to be generated anyway.)

src/sensors/ariel-os-sensor-lis2mdl/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+repository.workspace = true
+license.workspace = true
+
+[dependencies]
+ariel-os-hal = { workspace = true, features = ["i2c"] }
+ariel-os-sensors = { workspace = true, features = ["max-sample-min-count-3"] }
+ariel-os-sensors-utils = { workspace = true }
+embassy-sync = { workspace = true }
+embassy-time = { workspace = true }
+embedded-hal-async = { workspace = true }
```
ROMemories, line 17: Unlike other sensor drivers, this one doesn't need to depend on `portable-atomic`, because it's only used for storing the I2C address, and this sensor only has one.

src/sensors/ariel-os-sensor-lis2mdl/src/i2c.rs
```diff
@@ -0,0 +1,264 @@
...
+        // Software reset can take 5 microseconds (AN5069 §7).
+        Timer::after_micros(10).await;
+
+        // Always enable Block Data Update.
+        let cfg = crate::BDU_BITS;
+        i2c_device
+            .write(TARGET_I2C_ADDR, &[Register::CfgRegC as u8, cfg])
+            .await
+            .map_err(|_| ())?;
+
```
ROMemories, line 91: Should we enable the low-pass filter as well? This might reduce the RMS noise a bit (see Table 9 of AN5069), but I'm not sure it would work at all considering it's a digital filter (therefore after the ADC, see Figure 2 of AN5069) and we're only using the single-shot mode, with no guarantee that measurements (and therefore ADC conversions I believe) would be triggered frequently enough to be useful.

Cargo.toml
```diff
@@ -29,6 +29,7 @@ members = [
   "src/lib/rbi",
   "src/lib/ringbuffer",
   "src/sensors/ariel-os-sensor-lis2du12",
+  "src/sensors/ariel-os-sensor-lis2mdl",
```
ROMemories, line 32: Since #1841, we now add the sensor drivers as workspace members.

src/sensors/ariel-os-sensor-lis2mdl/src/i2c.rs
```diff
@@ -0,0 +1,258 @@
...
+            }
+            State::Enabled => ReadingWaiter::new_err(ReadingError::NotMeasuring),
+            State::Uninitialized | State::Disabled | State::Sleeping => {
+                ReadingWaiter::new_err(ReadingError::NonEnabled)
+            }
+        }
+    }
+
+    fn set_mode(&self, mode: SensorMode) -> Result<State, SetModeError> {
+        self.state.set_mode(mode)
```
ROMemories, line 212: Updated following #1960

PR #1422, opened 2025-10-06
docs(book): document the current memory layout

ROMemories:
# Description

This adds information documentation of the memory layout, across various configuration settings (esp. async executors and multithreading). Currently this is added to the "Async Executors" page, but we may want to move this elsewhere later.

## Open Questions

- Should we render the diagrams with [svgbob](https://ivanceras.github.io/svgbob-editor/) or keep them as is?

## Changelog Entry

Some documentation of the memory layout has been added to the book.

## Change checklist

### DIFF
--- a/book/src/memory-layout.md
+++ b/book/src/memory-layout.md
@@ -1,0 +1,162 @@
+# Memory Layout
+
+This page is intended to give a relatively high-level overview of the memory layout used by Ariel OS across processor architectures and MCU families.
+Minor variations across them are omitted for clarity.
+Additionally, no stability guarantees are currently provided regarding the memory layout.
+
+> [!TIP]
+> If needed, the following command can be used to obtain the exact locations of sections:
+>
+> ```sh
+> readelf --sections <path-to-elf>
+> ```
+
+- On architectures and MCUs that allow it[^flip-link-hw-support], Ariel OS places the ISR stack, `.isr_stack`, at the very beginning of the RAM.
+  This provides a form of stack overflow protection, as write attempts would then collide with the boundary of the physical RAM and trigger a fault, instead of overwriting `static` data.
+  See the [`flip-link` crate][flip-link-readme] for an explanation of the technique.
+  The ISR stack is named `.stack` on Xtensa, and may not be placed at the beginning of the RAM nor before the `.data` section.
+- Async tasks are allocated statically, as individual `static`s, anywhere in the `.bss` section.
+- The thread stacks, if [multithreading][multithreading-book] is enabled, are currently allocated as individual `static`s, anywhere in the `.bss` section.
+- If [multithreading][multithreading-book] is enabled, and if the MCU has two symmetrical cores and [`multi-core`][multicore-support-book] is enabled, the ISR stack of the second core is allocated as a `static`, anywhere in the `.bss` section.
+- Depending on the architecture, the uninitialized section is either called `.uninit` or `.noinit`.
+- The heap, if enabled, may take up to the remaining space.
+  `heapsize_required` ensures that there is enough space for the heap, at link time.
+
+[^flip-link-hw-support]: Currently that is Cortex-M-based MCUs and *some* RISC-V based ESP32s (the ESP32-C6, but not the ESP32-C3).
+
+<!--
+The diagrams are based on the following resources:
+
+- Location of the `.isr_stack` section (inserted by Ariel OS):
+    - `build.rs`: https://github.com/ariel-os/ariel-os/blob/2002a036f49848e9c049b735ed0053bce23b6172/src/ariel-os-rt/build.rs#L17-L34
+    - `isr_stack.ld.in`: https://github.com/ariel-os/ariel-os/blob/2002a036f49848e9c049b735ed0053bce23b6172/src/ariel-os-rt/isr_stack.ld.in#L1-L12
+- Location of the `.stack` section (defined upstream):
+    - https://github.com/ariel-os/ariel-os/blob/2002a036f49848e9c049b735ed0053bce23b6172/src/ariel-os-rt/isr_stack_xtensa.ld.in#L2-L3
+    - https://github.com/esp-rs/esp-hal/blob/909db474ae7df254bb0e51cc5ba94c13411813d4/esp-hal/ld/sections/stack.x#L3
+- Location of the ISR stack of the second SMP core:
+    - https://github.com/ariel-os/ariel-os/blob/2002a036f49848e9c049b735ed0053bce23b6172/src/ariel-os-threads/src/lib.rs#L565-L567
+- Location of the `.data` section
+    - Cortex-M: https://github.com/rust-embedded/cortex-m/blob/84e5c011068f01b7716684d20f45841cdfe3f285/cortex-m-rt/link.x.in#L126
+    - RISC-V ESP32: https://github.com/rust-embedded/riscv/blob/187453b7904f997c4aa679d5bcf135f16c8853f8/riscv-rt/link.x.in#L129
+    - Xtensa ESP32: https://github.com/esp-rs/esp-hal/blob/d9815b205115b3403d798c28f1bfee4c9eea8fd1/xtensa-lx-rt/xtensa.in.x#L36
+- Location of the `.bss` section
+    - Cortex-M: https://github.com/rust-embedded/cortex-m/blob/84e5c011068f01b7716684d20f45841cdfe3f285/cortex-m-rt/link.x.in#L161
+    - RISC-V ESP32: https://github.com/rust-embedded/riscv/blob/187453b7904f997c4aa679d5bcf135f16c8853f8/riscv-rt/link.x.in#L150
+    - Xtensa ESP32: https://github.com/esp-rs/esp-hal/blob/d9815b205115b3403d798c28f1bfee4c9eea8fd1/xtensa-lx-rt/xtensa.in.x#L47
+- Location and name of the `.uninit`/`.noinit` section
+    - Cortex-M: https://github.com/rust-embedded/cortex-m/blob/84e5c011068f01b7716684d20f45841cdfe3f285/cortex-m-rt/link.x.in#L175
+    - RISC-V ESP32: https://github.com/rust-embedded/riscv/blob/187453b7904f997c4aa679d5bcf135f16c8853f8/riscv-rt/link.x.in#L168
+    - Xtensa ESP32: https://github.com/esp-rs/esp-hal/blob/d9815b205115b3403d798c28f1bfee4c9eea8fd1/xtensa-lx-rt/xtensa.in.x#L55
+-->
+
+<!-- These diagrams can be rendered with Svgbob https://github.com/ivanceras/svgbob -->
+
+<figure>
+<pre>
+           .-------------. - beginning of RAM
+           |      ⋮      |
+           +-------------+ -
+           |             |
+         | |             | ^
+Addresses| | .isr_stack  | | ≥ isr_stacksize_required + executor_stacksize_required
+         v |             | v
+           |             |
+           +-------------+ -
+           |             |
+           |             |
+           |             |    .- - .-------------.
+           |             |    |    |      ⋮      |
+           |    .data    |    |    +-------------+
+           |             |    |    |    Async    |
+           |             |    |    |    tasks    |
+           |             |    |    |      ...    |
+           |             |    |    +-------------+
+           +-------------+ - -'    |      ⋮      |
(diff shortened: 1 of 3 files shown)


book/src/async-support.md
```diff
@@ -35,6 +35,110 @@ Running multiple executors allows running them with different priorities.
...
+</pre>
+    <figcaption class="text-center">Memory layout when using <code>executor-thread</code></caption>
+</figure>
+
+<style>
+.text-center {
+    text-align: center;
+}
+</style>
+
```
ROMemories: I would also like to later add links to everything that allows to check this (e.g., links to the `cortex-m-rt`, `riscv-rt`, `xtensa-lx-rt` crates, and to our linker scripts).

book/src/async-support.md
```diff
@@ -36,6 +36,145 @@ Running multiple executors allows running them with different priorities.
...
+- The thread stacks, if [multithreading][multithreading-book] is enabled, are currently declared as individual `static`s, they are therefore likely not contiguously allocated in the `.bss` section.
+- Depending on the architecture the uninitialized section is either called `.uninit` or `.noinit`.
+- The heap, if enabled, takes the remaining space.
+  `heapsize_required` allows link-time checking that there is enough space for the heap.
+
+<!--
+The diagrams are based on the following resources:
+
+- Location of the `.isr_stack` section (inserted by Ariel OS):
+    - `build.rs`: https://github.com/ariel-os/ariel-os/blob/2002a036f49848e9c049b735ed0053bce23b6172/src/ariel-os-rt/build.rs#L17-L34
```
ROMemories: Based on this, the location of `.isr_stack` on ESP32-C3 ~~may~~ is not ~~be~~ correct in the diagrams and needs to be updated.

book/src/async-support.md
ROMemories: I've added the links.

book/src/async-support.md
ROMemories: I've added an introductory paragraph to the book section that hopefully allows us not to have to document this.

book/src/async-support.md
```diff
@@ -36,8 +36,160 @@ Running multiple executors allows running them with different priorities.
...
+- Async tasks are allocated statically, as individual `static`s, anywhere in the `.bss` section.
+- The thread stacks, if [multithreading][multithreading-book] is enabled, are currently allocated as individual `static`s, anywhere in the `.bss` section.
+- Depending on the architecture, the uninitialized section is either called `.uninit` or `.noinit`.
+- The heap, if enabled, takes the remaining space.
+  `heapsize_required` allows link-time checking that there is enough space for the heap.
+
+<!--
+The diagrams are based on the following resources:
+
+- Location of the `.isr_stack` section (inserted by Ariel OS):
```
ROMemories: It's not used on Xtensa, [Ariel OS instead uses](https://github.com/ariel-os/ariel-os/blob/2002a036f49848e9c049b735ed0053bce23b6172/src/ariel-os-rt/isr_stack_xtensa.ld.in#L2-L3) the [upstream stack definition](https://github.com/esp-rs/esp-hal/blob/909db474ae7df254bb0e51cc5ba94c13411813d4/esp-hal/ld/sections/stack.x#L3) (named `.stack`).

kaspar030:
(random thought: on multicore MCUs, second core's ISR stack will also land anywhere in .bss IIRC)

ROMemories:
> on multicore MCUs, second core's ISR stack will also land anywhere in .bss IIRC

@kaspar030 Do you have some pointers of where this happens (so I can add the permlink to the list)?

kaspar030:
> > on multicore MCUs, second core's ISR stack will also land anywhere in .bss IIRC
> 
> @kaspar030 Do you have some pointers of where this happens (so I can add the permlink to the list)?

Here's where the second core stack gets defined:

https://github.com/ariel-os/ariel-os/blob/6b6f4eb9bb8bbb2df6ade60c5d17824d95daf7ae/src/ariel-os-threads/src/lib.rs#L565-L567

It is a regular static, so, .bss.

book/src/async-support.md
```diff
@@ -36,8 +36,164 @@ Running multiple executors allows running them with different priorities.
...
+> ```sh
+> readelf --sections <path-to-elf>
+> ```
+
+- Ariel OS places the ISR stack (`.isr_stack`) before the `.data` section and at the very beginning of the RAM on architectures and MCUs that allow it.
+  This provides a form of stack overflow protection, as write attempts would then collide with the boundary of the physical RAM and trigger a fault, instead of overwriting `static` data.
+  See the [`flip-link` crate][flip-link-readme] for an explanation of the technique.
+- Async tasks are allocated statically, as individual `static`s, anywhere in the `.bss` section.
+- The thread stacks, if [multithreading][multithreading-book] is enabled, are currently allocated as individual `static`s, anywhere in the `.bss` section.
+- If [multithreading][multithreading-book] is enabled, and if the MCU has two symmetrical cores and [`multi-core`][multicore-support-book] is enabled, the ISR stack of the second core is allocated as a `static`, anywhere in the `.bss` section.
```
ROMemories: I've added an item documenting the location of the second ISR stack if present, as clarified in https://github.com/ariel-os/ariel-os/pull/1422#issuecomment-4162021541. I'm not adding this to the diagrams (at least for now), as this depends on the hardware, currently applies to a small number of MCUs, and I think the text is clear enough for now (and adding it would make the diagrams less so).

book/src/async-support.md
```diff
@@ -36,8 +36,164 @@ Running multiple executors allows running them with different priorities.
...
+- Depending on the architecture, the uninitialized section is either called `.uninit` or `.noinit`.
+- The heap, if enabled, takes the remaining space.
```
ROMemories: Later I'd like to have a "Heap Allocator" book page, and add a back link here.

book/src/async-support.md
```diff
@@ -36,8 +36,167 @@ Running multiple executors allows running them with different priorities.
...
+
+- Ariel OS places the ISR stack, `.isr_stack` (`.stack` on Xtensa), before the `.data` section and at the very beginning of the RAM on architectures and MCUs that allow it.
+  This provides a form of stack overflow protection, as write attempts would then collide with the boundary of the physical RAM and trigger a fault, instead of overwriting `static` data.
+  See the [`flip-link` crate][flip-link-readme] for an explanation of the technique.
+- Async tasks are allocated statically, as individual `static`s, anywhere in the `.bss` section.
+- The thread stacks, if [multithreading][multithreading-book] is enabled, are currently allocated as individual `static`s, anywhere in the `.bss` section.
+- If [multithreading][multithreading-book] is enabled, and if the MCU has two symmetrical cores and [`multi-core`][multicore-support-book] is enabled, the ISR stack of the second core is allocated as a `static`, anywhere in the `.bss` section.
+- Depending on the architecture, the uninitialized section is either called `.uninit` or `.noinit`.
+- The heap, if enabled, takes the remaining space.
+  `heapsize_required` allows link-time checking that there is enough space for the heap.
```
ROMemories: On ESP, it *also* [actually sizes the heap, if the value provided is > 2 KiB](https://github.com/ariel-os/ariel-os/blob/6b6f4eb9bb8bbb2df6ade60c5d17824d95daf7ae/laze-project.yml#L954), but this may be better documented where we'll document the `isr_stacksize_required`, `executor_stacksize_required`, and `heapsize_required` variables.

book/src/async-support.md
```diff
@@ -36,8 +36,167 @@ Running multiple executors allows running them with different priorities.
 
 <!-- TODO: Power consumption optimization -->
 
+## Memory Layout
```
kaspar030: yeah, `async-support.md` really seems like the wrong place. how about `memory-layout.md`?

book/src/async-support.md
kaspar030: `heapsize_required` should always be set if heap is needed, that non-esp uses the remaining space is more of a quirk, I think. We need to teach users early here. Can reword/move later.

```suggestion
  `heapsize_required` ensures that there is enough space for the heap (at link time).
   On some platforms (`esp`), the heap will *only* take `heapsize_required`.
```

book/src/async-support.md
```diff
@@ -36,8 +36,167 @@ Running multiple executors allows running them with different priorities.
...
+    - Xtensa ESP32: https://github.com/esp-rs/esp-hal/blob/d9815b205115b3403d798c28f1bfee4c9eea8fd1/xtensa-lx-rt/xtensa.in.x#L36
+- Location of the `.bss` section
+    - Cortex-M: https://github.com/rust-embedded/cortex-m/blob/84e5c011068f01b7716684d20f45841cdfe3f285/cortex-m-rt/link.x.in#L161
+    - RISC-V ESP32: https://github.com/rust-embedded/riscv/blob/187453b7904f997c4aa679d5bcf135f16c8853f8/riscv-rt/link.x.in#L150
+    - Xtensa ESP32: https://github.com/esp-rs/esp-hal/blob/d9815b205115b3403d798c28f1bfee4c9eea8fd1/xtensa-lx-rt/xtensa.in.x#L47
+- Location and name of the `.uninit`/`.noinit` section
+    - Cortex-M: https://github.com/rust-embedded/cortex-m/blob/84e5c011068f01b7716684d20f45841cdfe3f285/cortex-m-rt/link.x.in#L175
+    - RISC-V ESP32: https://github.com/rust-embedded/riscv/blob/187453b7904f997c4aa679d5bcf135f16c8853f8/riscv-rt/link.x.in#L168
+    - Xtensa ESP32: https://github.com/esp-rs/esp-hal/blob/d9815b205115b3403d798c28f1bfee4c9eea8fd1/xtensa-lx-rt/xtensa.in.x#L55
+-->
```
kaspar030: (did not check links)

book/src/async-support.md
```diff
@@ -36,8 +36,167 @@ Running multiple executors allows running them with different priorities.
...
+<!-- These diagrams can be rendered with Svgbob https://github.com/ivanceras/svgbob -->
+
+<figure>
+<pre>
+           .-------------. - beginning of RAM
+           |      ⋮      |
+           +-------------+ -
+           |             |
+         | |             | ^
+Addresses| | .isr_stack  | | ≥ isr_stacksize_required + executor_stacksize_required
```
kaspar030: As noted, this is only on Cortex-M (and esp32-c3/6?). Mark somehow, like, `(*)` + footnote, or `[.isr_stack]`, or both?

kaspar030 (changes requested):

book/src/async-support.md
```diff
@@ -36,8 +36,167 @@ Running multiple executors allows running them with different priorities.
...
+Additionally, no stability guarantees are currently provided regarding the memory layout.
+
+> [!TIP]
+> If needed, the following command can be used to obtain the exact locations of sections:
+>
+> ```sh
+> readelf --sections <path-to-elf>
+> ```
+
+- Ariel OS places the ISR stack, `.isr_stack` (`.stack` on Xtensa), before the `.data` section and at the very beginning of the RAM on architectures and MCUs that allow it.
```
kaspar030: `on architectures and MCUs that allow it` -> footnote which platforms currently do that.

book/src/async-support.md
ROMemories: I've moved the section to a dedicated page.

book/src/async-support.md
ROMemories: Done.

book/src/async-support.md
ROMemories: I've rephrased the previous sentence as it wasn't correct on ESPs, and got the suggestion in.

book/src/memory-layout.md
```diff
@@ -0,0 +1,162 @@
...
+> If needed, the following command can be used to obtain the exact locations of sections:
+>
+> ```sh
+> readelf --sections <path-to-elf>
+> ```
+
+- On architectures and MCUs that allow it[^flip-link-hw-support], Ariel OS places the ISR stack, `.isr_stack`, at the very beginning of the RAM.
+  This provides a form of stack overflow protection, as write attempts would then collide with the boundary of the physical RAM and trigger a fault, instead of overwriting `static` data.
+  See the [`flip-link` crate][flip-link-readme] for an explanation of the technique.
+  The ISR stack is named `.stack` on Xtensa, and may not be placed at the beginning of the RAM nor before the `.data` section.
```
ROMemories, line 17: I've added this sentence to still mention the name of the section on Xtensa, and mention that it may not be placed before the `.data` section, which might address your other comment.

kaspar030 (approved):
LGTM

PR #1427, opened 2025-10-07
chore: Unsafety policy

chrysn:
# Description

As discussed in today's meeting, it'd be good to have clear unsafety policy.

This Pr does two things:

* [x] Add lints for things we already do (or decided would be good to do) where this is easy. (CI may tell me that some things are not as easy or already-done as I thought).
* [ ] Write up which checks we use, and how this means that we limit unsafe code to where it makes sense.

## Change checklist


Cargo.toml
```diff
@@ -167,6 +167,8 @@ unused_must_use = "deny"
 # ... until we figure out a way to generate a list of all used cfg variables
 # across build configurations ...
 unexpected_cfgs = "allow"
+# When writing code, explicitly enable this per module.
```
kaspar030:
```suggestion
# When writing code, explicitly opt-out per module by adding `#[expect(unsafe_code)]`
```

ROMemories:
Can you keep the documentation out of this PR, so the doc-only PR can use `ci-build:skip`?

chrysn:
I'd develop them side-by-side here but be careful about not blocking CI on it. (Documenting it may uncover things we'll want to adjust in the checks – at least that was what happened right after the meeting that started on the doucmentation).

book/src/coding-conventions.md
```diff
@@ -68,9 +68,12 @@ This is useful for readers, to quickly access the mentioned item, but it also he
 
 ### `unsafe` Code
 
-When `unsafe` is used, a `SAFETY` comment MUST be added, in the style [supported by Clippy](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks).
+Code containing `unsafe` is denied outside of modules where the [`unsafe_code`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-code) lint is explicitly expected
```
chrysn:
```suggestion
Code containing `unsafe` is denied outside of modules where the [`unsafe-code`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-code) lint is explicitly expected
```

chrysn:
I think this is ready for review now; my impression is that the ci-build:skip setting should suffice (as it's all just modifying lints or docs anyway).

A follow-up at some time should be to replace the grandfathered-in `expect(clippy::undocumented_unsafe_blocks)` with actual comments, but that's best done by the respective authors; I've only added it for very simple cases.

chrysn:
What is *not* in here is [`multiple_unsafe_ops_per_block`](https://rust-lang.github.io/rust-clippy/stable/index.html#multiple_unsafe_ops_per_block) -- feel free to start adding it, but I think it's excessive.

book/src/coding-conventions.md
ROMemories:
```suggestion
Code containing `unsafe` is denied outside of modules where the [`unsafe_code`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-code) lint is explicitly `#[expect]`ed
```

book/src/coding-conventions.md
```diff
@@ -68,9 +68,12 @@ This is useful for readers, to quickly access the mentioned item, but it also he
...
+(or, in complex situations, allowed).
```
ROMemories:
```suggestion
(or, in complex situations, `#[allow]`ed, in which case a `reason` MUST be provided).
```

book/src/coding-conventions.md
```diff
@@ -68,9 +68,12 @@ This is useful for readers, to quickly access the mentioned item, but it also he
...
 
-TODO: enforce it in CI
+For all `unsafe` blocks, a `SAFETY` comment MUST be added, in the style [of the `undocumented-unsafe-blocks` Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks).
+
+Any `unsafe` function MUST be documented with the preconditions for sound use in a `# Safety` section, in the style [style of the `missing-safety-doc` Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc).
```
ROMemories:
```suggestion
Any `unsafe` function MUST be documented with the preconditions for sound use in a `# Safety` section, in the [style of the `missing-safety-doc` Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc).
```

src/lib/coapcore/src/oluru.rs
```diff
@@ -12,7 +12,6 @@
 //! Throughout the documentation, "high" and "low" indicate priorities, whereas "small" and "large"
 //! indicate numeric values (including of priorities, where "high" corresponds to "small" and "low"
 //! to "large").
-#![forbid(unsafe_code)]
```
ROMemories, line 15: Do we *need* to remove this? This revokes the promise that no modules of this crate use `unsafe`.

Cargo.toml
```diff
@@ -177,6 +179,7 @@ pedantic = { level = "warn", priority = -1 }
 too_long_first_doc_paragraph = "warn"
 unused_result_ok = "deny"
 unused_trait_names = "warn"
+undocumented_unsafe_blocks = "warn"
```
ROMemories, line 182: To preserve the lexicographical order of these:
```suggestion
undocumented_unsafe_blocks = "warn"
unused_result_ok = "deny"
unused_trait_names = "warn"
```

Cargo.toml
```diff
@@ -167,6 +167,8 @@ unused_must_use = "deny"
 # ... until we figure out a way to generate a list of all used cfg variables
 # across build configurations ...
 unexpected_cfgs = "allow"
+# When writing code, explicitly opt-out per module by adding `#[expect(unsafe_code)]`
+unsafe_code = "warn"
```
ROMemories, line 171:
```suggestion
# When writing code, explicitly opt out per module by adding `#[expect(unsafe_code)]`
unsafe_code = "warn"
unsafe_op_in_unsafe_fn = "deny"
unused_must_use = "deny"
# ... until we figure out a way to generate a list of all used cfg variables
# across build configurations ...
unexpected_cfgs = "allow"
```

src/ariel-os-threads/src/sync/channel.rs
```diff
@@ -1,5 +1,8 @@
 //! Synchronous channel implementation for sending data between threads.
 
+#![expect(unsafe_code)]
+#![expect(clippy::undocumented_unsafe_blocks)]
```
ROMemories: We've used the following reason in other places:
```suggestion
#![expect(clippy::undocumented_unsafe_blocks, reason = "should be addressed eventually")]
```

src/ariel-os-rt/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![cfg_attr(test, no_main)]
 //
 #![allow(incomplete_features)]
+#![allow(unsafe_code)]
```
ROMemories, line 5: I suppose we can't easily use `expect`? If so it would be useful to add a `reason` (e.g., `conditional compilation`, which we use in other places). Same in other places that use `allow`.

src/ariel-os-random/src/lib.rs
```diff
@@ -23,6 +23,10 @@
 //! [`rand_pcg::Pcg32`] is decided yet for the fast one. Neither the algorithm nor the size of
 //! [`FastRng`] or [`CryptoRng`] is guaranteed.
 #![no_std]
+#![allow(
+    unsafe_code,
+    reason = "providing the getrandom() custom implementation needs unsafe operations"
+)]
```
ROMemories: Alternatively, it could be clearer to move `__getrandom_v03_custom` into an *inline* module (whose name doesn't matter), and `expect` inside it.

ROMemories:
> my impression is that the ci-build:skip setting should suffice (as it's all just modifying lints or docs anyway)

`ci-build:skip` is indeed enough for what the PR currently contains.

src/lib/coapcore/src/oluru.rs
chrysn: This crate is on its way out anyway on the next run; no changes other than mechanical ones are happening there, and the forbid was a relic of when this was intended to be a standalone thing.

Cargo.toml
chrysn: Only if you insist; generally I don't think we should have this kind of soft-code-style things that are not covered by linters and thus automated.

Otherwise this creates a lot of busy work (retroactively sorting on every commit in a history, esp. because the GitHub GUI changes are not just click-and-apply) for the marginal benefit of automatic merging if ever two racing PRs add the same item in different places.

book/src/coding-conventions.md
chrysn: I think the reason argument is more for why something is allowed in a place in the first place, not for arguing the allow/expect trade-off.

If we want to require reasons, please enable `allow_attributes_without_reason` throughout the codebase.

(The `#[` part is applied.)

src/ariel-os-threads/src/sync/channel.rs
chrysn: Applied to all undocumented_unsafe_blocks.

src/ariel-os-rt/src/lib.rs
chrysn: The *only* reason we ever do an allow is because some conditional compilation doesn't make it trigger unconditionally (else we'd do an expect), so I don't see much value in documenting this.

(Checking at PR time yes, so I confirm: This PR only has allow() where manual inspection or CI revealed that an expect() wouldn't do).

src/ariel-os-random/src/lib.rs
chrysn: In general, we could always narrow it down, but then again, we could *always* move the allow(unsafe) in so far that it just narrowly encapsulates the unsafe, at which point it becomes just a restatement of using the unsafe comment there.

I think the point of having this file-level is to alert the developer and reviewer to the fact that unsafe happens in this module, and thus all code in it potentially contributes to unsafety requirements.

In this concrete case, full separation *is* a viable option, and thus moved into a separate-file module.

Cargo.toml
chrysn: See [above](https://github.com/ariel-os/ariel-os/pull/1427#discussion_r2425462312)

src/lib/coapcore/src/oluru.rs
ROMemories: I mean I still don't see why this PR has to *remove* it, but fine.

book/src/coding-conventions.md
ROMemories: "style" is still there twice.

chrysn: Thanks, absorbed and force-pushed.

PR #1428, opened 2025-10-07
docs(gen-book): introduce chip pages

jkr-ariel:
# Description

This PR aims to introduce dedicated chip pages in the ArielOS book. They will be similar in structure to the board pages introduced in #1327.

## Issues/PRs references

Part of #1182.

## Change checklist


ROMemories:
Can you rebase some more? There's still commits from #1433 in there.

.github/workflows/main.yml
```diff
@@ -442,6 +442,9 @@ jobs:
       - name: Check that the board pages are all up-to-date
         run: ./doc/gen_book.rs board-pages doc/support_matrix.yml book/src/boards/ --check --template-path book/templates/board-page.md.tmpl
 
+      - name: Check that the chip pages are all up-to-date
+        run: ./doc/gen_book.rs chip-pages doc/support_matrix.yml book/src/chips/ --check --template-path book/templates/chip-page.md.tmpl
```
ROMemories, line 445: Following #1459, these need to be added to the `update-book` task.

book/templates/chip-index.md.tmpl
```diff
@@ -0,0 +1,6 @@
+# Chip Pages
+
+Here's the list of supported chips in ArielOS.
```
ROMemories: We don't really use contractions in the book:

```suggestion
Here is the list of supported chips in Ariel OS.
```

book/templates/chip-index.md.tmpl
```diff
@@ -0,0 +1,6 @@
...
+{% for chip_technical_name, chip_info in chips|dictsort %}
+- [{{ chip_info.name }}](./{{ chip_technical_name }}.md)
+{%- endfor %}
```
ROMemories, line 6: It would likely be useful and more readable to group these by manufacturers, but that requires introducing a manufacturer key and is better done in a follow-up.

book/templates/chip-page.md.tmpl
```diff
@@ -0,0 +1,29 @@
+# {{ chip_info.name }}
+
```
ROMemories, line 2: It would be useful to document the Ariel OS chip name here, and maybe remove it from the board page, but this can be done in a follow-up.

ROMemories, line 2: In the same vein, it would be useful to add a link to the chip page from the board pages that use that chip (but, again, follow-up).

doc/gen_book.rs
```diff
@@ -370,6 +440,79 @@ impl SubCommandPages {
...
+                source,
+            }
+        })?;
+
+        let chip_info = gen_chip_functionalities(&matrix)?;
+
+        for chip in &chip_info {
+            let mut env = Environment::new();
+            env.add_filter("slugify", slugify::<String>);
+            env.add_filter("capitalize_first", capitalize_first);
```
ROMemories: Seems like this could be factored out with the other occurrence in `SubCommandBoardPages::run()`.

book/templates/chip-index.md.tmpl
ROMemories: There needs to be a (non-breakable) space between Ariel and OS.

PR #1433, opened 2025-10-08
feat(esp)!: introduce laze contexts for ESP32 chips with on-chip flash

ROMemories:
# Description

This introduces/renames laze contexts of ESP32 chips that feature on-chip/in-package flash.
This will be needed for the storage API.

BREAKING CHANGE: this effectively renames documented laze contexts.

## Change checklist


laze-project.yml
```diff
@@ -1824,7 +1830,7 @@ builders:
       - has_usb_device_port
 
   - name: heltec-wifi-lora-32-v3
-    parent: esp32s3
+    parent: esp32s3fx8
```
ROMemories: The [manufacturer page](https://heltec.org/project/wifi-lora-32-v3/) mentions the following:
> Microprocessor: ESP32-S3FN8 (Xtensa® 32-bit LX7 dual-core processor, five-stage pipeline rack Structure, frequency up to 240 MHz).

and there is no visible flash chip in the pictures.

laze-project.yml
```diff
@@ -413,9 +413,12 @@ contexts:
       PROBE_RS_PROTOCOL: jtag
       CARGO_TARGET_PREFIX: CARGO_TARGET_RISCV32IMC_UNKNOWN_NONE_ELF
 
-  - name: esp-c3-01m
+  - name: esp32c3fx4
     parent: esp32c3
 
+  - name: esp-c3-01m
+    parent: esp32c3fx4
```
ROMemories: Documentation for this hardware module is scarce; I couldn't find this in the official datasheet but I've [found this](https://templates.blakadder.com/ESP-C3-01M.html) which states:

> C3FN4 with 4Mb [sic] flash

I'm not exactly sure this is correct.

doc/support_matrix.yml
```diff
@@ -192,6 +192,25 @@ chips:
         comments:
           - not currently compatible with threading
 
+  # Currently identitical to esp32c3.
```
ROMemories: We don't currently have a way to de-duplicate this here, but this can be tackled separately.

laze-project.yml
```diff
@@ -1775,7 +1784,7 @@ builders:
     parent: nrf52833
 
   - name: dfrobot-firebeetle2-esp32-c6
-    parent: esp32c6
+    parent: esp32c6fx4
```
ROMemories: Doesn't require a doc update because we don't currently document support for the `dfrobot-firebeetle2-esp32-c6` (but we could as a follow-up).

Edit: the follow-up is #1436.

laze-project.yml
ROMemories: We've dropped this laze context in #1445.

kaspar030 (approved):

PR #1367, opened 2025-09-18
feat(esp): Add UART HAL support

bergzand:
# Description

This adds support for the UART HAL on the Espressif ESP series devices.

The Espressif devices are built different as they have an internal fifo for uart and do not need a provided buffer.

## Issues/PRs references

depends on #1365

## Change checklist

- [ ] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1436, opened 2025-10-09
docs(support): document support for the DFRobot FireBeetle 2 ESP32-C6

ROMemories:
# Description

Support was added in #983 but left undocumented mostly because we didn't have a tier system at the time.
Following #1433, it's also more interesting to document, as it uses an ESP chip with in-package flash.

## Issues/PRs references

Depends on #1433.
Depends on #1367.

## Change checklist


kaspar030:
I pushed the updated c6 wifi support.

PR #1441, opened 2025-10-09
fix(rt): consider `executor-single-thread` for isr_stacksize

kaspar030:
# Description

On esp, with `executor-single-thread` mode, there's only one stack used. This needs to be taken into account for stack size calculations.

## Change checklist

### DIFF
--- a/src/ariel-os-rt/src/lib.rs
+++ b/src/ariel-os-rt/src/lib.rs
@@ -58,7 +58,7 @@ mod isr_stack {
             "ISR stack size (in bytes)"
         );
 
-        #[cfg(feature = "executor-interrupt")]
+        #[cfg(any(feature = "executor-interrupt", feature = "executor-single-thread"))]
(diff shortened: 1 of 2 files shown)


src/ariel-os-rt/src/lib.rs
```diff
@@ -69,7 +69,7 @@ mod isr_stack {
             CONFIG_ISR_STACKSIZE + CONFIG_EXECUTOR_STACKSIZE
         }
 
-        #[cfg(not(feature = "executor-interrupt"))]
+        #[cfg(not(any(feature = "executor-interrupt", feature = "executor-single-thread")))]
```
ROMemories, line 72: (As the CI failure requires a change anyway) Could we use `cfg_if!` instead here? I think it's clearer that having two negated conditions.

ROMemories:
Superseded by #1454.

PR #1369, opened 2025-09-18
feat(stm32): Add UART HAL support

bergzand:
# Description

This adds support for the UART HAL on the ST stm32 series devices

## Issues/PRs references

depends on #1365
Depends on #1456.
Depends on #1457.

## Open Questions

The Seeedstudio-lora-e5-mini uses USART2 for the SWI, while it is exposed on the board. I've added a comment to the support matrix, but the better solution would be to use something that is not exposed as SWI.

## Change checklist

- [ ] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1453, opened 2025-10-17
docs(support-matrix): add uart functionality status for some chips

jkr-ariel:
# Description

This PR adds uart functionality status for the ESP32-S3Fx8, ESP32-C3Fx4 and STM32U073KC chips which are currently missing it. They are marked as `needs_testing`.

## Change checklist

### DIFF
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -229,6 +229,7 @@ chips:
       hwrng: supported
       i2c_controller: supported
       spi_main: supported
+      uart: supported
       logging: supported
       storage:
         status: not_currently_supported
@@ -299,6 +300,7 @@ chips:
       hwrng: supported
       i2c_controller: needs_testing
       spi_main: needs_testing
+      uart: supported
       logging: supported
       storage:
         status: not_currently_supported


doc/support_matrix.yml
```diff
@@ -443,6 +445,7 @@ chips:
         comments:
           - removing items not supported
       wifi: not_available
+      uart: needs_testing
```
ROMemories: Will be added by #1369:
```suggestion
```

doc/support_matrix.yml
```diff
@@ -242,6 +242,7 @@ chips:
         status: not_available
         comments:
           - no generic USB peripheral
+      uart: needs_testing
```
ROMemories: Really the only difference with esp32c3 is that esp32c3fx4 has in-package flash:
```suggestion
      uart: supported
```
It would be better to move the `uart` key just after `spi_main` for consistency.

doc/support_matrix.yml
```diff
@@ -310,6 +311,7 @@ chips:
           - USB does not enumerate
           - "See also: https://github.com/ariel-os/ariel-os/issues/903"
       ethernet_over_usb: not_currently_supported
+      uart: needs_testing
```
ROMemories: Same:
```suggestion
      uart: supported
```

ROMemories (approved):

PR #1451, opened 2025-10-17
fix(esp): fixup some issues in preempt

kaspar030:
# Description

1. preempt.rs wasn't updated for changed thread `create_raw()` API
2. `THREAD_SEMAPHORE` was placed in flash due to not containing anything mutable
3. removed some ptr cast through `*const`

I tested this with `CONFIG_WIFI_NETWORK=... CONFIG_WIFI_PASSWORD=... laze -Cexamples/udp-echo build -b espressif-esp32-s3-devkitc-1 -s executor-thread run`.

## Issues/PRs references

Fixes #874 and #289.

## Change checklist

- [ ] I have cleaned up my commit history and squashed fixup commits.
- [ ] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [ ] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

PR #1454, opened 2025-10-17
docs(async-executor)!: remove the `executor-single-thread` from the book

ROMemories:
# Description

This removes the `executor-single-thread` executor flavor from the book, so its implementation can be removed in #1328.

BREAKING CHANGE: this removes a piece of functionality

## How to review this PR

Only the documentation change needs to be reviewed (see https://github.com/ariel-os/ariel-os/pull/1454#issuecomment-3670288715). The other commit will be removed before merging.

## Testing

Successfully tested `http-client` on espressif-esp32-c6-devkitc-1 (focused on ESPs as they were the only ones selecting `executor-single-thread ` by default).

## Issues/PRs references

- Depends on #1451
- Depends on #1455
- Supersedes #1441
- Depends on #1559

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2438,7 +2438,6 @@ source = "git+https://github.com/ariel-os/esp-hal?rev=aa8d1995fc06eb5e6d04a2022e
 dependencies = [
  "critical-section",
  "document-features",
- "embassy-executor",
  "embassy-sync 0.6.2",
  "embassy-time",
  "embassy-time-driver",
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -376,7 +376,6 @@ contexts:
       - ?debug-console
       - ?esp-println
     provides:
-      - has_executor_single_thread_support
       - has_hwrng
     env:
       esp_wifi_heapsize_required: $(72*1024)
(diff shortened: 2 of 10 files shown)


ROMemories:
I think the `.isr_stack` is simply not inserted at the right location on (plain) ESP32 and that it's not caused by this PR, this PR only makes this visible.

ROMemories:
I'm reliably getting a panic *after* the web page is fetched with the `http-client` example on ESP32-S3:

```
[…]
</body>
</html>
ERROR panicked at 'Unhandled level 1070278064 interrupt: WIFI_MAC'
panicked at [~]/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/defmt-1.0.1/src/lib.rs:385:5:
explicit panic
```

@kaspar030 Can you reproduce this?

kaspar030:
> I'm reliably getting a panic _after_ the web page is fetched with the `http-client` example on ESP32-S3:
> @kaspar030 Can you reproduce this?

Yes, I'm getting the same.

kaspar030:
> @kaspar030 Can you reproduce this?

I rebased #1328 on this, there it doesn't reproduce. I suggest we go with it as is.

kaspar030:
@ROMemories How about I ACK this here, we close it, and I cherry-pick it into #1328?

ROMemories:
> @ROMemories How about I ACK this here, we close it, and I cherry-pick it into #1328?

I've updated the PR to only keep the documentation change in this PR (for now the other implementation commit is still in this PR so you easily move it into #1328). Only the documentation change is a breaking change, so we can land this as early as possible, before landing #1328.

ROMemories:
Done as part of #1328.

PR #1459, opened 2025-10-20
docs(book): add `update-book` laze target

kaspar030:
# Description

PR creates a laze target for updating all generated book pages.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1757,6 +1757,16 @@ builders:
           - echo "Installing c2rust..."
           - 'echo "WARNING: This uses *a lot* of memory!"'
           - cargo install c2rust $@
+      update-book:
+        build: false
+        cmd:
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -1757,6 +1757,17 @@ builders:
           - echo "Installing c2rust..."
           - 'echo "WARNING: This uses *a lot* of memory!"'
           - cargo install c2rust $@
+      update-book:
+        build: false
+        cmd:
+          - echo "Updating book..."
```
bergzand: Does this really need an extra `echo`? The command is almost instant here.

kaspar030: nope, removed

bergzand (approved):

ROMemories:
@kaspar030 I refrained from doing this kind of thing before because I was fearing the task would also show up in OoT applications, where it doesn't make sense. Would it be possible to do that in a way that doesn't create a task for OoT applications? Does laze need a concept of "private tasks", that don't get imported?

Issue #1189, opened 2025-07-24
Tracking our crates patches


ROMemories:
Currently we patch some of our most central dependencies using [`[patch.crates-io]`](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) in [`ariel-os-cargo.toml`](https://github.com/ariel-os/ariel-os/pull/596).
Nightly Cargo is required for this file to be properly picked, as this relies on the unstable `config-include` Cargo feature.

This is an issue to track the list of patches we have, so we can work toward reducing it:

- [x] `embassy-embedded-hal`: https://github.com/embassy-rs/embassy/pull/4453, should be included in v0.4.0 even if not in the changelog
- [x] `embassy-executor`: v0.8.0 was released
- [x] `embassy-executor-macros`: just used by `embassy-executor`
- [ ] `embassy-hal-internal`: **`optfield` to provide `OptionalPeripherals`** (we may want to explore alternative strategies)
- [ ] `embassy-nrf`: `OptionalPeripherals`
- [ ] `embassy-net`: #1211
  - Changing IP address logs to `info`, from `debug`
  - https://github.com/ariel-os/ariel-os/pull/1188
- [ ] `embassy-rp`
  - `OptionalPeripherals`
  - **Multicore-related FIFO handler**
- [ ] `embassy-stm32`: `OptionalPeripherals`
- [ ] `embassy-time`: native (std) patches: https://github.com/embassy-rs/embassy/pull/4978
- [ ] `embassy-usb-synopsys-otg`: backported a [fix](https://github.com/embassy-rs/embassy/pull/5552) (merged upstream, waiting for release)
- [x] `try-lock`: https://github.com/seanmonstar/try-lock/pull/11 if we were to publish our own version of it, the deprecated, unsound `try_lock_order()` method could be removed. Forked in #1461
- [x] `embedded-test`: v0.7.0 was released

(Patches for ESP are not yet listed, feel free to add them. For Xtensa the nightly *compiler* is required anyway.)

In bold are the patches that still require action from us, to either upstream them or find alternative solutions.

chrysn:
We might want to group this into two tiers:

* Crates that are part of public APIs (of any other crate we use in different places)
* Crates that are only used internally, or in a coordinated set of crates that all take the patched version

[...]

PR #1461, opened 2025-10-21
chore: replace try-lock with version that supports portable-atomics

chrysn:
# Description

Some CoAP tests didn't build everywhere because those platforms need the portable-atomics fill.

We now have a soft-fork maintained in the project, https://crates.io/crates/try-lock-portable-atomic. Note that we don't *directly* depend on try-lock, but some crates I've written around CoAP do.

## Issues/PRs references

Contributes-to: https://github.com/ariel-os/ariel-os/issues/1189
… and if there is anywhere an issue about the trouble with try-lock, please let me know, I didn't find it.

## Change checklist

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -12,6 +12,12 @@ criteria = "safe-to-deploy"
 delta = "0.1.2 -> 0.1.3"
 notes = "Changes affect only a dependency version, whose safety is out of scope (vetted independently), and will not affect the operation of this crate."
 
+[[audits.try-lock-portable-atomic]]
+who = "chrysn <chrysn@fsfe.org>"
+criteria = "safe-to-deploy"
+version = "0.2.5-patch0"
+notes = "Note that there are several deprecated functions that are documented to not be sound and deprecated; users need to heed those deprecation warnings to use the crate soundly."
+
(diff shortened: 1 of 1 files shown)


kaspar030:
I guess we can also drop the ariel-os-cargo.toml entry then?

kaspar030:
> and if there is anywhere an issue about the trouble with try-lock, please let me know, I didn't find it

no issue, I just stumbled over this yesterday night testing #1055.

chrysn:
Note: @ROMemories requested in #1189: "if we were to publish our own version of it, the deprecated, unsound try_lock_order() method could be removed."

I'm not doing this here because doing that would make the soft fork way harder to maintain. If you insist, feel free to alter the fork, document those changes in PRs to upstream, and be prepared to be tagged every time try-lock undergoes an update.

… and I'm in the process of hunting down the required cargo-vet steps.

chrysn:
All vetted now, even with proper reviews. In hindsight, I should have uploaded try-lock 0.2.5 unmodified and then could have done a delta-review; in double hindsight I could probably have even done that after publishing 0.2.5.patch0. Anyway, I went through the whole crate and reviewed it. It's not long anyway.

PR #1462, opened 2025-10-22
docs(gen-book): add board page url to the matrix

bergzand:
# Description

This adds the URL for the board page to the boards in the support matrix.

## Issues/PRs references

Suggested by @bugadani

## Open Questions

None

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1463, opened 2025-10-22
feat: add board nrf9151-dk

chrysn:
# Description

Another board using nrf9151.

## Open Questions

* [x] <del>Not added to the support matrix (but then again, [the book](https://ariel-os.github.io/ariel-os/dev/docs/book/adding-board-support.html) doesn't say I should)</del>

## Changelog entry

The nRF9151-DK board is now supported.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -155,7 +155,7 @@ jobs:
               ;;
             'ci-build:nrf')
               # Purely redundant boards are excluded for now.
-              echo "LAZE_BUILDERS=bbc-microbit-v1,bbc-microbit-v2,nordic-thingy-91-x-nrf9151,nrf52840dk,nrf52dk,nrf5340dk,nrf5340dk-net,nrf9160dk-nrf9160" >> "$GITHUB_ENV"
+              echo "LAZE_BUILDERS=bbc-microbit-v1,bbc-microbit-v2,nordic-thingy-91-x-nrf9151,nrf52840dk,nrf52dk,nrf5340dk,nrf5340dk-net,nrf9160dk-nrf9160,nrf9151-dk" >> "$GITHUB_ENV"
               ;;
             'ci-build:rp')
               echo "LAZE_BUILDERS=rpi-pico,rpi-pico2,rpi-pico2-w,rpi-pico-w" >> "$GITHUB_ENV"


chrysn:
<del>Careful when testing so far: I'm locked out of the board after an Ariel flash (`WARN probe_rs::vendor::nordicsemi::sequences::nrf: Core 0 is locked. Erase procedure will be started to unlock it.`), and nrf-recover is not helping me.</del>

Adding probe-rs `--allow-erase-all` did help after the 2nd attempt, so while I yet have some discovery to do, at least things don't seem to get bricked.

nponsard:
> Careful when testing so far: I'm locked out of the board after an Ariel flash (`WARN probe_rs::vendor::nordicsemi::sequences::nrf: Core 0 is locked. Erase procedure will be started to unlock it.`), and nrf-recover is not helping me.

On the Thingy:91X I just end up spamming the flash command with `-- --allow-erase-all` appended at the end and after 3 or 4 attempts it flashes.

nponsard:
I also noticed that running `nrfutil device recover` before flashing makes it more reliable.

chrysn:
@nponsard, is that documented anywhere for thingy:91X?

If so, I'll be happy to add this here too.

If not, with the recent addition in the support matrix, I think this is done and ready for review.

nponsard:
> @nponsard, is that documented anywhere for thingy:91X?

I haven't found anywhere where this is properly documented.

chrysn:
Then let's figure this out together later; would you mind reviewing this?

doc/support_matrix.yml
```diff
@@ -610,6 +610,13 @@ boards:
       user_usb: supported
       ethernet_over_usb: supported
 
+  nrf9151-dk:
+    name: nRF9151 DK
```
nponsard: The boards above and below have a dash instead of a space before the `DK`. I think it would make more sense to rename this to `nRF9151-DK`

chrysn: This is following the vendors' website description as well as [hardware documentation](https://docs.nordicsemi.com/bundle/ug_nrf9151_dk/page/UG/nrf91_DK/intro.html). The guidance we (so far, silently) have used is [RDM0003 which says](github.com/RIOT-OS/RIOT/blob/master/doc/memos/rdm0003.md#:~:text=For%20the%20friendly%20name) which says that "vendor's name and the product name are taken verbatim from the vendor, including stylization and capitalization, as far as that is possible within Unicode".

But admittedly, the vendor uses the dash on the packaging of the kit (but not inside on the radio compliance sheet) and also uses a dash on the silkscreen. \*sigh\*.

Adjusted.

boards/nrf9151-dk.yaml
```diff
@@ -0,0 +1,29 @@
+boards:
+  nrf9151-dk:
+    chip: nrf9151
+    leds:
+      led1:
```
nponsard: leds start indexing at 0, not having a `led0` breaks `examples/blinky`

boards/nrf9151-dk.yaml
```diff
@@ -0,0 +1,29 @@
...
+      led3:
+        pin: P0_04
+        active: low
+        color: green
+      led4:
+        pin: P0_05
+        active: low
+        color: green
+    buttons:
+      button1:
```
nponsard: `buttons` also start indexing at 0, this breaks the `gpio` example.

nponsard (changes requested):

boards/nrf9151-dk.yaml
chrysn: This can't possibly be the first device with that issue (IIRC it's the same across the nRF DKs), looking up there…

nponsard: I think the names in the sbd are arbitrary and does not follow the labels on the board

boards/nrf9151-dk.yaml
chrysn: Resolving as duplicate of parallel discussion on buttons.

boards/nrf9151-dk.yaml
chrysn: As per [chat discussion](https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$z2kbkxbQNhLBP-FYliAlZcFEEUKNXFF8GyHmu-ipTno?via=matrix.org&via=utwente.io&via=club1.fr), resolving by setting the correct names as aliases and using synthetic values as keys. The plan as a whole is that sbd will drop those labels anyway.

boards/nrf9151-dk.yaml
```diff
@@ -0,0 +1,29 @@
+boards:
+  nrf9151-dk:
+    chip: nrf9151
+    leds:
+      led1:
+        pin: P0_00
+        active: low
```
kaspar030: the leds are active high: https://docs.nordicsemi.com/bundle/ug_nrf9151_dk/page/UG/nrf91_DK/hw_description/hw_buttons_leds.html "The LEDs are active high, meaning that writing a logical one (1) to the output pin illuminates the LED."

boards/nrf9151-dk.yaml
```diff
@@ -0,0 +1,29 @@
...
+        color: green
+      led3:
+        pin: P0_04
+        active: low
+        color: green
+      led4:
+        pin: P0_05
+        active: low
+        color: green
+    buttons:
```
kaspar030, line 25: the buttons are all active low:

"The buttons are active low, meaning that the input is connected to ground when the buttons are pushed. They have no external pull-up resistor, and therefore the P0.08, P0.09, P0.18, and P0.19 pins must be configured as an input with an internal pull-up resistor."

chrysn: (see other discussion)

boards/nrf9151-dk.yaml
chrysn: Thanks -- missed that in my tests because I had *both* wrong.

Also, I think this has no effect yet (at least nothing visible after sbd update).

nponsard:
Following [this discussion](https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$9tCFCIDCR5_JvEPWjqsj0T1zyTmUPRQijdAiqMJ1LZY?via=matrix.org&via=utwente.io&via=beeper.com) this board's name should be added to the builders in `.github/workflows/build.yml`, under `'ci-build:nrf')` in step `"limit build unless nightly build"`.

chrysn:
Thanks; all addressed, and rebased along with indentation fixes to make CI happy.

nponsard:
@chrysn looks like you have to regenerate the support matrix.

bergzand:
> @chrysn looks like you have to regenerate the support matrix.

Caused by the merge of #1462 :innocent:

chrysn:
Had to rebase to main and re-do all update-book after #1462 raced this to main and created merge conflicts in the generated files. [edit, correction: merge went fine but it was not the product of the merged generators]

PR #1465, opened 2025-10-22
feat: add board espressif-esp32-s2-devkitc-1

bugadani:
# Description

This PR adds support for the ESP32-S2 devkit, as well as the SOLO-2 module and ESP32-S2 in general. I tried to include the chip variants and the module on the devkit, to the level of how the S3 is supported.

I've tested the http-server and usb-serial examples.

## Changelog entry

The ESP32-S2, ESP32-S2Fx2, ESP32-S2Fx4, ESP32-S2Fx4R2 MCUs, the ESP32-S2-SOLO-2 hardware module, and the Espressif ESP32-S2-DevKitC-1 board are now supported.

## Change checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -379,6 +379,7 @@ contexts:
       - has_executor_single_thread_support
       - has_hwrng
     env:
+      esp_wifi_heapsize_required: $(72*1024)
       RUSTFLAGS:
         # linkall first
         - -Clink-arg=-Tlinkall.x
@@ -434,6 +435,33 @@ contexts:
 
   - name: esp32-c6-wroom-1
     parent: esp32c6
+
+  - name: esp32s2
+    parent: esp
+    selects:
+      - xtensa
+    provides:
+      - sw/benchmark
+    provides_unique: [c-function-abort]
+    env:
+      RUSTC_TARGET: xtensa-esp32s2-none-elf
+      PROBE_RS_CHIP: esp32s2
+      PROBE_RS_PROTOCOL: jtag
+      CARGO_TARGET_PREFIX: CARGO_TARGET_XTENSA_ESP32S2_NONE_ELF
+      CC: xtensa-esp32s2-elf-gcc
+      esp_wifi_heapsize_required: $(64*1024)
+
+  - name: esp32s2fx2
+    parent: esp32s2
+
+  - name: esp32s2fx4
+    parent: esp32s2
+
+  - name: esp32s2fx4r2
+    parent: esp32s2
+
+  - name: esp32-s2-solo-2
+    parent: esp32s2
 
   - name: esp32s3
     parent: esp
@@ -784,6 +812,7 @@ modules:
   - name: xtensa
     context:
       - esp32
+      - esp32s2
       - esp32s3
     provides_unique:
       - cargo-toolchain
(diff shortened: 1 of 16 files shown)


kaspar030:
> The chip triggers the following warning. It seems expect(unsafe_code) isn't working as expected.

This one I've seen before, unrelated to this PR.

kaspar030:
> * Should we consider the SOLO-2 and SOLO-2U modules to be different? They are essentially the same, but have different antenna options.

Let's go with one module!

> I had to reduce the esp-wifi heap requirement for examples to build, because the S2 has less RAM available. 

hm, upstream says somewhere "72k are needed for wifi"?

> I don't know how this could be done on a per-chip basis. 

You could add a variable `esp_wifi_heapsize_required` to both the `esp` and `esp-s2`context,
and in `esp-wifi`, change the literal to `${esp_wifi_heapsize_required}`. As `esp-s2` is a descendant of `esp`, laze will chose the right one.

> Additionally, the latest esp-hal releases allow placing heap to a currently (in Ariel OS) unused section of the RAM, helping with this issue a bit.

I've seen this and will use that in the branch that updates esp-hal.

laze-project.yml
```diff
@@ -1484,9 +1511,9 @@ modules:
         # esp-wifi needs a lot of ISR stack.
         isr_stacksize_required:
           - "32768"
-        # esp-wifi needs 72KiB
+        # esp-wifi needs a large-ish heap
         heapsize_required:
-          - $(72*1024)
+          - $(64*1024)
```
```suggestion
          - ${esp_wifi_heapsize_required}
```

laze-project.yml
```diff
@@ -435,6 +435,32 @@ contexts:
...
+      - xtensa
+    provides:
+      - sw/benchmark
+    provides_unique: [c-function-abort]
+    env:
+      RUSTC_TARGET: xtensa-esp32s2-none-elf
+      PROBE_RS_CHIP: esp32s2
+      PROBE_RS_PROTOCOL: jtag
+      CARGO_TARGET_PREFIX: CARGO_TARGET_XTENSA_ESP32S2_NONE_ELF
+      CC: xtensa-esp32s2-elf-gcc
```
kaspar030, line 451:
```suggestion
      CC: xtensa-esp32s2-elf-gcc
      esp_wifi_heapsize_required: $(64*1024)
```

laze-project.yml
kaspar030, line 388:
```suggestion
		esp_wifi_heapsize_required: $(72*1024)
```

(watch indentation, gh interface is not helping)

laze-project.yml
bugadani: NGL that this requires a curly braces instead of parentheses was very hard to notice 😅

kaspar030:
LGTM, I guess this is ready for review?

bugadani:
Depends a bit on https://github.com/ariel-os/ariel-os/pull/1464 (the first few commits are that PR just to simplify my testing), whether you want them separated or not better, or if this should be rebased off of that one.

bugadani:
I've removed the first two commits, this should now be ready unless I've messed up in the process.

kaspar030:
Thanks a bunch @bugadani!

PR #1467, opened 2025-10-23
fix(esp): add missing peripheral support for esp32s2

kaspar030:
# Description

#1465 added basic esp32s2 support, but did not wire up the peripherals support.

Also, disables the http-client for now, there's not enough memory.

Compile-tested with `CONFIG_WIFI_PASSWORD=doesnt CONFIG_WIFI_NETWORK=matter laze build -b espressif-esp32-s2-devkitc-1 -g`.

## Change checklist

### DIFF
--- a/examples/http-client/laze.yml
+++ b/examples/http-client/laze.yml
@@ -9,3 +9,5 @@ apps:
       - random
     conflicts:
       - ram-small
+      # not enough ram
+      - context::esp32s2
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -2,7 +2,10 @@
 
 #![cfg_attr(not(any(test, feature = "std")), no_std)]
 #![cfg_attr(test, no_main)]
-#![cfg_attr(not(target_has_atomic = "ptr"), expect(unsafe_code))]
+#![cfg_attr(
+    all(feature = "log", not(target_has_atomic = "ptr")),
+    expect(unsafe_code)
+)]
 #![deny(missing_docs)]
 
 #[featurecomb::comb]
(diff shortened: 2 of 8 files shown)


src/ariel-os-debug/src/lib.rs
```diff
@@ -2,7 +2,10 @@
 
 #![cfg_attr(not(any(test, feature = "std")), no_std)]
 #![cfg_attr(test, no_main)]
-#![cfg_attr(not(target_has_atomic = "ptr"), expect(unsafe_code))]
+#![cfg_attr(
```
kaspar030, line 5: piggyback

src/ariel-os-esp/src/i2c/controller/mod.rs
```diff
@@ -115,7 +115,8 @@ macro_rules! define_i2c_drivers {
                     twim_config.frequency = config.frequency.into();
                     #[cfg(any(context = "esp32c3", context = "esp32c6", context = "esp32s3"))]
                     let disabled_timeout = BusTimeout::Disabled;
-                    #[cfg(context = "esp32")]
+                    // TODO: use `BusTimeout::Disabled` on s2 after esp-hal bump to 1.0.0
```
kaspar030, line 118: upstream has `BusTimeout::Disabled` for the s2 already, but not in our branch

examples/http-client/laze.yml
```diff
@@ -9,3 +9,5 @@ apps:
       - random
     conflicts:
       - ram-small
+      # not enough ram
+      - context::esp32s2
```
nponsard, line 13: This is concerning, the esp32s2 has 320KB of RAM, this sounds like it would be enough (the nrf52840 runs this example and has 256KB of RAM)

kaspar030: Well, there's a bunch of RAM (~70k IIUC) that is untapped (dram2_seg?), but will be used once we upgrade esp-hal. Here we're missing ~12k RAM, so this is temporary.

Also, esp is using a full wifi stack. That just needs more RAM.

nponsard: I see, I hope the esp-hal upgrade will allow us to use HTTP on this chip.

src/ariel-os-debug/src/lib.rs
nponsard: I would have preferred this to be in another commit, but this is useful it removes a warning appearing in some builds.

PR #1470, opened 2025-10-24
chore: bump rand crates

bugadani:
# Description

This PR extracts the version bump from https://github.com/ariel-os/ariel-os/pull/1328, and adds an adapter type for rand_core 0.6 implementations, which is needed because the currently supported HALs do not implement rand_core 0.9.

## Change checklist

### DIFF
--- a/src/ariel-os-native/src/hwrng.rs
+++ b/src/ariel-os-native/src/hwrng.rs
@@ -1,4 +1,27 @@
+/// Implement `RngCore` on top of an older `getrandom` crate.
+struct Getrandom02Rng;
+
+impl rand::rand_core::RngCore for Getrandom02Rng {
+    fn next_u32(&mut self) -> u32 {
+        let mut buf = [0; 4];
+        self.fill_bytes(&mut buf);
+        u32::from_le_bytes(buf)
+    }
+
+    fn next_u64(&mut self) -> u64 {
+        let mut buf = [0; 8];
+        self.fill_bytes(&mut buf);
+        u64::from_le_bytes(buf)
+    }
+
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        if let Err(e) = getrandom::getrandom(dest) {
+            panic!("Error: {}", e);
+        }
+    }
+}
+
 /// Constructs the hardware random number generator (RNG).
 pub fn construct_rng(_peripherals: &mut crate::OptionalPeripherals) {
-    ariel_os_random::construct_rng(rand::rngs::OsRng::default());
+    ariel_os_random::construct_rng(&mut Getrandom02Rng);
 }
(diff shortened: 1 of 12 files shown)


src/ariel-os-native/src/hwrng.rs
```diff
@@ -1,4 +1,27 @@
+/// Implement `RngCore` on top of an older `getrandom` crate.
+struct Getrandom02Rng;
```
bugadani, line 2: This is required because `OsRand` in rand_core 0.9 relies on `getrandom@0.3`. Because `ariel_os_random` defines its own `getrandom` backend (for `csprng`), the result would compile, but panic in runtime.

This hasn't been a problem until now because the old version of rand_core still depended on `getrandom@0.2`

chrysn: Any reason not to just also update the native crate's getrandom dependency? (I'm assembling a branch with a few minor enhancements rather than ping-ponging, so I'm looking more for a "no because" than us editing simultaneously)

chrysn: Hm, leaving this on 0.2 (as it was) neatly sidesteps an issue we're facing when updating everything: Our native backend wants getrandom to full run own "fake" HWRNG, but we also provide getrandom. So our options on native (other than always using an outdated getrandom as a backend and the latest to be provided) are:
1. Work around the `__getrandom_v03_custom` name and use a different source (we can't easily make our own `OsRng` use a different getrandom symbol and make the host side provide that, but maybe we could more directly tap into Linux)
2. Not provide getrandom ourselves on native (and thus not set getrandom_backend="custom"), but use it as our "HWRNG"; crates that do use getrandom then will find themselves not using Ariel's RNG but really the OS's. As both are supposed to be crytpo secure, and native doesn't give any latency guarantees, that should all work just as fine, and I think that's what I'll try.

(I'd prefer if we could do 1 over 2, just to get more similar code paths, but if that means getting less portability for native, then maintenance effort wins -- and native *will* be a bit of a mixed bag, possibly selectable, in how far it uses the host and Ariel facilities anyway.)

bugadani: I'm unsure if you're asking me to do something specific, but I don't feel familiar enough with the codebase to try and redesign the involved code.

chrysn: No need to, I just wanted to check whether there was anything you explored and then decided against.

Updating all rand traits means that what we used to have as a happy coincidence doesn't work any more; your workaround restores functionality, and I'm exploring how to restructure things to not need a workaround (in is probably better suited for a follow-up PR that I'm preparing while reviewing yours).

chrysn (approved):
Concluding my review, the `RngAdapter` is a workaround that we need for these updates, and fixing which is beyond the scope of the update (esp. from an external contributor).

So: Thanks, I'm approving this PR. It doesn't pass CI due to an upgraded dependency that'd need vetting, but I did that in #1477 that is a follow-up to this, so once that gets merged, this one will be merged along with it automatically.

PR #1477, opened 2025-10-29
test(getrandom): add test and fix small issues

chrysn:
# Description

<del>During https://github.com/ariel-os/ariel-os/pull/1470 it became apparent that we are using different versions of getrandom to back the native RNG (0.2) and to provide (0.3).</del>

<del>While this conveniently solved the issue of keeping apart our provided APIs and the host system, this is not sustainable and required workarounds introduced in #1470.</del>

This PR takes up things from there that should be fixed but also don't quite fit in there.

The big "fix native" is a larger PR than I want to hold #1470 for -- so this one just adds a test, and fixes vetting issues from #1470. The first commit in the series is from there (and IMO good to go); please review my changes on top of it so they can go in.

## Issues/PRs references

Follow-up-for: #1470
Follow-up-for: #1339

## Necessary-follow-ups

During https://github.com/ariel-os/ariel-os/pull/1470 it became apparent that we are using different versions of getrandom to back the native RNG (0.2) and to provide (0.3).

While this conveniently solved the issue of keeping apart our provided APIs and the host system, this is not sustainable and required workarounds introduced in #1470.

This still sits on my stash on top of this commit, and will be a follow-up – but I maintain as in [my ACK there](https://github.com/ariel-os/ariel-os/pull/1470#pullrequestreview-3392560646) that the workaround introduced is justified for the moment.

[edit] The follow-up is #1480.

## Change checklist

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -6,6 +6,11 @@ who = "chrysn <chrysn@fsfe.org>"
 criteria = "safe-to-deploy"
 delta = "0.4.27 -> 0.4.28"
 
+[[audits.rand_pcg]]
+who = "chrysn <chrysn@fsfe.org>"
+criteria = "safe-to-deploy"
+delta = "0.3.1 -> 0.9.0"
+
 [[audits.scroll-ring]]
 who = "chrysn <chrysn@fsfe.org>"
 criteria = "safe-to-deploy"
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -2285,6 +2285,26 @@ who = "David Cook <dcook@divviup.org>"
 criteria = "safe-to-deploy"
 version = "0.3.0"
 
+[[audits.isrg.audits.rand]]
+who = "David Cook <dcook@divviup.org>"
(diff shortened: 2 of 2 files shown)


tests/random-getrandom/README.md
```diff
@@ -0,0 +1,8 @@
+# random
+
+## About
+
+This application is testing whether getrandom works.
+
+Its main value in testing is at the linker stage,
+as the native board's getrandom as provided by Ariel OS may interfere with the host side getrandom implementation.
```
nponsard, line 8: I don't understand why there's a line break here.

nponsard (commented):
Small comment about the test readme.
Otherwise looks good to me, the test works on my esp32c6.

tests/random-getrandom/README.md
chrysn: [Semantic linebreaks](https://sembr.org/). Have been discussed time and again on chat, they make diffs (and I argue also source) easier to read and do not affect the resulting HTML.

nponsard: Oh okay, thanks for the link !

PR #1483, opened 2025-10-31
docs(book): document BLE usage

nponsard:
# Description

This PR adds a book page describing how to use and configure BLE in Ariel OS.

## Change checklist

### DIFF
--- a/book/src/bluetooth.md
+++ b/book/src/bluetooth.md
@@ -1,0 +1,56 @@
+# Bluetooth Low Energy (BLE)
+
+Ariel OS supports BLE through [TrouBLE][trouble-github].
+The hardware-specific initialization is handled by Ariel OS.
+
+## Enabling the BLE Stack
+
+BLE is enabled by selecting one or both of the BLE [laze modules][laze-modules-book] depending on your use case:
+
+- `ble-peripheral`: peripheral BLE role, allows for advertising the device's presence.
+- `ble-central`: central BLE role, allows for scanning and creating connections to devices.
+
+If you want to learn more about BLE concepts, you can read the [TrouBLE documentation][trouble-doc].
+
+## Configuring the BLE Stack
+
+The ability to configure which Bluetooth address is used and other capacity parameters like the MTU is planned in future updates.
+
+> [!IMPORTANT]
+> For compatibility reasons the MTU is fixed at 27 bytes and the address is fixed as a static random address of `FF:E4:05:1A:8F:FF`.
+
+## Using the BLE Stack
+
+To use BLE in your application, you need to get the pre-configured stack and build the host instance from it:
+
+```rust
+let stack = ariel_os::ble::ble_stack().await;
+let host = stack.build();
+```
+
+Here `stack` is of type [`trouble_host::Stack`][trouble-host-stack] and `host` is of type [`trouble_host::Host`][trouble-host-host].
+
+You then need to run the runner background task in your application.
+
+```rust
+host.runner.run().await
+```
+
+> [!NOTE]
+> The runner has other options like `run_with_handler(<handler>)` that you may want to use (when scanning for example).
+
+This function will return a result only when a critical error happened in the BLE stack. For most use cases you will need to execute code alongside the runner in the same task (that accesses elements of the `stack` or `host`), you can do that using `embassy_futures::join::join`:
+
+```rust
+embassy_futures::join::join(host.runner.run(), async {
+  // Execute BLE-related code here that uses `host` or `stack`
+}).await;
+```
+
+See [the Ariel OS examples](https://github.com/ariel-os/ariel-os/tree/main/examples) for more details.
+
+[laze-modules-book]: ./build-system.md#laze-modules
+[trouble-github]: https://github.com/embassy-rs/trouble
+[trouble-doc]: https://embassy.dev/trouble/#_concepts
+[trouble-host-stack]: https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Stack.html
+[trouble-host-host]: https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Host.html
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -80,6 +80,7 @@
 - [Async Executors](./async-support.md)
 - [Debug Console](./debug-console.md)
 - [Networking](./networking.md)
+- [Bluetooth Low Energy](./bluetooth.md)
 - [Randomness and Entropy](./randomness.md)
 - [Multithreading](./multithreading.md)
 - [Persistent Storage](./storage.md)
--- a/book/templates/SUMMARY.md.tmpl
+++ b/book/templates/SUMMARY.md.tmpl
@@ -22,6 +22,7 @@
 - [Async Executors](./async-support.md)
 - [Debug Console](./debug-console.md)
 - [Networking](./networking.md)
+- [Bluetooth Low Energy](./bluetooth.md)
 - [Randomness and Entropy](./randomness.md)
 - [Multithreading](./multithreading.md)
 - [Persistent Storage](./storage.md)


book/src/ble.md
```diff
@@ -0,0 +1,55 @@
+## Bluetooth Low Energy
+
+Ariel os provides an abstraction over [TrouBLE](https://github.com/embassy-rs/trouble). The hardware specific initialization is handled by Ariel OS for supported chips.
+
+### Enabling the BLE stack
+
+BLE is enabled by selecting one or both of the BLE [laze modules][laze-modules-book] depending on your use case:
+
+- `ble-peripheral`: for advertising data, acting as a "sever".
```
ROMemories:
```suggestion
- `ble-peripheral`: for advertising data, acting as a "server".
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
+## Bluetooth Low Energy
+
+Ariel os provides an abstraction over [TrouBLE](https://github.com/embassy-rs/trouble). The hardware specific initialization is handled by Ariel OS for supported chips.
```
ROMemories: I don't think we really provide an *abstraction over* TrouBLE. TrouBLE provides an implementation, that we expose. I've removed "for supported chips" because it could lead to believe that initialization is not provided in some cases:
```suggestion
Ariel OS supports BLE through [TrouBLE](https://github.com/embassy-rs/trouble).
The hardware-specific initialization is handled by Ariel OS.
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
+## Bluetooth Low Energy
```
ROMemories: Heading levels also need to be decremented in the whole file.

```suggestion
# Bluetooth Low Energy (BLE)
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+
+Ariel os provides an abstraction over [TrouBLE](https://github.com/embassy-rs/trouble). The hardware specific initialization is handled by Ariel OS for supported chips.
+
+### Enabling the BLE stack
```
ROMemories: Title case:
```suggestion
## Enabling the BLE Stack
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+
+BLE is enabled by selecting one or both of the BLE [laze modules][laze-modules-book] depending on your use case:
+
+- `ble-peripheral`: peripheral BLE role, allows for advertising the device's presence.
+- `ble-central`: central BLE role, allows for scanning and creating connections to devices.
+
+If you want to learn more about BLE concepts, you can read the [TrouBLE documentation](https://embassy.dev/trouble/#_concepts).
+
+### Using the BLE stack
```
ROMemories:
```suggestion
## Using the BLE Stack
```

book/src/bluetooth.md
ROMemories, line 1: I think this file should be renamed to `bluetooth.md` in case some time in the future we happen to ship Bluetooth Classic support. That's important to do now because this ends up in the URL. It's fine to keep the page title as-is however.

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+
+To use BLE in your application, you need to get the pre-configured stack and build the host object from it:
```
ROMemories:
```suggestion
To use BLE in your application, you need to get the pre-configured stack and build the host instance from it:
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+
+```rust
+let stack = ariel_os::ble::ble_stack().await;
+let host = stack.build();
+```
+
+Here `stack` is of type [`trouble_host::Stack`](https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Stack.html) and `host` is of type [`trouble_host::Host](https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Host.html)
```
ROMemories:
```suggestion
Here `stack` is of type [`trouble_host::Stack`](https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Stack.html) and `host` is of type [`trouble_host::Host`](https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Host.html).
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+
+You then need to run the runner background task in your application.
+
+```rust
+host.runner.run().await;
+```
+
+> Note: the runner has other options like `run_with_handler(&lt;handler&gt;) that you may want to use (when scanning for example).
```
ROMemories: Available since #1502 (not sure about the other corrections):
```suggestion
> [!NOTE]
> The runner has other options like `run_with_handler(<handler>)` that you may want to use (when scanning for example).
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+
+This function will return a result only when a critical error happened in the BLE stack. For most use cases you will need to execute code alongside the runner in the same task (that accesses elements of the `stack` or `host`), you can do that using `embassy_futures::join::join`:
+
+```rust
+embassy_futures::join::join(host.runner.run(), async {
+  // Execute BLE-related code here that use `host` or `stack`
```
ROMemories:
```suggestion
  // Execute BLE-related code here that uses `host` or `stack`
```

book/src/ble.md
```diff
@@ -0,0 +1,43 @@
...
+}).await;
+```
+
+See [examples](https://github.com/ariel-os/ariel-os/tree/main/examples) for more details.
```
ROMemories:
```suggestion
See [the Ariel OS examples](https://github.com/ariel-os/ariel-os/tree/main/examples) for more details.
```

book/src/ble.md
ROMemories: Generally in the book we use [link reference definitions](https://spec.commonmark.org/0.31.2/#link-reference-definitions) ("link footnotes") now to avoid breaking the reading flow too much (not every place in the book may be doing this yet however).

book/src/bluetooth.md
```diff
@@ -0,0 +1,43 @@
...
+let stack = ariel_os::ble::ble_stack().await;
+let host = stack.build();
+```
+
+Here `stack` is of type [`trouble_host::Stack`](https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Stack.html) and `host` is of type [`trouble_host::Host](https://docs.embassy.dev/trouble-host/0.1.0/default/struct.Host.html)
+
+You then need to run the runner background task in your application.
+
+```rust
+host.runner.run().await;
```
ROMemories: I don't think we should encourage to ignore the returned `Result` (these things tend to be cargo-culted, [C-QUESTION-MARK](https://rust-lang.github.io/api-guidelines/documentation.html#c-question-mark)):
```suggestion
host.runner.run().await
```

book/src/bluetooth.md
nponsard: I renamed the file but kept the name "Bluetooth Low Energy" in `SUMMARY.md`

book/src/bluetooth.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Enabling the BLE Stack
+
+BLE is enabled by selecting one or both of the BLE [laze modules][laze-modules-book] depending on your use case:
+
+- `ble-peripheral`: peripheral BLE role, allows for advertising the device's presence.
+- `ble-central`: central BLE role, allows for scanning and creating connections to devices.
+
+If you want to learn more about BLE concepts, you can read the [TrouBLE documentation][trouble-doc].
+
```
ROMemories, line 14: We should say something about the configuration, even if it means saying that it's not currently supported in any way:
```suggestion
## Configuring the BLE Stack
```

This page should also say something about the `BD_ADDR` (aka the BLE MAC address) which is used by default.

ROMemories, line 14: In the same vein, should the `ble-config-override` Cargo feature (and the corresponding `#[config]` macro parameter) be documented now? If so that should be done *before* landing this PR.

nponsard: `ble-config-override` is actually broken as the `trouble_host::prelude::BdAddr` cannot be constructed with const functions. That's why I'm replacing it in #1469

nponsard: Added a section explaining plans to allow configuration of some BLE parameters.

book/src/bluetooth.md
```diff
@@ -0,0 +1,55 @@
...
+## Configuring the BLE Stack
+
+The ability to configure which Bluetooth address is used and other capacity parameters like the MTU is planned in future updates.
+
+For compatibility reasons the MTU is fixed at 27 bytes and the address is fixed as a static random address of `FF:E4:05:1A:8F:FF`
```
ROMemories: I think this warrants some more visibility, as that means it's not (yet) usable in production:
```suggestion
> [!IMPORTANT]
> For compatibility reasons the MTU is fixed at 27 bytes and the address is fixed as a static random address of `FF:E4:05:1A:8F:FF`.
```

PR #1484, opened 2025-11-01
chore: update nightly version

chrysn:
# Description

Many modern crates don't even work with the nightlies any more (eg. wasmtime, but also CoAP crates).

We should do this more regularly anyway. It's probably fine to have this pinned so that nightly changes don't break CI immediately, but then we should still not let it fall behind.

## Change checklist


chrysn:
Most of the fixes are self-explanatory and address new things that cargo-doc and clippy learned. The doc link fix is noteworthy because that issue was really not caught even though our CI is configured to be rather strict ([the current docs](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os_nrf/uart/struct.UARTE0.html#method.new) show literal square brackets), and the new cargo-doc just detects the issue.

PR #1487, opened 2025-11-02
feat(coapcore): send credential by value if requested

chrysn:
# Description

By default, coapcore sends credentials only by reference to keep data low.

When a client does *not* know the credentials yet (eg. because it is doing TOFU or unilateral authentication), [draft-ietf-ace-edhoc-oscore-profile](https://www.ietf.org/archive/id/draft-ietf-ace-edhoc-oscore-profile-09.html#name-requesting-authentication-c) has us covered with an EAD option, support for which is added here: If the option is present, the full credential is sent.

## Issues/PRs references

This only has an effect with a client that sends it, eg. aiocoap after its [!44](https://codeberg.org/aiocoap/aiocoap/pulls/44).

Therefore, the README, examples and credentials are not yet updated to do unilateral authentication more often.

Adding this allows simplifying the trevm CoAP example, as we can then do away with copying the server's credential, as for a firmware upload, unilateral authentication can be fine.

## Change checklist

### DIFF
--- a/src/lib/coapcore/src/iana.rs
+++ b/src/lib/coapcore/src/iana.rs
@@ -13,6 +13,13 @@ pub(crate) mod edhoc_ead {
     /// [Section 10.8 of the ACE-EDHOC
     /// profile](https://datatracker.ietf.org/doc/html/draft-ietf-ace-edhoc-oscore-profile-06#section-10.8).
     pub(crate) const ACETOKEN: u16 = 20;
+
+    /// (Requesting an) EAD Credential by value
+    ///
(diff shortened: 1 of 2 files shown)


chrysn:
Marking as self-approved as it's a coapcore change, scope is limited and behavior differs only when triggered.

As a reviewer I might have asked about whether sending a credential by value can have privacy implications, but that is explained in general in EDHOC: The responder is the party with the less vulnerable identity. We could make this configurable, but generally, unless all keys are known in advance, one party has to show their identity at least to an active attacker, and that is the responder who is now opening up here.

PR #1489, opened 2025-11-03
docs(support): document the storage support caveat on STM32WLE5JC

ROMemories:
# Description

This adds the missing caveat on storage support, as for every other STM32 chip, since the `embassy-stm32` flash driver doesn't implement `MultiwriteNorFlash`.

## Change checklist


ROMemories:
Self-approving as the change is only docs and trivial.

PR #1493, opened 2025-11-04
docs(adding-board): document adding the board to the list built in CI

ROMemories:
# Description

Documents that new boards need to be added to the list of builders for them to be built by the CI.

## Change checklist

### DIFF
--- a/book/src/adding-board-support.md
+++ b/book/src/adding-board-support.md
@@ -61,6 +61,12 @@ sbd-gen generate-ariel boards -o src/ariel-os-boards --mode update
 > CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build --global -b <builder>
 > ```
 
+<div class="warning">
+
+When adding support for a board *in tree*, its name needs to be added to the [list of builders of its corresponding MCU family][ci-builder-lists], so that it gets appropriatetly built for in CI.
+
+</div>
+
 ## Extra Steps for Some MCU Families
 
 ### `stm32`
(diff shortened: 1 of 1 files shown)


book/src/adding-board-support.md
```diff
@@ -61,6 +61,12 @@ sbd-gen generate-ariel boards -o src/ariel-os-boards --mode update
 > CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build --global -b <builder>
 > ```
 
+<div class="warning">
+
+When adding support for a board *in tree*, its name needs to be added to the [list of builders of its corresponding MCU family][ci-builder-lists], so that it gets appropriatetly built for in CI.
+
```
ROMemories, line 67: The blank lines are needed so that the inner paragraph is interpreted as Markdown.

bergzand:
What's the reason for rendering this as a warning? In my opinion it is sufficient to inform the developer that this should be done in a regular paragraph.

ROMemories:
> What's the reason for rendering this as a warning?

I think there's two reasons: this is only needed when adding support for a board *in tree*, it's not intrinsic to the process of adding a board itself, so it's a way of marking that as an aside and, because it follows a blockquote, I was fearing this would otherwise be missed if only shown as regular text. Forgetting to add the new board to the list of builders means it will silently *not* be compiled for, so I thought that warranted a warning.

bergzand:
> I was fearing this would otherwise be missed if only shown as regular text.

To be honest, I missed it when looking through the book render, and had to go back and look where the warning was supposed to show up to find it. :sweat_smile:

PR #1498, opened 2025-11-05
docs(gen-book): add link to chip page from board page

jkr-ariel:
# Description

This PR adds a link to the chip page from the corresponding board page.

## Issues/PRs references

Follow up to #1428.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1501, opened 2025-11-05
docs(gen-book): add link to chip page from overview tables

jkr-ariel:
# Description

This PR adds a link to the chip page from the Ariel OS chip name column in the overview tables.

## Issues/PRs references

Follow up to #1498.

## Change checklist

### DIFF
--- a/book/src/support_matrix_tier1.html
+++ b/book/src/support_matrix_tier1.html
@@ -27,7 +27,7 @@
   <tbody>
     <tr>
       <td>nRF52833</td>
-      <td><a href="/chips/nrf52833.html"><code>nrf52833</code></a></td>
+      <td><a href="chips/nrf52833.html"><code>nrf52833</code></a></td>
       <td><a href="https://web.archive.org/web/20250109121140/https://microbit.org/new-microbit/">BBC micro:bit V2</a></td>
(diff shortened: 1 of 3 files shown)


doc/gen_book.rs
```diff
@@ -670,7 +671,8 @@ fn gen_board_functionalities(matrix: &schema::Matrix) -> Result<Vec<BoardSupport
                     comments,
                 })
             });
-            let board_doc_page = ["boards/", board_technical_name, ".html"].concat();
+            let board_doc_page = ["/boards/", board_technical_name, ".html"].concat();
```
jkr-ariel: I added a `/` here to make the URL path absolute. Not including it is not a problem (yet) because the generated html file is at the root of the directory structure, so the relative and absolute paths are equivalent in this case.

ROMemories: That doesn't work though, because the book is not hosted at the root of the website. It might be possible to make it work by fiddling with [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/base) (maybe there's an mdbook feature for that?), but I'm not sure it's worth the trouble. Same for the chip links.

PR #1512, opened 2025-11-07
ci(build): avoid the need for listing every `rp` laze builder

ROMemories:
# Description

This revisits how we select the set of laze builders for the `ci-build:rp` PR build label, to avoid having to manually list the laze builders.

## Issues/PRs references

Depends on #1513.

## Change checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -143,6 +143,11 @@ jobs:
       - name: "limit build unless nightly build"
         if: github.event_name != 'schedule' && steps.should-skip.outputs.result != 'true'
         run: |
+          allow_context() {
+            hal_contexts='context::esp,context::host,context::native,context::nrf,context::rp,context::stm32'
+
+            echo "$hal_contexts" | tr ',' '\n' | grep -v "^context::${1}$" | paste -sd ','
+          }
           case '${{ needs.check-labels.outputs.label }}' in
             'ci-build:esp')
               echo "NEEDS_XTENSA_TOOLCHAIN=1" >> "$GITHUB_ENV"
@@ -160,7 +165,7 @@ jobs:
               echo "LAZE_BUILDERS=bbc-microbit-v1,bbc-microbit-v2,nordic-thingy-91-x-nrf9151,nrf52840dk,nrf52dk,nrf5340dk,nrf5340dk-net,nrf9160dk-nrf9160,nrf9151-dk" >> "$GITHUB_ENV"
               ;;
             'ci-build:rp')
-              echo "LAZE_BUILDERS=rpi-pico,rpi-pico2,rpi-pico2-w,rpi-pico-w" >> "$GITHUB_ENV"
+              echo "LAZE_DISABLE=$(allow_context rp)" >> "$GITHUB_ENV"
               ;;
             'ci-build:small')
               echo "NEEDS_XTENSA_TOOLCHAIN=1" >> "$GITHUB_ENV"
@@ -249,6 +254,8 @@ jobs:
         if: steps.should-skip.outputs.result != 'true'
         run: |
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
+          echo "LAZE_DISABLE=${LAZE_DISABLE}"
+          laze build -mg info-boards | uniq
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }}
 
       # cleanup previous build artifacts. do in background (no need to wait).


.github/workflows/build.yml
ROMemories: This would need to be revisited.

.github/workflows/build.yml
```diff
@@ -158,7 +158,7 @@ jobs:
               echo "LAZE_BUILDERS=bbc-microbit-v1,bbc-microbit-v2,nordic-thingy-91-x-nrf9151,nrf52840dk,nrf52dk,nrf5340dk,nrf5340dk-net,nrf9160dk-nrf9160,nrf9151-dk" >> "$GITHUB_ENV"
               ;;
             'ci-build:rp')
-              echo "LAZE_BUILDERS=rpi-pico,rpi-pico2,rpi-pico2-w,rpi-pico-w" >> "$GITHUB_ENV"
+              echo "LAZE_DISABLE=context::esp,context::host,context::native,context::nrf,context::stm32" >> "$GITHUB_ENV"
```
kaspar030: maybe have one list of contexts and then do sth like this:

`echo "word1 word2 word3" | tr ' ' '\n' | grep -v word2 | paste -sd ','`

.github/workflows/build.yml
kaspar030: This will go when we switch to self-hosted workers. until then, I think we could just do this even if no esps are built.

.github/workflows/build.yml
```diff
@@ -158,7 +159,7 @@ jobs:
               echo "LAZE_BUILDERS=bbc-microbit-v1,bbc-microbit-v2,nordic-thingy-91-x-nrf9151,nrf52840dk,nrf52dk,nrf5340dk,nrf5340dk-net,nrf9160dk-nrf9160,nrf9151-dk" >> "$GITHUB_ENV"
               ;;
             'ci-build:rp')
-              echo "LAZE_BUILDERS=rpi-pico,rpi-pico2,rpi-pico2-w,rpi-pico-w" >> "$GITHUB_ENV"
+              echo "LAZE_DISABLE=$(echo "$hal_contexts" | tr ' ' '\n' | grep -v 'context:rp' | paste -sd ',')" >> "$GITHUB_ENV"
```
kaspar030:
```suggestion
              echo "LAZE_DISABLE=$(echo "$hal_contexts" | tr ',' '\n' | grep -v 'context:rp' | paste -sd ',')" >> "$GITHUB_ENV"
```

kaspar030: how about moving the guts into a function?

```sh
allow_context() {
hal_contexts='context::esp,context::host,context::native,context::nrf,context:rp,context::stm32'
  
  echo "$hal_contexts" | tr ',' '\n' | grep -v "context::$1" | paste -sd ',')"
}
```

And then `LAZE_DISABLE=$(allow_context rp) >> $GITHUB_ENV`.
Maybe even move the assignment into the function.

.github/workflows/build.yml
```diff
@@ -246,6 +247,8 @@ jobs:
         if: steps.should-skip.outputs.result != 'true'
         run: |
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
+          echo "LAZE_DISABLE=${LAZE_DISABLE}"
+          laze build -g info-boards
```
ROMemories: I'm not sure why this errors out, that would be a useful debugging tool.

kaspar030: you'll need to add `-m`/`--multiple` I think

kaspar030: and pipe through `uniq` :/

ROMemories: That's also very slow (when running locally)

.github/workflows/build.yml
ROMemories: Added the function, but that'll need revisiting if we want to support additive labels later.

.github/workflows/build.yml
ROMemories: We could just introduce a `NEEDS_XTENSA_TOOLCHAIN` env var in the switch-case, and use that here (that would even be more readable). Installing the toolchain takes a bit more than one minute for each job, which is not nothing.

ROMemories: Introduced in #1513.

.github/workflows/build.yml
kaspar030: same here. it is slow to start with b/c for every build an "echo" is executed, sequentially. This'll get better with https://github.com/kaspar030/laze/pull/807 ...

.github/workflows/build.yml
```diff
@@ -246,6 +254,8 @@ jobs:
         if: steps.should-skip.outputs.result != 'true'
         run: |
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
+          echo "LAZE_DISABLE=${LAZE_DISABLE}"
+          laze build -mg info-boards | uniq
```
kaspar030, line 258: so you'd like to keep this in?

ROMemories: I mean it's "slow" but only costs < 1 s on my machine. However I can't easily spot it in the CI logs, but maybe it's lost in the sea of "Caused by:".

kaspar030: ok, let's keep it for now.

(It takes ~2s on my machine, which is supposed to be quite fast actually.)

kaspar030 (approved):

.github/workflows/build.yml
```diff
@@ -143,6 +143,11 @@ jobs:
       - name: "limit build unless nightly build"
         if: github.event_name != 'schedule' && steps.should-skip.outputs.result != 'true'
         run: |
+          allow_context() {
+            hal_contexts='context::esp,context::host,context::native,context::nrf,context::rp,context::stm32'
```
ROMemories, line 147: `context:rp` was missing a colon.

.github/workflows/build.yml
```diff
@@ -143,6 +143,11 @@ jobs:
...
+
+            echo "$hal_contexts" | tr ',' '\n' | grep -v "^context::${1}$" | paste -sd ','
```
ROMemories, line 149: I've adjusted the `grep` regex to match against the entire line.

PR #1514, opened 2025-11-07
docs(gen-book): automatically track functionality support

### LINKED ISSUE #1509: Specify criteria sufficient for claiming support for functionality
Our support information is currently manually filled in in the support matrix. However, it's not always clear what the requirements are to be able to mark a piece of functionality as `supported`.

A way to list testing criteria should be introduced in the [definition of these pieces of functionality](https://github.com/ariel-os/ariel-os/blob/2e45f11500926608f6aa194a7d6aa7524319b9fb/doc/support_matrix.yml#L22). These criteria should be *sufficient* to be able to claim support. Examples of these criteria would be:

- `hwrng`: the `random` example passes
- `spi_main`: the `spi-loopback` test passes
- `storage`: the `storage` example passes

It should be possible to list multiple, cumulative criteria.

These criteria don't need to be rendered in the docs, at least for now.

jkr-ariel:
# Description

Currently, functionality support status is manually filled in the support matrix.

This PR aims to change that by introducing the concept of functionality support criteria in the support matrix: requires a set of tests or examples to be built for a specific piece of hardware for the functionality to be considered supported on that piece of hardware.

## Issues/PRs references

Closes #1509.
Depends on #1524.

## Change checklist

### DIFF
--- a/book/src/boards/bbc-microbit-v1.md
+++ b/book/src/boards/bbc-microbit-v1.md
@@ -15,7 +15,7 @@
 
 |Functionality|Support Status|
 |---|:---:|
-|GPIO|<span title="supported">✅</span>|
+|GPIO|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>|
 |Debug Output|<span title="supported">✅</span>|
 |I2C Controller Mode|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>|
 |SPI Main Mode|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>|
@@ -23,7 +23,6 @@
 |Logging|<span title="supported">✅</span>|
 |User USB|<span title="not available on this piece of hardware">–</span>|
 |Wi-Fi|<span title="not available on this piece of hardware">–</span>|
-|Bluetooth Low Energy|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>|
 |Ethernet over USB|<span title="not available on this piece of hardware">–</span>|
 |Hardware Random Number Generator|<span title="supported">✅</span>|
 |Persistent Storage|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>|
--- a/book/src/boards/bbc-microbit-v2.md
+++ b/book/src/boards/bbc-microbit-v2.md
@@ -23,7 +23,6 @@
 |Logging|<span title="supported">✅</span>|
 |User USB|<span title="not available on this piece of hardware">–</span>|
 |Wi-Fi|<span title="not available on this piece of hardware">–</span>|
-|Bluetooth Low Energy|<span title="supported">✅</span>|
 |Ethernet over USB|<span title="not available on this piece of hardware">–</span>|
 |Hardware Random Number Generator|<span title="supported">✅</span>|
 |Persistent Storage|<span title="supported">✅</span>|
--- a/book/src/boards/dfrobot-firebeetle2-esp32-c6.md
+++ b/book/src/boards/dfrobot-firebeetle2-esp32-c6.md
@@ -21,9 +21,8 @@
 |SPI Main Mode|<span title="supported">✅</span>|
 |UART|<span title="supported">✅</span>|
 |Logging|<span title="supported">✅</span>|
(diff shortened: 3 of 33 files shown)


doc/support_matrix.yml
```diff
@@ -22,36 +22,62 @@ support_keys:
...
     description:
   - name: user_usb
     title: User USB
+    support_criteria:
+      - example-usb-serial
     description:
   - name: wifi
     title: Wi-Fi
+    support_criteria:
+      - udp-echo
```
jkr-ariel: Not really sure what to put here since the network examples work transparently whether using wifi or ethernet over usb.

My first idea is that we can require a specific laze context/modules to be present in the output of `laze build -g -G -i out.json`. Draft of what it could look like:

```yaml
- name: wifi
  title: Wi-Fi
  support_criteria:
    - udp-echo:
      any:
        - wifi-cyw43
        - wifi-esp
```

This would mark the `wifi` functionality as supported if the laze builder contains either `wifi-cyw43` or `wifi-esp` in the JSON output of the command above.

This could also be used to check if a chip supports a given functionality.

doc/gen_book.rs
```diff
@@ -557,6 +573,26 @@ struct FunctionalitySupport {
...
 
+fn parse_laze_info_json(laze_info_path: &PathBuf) -> miette::Result<LazeInfo> {
+    let contents = std::fs::read(&laze_info_path).map_err(|source| Error::LazeInfoFile {
+        path: laze_info_path.clone(),
+        source,
+    })?;
+
+    let laze_info = serde_json::from_slice::<LazeInfo>(&contents).map_err(|source| {
+        let src = String::from_utf8(contents).expect("failed to build String from JSON bytes");
+        let err_span = miette::SourceOffset::from_location(&src, source.line(), source.column());
```
kaspar030, line 584: So this is how you get the nice miette errors from a serde parsed json? neat!

doc/support_matrix.yml
```diff
@@ -22,36 +22,62 @@ support_keys:
 functionalities:
   - name: gpio
     title: GPIO
+    # Name(s) of the laze app(s) (as defined in the app's `laze.yml`)
+    # that must be built for the functionality to be marked
+    # as supported for a specific piece of hardware.
+    # Can be tests or examples.
+    support_criteria:
+      - tests_gpio
```
kaspar030: I think it would be nifty if not only applications but also modules could be used here.

jkr-ariel:
Currently, if the laze app(s) specified in `support_criteria` is not found the the JSON output for a given builder, we mark the functionality as "not currently supported", but sometimes the correct status is "not available on this piece of hardware".

I do not see an easy way to set the correct support status by only looking at the laze info.

doc/support_matrix.yml
kaspar030: > Not really sure what to put here since the network examples work transparently whether using wifi or ethernet over usb.

1. Unless the `any` is trivial to implement, we could also make the wifi modules all `provide: wifi`, as a marker. That way just `wifi` would be a valid criteria.

2. laze currently chooses wifi over usb-ethernet if both are available, so on wifi capable boards, usb-ethernet is not built. => this could be solved by running the `laze -i output.json` multiple times, one time with `--disable wifi`.

ROMemories:
Honestly I would prefer we split this into two steps: first only listing the criteria in the `support_matrix.yml` and adjusting the schema to allow that, so that humans can read that, and then we'll reconsider how the support info is possibly generated from these criteria (even if that means revisiting the schema again).

jkr-ariel:
~~To track the support status of chips, the process is a bit more cumbersome. For each functionality of each chip:~~
- ~~Go into each laze builder object~~
- ~~Check if the laze app of the criteria is present in the builder~~
- ~~Look in the modules of the laze app for a module with the name `context::<name_of_chip>`~~

~~We would still encounter the issue mentioned above with functionalities being reported as not supported whereas they are in fact not available on hardware.~~

Actually this does not because there could be a module in the builder's app that was not set by the chip context.  
For example: for the `rpi-pico-w` builder, the `wifi-cyw43` is present along with `context::rp2040` but that does not mean that this MCU supports wifi.

It is unclear how to track support status for chips.

PR #1399, opened 2025-09-29
feat(sensors): give access to the sensor from samples

nponsard:
# Description

This PR adds the ability to access the sensor that generated the `Samples`.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1520, opened 2025-11-13
feat(sensors): tie channels to samples in return types

ROMemories:
# Description

Following #1399, it's now possible to rework the `Reading` trait to return tuples of `(ReadingChannel, Sample)` so that users don't have to awkwardly zip `ReadingChannels` with `Samples` themselves.
It was necessary to introduce a custom iterator type for type erasure (it's not possible to simply use `Iterator::zip()`, because this would require to use `array::IntoIter`, which has the array size as a generic, which makes the march arms have different types).

Using `ci-build:skip` as there is no users of that interface in the tree yet.

This is not considered a breaking change as the sensor API is currently undocumented on purpose.

## Documentation

The documentation of the sensor API can be generated locally with:

```sh
cargo +nightly doc -p ariel-os-sensors --open 
```

The only visible difference should be on the return types of `Reading::sample()` and `Reading::samples()`. It even seems fine to keep the custom iterator private, as the type is only instantiated internally and only returned as an opaque type.

## Testing

Tested by merging against #1412: the only required adjustment is where you would expect, when accessing `Reading::samples()`.

## Change checklist

### DIFF
--- a/src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
+++ b/src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
@@ -36,7 +36,8 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
         quote! {
             InnerSamples::#variant(samples) => {
                 if let Some(sample) = samples.first() {
-                    *sample
+                    let reading_channel = self.sensor.reading_channels().first();
+                    (reading_channel, *sample)
                 } else {
                     // NOTE(no-panic): there is always at least one sample
                     unreachable!();
@@ -62,7 +63,7 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
 
     let samples_iter = (1..=count).map(|i| {
         let variant = variant_name(i);
-        quote! { InnerSamples::#variant(ref samples) => samples.iter().copied() }
+        quote! { InnerSamples::#variant(samples) => samples.iter().copied() }
     });
     let reading_channels_iter = (1..=count).map(|i| {
         let variant = variant_name(i);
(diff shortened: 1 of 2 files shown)


src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -156,8 +164,6 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
             ///
             /// For a given sensor driver, the number and order of items match the one of
             /// [`Samples`].
-            /// [`Iterator::zip()`] can be useful to zip the returned iterator with the one
-            /// obtained with [`Reading::samples()`].
```
ROMemories, line 160: Removed this as users don't have to do that themselves anymore.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -112,7 +114,7 @@ pub trait Reading: core::fmt::Debug {
     ///
     /// The default implementation must be overridden on types containing multiple
     /// [`Sample`]s.
-    fn samples(&self) -> impl ExactSizeIterator<Item = Sample> {
+    fn samples(&self) -> impl ExactSizeIterator<Item = (ReadingChannel, Sample)> {
```
ROMemories, line 117: Changing this return type is the whole point.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -179,6 +185,47 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
...
+                        Some((reading_channel, sample))
+                    }
+                    _ => None,
+                }
+            }
+        }
+
+        impl ExactSizeIterator for ChannelsSamplesZip {
+            fn len(&self) -> usize {
+                self.reading_channels.iter().len().min(self.samples.iter().len())
```
nponsard, line 224: Is comparing both lengths necessary ? If `reading_channels` and `samples` don't have the same length this means something already went wrong.

ROMemories: > If `reading_channels` and `samples` don't have the same length this means something already went wrong.

I agree that's an assumption, but we don't really enforce that equality anywhere in the API, and picking the minimum avoids having to take care of this in this trait implementation (and I'd rather not panic here).

PR #1521, opened 2025-11-13
refactor(sensors): rename `Accuracy` to `SampleMetadata`

ROMemories:
# Description

This makes the name of the type more generic, so we can store other sorts of metadata without it looking odd, like in #1354 (and we'll likely introduce another metadata variant for channels disabled by configuration, e.g., for IMUs, see #1522).

This is not considered a breaking change as the sensor API is undocumented on purpose.

## Testing

Merged against #1412, the required adjustments were as expected.

## Documentation

The documentation of the sensor API can be generated locally with:

```sh
cargo +nightly doc -p ariel-os-sensors --open 
```

## Change checklist

### DIFF
--- a/src/ariel-os-sensors/src/sample.rs
+++ b/src/ariel-os-sensors/src/sample.rs
@@ -10,7 +10,7 @@ pub enum SampleError {
 }
 
 #[expect(clippy::doc_markdown)]
-/// Represents a value obtained from a sensor device, along with its accuracy.
+/// Represents a value obtained from a sensor device, along with its metadata.
 ///
 /// # Scaling
 ///
@@ -22,7 +22,7 @@ pub enum SampleError {
 ///
 /// For instance, in the case of a temperature sensor, if [`Self::value()`] returns `2225` and the
 /// scaling value is `-2`, this means that the temperature measured and returned by the sensor
-/// device is `22.25` (the [measurement error](Accuracy) must additionally be taken into
+/// device is `22.25` (the [measurement error](SampleMetadata) must additionally be taken into
 /// account).
 /// This is required to avoid handling floats.
 ///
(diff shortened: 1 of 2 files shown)


src/ariel-os-sensors/src/sample.rs
```diff
@@ -50,25 +48,27 @@ impl Sample {
...
+/// Metadata associated with a [`Sample`].
+///
+/// Includes the measurement accuracy if available.
 #[derive(Debug, Copy, Clone, PartialEq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub enum Accuracy {
+pub enum SampleMetadata {
     /// Unknown accuracy.
-    Unknown,
+    UnknownAccuracy,
```
ROMemories, line 86: This variant also needed renaming so its name makes sense.

src/ariel-os-sensors/src/sample.rs
```diff
@@ -50,25 +48,27 @@ impl Sample {
...
     /// No measurement error (e.g., boolean values from a push button).
     NoError,
```
nponsard: Now that its in a metadata enum, this can be read like it refers to an Error type (from a rust `Result`). Could there be a better naming and wording of the comment ?
Tried finding an alternative but couldn't come with one, if there's no better option it can stay like that.

ROMemories: Agreed, I'll try to come up with something better.

ROMemories: Simply went with `NoMeasurementError` and tried to clarify the comment a bit. Observational/measurement error is a [well-defined concept](https://en.wikipedia.org/wiki/Observational_error). We can always refine the comment later.

PR #1524, opened 2025-11-14
docs(support-matrix): introduce criteria for functionality support

jkr-ariel:
# Description

This PR adds a new `support_criteria` key for the `functionalities` in `support_matrix.yml`. The goal is to document the requirements for a functionality to considered supported on a piece of hardware.

## Issues/PRs references

Part of #1509.
Extracted from #1514.

## Change checklist

### DIFF
--- a/doc/schema.rs
+++ b/doc/schema.rs
@@ -24,6 +24,7 @@ pub struct SupportKeyInfo {
 pub struct FunctionalityInfo {
     pub name: String,
     pub title: String, // FIXME: rename this
+    pub support_criteria: String,
     pub description: String,
 }
 


kaspar030:
Would it make sense to drop the distinction between app and module here? every app is also a module.

jkr-ariel:
> Would it make sense to drop the distinction between app and module here? every app is also a module.

I agree this enum is not very practical to work with, but how would you go about parsing a support criteria that can either be:

```yaml
support_criteria:
  - laze_app
 ```
 
 or
 
 ```yaml
 support_criteria:
   - laze_app: required_module
  ```

**Comment by kaspar030** [comment]
> I agree this enum is not very practical to work with, but how would you go about parsing a support criteria that can either be:
> 
> ```yaml
> support_criteria:
>   - laze_app: required_module
> ```

This boils down to whether combinations are possible, and/or different sets. As `support_criteria` is a list, I'd assume that all entries must match (logical `AND`). So to do both `laze_app` and `required_module`, why not just list both:

```yaml
support_criteria:
  - laze_app
  - required_module
```

Or do we need more complex criteria like `(laze_app1 AND some_module) OR (laze_app2 AND other_module)`?

If yes, the criteria list could be OR by default:

```yaml
support_criteria:
  - laze_app1 AND some_module
  - laze_app2 AND other_module
```

jkr-ariel:
>Or do we need more complex criteria like (laze_app1 AND some_module) OR (laze_app2 AND other_module)?

I think we want something like `(laze_app1 AND some_module) AND (laze_app2 AND other_module)` but not `laze_app1 AND other_module`.

Currently the way we parse the JSON of `laze -g -G -i output.json` is that we first check if the `laze_app` is present in a given builder:
```bash
$ jq '.builds."rpi-pico-w"|keys' /tmp/laze-info.json
[
  "bench_sched_flags",
  "bench_sched_yield",
  "ble-advertiser",
  "ble-scanner",
  "crates/ariel-os-alloc",
  "crates/ariel-os-identity",
  "device-metadata",
  "example-alloc",
  "example-benchmark",
  "example-coap-client",
  "example-coap-server",
  "example-log",
  "example-power",
  "example-random",
  "example-storage",
  "example-testing",
  "example-threading",
  "example-usb-serial",
  "hello-world",
  "hello-world-threading",
  "http-client",
  "http-server",
  "i2c-controller",
  "minimal",
  "spi-loopback",
  "spi-main",
  "tcp-echo",
  "test-coap",
  "test-random-getrandom",
  "tests_gpio",
  "thread-async-interop",
  "threading-channel",
  "threading-dynamic-prios",
  "threading-event",
  "threading-fpu",
  "threading-lock",
  "threading-multicore",
  "threading-mutex",
  "uart-loopback",
  "udp-echo"
]
```

And then check if the module is present in the app:
```bash
$ jq '.builds."rpi-pico-w"."udp-echo".modules|keys' /tmp/laze-info.json
[
  "context::ariel-os",
  "context::default",
  "context::rp",
  "context::rp2040",
  "context::rpi-pico-w",
  "cortex-m",
  "cortex-m0-plus",
  "debug-console",
  "debug-logging-facade-default",
  "defmt",
  "executor-default",
  "executor-interrupt",
  "lto",
  "network",
  "network-config-dhcp",
  "network-config-ipv4-default",
  "network-config-ipv4-dhcp",
  "panic-printing",
  "probe-rs",
  "rp-link-arg",
  "rtt-target",
  "semihosting",
  "single-core",
  "stable",
  "thumbv6m-none-eabi",
  "udp-echo",
  "wifi-cyw43"
]
```

Which means that if we have a list like:
```yaml
support_criteria:
  - laze_app
  - required_module
```

We would need to know if the element is a laze app or a module (laze apps are modules but this distinction refers to how they are structured in the JSON).

Do you think there a better way to parse this JSON?

kaspar030:
> Do you think there a better way to parse this JSON?

I'd ignore the application field, and build a list of sets for each application with just the modules (which include the application module).

I don't think we'll need cross-application rules (like, something is supported if there is one application with some modules AND another application with other modules), only module combinations (including maybe application modules), and I hope that most support criteria will be just one module (e.g., `udp-echo` is sufficient for networking, `i2c-loopback` for i2c, ...).

Not sure how to query this best. Maybe for each builder, have a map from each module to the (builder/application) module sets it is in.

Maybe actually single module queries are sufficient, and simplest to implement.

jkr-ariel:
> I'd ignore the application field, and build a list of sets for each application with just the modules (which include the application module).

Oops I did not realize that the app's name was inside the modules, that makes it a bit easier indeed.

> Maybe actually single module queries are sufficient, and simplest to implement.

I think this could work. If a laze app has a `wifi` module, it does not really matter which app it is and we can just set the functionality as supported for that builder.

PR #1530, opened 2025-11-17
feat(sensors): Add a thermometer example for the STM32U083C-DK

anlavandier:
# Description

This PR adds a example that uses the STTS22H temperature sensor and the LCD screen of the STM32U083C-DK to run a simple thermometer

## Issues/PRs References

Depends on #1363 

## Changelog entry

A thermometer example is now available: it demonstrates usage of the new sensor API and, on the STM32U083C-DK, displays the reading on the onboard LCD.

## Change Checklist

### DIFF
--- a/examples/thermometer/Cargo.toml
+++ b/examples/thermometer/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "thermometer"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "i2c",
+  "sensors",
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os-sensor-stts22h = { path = "../../src/sensors/ariel-os-sensor-stts22h" }
+ariel-os-sensors = { workspace = true }
+
+stm32-lcd-driver = { git = "https://codeberg.org/AudaciousAxiom/stm32-lcd-driver.git", rev = "24482c9d80793e97c4fe457a99102650c2152f41" }
+
+embassy-sync = { workspace = true }
+once_cell = { workspace = true }
+
+[lints]
+workspace = true
--- a/examples/thermometer/README.md
+++ b/examples/thermometer/README.md
@@ -1,0 +1,11 @@
+# thermometer
+
+## About
+
+This application is using the STTS22H temperature sensor and the LCD of the STM32U083C-DK to build a simple thermometer.
+
+## How to run
+
+In this directory, run
+
+    laze build -b stm32u083c-dk run
--- a/examples/thermometer/src/main.rs
+++ b/examples/thermometer/src/main.rs
@@ -1,0 +1,153 @@
+#![no_main]
+#![no_std]
+
+mod i2c_bus;
+mod pins;
+mod sensors;
+
+use ariel_os::{
+    debug::log::{error, info},
+    sensors::{
+        Category, Label, MeasurementUnit, REGISTRY, Reading as _,
+        sensor::{ReadingChannel, Sample},
+    },
+    time::Timer,
+};
+
+use stm32_lcd_driver::{Digit, Lcd};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    i2c_bus::init(peripherals.i2c);
+    sensors::init().await;
+
+    info!("This will print the readings of the temperature sensor on the LCD");
+
+    let mut lcd = Lcd::new(peripherals.lcd.lcd, peripherals.pins.into_pins());
+    lcd.initialize().await;
+
+    loop {
+        let Some(sensor) = REGISTRY
+            .sensors()
+            .find(|s| s.categories().contains(&Category::Temperature))
+        else {
+            info!("There aren't any registered temperature sensors");
+            break;
+        };
+
+        if let Err(err) = sensor.trigger_measurement() {
+            error!("Error when triggering a measurement: {}", err);
+            Timer::after_secs(2).await;
+            continue;
+        }
+        let reading = sensor.wait_for_reading().await;
+
+        match reading {
+            Ok(samples) => {
+                for (reading_channel, sample) in samples
+                    .samples()
+                    .filter(|(reading_channel, _)| reading_channel.label() == Label::Temperature)
+                {
+                    // Our code only supports Celsius right now
+                    match reading_channel.unit() {
+                        MeasurementUnit::Celsius => {
+                            print_temp_to_lcd(&mut lcd, sample, reading_channel)
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            Err(err) => {
+                error!("Error when reading: {}", err);
+            }
+        }
+        Timer::after_secs(2).await;
+    }
+}
+
+fn print_temp_to_lcd(lcd: &mut Lcd, sample: Sample, reading_channel: ReadingChannel) {
+    let value = match sample.value() {
+        Ok(value) => value,
+        Err(_) => {
+            info!("Error when sampling");
+            return;
+        }
+    };
+
+    let channel_scaling = i32::from(reading_channel.scaling());
+    let factor = 10_i32.pow(channel_scaling.unsigned_abs());
+    let (integer_part, decimal_part) = if channel_scaling < 0 {
+        // Fixed point arithmetic
+        let int_part = value / factor;
+        (
+            int_part,
+            value.unsigned_abs() - int_part.unsigned_abs() * factor.unsigned_abs(),
+        )
+    } else {
+        // Just multiply
+        (value * factor, 0)
+    };
+
+    if integer_part >= 1000 {
+        unreachable!();
+    } else if integer_part <= -100 {
+        unreachable!();
+    }
+
+    lcd.clear();
+
+    if integer_part <= 0 {
+        lcd.write_digit(Digit::Minus, 0).unwrap();
+    }
+
+    let integer_part = integer_part.unsigned_abs();
+
+    // hundreds digit
+    let h = integer_part / 100;
+    // tens digit
+    let t = integer_part / 10 - 10 * h;
+    // units digit
+    let u = integer_part - 10 * t + 100 * h;
+    // deci digit
+    let d = if decimal_part == 0 {
+        0
+    } else {
+        decimal_part / 10_u32.pow(decimal_part.ilog10())
+    };
+
+    if h != 0 {
+        // "htu"
+        lcd.write_digit(digit(h), 1).unwrap();
+        lcd.write_digit(digit(t), 2).unwrap();
+        lcd.write_digit(digit(u), 3).unwrap();
+    } else {
+        // "tu.d"
+        if t != 0 {
+            lcd.write_digit(digit(t), 1).unwrap();
+        }
+        lcd.write_digit(digit(u), 2).unwrap();
+        // Decimal Point is written at the same digit as the units digit
+        lcd.write_digit(Digit::Dp, 2).unwrap();
+        lcd.write_digit(digit(d), 3).unwrap();
+    };
+
+    lcd.write_digit(Digit::Degree, 4).unwrap();
+    lcd.write_digit(Digit::C, 5).unwrap();
+    lcd.display();
+}
+
+fn digit(a: u32) -> Digit {
+    match a {
+        1 => Digit::_1,
+        2 => Digit::_2,
+        3 => Digit::_3,
+        4 => Digit::_4,
+        5 => Digit::_5,
+        6 => Digit::_6,
+        7 => Digit::_7,
+        8 => Digit::_8,
+        9 => Digit::_9,
+        0 => Digit::_0,
+        a => unreachable!("{}", a),
+    }
+}
--- a/examples/thermometer/src/sensors.rs
+++ b/examples/thermometer/src/sensors.rs
@@ -1,0 +1,46 @@
+//! This module is intended to contain the auto-@generated instantiation and registration of sensor
+//! drivers.
+
+pub async fn init() {
+    #[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
+    stts22h::init().await;
+}
+
+#[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
+mod stts22h {
+    use ariel_os::i2c::controller::I2cDevice;
+
+    pub static STTS22H_I2C: ariel_os_sensor_stts22h::i2c::Stts22h<I2cDevice> =
+        const { ariel_os_sensor_stts22h::i2c::Stts22h::new(Some("onboard")) };
+    #[ariel_os::reexports::linkme::distributed_slice(ariel_os::sensors::SENSOR_REFS)]
+    #[linkme(crate = ariel_os::reexports::linkme)]
+    static STTS22H_I2C_REF: &'static dyn ariel_os::sensors::Sensor = &STTS22H_I2C;
+
+    #[ariel_os::task(autostart)]
+    pub async fn stts22h_i2c_runner() {
+        STTS22H_I2C.run().await
+    }
+
+    pub(super) async fn init() {
+        let mut config = ariel_os_sensor_stts22h::i2c::Config::default();
+        config.address = {
+            #[cfg(context = "st-steval-mkboxpro")]
+            let address = ariel_os_sensor_stts22h::i2c::I2cAddress::AddrVdd;
+            #[cfg(context = "stm32u083c-dk")]
+            let address = ariel_os_sensor_stts22h::i2c::I2cAddress::AddrGnd;
+            address
+        };
+
+        STTS22H_I2C
+            .init(
+                ariel_os_sensor_stts22h::i2c::Peripherals {},
+                I2cDevice::new(crate::i2c_bus::I2C_BUS.get().unwrap()),
+                config,
+            )
+            .await;
+    }
+}
+
+#[allow(unused, reason = "should be directly accessible without going through the registry")]
+#[cfg(any(context = "st-steval-mkboxpro", context = "stm32u083c-dk"))]
+pub use stts22h::STTS22H_I2C;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5355,6 +5355,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "stm32-lcd-driver"
+version = "0.1.0"
+source = "git+https://codeberg.org/AudaciousAxiom/stm32-lcd-driver.git?rev=24482c9d80793e97c4fe457a99102650c2152f41#24482c9d80793e97c4fe457a99102650c2152f41"
+dependencies = [
+ "critical-section",
+ "embassy-stm32",
+ "embassy-time",
+ "stm32-metapac",
+]
+
+[[package]]
 name = "stm32-metapac"
 version = "16.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5548,6 +5559,19 @@ dependencies = [
  "ariel-os",
  "ariel-os-boards",
  "embedded-test",
+]
+
+[[package]]
+name = "thermometer"
+version = "0.0.0"
+dependencies = [
+ "ariel-os",
+ "ariel-os-boards",
+ "ariel-os-sensor-stts22h",
+ "ariel-os-sensors",
+ "embassy-sync 0.6.2",
+ "once_cell",
+ "stm32-lcd-driver",
 ]
 
 [[package]]
--- a/deny.toml
+++ b/deny.toml
@@ -62,6 +62,8 @@ allow-git = [
   "https://gitlab.com/etonomy/riot-module-examples",
   # while https://github.com/twittner/minicbor/pull/9 is open
   "https://github.com/chrysn-pull-requests/minicbor",
+  # for the thermometer example
+  "https://codeberg.org/AudaciousAxiom/stm32-lcd-driver",
 ]
 
 [sources.allow-org]
--- a/examples/laze.yml
+++ b/examples/laze.yml
@@ -26,6 +26,7 @@ subdirs:
   - storage
   - tcp-echo
   - testing
+  - thermometer
   - thread-async-interop
   - threading
   - threading-channel
--- a/examples/thermometer/laze.yml
+++ b/examples/thermometer/laze.yml
@@ -1,0 +1,4 @@
+apps:
+  - name: thermometer
+    context:
+      - stm32u083c-dk
--- a/examples/thermometer/src/i2c_bus.rs
+++ b/examples/thermometer/src/i2c_bus.rs
@@ -1,0 +1,20 @@
+//! This module is intended to be @generated.
+use ariel_os::{
+    debug::log::debug,
+    hal,
+    i2c::controller::{Kilohertz, highest_freq_in},
+};
+use embassy_sync::mutex::Mutex;
+
+pub static I2C_BUS: once_cell::sync::OnceCell<
+    Mutex<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, hal::i2c::controller::I2c>,
+> = once_cell::sync::OnceCell::new();
+
+pub fn init(peripherals: crate::pins::I2CPins) {
+    let mut i2c_config = hal::i2c::controller::Config::default();
+    i2c_config.frequency = const { highest_freq_in(Kilohertz::kHz(100)..=Kilohertz::kHz(400)) };
+    debug!("Selected frequency: {:?}", i2c_config.frequency);
+
+    let i2c_bus = crate::pins::SensorI2c::new(peripherals.i2c_sda, peripherals.i2c_scl, i2c_config);
+    let _ = I2C_BUS.set(Mutex::new(i2c_bus));
+}
--- a/examples/thermometer/src/pins.rs
+++ b/examples/thermometer/src/pins.rs
@@ -1,0 +1,81 @@
+use ariel_os::hal::{i2c, peripherals};
+
+ariel_os::hal::group_peripherals!(Peripherals {
+    lcd: LcdPeripherals,
+    pins: Pins,
+    i2c: I2CPins,
+});
+
+pub type SensorI2c = i2c::controller::I2C1;
+
+ariel_os::hal::define_peripherals!(LcdPeripherals { lcd: LCD });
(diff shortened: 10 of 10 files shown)


examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+            // - '.' takes no space on the LCD display
+            // - '°' takes up 2 bytes
+            // so the buffer has to hold 8 bytes;
+            let mut lcd_bytes = [0_u8; 8];
+
+            lcd_bytes[5..8].copy_from_slice("°C".as_bytes());
+
+            let start = if value >= 1000_f32 {
+                unreachable!("No way that this sensor survives 1000 °C");
+            } else if value >= 100_f32 {
```
ROMemories: Ideally this example wouldn't be using floats at all, especially as the Cortex-M0+ on the target MCU doesn't have an FPU. No need to keep the logging logic from the original example, you should see a reduction in flash size since the float implementation has to be provided by the compiler (as there is no FPU).

examples/sensors-debug/src/main.rs
```diff
@@ -0,0 +1,151 @@
...
+    loop {
+        // Trigger measurements for each sensor driver in parallel.
+        for sensor in REGISTRY.sensors() {
+            if let Err(err) = sensor.trigger_measurement() {
+                error!("Error when triggering a measurement: {}", err);
+            }
+        }
+
+        // Then, collect and display the readings one at a time.
+        for sensor in REGISTRY.sensors() {
```
ROMemories, line 33: Instead of looping over all registered sensor driver instances, this should instead "query" a sensor driver instance of `Category::Temperature` (and *possibly* filter by label as well) through the registry (using `REGISTRY.sensors()` and `Iterator::filter()`).

examples/thermometer/src/main.rs
anlavandier: I've changed it to not use any floating point logic (I think) but I didn't see any changes in code size, it's still 31KiB...

examples/sensors-debug/src/main.rs
anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/c8566be76b97684c70b34c139d88388a4fa0afa4

examples/thermometer/src/main.rs
ROMemories: By compiling with `-d debug-console`, I do see a reduction in text size from 16280 bytes (`cd4ec2bb`) down to 14312 bytes (`556cf893`). By using the `bloat` laze task, you can also see that compiler-provided `compiler_builtins::float::div::__divsf3` (528B) and `compiler_builtins::float::mul::__mulsf3` (416B) float functions are removed from the binary. :confetti_ball:

examples/sensors-debug/src/main.rs
ROMemories: Actually that `for` loop should be removed entirely, and `Iterator::find()` can be used instead of `Iterator::filter()` (otherwise the LCD screen would be refreshed for every registered temperature sensor, but we'd only see the last one due to the 2 s pause). No need to keep the split between `sensor.trigger_measurement()` and `sensor.wait_for_reading().await`, this is only useful when triggering the measurements of multiple sensor devices.

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+        unreachable!();
+    } else if integer_part <= -100 {
+        unreachable!();
+    }
+
+    // 6 "Digits" available on the LCD display but
+    // - '.' takes no space on the LCD display
+    // - '°' takes up 2 bytes
+    // so the buffer has to hold 8 bytes;
+    let mut lcd_bytes = [0u8; 8];
```
ROMemories: Could you try to use `Lcd::write_digit()` and `Digit` instead of building a string slice? I'd be curious to know whether this further reduces the binary size.

anlavandier:
Running `laze build -d panic-printing size -A` gives me `15080` bytes in `.text` for https://github.com/ariel-os/ariel-os/pull/1530/commits/507ac91ac3a132888b5c903bcd02eb74aba29969 and `14438` bytes for 27cae8c

examples/thermometer/src/main.rs
anlavandier: Done in 94f479e

examples/sensors-debug/src/main.rs
anlavandier: Done in 94f479e

anlavandier:
Trying to use `bloat` is inconclusive, on 556cf89 I don't see any methods coming from `stm32_lcd_driver`. On 27cae8c I see `648B` from `Lcd::write_digit`

examples/sensors-debug/src/main.rs
anlavandier: By the way, this chain of comment is linked to the `sensors-debug` example and not the thermometer.

ROMemories: Ah right. Actually you can now rebase this PR on `main` as we've merged the `sensors-debug` example; that should clean up the diff.

anlavandier: I think that it's now correctly rebased

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+    sensors::init().await;
+
+    info!("Will print the readings of temperature sensor on the LCD Screen");
+
+    let mut lcd = Lcd::new(lcd_peris.lcd, pin_peris.into_pins());
+    lcd.initialize().await;
+
+    loop {
+        // Trigger measurements for each sensor driver in parallel.
+        match REGISTRY
```
ROMemories: This `match` should be replaced with a let-else for clarity.

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+            .sensors()
+            .find(|s| s.categories().contains(&Category::Temperature))
+        {
+            Some(sensor) => {
+                if let Err(err) = sensor.trigger_measurement() {
+                    error!("Error when triggering a measurement: {}", err);
+                }
```
ROMemories: If triggering a measurement failed, this shouldn't wait for a reading:
```suggestion
                    continue;
                }
```
We might want to have a delay as well to avoid busy-looping.

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+                let reading = sensor.wait_for_reading().await;
+
+                match reading {
+                    Ok(samples) => {
+                        for (reading_channel, sample) in samples.samples() {
+                            // Even though the sensor is guaranteed to be a temperature sensor,
+                            // a single sensor could provide multiple readings including ones that aren't temperature
+                            match reading_channel.unit() {
```
ROMemories: This shouldn't be using the `.unit()` but the `.label()` instead. Currently: if there is only one channel, the label must be `Label::Main`, otherwise you need to look for `Label::Temperature`. However seeing this I'm realizing this isn't really ergonomic and we should consider removing `Label::Main` entirely as it's actually not solving any problem (and you'd then only need to look for `Label::Temperature`). Thanks for this making this issue visible!

anlavandier: It's also that my code currently wouldn't work with temperatures expressed in Kelvin

ROMemories: Yeah that's right, but I guess that should be an additional, explicit check.

anlavandier: > f there is only one channel, the label must be Label::Main, otherwise you need to look for Label::Temperature.

I don't understand what I should do. Filter out reading channels that don't have `Label::Main` or `Label::Temperature` ?

ROMemories: I've updated #1363 to use `Label::Temperature` instead of `Label::Main` and opened #1533 to simply remove `Label::Main` entirely. You should be able to rebase on #1363 and simply filter on `Label::Temperature` now.

examples/thermometer/src/main.rs
anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/b2417d6d8d2eb15021efca41b8023dc8f8c1e0e0

examples/thermometer/src/main.rs
anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/ce2d82a61be9cc8f51917a5ceb9b4151cbad8000. I added a 2 sec timeout as is done in the happy path

examples/thermometer/src/main.rs
ROMemories: I think that's an if-let, a [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) would be even clearer.

examples/thermometer/src/main.rs
anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/1b0be081e2109d3992e9a590f25341fcccf48c54

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+                Timer::after_secs(2).await;
+                continue;
+            }
+            let reading = sensor.wait_for_reading().await;
+
+            match reading {
+                Ok(samples) => {
+                    for (reading_channel, sample) in
+                        samples.samples().filter(|(reading_channel, _)| {
+                            matches!(reading_channel.label(), Label::Temperature)
```
ROMemories: `PartialEq` is impled on `Label`:

```suggestion
                            reading_channel.label() == Label::Temperature
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+
+use stm32_lcd_driver::{Digit, Lcd};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    let pins::Peripherals {
+        lcd: lcd_peris,
+        pins: pin_peris,
+        i2c: i2c_peris,
+    } = peripherals;
```
ROMemories: Not sure pattern matching/desctructuring is really needed here: you can simply access each field where you need it.

examples/thermometer/src/main.rs
anlavandier: Oh, I didn't know this was a thing. It's now done in https://github.com/ariel-os/ariel-os/pull/1530/commits/7d72ba64242f3696305481289ffe0a814fd15530

examples/thermometer/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+stm32-lcd-driver = { git = "https://codeberg.org/AudaciousAxiom/stm32-lcd-driver.git" }
+
+ariel-os-sensor-stts22h = { path = "../../src/sensors/ariel-os-sensor-stts22h" }
+ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
+
+
+cfg-if = { workspace = true }
```
ROMemories: I don't think this dependency is needed.

examples/thermometer/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "i2c",
+  "sensors",
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+stm32-lcd-driver = { git = "https://codeberg.org/AudaciousAxiom/stm32-lcd-driver.git" }
```
ROMemories: Can you pin to the exact commit? Otherwise it's only pinned by the lockfile.

ROMemories: The git source also needs to be allowed in the `deny.toml` configuration (I think that's fine). That'll allow Clippy to run in CI.

examples/thermometer/README.md
```diff
@@ -0,0 +1,11 @@
+# sensors-debug
```
ROMemories:
```suggestion
# thermometer
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+            Timer::after_secs(2).await;
+            continue;
+        }
+        let reading = sensor.wait_for_reading().await;
+
+        match reading {
+            Ok(samples) => {
+                for (reading_channel, sample) in
+                    samples.samples().filter(|(reading_channel, _)| {
+                        matches!(reading_channel.label(), Label::Temperature)
```
ROMemories: This is functionally equivalent, but otherwise it looks like `Label` doesn't implement `PartialEq`:
```suggestion
                        reading_channel.label() == Label::Temperature
```

examples/thermometer/src/main.rs
anlavandier: Removed in https://github.com/ariel-os/ariel-os/pull/1530/commits/36ef78bec88098c4e06d25f82b1baeafd858090f

examples/thermometer/Cargo.toml
anlavandier: Removed in https://github.com/ariel-os/ariel-os/pull/1530/commits/53b4005868ed91b58eabd0f18bc4c5cfa7d40a8e

examples/thermometer/Cargo.toml
anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/53b4005868ed91b58eabd0f18bc4c5cfa7d40a8e

anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/1d0c436016d1a184e5c145d86754e49d64e2dd6f

examples/thermometer/README.md
anlavandier: Fixed in https://github.com/ariel-os/ariel-os/pull/1530/commits/06741570549e3b0ad3d622ad49ea6e22116b6545

examples/thermometer/src/main.rs
anlavandier: Done in https://github.com/ariel-os/ariel-os/pull/1530/commits/cd00b667cfc875b38f4c604b96b5bfb8576eccac

anlavandier:
[Seems like `ariel-os-sensors-stts22h`'s test are failing](https://github.com/ariel-os/ariel-os/actions/runs/19498032945/job/55804988552?pr=1530#step:8:800)

ROMemories:
> [Seems like `ariel-os-sensors-stts22h`'s test are failing](https://github.com/ariel-os/ariel-os/actions/runs/19498032945/job/55804988552?pr=1530#step:8:800)

Yes, that's not really an issue for now; I'll refactor later to avoid the dependency on `ariel_os_embassy` and use `ariel_os_hal` and `embassy_time` instead.

ROMemories:
I think this is in a good state now, great job on this! When we later land the STTS22H driver, I'll do a last review, testing this further and looking into the business logic in detail, but for now there isn't much more to do on this PR.

*Much* later this could serve as a playground for power optimization: we could set the processor in deep sleep and only wake up based a timer to refresh the display, as the LCD *hardware driver* can run in deep sleep.

anlavandier:
Should be good now @ROMemories

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+    let (integer_part, decimal_part) = if channel_scaling < 0 {
+        // Fixed point arithmetic
+        let int_part = value as i32 / 10_i32.pow(-channel_scaling as u32);
+        (
+            int_part,
+            value.unsigned_abs() - int_part.unsigned_abs() * 10_u32.pow(-channel_scaling as u32),
+        )
+    } else {
+        // Just multiply
+        (value as i32 * 10_i32.pow(channel_scaling as u32), 0)
```
ROMemories: These casts could use `cast_unsigned()`, as in #1621.

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+    lcd.clear();
+
+    if integer_part >= 0 {
+        lcd.write_digit(Digit::Space, 0).unwrap();
+    } else {
+        lcd.write_digit(Digit::Minus, 0).unwrap();
+    }
+
+    let decimal_part = decimal_part as u32;
+    let integer_part = integer_part.abs() as u32;
```
ROMemories: Can these casts be replaced with something else? It's best to only use `as` when there's no other option.

examples/thermometer/src/main.rs
anlavandier: Done in 135bc45

examples/thermometer/src/main.rs
anlavandier: The first one was useless and removed, the second one was removed by using [`unsigned_abs()`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.unsigned_abs). 
Done in 135bc45

anlavandier:
Should be good now @ROMemories

anlavandier:
Should I mark this PR as ready for review ?

examples/thermometer/src/sensors.rs
ROMemories, line 1: Can you update this file based on `examples/sensors-debug`? Copy-pasting should do for now.

examples/thermometer/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "i2c",
+  "sensors",
+  "time",
+] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os-sensor-stts22h = { path = "../../src/sensors/ariel-os-sensor-stts22h" }
+ariel-os-sensors = { path = "../../src/ariel-os-sensors" }
```
ROMemories:
```suggestion
ariel-os-sensors = { workspace = true }
```

examples/thermometer/README.md
```diff
@@ -0,0 +1,11 @@
+# thermometer
+
+## About
+
+This application is using the STTS22H temperature sensor and the LCD Screen of the STM32U083C-DK to run a simple thermometer
```
ROMemories:
```suggestion
This application is using the STTS22H temperature sensor and the LCD of the STM32U083C-DK to build a simple thermometer.
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+
+use stm32_lcd_driver::{Digit, Lcd};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn main(peripherals: pins::Peripherals) {
+    i2c_bus::init(peripherals.i2c);
+    sensors::init().await;
+
+    info!("Will print the readings of temperature sensor on the LCD Screen");
```
ROMemories:
```suggestion
    info!("Will print the readings of temperature sensor on the LCD");
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    let mut lcd = Lcd::new(peripherals.lcd.lcd, peripherals.pins.into_pins());
+    lcd.initialize().await;
+
+    loop {
+        // Trigger measurements for each sensor driver in parallel.
```
ROMemories: There is no loop anymore:
```suggestion
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+            info!("Error when sampling");
+            return;
+        }
+    };
+
+    let channel_scaling = reading_channel.scaling();
+
+    let (integer_part, decimal_part) = if channel_scaling < 0 {
+        // Fixed point arithmetic
+        let int_part = value as i32 / 10_i32.pow(u32::from((-channel_scaling).cast_unsigned()));
```
ROMemories: `value` is already an `i32`:
```suggestion
        let int_part = value / 10_i32.pow(u32::from((-channel_scaling).cast_unsigned()));
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+        (
+            int_part,
+            value.unsigned_abs()
+                - int_part.unsigned_abs()
+                    * 10_u32.pow(u32::from((-channel_scaling).cast_unsigned())),
+        )
+    } else {
+        // Just multiply
+        (
+            value as i32 * 10_i32.pow(u32::from(channel_scaling.cast_unsigned())),
```
ROMemories:
```suggestion
            value * 10_i32.pow(u32::from(channel_scaling.cast_unsigned())),
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+fn print_temp_to_lcd(lcd: &mut Lcd, sample: Sample, reading_channel: ReadingChannel) {
+    let value = match sample.value() {
+        Ok(value) => value,
+        Err(_) => {
+            info!("Error when sampling");
+            return;
+        }
+    };
+
+    let channel_scaling = reading_channel.scaling();
```
ROMemories: You're right, we can actually replace these `cast_unsigned()`s with `unsigned_abs()`s below, which is much cleaner:
```suggestion
    let channel_scaling = i32::from(reading_channel.scaling());
```

(I'll do the same for `sensors-debug`.)

ROMemories: Actually I think `10_i32.pow(channel_scaling.unsigned_abs()` can then be factored out.

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+    let h = integer_part / 100;
+    // tens digit
+    let t = integer_part / 10 - 10 * h;
+    // units digit
+    let u = integer_part - 10 * t + 100 * h;
+    // deci digit
+    let d = if decimal_part == 0 {
+        0
+    } else {
+        decimal_part / 10_u32.pow(decimal_part.ilog10())
```
ROMemories, line 115: I think this will always round down, but given the accuracy of temperature sensors I don't think that matters anyway.

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+    if h != 0 {
+        // "htu"
+        lcd.write_digit(digit(h), 1).unwrap();
+        lcd.write_digit(digit(t), 2).unwrap();
+        lcd.write_digit(digit(u), 3).unwrap();
+    } else {
+        // "tu.d"
+        lcd.write_digit(digit(t), 1).unwrap();
+        lcd.write_digit(digit(u), 2).unwrap();
+        // Decimal Point is written at the same point the units number
```
ROMemories:
```suggestion
        // Decimal Point is written at the same digit as the units digit
```

examples/thermometer/src/main.rs
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+        decimal_part / 10_u32.pow(decimal_part.ilog10())
+    };
+
+    if h != 0 {
+        // "htu"
+        lcd.write_digit(digit(h), 1).unwrap();
+        lcd.write_digit(digit(t), 2).unwrap();
+        lcd.write_digit(digit(u), 3).unwrap();
+    } else {
```
ROMemories, line 123: I could be useful to also handle the case where `d == 0`, currently this prints `-01.2` otherwise (instead of `- 1.2`). It's fine to keep the minus sign always in the same place however.

examples/thermometer/src/sensors.rs
anlavandier: Done in [0a822dd](https://github.com/ariel-os/ariel-os/pull/1530/commits/0a822dd3c24d9af2adb593264d7615d90999b62b). I removed the context cfgs as they are already in the `laze.yml`

examples/thermometer/Cargo.toml
anlavandier: Done in [cac8c4a](https://github.com/ariel-os/ariel-os/pull/1530/commits/cac8c4a481b84dacb559adf30fec9d440bae42c3)

examples/thermometer/README.md
anlavandier: Done in [15a82c7](https://github.com/ariel-os/ariel-os/pull/1530/commits/15a82c75e8f9c97f8d77f6eec8948b6abab3333c)

examples/thermometer/src/main.rs
anlavandier: Done with a slightly different message in 4222c07

examples/thermometer/src/main.rs
anlavandier: Removed in 4222c07

examples/thermometer/src/main.rs
anlavandier: Removed in the casts in 4222c07

examples/thermometer/src/main.rs
anlavandier: 4222c07

examples/thermometer/src/main.rs
anlavandier: I factored out the `10_i32` in 4222c07

examples/thermometer/src/main.rs
anlavandier: ~~The first decimal digit is only wrong in `XX.196` type of temperatures where proper rounding would lead to `XX.20` but considering the accuracy of `+-.5` in the best of cases I agree that this doesn't matter.~~ Point still stands that I agree with you that it doesn't really matter

examples/thermometer/src/main.rs
anlavandier: Done in 4222c07

examples/thermometer/src/main.rs
anlavandier: Done in 4222c07 by putting a space if `t == 0` (the tens numbers which you were referring to)

examples/thermometer/src/main.rs
ROMemories: Opened #1625 to keep this logic consistent with `sensors-debug`'s.

examples/thermometer/src/main.rs
ROMemories: As the LCD is always cleared before being refreshed, I don't think we even need to put a space.

examples/thermometer/src/sensors.rs
ROMemories: I'd rather keep exactly the same file actually, so that it doesn't look like it's actually different and so that we can easily add other sensors from other boards later if needed (also the file should have a trailing new line character).

anlavandier:
@ROMemories  
I removed both spaces in 553e3a7, and copied the sensors registration from the `sensors-debug` example in d7ef2b3

ROMemories (approved):
Please squash!

anlavandier:
Done @ROMemories

PR #1535, opened 2025-11-18
feat(defmt): propagate the `defmt` feature to ecosystem trait crates

ROMemories:
# Description

This propagates the `demt` Cargo feature to `embedded-hal`, `embedded-hal-async`, `embedded-io`, `embedded-io-async`. This is especially useful when writing driver-agnostic drivers (generic over the traits these crates provide) and wanting to print the error types for debugging.

## Changelog entry

The `defmt` Cargo feature is now propagated to `embedded-hal`, `embedded-hal-async`, `embedded-io`, and `embedded-io-async`.

## Change Checklist


src/ariel-os-embassy/Cargo.toml
```diff
@@ -156,6 +156,10 @@ defmt = [
   "embassy-net?/defmt",
   "embassy-time?/defmt",
   "embassy-usb?/defmt",
+  "embedded-hal/defmt-03",
+  "embedded-hal-async/defmt-03",
+  "embedded-io/defmt-03",
```
ROMemories, line 161: The Cargo feature is renamed to `defmt` in `embedded-io` v0.7.1.

Cargo.lock
```diff
@@ -2135,6 +2135,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c4c685bbef7fe13c3c6dd4da26841ed3980ef33e841cddfa15ce8a8fb3f1884"
 dependencies = [
+ "defmt 0.3.100",
```
ROMemories, line 2138: [`defmt` v0.3.99+ is actually v1.0 in disguise](https://github.com/rust-embedded/embedded-hal/issues/707) (using [the semver trick](https://github.com/dtolnay/semver-trick) I suppose) and it's already in the lockfile anyway.

src/ariel-os-hal/Cargo.toml
```diff
@@ -145,5 +145,8 @@ defmt = [
   "ariel-os-nrf/defmt",
   "ariel-os-rp/defmt",
   "ariel-os-stm32/defmt",
+  "embedded-hal/defmt-03",
+  "embedded-hal-async/defmt-03",
+  "embedded-io?/defmt-03",
```
ROMemories, line 150: I'm propagating from this crate as well in case we decide later it makes sense to use this crate as a standalone HAL outside of Ariel OS.

PR #1538, opened 2025-11-19
docs(rt): enable stack-painting docs

ROMemories:
# Description

This enables the generation of docs related to stack painting implemented in #977.

## Issues/PRs References

- Depends on #1566
- Depends on #1576
- Depends on #1577
- Depends on #1617

## Open Questions

- ~~Stack painting is not currently supported on `native`, we need to document that in some way.~~

## Changelog Entry

Facilities for measuring stack usage, using stack painting, have been added. 

## Change Checklist

### DIFF
--- a/src/ariel-os-rt/src/stack.rs
+++ b/src/ariel-os-rt/src/stack.rs
@@ -40,6 +40,12 @@ const STACK_PAINT_IGNORE: usize = 16;
 /// In the current implementation, and assuming the stack data follows a uniform distribution, this
 /// is unlikely to result in an underestimation of more than one byte.
 ///
+/// # Note
+///
+/// On [native][native-book], the stack *modeled by this type* is currently empty, so its usage and
+/// free space will both be reported as zero.
+///
+/// [native-book]: https://ariel-os.github.io/ariel-os/dev/docs/book/native-target.html
 // # Note
 //
 // The machinery for stack painting has a couple of assumptions:


src/ariel-os-rt/src/stack.rs
```diff
@@ -170,12 +170,14 @@ impl Stack {
     }
 
     /// Returns this [`Stack`]'s `lowest` address.
+    #[doc(hidden)]
```
ROMemories: To me this is otherwise leaking implementation details that users cannot even make sense of.

src/ariel-os-rt/src/stack.rs
ROMemories, line 169: I still think this should be `&mut`.

kaspar030: Not mutating self?

ROMemories: Don't you think the following *looks* weird?

```rust
let stack = ariel_os::rt::stack::Stack::get();

embassy_futures::join::join(
    async { stack.repaint() },
    async { stack.repaint() },
).await;
```

Taking `&mut` would prevent this.

ROMemories, line 1: The behavior on native still needs to be documented.

ROMemories: Done now.

ROMemories:
I'll pick a `ci-build:*` label when the scope of the PR is clearer.

src/ariel-os-rt/src/stack.rs
kaspar030: It is also not an issue at all, or unsafe.

Also, this makes actual sense (repainting and measuring embedded in other code):

```
let stack = ariel_os::rt::stack::Stack::get();

embassy_futures::join::join(
    async { stack.repaint(); do_something(); printlns("{}", stack.max_used() },
    async { stack.repaint(); ... },
).await;

src/ariel-os-rt/src/stack.rs
kaspar030: Why not? Users will work with stacks and need to reason about them. Why hide this from them?

ROMemories: Because `lowest`/`highest` (with backticks) are private fields (that users therefore don't (need to) know about), which are under-defined (e.g., is the lowest stack address included in the stack? How do these two addresses behave on each architecture?). These methods are also just not needed at all to get to the stack size, which is the main goal here, so we could always expose these methods later, with better documentation.

(I'm also realizing now that the docs of the two methods are actually inverted.)

src/ariel-os-rt/src/stack.rs
ROMemories: That functionally looks like the same snippet as above? Independent functions should just get their own `Stack` instance.

kaspar030: Is there some correctness or safety issue with non-mut self on `repaint()` that is fixed by making self `&mut`?

ROMemories: I don't exactly have one in mind, but stack painting is of course pretty intrusive. I actually approach this the other way around: is there a legitimate use case that would be prevented by taking `&mut`, since removing that restriction later is not a breaking change (and is very easy to do), while adding it later is.

ROMemories: I've added a `refactor` commit that makes this function take `&mut self` instead of `&self`. (Technically this makes the PR non-doc-only, but that seems acceptable enough, otherwise I can split this.) AFAIK, going from `&mut self` to `&self` is not a breaking change, so we could always *relax* this later if we do want to.

kaspar030: Why? `repaint()` doesn't mutate `self`.

ROMemories: You already wrote this here: https://github.com/ariel-os/ariel-os/pull/1538#discussion_r2542620618. Everything I wrote since still stands: not mutating `self` in a method does not preclude taking `&mut self`, and sometimes it can be a good idea design-wise. Additionally (and more importantly in this case), this method contains an `unsafe` block that mutates memory, so the fact that the method does not mutate `self` directly is not a *sufficient* argument. In this case I still think it would be better to take `&mut self` for the design reasons I've exposed above.

ROMemories:
> please don't hide `highest()` / `lowest()`

This would require properly introducing the concept of "lowest stack address" and "highest stack address," which are not currently part of these docs, and I'm not sure what properties you would want to guarantee about them (especially across architectures). If you write them, we can have these methods in the docs.

src/ariel-os-rt/src/stack.rs
kaspar030: Then please split out as to not block the doc enablement on this conflict.

ROMemories: Even though I still think it would the right move to take an exclusive reference here, both for defense-in-depth/avoiding future breaking changes, and because it would make for a less surprising API for users, I've removed the commit from this PR.

kaspar030:
> This would require properly introducing the concept of "lowest stack address" and "highest stack address," which are not currently part of these docs, and I'm not sure what properties you would want to guarantee about them (especially across architectures). If you write them, we can have these methods in the docs.

It does not require that kind of definition or guarantee. It is informational. `lowest address <= highest address` is implicit. A usize number falls out. *doing* anything with it is probably dangerous, but that's not up to us. It is useful to get an idea of where things are in memory, maybe relative to each other. *I* use these when debugging. Just don't hide them.

ROMemories:
> It is informational.

Then the docs should say so.

> A usize number falls out. _doing_ anything with it is probably dangerous, but that's not up to us.

If the returned integer has no associated semantics, I don't see how it can be useful at all.

Please have another look at what actually is in the docs right now, all the visible concepts are about the size of the stack and its usage, not its location. Also, as explained in https://github.com/ariel-os/ariel-os/pull/1538#discussion_r2878431443, the docs of these methods specifically refer to private fields and to the `Stack` type itself, not to the general concept of stack, so we can't even rely on the user's knowledge of it.

src/ariel-os-rt/src/stack.rs
ROMemories: After some out-of-band discussion, I've reworded the documentation of these methods so we can have them in the documentation, while limiting the set of guarantees we provide.

PR #1544, opened 2025-11-20
docs: add missing punctuation at the end of paragraphs

ROMemories:
# Description

This adds missing punctuation at the end of doc paragraphs. These were found using [`clippy::doc_paragraphs_missing_punctuation`](https://rust-lang.github.io/rust-clippy/master/index.html?search=doc_pa#doc_paragraphs_missing_punctuation).

~~This is marked as draft until [`clippy::doc_paragraphs_missing_punctuation`](https://rust-lang.github.io/rust-clippy/master/index.html?search=doc_pa#doc_paragraphs_missing_punctuation) lands on nightly (in less than two weeks) or we want to release (whichever comes first).~~ The lint has now landed on nightly.

## Issues/PRs References

- Depends on #1598

## Open Questions

Are we ok with bumping nightly (in a separate PR) for this, or should we wait until the lint lands on stable ([ETA 2026-01-22](https://releases.rs/docs/1.93.0/))?

## Change Checklist

### DIFF
--- a/src/ariel-os-threads/src/arch/cortex_m.rs
+++ b/src/ariel-os-threads/src/arch/cortex_m.rs
@@ -42,6 +42,7 @@ impl Arch for Cpu {
     /// The exact order in which Cortex-M pushes the registers to the stack when
     /// entering the ISR is:
     ///
+    /// ```text
     /// +---------+ <- sp
     /// |   r0    |
     /// |   r1    |
@@ -52,6 +53,7 @@ impl Arch for Cpu {
     /// |   PC    |
     /// |   PSR   |
     /// +---------+
+    /// ```
     fn setup_stack(thread: &mut Thread, stack: &mut [u8], func: fn(), arg: Option<usize>) {
         let stack_start = stack.as_ptr() as usize;
 
--- a/examples/ble-advertiser/src/main.rs
+++ b/examples/ble-advertiser/src/main.rs
@@ -1,4 +1,4 @@
-//! Adapted from the example in `trouble_host`
+//! Adapted from the example in `trouble_host`.
 #![no_main]
 #![no_std]
 
--- a/examples/ble-scanner/src/main.rs
+++ b/examples/ble-scanner/src/main.rs
@@ -1,4 +1,4 @@
-//! Adapted from the example in `trouble_host`
+//! Adapted from the example in `trouble_host`.
 #![no_main]
 #![no_std]
 
--- a/examples/storage/src/main.rs
+++ b/examples/storage/src/main.rs
@@ -12,7 +12,7 @@ use serde::{Deserialize, Serialize};
 
 /// Example object.
 ///
-/// The serde Serialize / Deserialize traits are required for storage
+/// The serde Serialize / Deserialize traits are required for storage.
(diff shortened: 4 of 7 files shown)


chrysn:
> Are we ok with bumping nightly (in a separate PR) for this

AIU our lints run on a pinned version anyway (otherwise any `expec()` might explode randomly), and if that is nightly, why not start using it before stabilization, as long as any fall-out between now and 2026-01-22 is taken care of.

As for updating the nightly, I think we should update those liberally whenever needed; better to do it often and fix small things (or nudge those who are experts in those nightly parts) than let it spool up and let it become big.

ROMemories:
> As for updating the nightly, I think we should update those liberally whenever needed; better to do it often and fix small things (or nudge those who are experts in those nightly parts) than let it spool up and let it become big.

One possible concern of having a nightly that's too new compared to the current stable is that some Clippy false positive could have been fixed on nightly while still triggering on stable, and we could be missing that as our Clippy only runs on nightly in CI.

Anyway I opened #1598 to bump the nightly.

nponsard:
I'm trying to come up with a command to know if we missed some. 

Trying `laze build -g -m clippy` but it's hard to parse (and there's an error in `examples/alloc`).

Seems that there's punctuation missing in `src/ariel-os-rt/src/riscv.rs:12:44`.

ROMemories:
> Trying `laze build -g -m clippy` but it's hard to parse

That definitely seems like a command we could run in CI!

---

I think it's fine if this PR only addresses cases that are found by our current Clippy checks, it already addresses every items that are public, which are the most important considering this is only about documentation.

I agree though that we should also go over the lint errors found by the above command and fix them.

nponsard:
> > Trying `laze build -g -m clippy` but it's hard to parse
> 
> That definitely seems like a command we could run in CI!

We may need an architecture similar to the build CI as there are a lot of combinations generated by this command.

nponsard:
Been manually fixing errors and re-running it, here's the missing punctuation I found so far: 

- tests/uart-loopback/src/main.rs:1:47
- src/ariel-os-threads/src/arch/cortex_m.rs:54:20
- src/ariel-os-rt/src/cortexm.rs:241:44
- src/ariel-os-rt/src/riscv.rs:12:44
- examples/ble-advertiser/src/main.rs:1:47
- examples/ble-scanner/src/main.rs:1:47
- examples/storage/src/main.rs:15:70

src/ariel-os-threads/src/arch/cortex_m.rs
```diff
@@ -42,6 +42,7 @@ impl Arch for Cpu {
     /// The exact order in which Cortex-M pushes the registers to the stack when
     /// entering the ISR is:
     ///
+    /// ```text
```
ROMemories, line 45: This should additionally preserve formatting of the diagram.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -100,7 +100,7 @@ impl<T> Signal<T> {
         })
     }
 
-    /// Removes any pending value from the signal
+    /// Removes any pending value from the signal.
```
ROMemories, line 103: I've added the missing period after rebasing.

PR #1550, opened 2025-11-21
ci(build): skip the workflow based on a partial tree hash

ROMemories:
# Description

This introduces a second check that carefully excludes a couple of directories when checking for whether the Build workflow can be skipped because it already ran successfully before. The goal is to increase cache hits when only modifying paths that are *very clearly* unrelated to what the Build workflow builds.

## How to review this PR

## Testing

Tested by modifying `supply-chain/config.toml` by adding a few blank lines and checking that the Build workflow was indeed skipped, while carefully looking at the docs: the first cache fails to restore (as expected, as the git tree hash is different), but the second one does hit as `supply-chain` is excluded from the partial hash calculation.

It also seems that sometimes the second check (and likely the first one as well) results in a cache miss because of the [limited size of GitHub's cache](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy). Depending on the number of jobs that has run between the two CI runs the caches may have been evicted.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,6 +128,10 @@ jobs:
             git rev-parse HEAD^{tree} > .git-tree-hash
             echo "git_hash=$(cat .git-tree-hash)" >> $GITHUB_OUTPUT
 
+            echo '${{ hashFiles('**', '!.git', '!book', '!supply-chain') }}' > .partial-tree-hash
+            echo "partial_hash=$(cat .partial-tree-hash)" # For debugging
+            echo "partial_hash=$(cat .partial-tree-hash)" >> $GITHUB_OUTPUT
+
       # Merely checks *whether* the workflow with the same hash has run
       # successfully in the past, restores nothing more than the file containing
       # the hash value.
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -119,13 +119,14 @@ jobs:
...
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
 
-      - name: Get git tree hash
+      - name: Get partial tree hash
         id: get-tree-hash
         run: |
-            git rev-parse HEAD^{tree} > .tree-hash
-            echo "hash=$(cat .tree-hash)" >> $GITHUB_OUTPUT
+            tree_hash='${{ hashFiles('**', '!book', '!supply-chain') }}'
```
ROMemories: - Excluding `book` is useful for allowing us to cheaply iterate on the book in a PR that cannot use the `ci-build:skip` label.
- Excluding `supply-chain` is very useful to add cargo-vet info to a PR opened by Renovate while avoiding rerunning the Build workflow pre-merge.

Edit: [Documentation of `hashFiles`](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#hashfiles)

kaspar030:
I don't think we should do this.

The tree hash is used for caching positive CI results.

1. git tree hash is *instant*, whereas this needs to actually re-calculate everything
2. it is easy to reason about the git tree hash: it is the same for identical file content, contrary to the commit hash, which might be different for different metadata (but same file content).

Let's keep the git tree hash as a fast "skip if unchanged" and handle individual parts of the tree properly (e.g., in the build system(s)).

ROMemories:
> Let's […] handle individual parts of the tree properly (e.g., in the build system(s)).

How do you propose to do this?

ROMemories:
After some offline discussion, we agreed to keep the Git-based check and to introduce another separate check that relies on the hash of the partial file tree.

.github/workflows/build.yml
```diff
@@ -122,22 +122,38 @@ jobs:
...
-        id: get-tree-hash
+      - name: Get tree hashes
+        id: get-tree-hashes
         run: |
-            git rev-parse HEAD^{tree} > .tree-hash
-            echo "hash=$(cat .tree-hash)" >> $GITHUB_OUTPUT
+            git rev-parse HEAD^{tree} > .git-tree-hash
+            echo "git_hash=$(cat .git-tree-hash)" >> $GITHUB_OUTPUT
 
+            echo '${{ hashFiles('**', '!.git', '!book', '!supply-chain') }}' > .partial-tree-hash
```
ROMemories, line 131: Excluding the Git repo itself seems to be also necessary as `**` also matches hidden files.

kaspar030 (approved):

PR #1551, opened 2025-11-21
ci: periodically build `ariel-os-hello` and `ariel-os-template`

### LINKED ISSUE #1172: More scheduled CI checks for increased reliability and anticipation
We could use more scheduled checks, similar to the the one running Build every night. This could make sure the very first things beginners will go through do work as expected.

The following would be particularly useful:

- [ ] Building [`ariel-os-hello`](https://github.com/ariel-os/ariel-os-hello) and [`ariel-os-template`](https://github.com/ariel-os/ariel-os-template), following closely our ["Getting Started" instructions](https://ariel-os.github.io/ariel-os/dev/docs/book/getting-started.html) to make sure these are always working
- [ ] Checking that the file trees of `ariel-os-hello` and `ariel-os-template` are byte-for-byte identical, to avoid https://github.com/ariel-os/ariel-os-hello/issues/8

Regarding scheduling frequency, I don't think they *have* to run every night, but they should still run regularly enough to be useful.

jkr-ariel:
# Description

This PR adds a CI job that periodically builds [ariel-os-hello](https://github.com/ariel-os/ariel-os-hello) and [ariel-os-template](https://github.com/ariel-os/ariel-os-template) to make sure instructions in the [Getting Started](https://ariel-os.github.io/ariel-os/dev/docs/book/getting-started.html) page are correct.

## Issues/PRs References

Closes #1172.

## Open Questions

At what frequency do we want this job to trigger? Nightly seems unnessecary so maybe twice a week could be enough?

## Change Checklist

### DIFF
--- a/.github/workflows/build-templates.yml
+++ b/.github/workflows/build-templates.yml
@@ -85,11 +85,7 @@ jobs:
           export LAZE_BUILDERS='espressif-esp32-c6-devkitc-1,espressif-esp32-s3-devkitc-1,nrf52840dk,rpi-pico-w,st-nucleo-wb55'
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build
 
-  report-failure:
-    runs-on: ubuntu-latest
-    needs: [build]
-    steps:
-      - name: Report nightly failure
+      - name: Report failure
         if: failure() && github.event_name == 'schedule' && github.repository == 'ariel-os/ariel-os'
         uses: s3krit/matrix-message-action@v0.0.3
         with:


nponsard:
It would be nice to have a comment in the book (not rendered to the user) asking people modifying the instructions to make sure the CI matches the instructions.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+      - run: |
+          result="${{ needs.build.result }}"
+          if [[ $result == "success" || $result == "skipped" ]]; then
+            exit 0
+          else
+            exit 1
+          fi
+
+      - name: Report nightly failure
+        if: failure() && github.event_name == 'schedule' && github.repository == 'ariel-os/ariel-os-hello'
```
nponsard: This looks like it will be never triggered.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+        uses: s3krit/matrix-message-action@v0.0.3
+        with:
+          room_id: ${{ secrets.MATRIX_ROOM_ID }}
+          access_token: ${{ secrets.MATRIX_ACCESS_TOKEN }}
+          message: "The nightly build in ${{ github.repository }} [failed](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})."
```
nponsard: If this is sent in the Ariel OS room, it would be better to indicate that the ariel-os-hello test failed.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
+name: Build templates
+
+on:
+  push:
+    branches:
+      - ci/scheduled-checks
```
nponsard: This needs to be removed before merging.

book/src/getting-started.md
```diff
@@ -7,6 +7,8 @@ It explains how to compile and run the `hello-word` example to verify your setup
 
 1. Install the needed build dependencies.
 
+    <!-- If you modify these intructions, you should update .github/workflows/build-templates.yml to reflect that. -->
```
nponsard: This comment should apply to the whole page IMO, not only the first build dependency installation step.

It would be better placed line 5.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: cargo install laze
+
+      - name: git clone ariel-os-hello repository
+        run: git clone https://github.com/ariel-os/ariel-os-hello
+
+      - name: use cargo generate to ariel-os-template
+        run: cargo generate --git https://github.com/ariel-os/ariel-os-template --name ariel-os-hello --destination /tmp
+
+      - name: verify that the file tree is identical between ariel-os-hello and the generated ariel-os-template
+        run: diff -r -x .git -x Cargo.lock ./ariel-os-hello /tmp/ariel-os-hello
```
nponsard, line 59: Is this a hard requirement ? 
I think we may want to showcase in the future some features in `ariel-os-hello` that don't appear in `ariel-os-template` because it doesn't fit as a base for a new project.

ROMemories: My understanding was that `ariel-os-hello` and `ariel-os-template` were meant to be identical (at least for some configuration of `ariel-os-template`), only different ways of bootstrapping a project (e.g., if you prefer not to use `cargo-generate` for instance). So I think we do need that check.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+          set +x
+          # Check which alternative versions are installed
+          IFS=":"; for p in $PATH; do ls $p/llvm-config-* $p/clang-* 2>/dev/null; done
+          sha256sum /usr/lib/llvm-*/lib/clang/*/include/stdint.h
+          true
+
+      # we verified previously that ariel-os-hello and the generated
+      # ariel-os-template have the same file tree, so we only build ariel-os-hello
+      - name: "ariel-os-hello compilation test"
+        run: cd ariel-os-hello && LAZE_BUILDERS=espressif-esp32-c6-devkitc-1,nrf52840dk,rpi-pico-w,st-nucleo-wb55 CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build
```
ROMemories: Can this be rewritten with a multiline string for clarity? Even splitting the `LAZE_BUILDERS` onto a separate line would help (with an `export`). Quoting that string would also not hurt.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
+name: Build templates
+
+on:
+  push:
+    branches:
+      - ci/scheduled-checks
+  schedule:
+    # run every morning at 3:17am
+    - cron: '17 3 * * *'
```
ROMemories: Can you add a [`workflow_dispatch`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatch)? The name isn't very explicit but it allows to manually trigger the workflow.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}-build
```
ROMemories: Otherwise this would conflict with the Build workflow (AFAIK this isn't scoped to the workflow file):
```suggestion
  group: ${{ github.head_ref || github.run_id }}-build-templates
```

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: install Rust toolchain for ESP devices
+        run: |
+          cargo install espup --locked
+          espup install
+
+      - name: install cargo-generate
+        run: cargo install cargo-generate
+
+      - name: install laze
+        run: cargo install laze
```
ROMemories, line 50: I'm torn on this one: using `cargo install` would compile from source every time. I suppose it's indeed best to follow the Getting Started instructions as closely as possible for now. If this turns out to take too long in CI, we'll switch to `esp-rs/xtensa-toolchain` and/or `taiki-e/install-action`.

.github/workflows/build-templates.yml
ROMemories: I don't think this actually does anything. It needs `--global` for that.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: Final Results
+    needs: [build]
+    steps:
+      - run: |
+          result="${{ needs.build.result }}"
+          if [[ $result == "success" || $result == "skipped" ]]; then
+            exit 0
+          else
+            exit 1
+          fi
```
ROMemories: If we're not using [a job matrix](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategymatrix) I think we just don't need this?

.github/workflows/build-templates.yml
ROMemories: Can you make sure to test this workflow in this PR so we can read the CI logs before merging?

jkr-ariel:
Needs https://github.com/ariel-os/ariel-os-hello/pull/13 to be merged for the "Build templates" workflow to succeed which is basically just https://github.com/ariel-os/ariel-os/pull/1572. I also made a PR for https://github.com/ariel-os/ariel-os-template but it is not blocking for the CI.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: "ariel-os-hello compilation test"
+        run: |
+          cd ariel-os-hello
+          export LAZE_BUILDERS='espressif-esp32-c6-devkitc-1,nrf52840dk,rpi-pico-w,st-nucleo-wb55'
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build
+
+  report-failure:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
```
ROMemories: I think these lines should be removed: the [`failure()` function](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#failure) returns true if a previous *step* of the CI job has failed. If the job has only one step and is gated on `failure()`, it will never run.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,101 @@
+name: Build templates
+
+on:
+  pull_request:
+    branches:
+      - main
```
ROMemories: We'll want to remove this trigger right before merging, so this workflow only runs in a scheduled manner.

.github/workflows/build-templates.yml
ROMemories: Still needs `--global`.

.github/workflows/build-templates.yml
ROMemories: Seeing that compiling `cargo-generate` takes 2 min 30 s, I think we should use `taiki-e/install-action` for that one (but keep `cargo install` for the others). `cargo-generate` is not specific to Ariel OS in any way.

.github/workflows/build-templates.yml
jkr-ariel: Speaking of scheduling, at what frequency do we want this workflow to run ? Nightly seems a bit too much, so maybe once or twice a week?

ROMemories: This depends on how long the workflow takes, which we'll only know after adding `--global`.

ROMemories: Actually without `--global`, the workflow seems to be taking < 15 min; that seems fine to run each night for now.

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,99 @@
...
+          IFS=":"; for p in $PATH; do ls $p/llvm-config-* $p/clang-* 2>/dev/null; done
+          sha256sum /usr/lib/llvm-*/lib/clang/*/include/stdint.h
+          true
+
+      # we verified previously that ariel-os-hello and the generated
+      # ariel-os-template have the same file tree, so we only build ariel-os-hello
+      - name: "ariel-os-hello compilation test"
+        run: |
+          cd ariel-os-hello
+          export CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password'
```
ROMemories: So we need to bring back the list of builders, and add one from the Xtensa architecture: `espressif-esp32-s3-devkitc-1`; otherwise the Xtensa toolchain installation won't be tested.

.github/workflows/build-templates.yml
ROMemories: Actually we indeed don't want `--global`: we only want the 4–5 builders on the one application; I guess I'm too used to seeing hundreds of builds, my bad :sweat_smile:

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,99 @@
...
+          IFS=":"; for p in $PATH; do ls $p/llvm-config-* $p/clang-* 2>/dev/null; done
+          sha256sum /usr/lib/llvm-*/lib/clang/*/include/stdint.h
+          true
+
+      # we verified previously that ariel-os-hello and the generated
+      # ariel-os-template have the same file tree, so we only build ariel-os-hello
+      - name: "ariel-os-hello compilation test"
+        run: |
+          cd ariel-os-hello
+          export LAZE_BUILDERS='espressif-esp32-c6-devkitc-1,nrf52840dk,rpi-pico-w,st-nucleo-wb55,espressif-esp32-s3-devkitc-1'
```
ROMemories: I'd rather keep these sorted; it makes spotting missing builders easier:
```suggestion
          export LAZE_BUILDERS='espressif-esp32-c6-devkitc-1,espressif-esp32-s3-devkitc-1,nrf52840dk,rpi-pico-w,st-nucleo-wb55'
```

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,95 @@
+name: Build templates
+
+on:
+  schedule:
+    # run every morning at 3:17am
```
ROMemories: The comment needs to be updated.

nponsard (approved):
Looks good to me !

PR #1552, opened 2025-11-21
fix(embassy): use generic timer queue when using threads

kaspar030:
# Description

Previously, using embassy timers from a thread (outside the executor), e.g. via `block_on()`, resulted in this panic:

```
panicked at /home/kaspar/.cargo/git/checkouts/embassy-0012190963c274d8-shallow/b552a87/embassy-executor/src/raw/waker.rs:38:9:                                 
Found waker not created by the Embassy executor. `embassy_time::Timer` only works with the Embassy executor.                                                   
```

This can be mitigated by enabling the "generic timer queue" in `embassy-time`, which this PR does.

Additionally, this adds an example showing how to use timers from threads, which can also be used to actually test this change.

This might have performance implications, also for timers not used by threads. The generic timer queue also needs preallocating timer slots. Feels suboptimal, but at least it works.

## Testing

Tested the example on `native` and `nrf52840dk`.

## Change Checklist

### DIFF
--- a/examples/threading-timers/Cargo.toml
+++ b/examples/threading-timers/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "example-threading-timers"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["time"] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+
+[lints]
+workspace = true
--- a/examples/threading-timers/src/main.rs
+++ b/examples/threading-timers/src/main.rs
@@ -1,0 +1,20 @@
+#![no_main]
+#![no_std]
+
+use ariel_os::{
+    asynch::blocker::block_on,
+    debug::{ExitCode, exit, log::info},
+    time::Timer,
+};
+
+#[ariel_os::thread(autostart)]
+fn main() {
+    info!("Hello from main thread!");
+
+    for i in 0..10 {
+        block_on(Timer::after_millis(100));
+        info!("Hello again #{}", i);
+    }
+
+    exit(ExitCode::SUCCESS);
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2732,6 +2732,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "example-threading-timers"
+version = "0.0.0"
+dependencies = [
+ "ariel-os",
+ "ariel-os-boards",
+]
+
+[[package]]
 name = "example-usb-serial"
 version = "0.0.0"
 dependencies = [
--- a/examples/laze.yml
+++ b/examples/laze.yml
@@ -31,6 +31,7 @@ subdirs:
   - threading-channel
   - threading-event
   - threading-multicore
+  - threading-timers
   - udp-echo
   - usb-keyboard
   - usb-serial
--- a/examples/threading-timers/README.md
+++ b/examples/threading-timers/README.md
@@ -1,0 +1,28 @@
+# threading-timers
+
+## About
+
+This application demonstrates how to use timers with threads.
+It starts a thread that prints a message ten times in 100ms intervals.
+
+## How to run
+
+In this directory, run
+
+    laze build -b nrf52840dk run
+
+## Expected output
+
+The output of this example is:
+
+    [INFO] Hello from main thread!
+    [INFO] Hello again #0
+    [INFO] Hello again #1
+    [INFO] Hello again #2
+    [INFO] Hello again #3
+    [INFO] Hello again #4
+    [INFO] Hello again #5
+    [INFO] Hello again #6
+    [INFO] Hello again #7
+    [INFO] Hello again #8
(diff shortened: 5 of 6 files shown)


src/ariel-os-embassy/Cargo.toml
```diff
@@ -130,20 +130,23 @@ eth-stm32 = ["ariel-os-hal/eth-stm32", "net", "eth"]
...
 ]
 ble-cyw43 = ["ble", "ariel-os-hal/ble-cyw43"]
 ble-peripheral = ["ble", "ariel-os-hal/ble-peripheral"]
 ble-central = ["ble", "ariel-os-hal/ble-central"]
 
-threading = ["dep:ariel-os-threads", "ariel-os-hal/threading"]
+threading = [
+  "dep:ariel-os-threads",
+  "ariel-os-hal/threading",
+  "embassy-time/generic-queue-128",
```
kaspar030: Going with the maximum here (128) for lack of precise information.

examples/threading-timers/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+    debug::{ExitCode, exit, log::info},
+    time::{Duration, Timer},
+};
+
+#[ariel_os::thread(autostart)]
+fn main() {
+    info!("Hello from main thread!");
+
+    for i in 0..10 {
+        block_on(Timer::after(Duration::from_millis(100)));
```
kaspar030: I'd say down the line we should wrap this into a thread-native timer api.

examples/threading-timers/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "example-threading-timers"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["time"] }
```
kaspar030, line 8: "time" is still needed as Cargo feature, compared to "threading", which happens to have a laze module. Which is confusing. But not this PR's problem.

examples/threading-timers/src/main.rs
ROMemories: Shorter and avoids importing `Duration` (we've switched to this in #1003):
```suggestion
        block_on(Timer::after_millis(100));
```

ROMemories (commented):
Tested successfully on espressif-esp32-c6-devkitc-1 (with and without `-s executor-thread`), rpi-pico2-w and st-nucleo-wb55.

However this reliably panics on stm32u83c-dk:

```
INFO  Hello from main thread!
└─ example_threading_timers::main @ src/main.rs:12  
panicked at src/ariel-os-threads/src/ensure_once.rs:42:33:
already borrowed: BorrowMutError
```

ROMemories:
I can reproduce the panic on an st-nucleo-c031c6, this likely happens on Cortex-M0+ chips (possibly because they lack atomics?).

examples/threading-timers/Cargo.toml
ROMemories: Seems fine to me

src/ariel-os-embassy/Cargo.toml
ROMemories: This seems to impact the size of `.data` not insignificantly:

| Cargo feature       | Size of `.data` (bytes) |
| ------------------- | ----------------------: |
| `generic-queue-8`   |  1144                   |
| `generic-queue-16`  |  1272                  |
| `generic-queue-32`  |  1528                   |
| `generic-queue-64`  |  2040                   |
| `generic-queue-128` |  3064                   |

Compiled `examples/threading-timers` for st-nucleo-c031c6.

kaspar030:
> I can reproduce the panic on an st-nucleo-c031c6, this likely happens on Cortex-M0+ chips (possibly because they lack atomics?).

I can reproduce it, too (on the `st-nucleo-c031c6`). Also on `thread-async-interop`, and also on main. :slightly_frowning_face:

That means there's an issue, but it is present in main, so unrelated to this PR.

src/ariel-os-embassy/Cargo.toml
kaspar030: 16b per timer queue entry :/

ROMemories: Also that increase in RAM usage comes from just enabling multithreading, even if not using time at all :/

kaspar030: Isn't the dependency to `embassy-time` only enabled if the `time` feature is enabled?

ROMemories: Ah, you need a question mark for that, otherwise this also enables `embassy-time`:

```suggestion
  "embassy-time?/generic-queue-128",
```

kaspar030:
> I can reproduce the panic on an st-nucleo-c031c6, this likely happens on Cortex-M0+ chips (possibly because they lack atomics?).

I can also reproduce it on the rpi-pico, which has an M0+. But only in single-core mode. That hints at an issue with critical-section, as multi-core mode uses a different impl.

PR #1555, opened 2025-11-22
refactor: make generic timer queue size configurable

kaspar030:
# Description

#1552 enabled the embassy generic timer queue with size 128 for all targets, which is too much for the smaller boards.

This PR:

- passes the generic-timer-queue-N features from ariel-os through ariel-os-embassy to the embassy crates
- uses just `_generic-timer-queue` to get the default
- thus, lowers the default to 64 entries
- sets `generic-timer-queue-8` for `ram-tiny` boards

This lacks a way to override the default for `ram-tiny` targets, but that can be tackled later.

## Changelog entry

Using `embassy-time` types (e.g., `Timer`) within threads is now supported. A generic timer queue is used, whose size can be configured using the `generic-timer-queue-*` Cargo features.

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -27,6 +27,7 @@ embassy-net = { workspace = true, optional = true, features = [
 ] }
 embassy-sync = { workspace = true }
 embassy-time = { workspace = true, optional = true }
+embassy-time-queue-utils = { workspace = true, optional = true }
 embassy-usb = { workspace = true, optional = true }
 
 embedded-hal = { workspace = true }
@@ -72,7 +73,7 @@ external-interrupts = [
   "ariel-os-embassy-common/external-interrupts",
   "ariel-os-hal/external-interrupts",
 ]
-time = ["dep:embassy-time"]
+time = ["dep:embassy-time", "dep:embassy-time-queue-utils"]
 
 ## Enables I2C support.
 i2c = [
@@ -143,8 +144,22 @@ ble-central = ["ble", "ariel-os-hal/ble-central"]
 threading = [
   "dep:ariel-os-threads",
   "ariel-os-hal/threading",
-  "embassy-time?/generic-queue-128",
+  # this sets the default timer queue size, which is currently 64. See `generic-queue-*` features.
+  "embassy-time-queue-utils?/_generic-queue",
 ]
+
+# See ariel-os/Cargo.toml for more details
+## Generic Queue with 8 timers
+timer-generic-queue-8 = ["embassy-time?/generic-queue-8"]
+## Generic Queue with 16 timers
+timer-generic-queue-16 = ["embassy-time?/generic-queue-16"]
+## Generic Queue with 32 timers
+timer-generic-queue-32 = ["embassy-time?/generic-queue-32"]
+## Generic Queue with 64 timers
+timer-generic-queue-64 = ["embassy-time?/generic-queue-64"]
+## Generic Queue with 128 timers
+timer-generic-queue-128 = ["embassy-time?/generic-queue-128"]
+
 network-config-ipv4-static = ["network-config-override"]
 network-config-override = []
 override-usb-config = []
--- a/src/ariel-os/Cargo.toml
+++ b/src/ariel-os/Cargo.toml
@@ -160,6 +160,21 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
 # Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
 ble-central = ["ble", "ariel-os-embassy/ble-central"]
 
+#! ## Timer queue size configuration
+#! Should only be enabled when threading is enabled, too.
+#! Default is 64 entries for regular boards and 8 entries for those with little ram.
+#! When more than this number of concurrent timers are used, the system will loop (and not sleep/idle anymore).
+## Enables support for up to 8 concurrent timers. (default on `ram-tiny` boards)
+timer-generic-queue-8 = ["ariel-os-embassy/timer-generic-queue-8"]
+## Enables support for up to 16 concurrent timers.
+timer-generic-queue-16 = ["ariel-os-embassy/timer-generic-queue-16"]
+## Enables support for up to 32 concurrent timers.
+timer-generic-queue-32 = ["ariel-os-embassy/timer-generic-queue-32"]
+## Enables support for up to 64 concurrent timers. (default)
+timer-generic-queue-64 = ["ariel-os-embassy/timer-generic-queue-64"]
+## Enables support for up to 128 concurrent timers.
+timer-generic-queue-128 = ["ariel-os-embassy/timer-generic-queue-128"]
+
 #! ## Development and debugging
 # Enables the debug console, required to use
 # [`println!`](ariel_os_debug::println).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -298,6 +298,7 @@ dependencies = [
  "embassy-stm32",
  "embassy-sync 0.6.2",
  "embassy-time",
+ "embassy-time-queue-utils",
  "embassy-usb",
  "embedded-hal 1.0.0",
  "embedded-hal-async",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ embassy-rp = { version = "0.4", default-features = false }
 embassy-stm32 = { version = "0.2", default-features = false }
 embassy-sync = { version = "0.6.1", default-features = false }
 embassy-time = { version = "0.4.0", default-features = false }
+embassy-time-queue-utils = { version = "0.1.0", default-features = false }
 embassy-usb = { version = "0.4.0", default-features = false }
 
 embedded-hal = { version = "1.0.0", default-features = false }
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -941,6 +941,8 @@ modules:
       global:
         isr_stacksize_required_default: "512"
         executor_stacksize_required_default: "512"
+        FEATURES:
+          - ariel-os/generic-queue-8
 
   - name: ram-small
     help: This module marks MCUs with little RAM (currently, the ones which cannot run HTTP or CoAP)


src/ariel-os-embassy/Cargo.toml
```diff
@@ -72,7 +73,7 @@ external-interrupts = [
   "ariel-os-embassy-common/external-interrupts",
   "ariel-os-hal/external-interrupts",
 ]
-time = ["dep:embassy-time"]
+time = ["dep:embassy-time", "dep:embassy-time-queue-utils"]
```
ROMemories, line 76: Do we actually need this dependency here? I've successfully tested the `blinky` example (which enables `time`) on the STM32C031 without this dependency.

src/ariel-os/Cargo.toml
```diff
@@ -160,6 +160,21 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
 # Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
 ble-central = ["ble", "ariel-os-embassy/ble-central"]
 
+#! ## Timer queue size configuration
+#! Should only be enabled when threading is enabled, too.
+#! Default is 64 entries for regular boards and 8 entries for those with little ram.
+#! While more than this number of concurrent timers are used, the system will loop (and not sleep/idle anymore).
+## Generic Queue with 8 timers
+generic-queue-8 = ["ariel-os-embassy/generic-queue-8"]
```
ROMemories: I think we should rename these Cargo features when exposing them (and maybe on `ariel-os-embassy` but it's less critical as it's internal). "generic-queue" doesn't make enough sense as a feature name on `ariel-os`.
```suggestion
timer-queue-8 = ["ariel-os-embassy/generic-queue-8"]
```

src/ariel-os/Cargo.toml
```diff
@@ -160,6 +160,21 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
 # Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
 ble-central = ["ble", "ariel-os-embassy/ble-central"]
 
+#! ## Timer queue size configuration
+#! Should only be enabled when threading is enabled, too.
+#! Default is 64 entries for regular boards and 8 entries for those with little ram.
+#! While more than this number of concurrent timers are used, the system will loop (and not sleep/idle anymore).
+## Generic Queue with 8 timers
```
ROMemories: For consistency with other Cargo features:
```suggestion
## Enables support for up to 8 concurrent timers.
```

src/ariel-os/Cargo.toml
```diff
@@ -160,6 +160,21 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
 # Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
 ble-central = ["ble", "ariel-os-embassy/ble-central"]
 
+#! ## Timer queue size configuration
+#! Should only be enabled when threading is enabled, too.
+#! Default is 64 entries for regular boards and 8 entries for those with little ram.
+#! While more than this number of concurrent timers are used, the system will loop (and not sleep/idle anymore).
```
ROMemories: Could this be clarified further? Does "loop"ing means busy looping? When exactly during runtime would it loop?
```suggestion
#! When more than this number of concurrent timers are used, the system will loop (and not sleep/idle anymore).
```

src/ariel-os-embassy/Cargo.toml
```diff
@@ -143,8 +144,22 @@ ble-central = ["ble", "ariel-os-hal/ble-central"]
 threading = [
   "dep:ariel-os-threads",
   "ariel-os-hal/threading",
-  "embassy-time?/generic-queue-128",
+  # this sets the default timer queue size, which is currently 64. See `generic-queue-*` features.
+  "embassy-time-queue-utils?/_generic-queue",
```
ROMemories, line 148: It's not exactly obvious to me why we need to go through this crate? Is it a way to have a default value that can still be overridden?

src/ariel-os/Cargo.toml
```diff
@@ -160,6 +160,21 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
...
+## Generic Queue with 8 timers
+generic-queue-8 = ["ariel-os-embassy/generic-queue-8"]
+## Generic Queue with 16 timers
+generic-queue-16 = ["ariel-os-embassy/generic-queue-16"]
+## Generic Queue with 32 timers
+generic-queue-32 = ["ariel-os-embassy/generic-queue-32"]
+## Generic Queue with 64 timers
+generic-queue-64 = ["ariel-os-embassy/generic-queue-64"]
+## Generic Queue with 128 timers
+generic-queue-128 = ["ariel-os-embassy/generic-queue-128"]
```
ROMemories: I tried to enable these Cargo features individually on `ariel-os` in the `threading-timers` example and checked the `.data` size. I was expecting it to vary depending on the enabled Cargo feature; however I could not observe that :/

src/ariel-os-embassy/Cargo.toml
kaspar030: Not necessarily, *if* we always set one of the `generic-queue-N` features.
The `_generic-queue` feature is not exposed on `embassy-time`. If it is set on `e-t-q-u`, it sets the default of 64.

src/ariel-os/Cargo.toml
kaspar030: I thought about that and find passing features through 3 crates even more confusing if they're renamed.

src/ariel-os/Cargo.toml
kaspar030: IIUC when no free slot can be found, a random `Waker` is popped and awoken. So the executor(s) and thread(s) for that waker would `poll()` that worker instead of waiting for the timer.

What wording would make that clearer?

src/ariel-os-embassy/Cargo.toml
kaspar030: see above
We can have a default value using laze.

src/ariel-os/Cargo.toml
kaspar030: Now that is an issue. lemme check

kaspar030: I built the `thread-async-interop` example, setting a `generic-queue-N` different than 64. It changed the size.
The feature only has effect if both `time` and `threading` are enabled.

kaspar030: also on threading-timers. it is `data` that changes

ROMemories: I can actually see the size change on an STM32, but I was testing on espressif-esp32-c6-devkitc-1 before and there is no change on this one. Same on espressif-esp32-s3-devkitc-1; is it expected to not change on ESP?

src/ariel-os/Cargo.toml
ROMemories: More confusing for whom? Users? I don't think the following is very clear:

```toml
[dependencies]
ariel-os = { path = "../../src/ariel-os", features = ["generic-queue-8", "time"] }
```
I would personally have no idea what/which queue that is referring to.

src/ariel-os-embassy/Cargo.toml
ROMemories: > Not necessarily, _if_ we always set one of the `generic-queue-N` features.

I'm not sure I'm following. The current line would enable the dependency as soon as `time` is enabled, even if `threading` isn't. Couldn't we move it into `threading` instead?

kaspar030: So `etqu` is a dependency of `embasssy-time` anyways. The `generic-queue-N` are features of `etqu` passed through by `e-t`. `_generic-queue` is not.

a) We want to enable the generic queue if both timers and threading are enabled.
b) We want to be able to override the default.

If we add `etqu` as direct dependency and enable it if `time` is enabled, then set `_generic-queue` if `threading` and the `etqu` dep is enabled, we've tackled a).

Now an issue I see is that while we can now override the default, doing so will enable the generic queue also if `threading` is disable.

I'm inclined to set all the `generic-queue-N` also directly on `etqu`.

`etqu` being a dep of `et` is an implementation detail, so that we have to fiddle with this is unfortunate. Do you see another way?

src/ariel-os/Cargo.toml
kaspar030: Ok. At which level should we do this? In `ariel-os` ofc, should the `ariel-os-embassy` features also have better names? I think yes.

Maybe prefix with `timer-`, so the features are called `timer-generic-queue-N`?

src/ariel-os/Cargo.toml
kaspar030: That's expected, esp-hal does its own timer queue IIUC.

src/ariel-os/Cargo.toml
ROMemories: > Ok. At which level should we do this? In `ariel-os` ofc, should the `ariel-os-embassy` features also have better names? I think yes.

Agreed.

> Maybe prefix with `timer-`, so the features are called `timer-generic-queue-N`?

I don't think we should have `generic` in the name, that's an implementation detail. The name should mention *what* the feature is for (timers, and it's some sort of queue so the queue needs sizing), but the *nature* of the queue is an implementation detail (if the nature of the queue is relevant for users to pick a sizing, the properties that are relevant should be explained, but not the implementation itself).

src/ariel-os-embassy/Cargo.toml
ROMemories: > a) We want to enable the generic queue if both timers and threading are enabled.
> b) We want to be able to override the default.

Agreed.

> So `etqu` is a dependency of `embasssy-time` anyways.

Not really, it's an optional dependency that's [only enabled when one of `generic-queue-*` or `std` or `wasm` is](https://lib.rs/crates/embassy-time) (linking to lib.rs here because it makes that pretty clear), which is not the case when only `time` is enabled (and `threading` isn't). This can be checked using the following (`embassy-time-queue-utils` is *not* a dep of `embassy-time` in that output):

```sh
laze -C examples/blinky/ build -b nrf52840dk tree --no-dedupe
```

However, it is indeed a dependency of `embassy-nrf`, `embassy-rp`, and `embassy-stm32` when the time driver is enabled so in practice I think it would always be in the build for us anyway, so I agree it would not change much to make it enabled by `time`.

I'm still not following though why we can't simply move `dep:embassy-time-queue-utils` in `threading`, as that's the only place we're using it directly.

src/ariel-os/Cargo.toml
ROMemories: Ok I think it's fine for now then, with the "when" in (we can always improve these docs later).

src/ariel-os-embassy/Cargo.toml
kaspar030: If we move it in 'threading', we'd always need to enable the generic queue, if threading is enabled, no? As it is now, we can enable the generic queue only when the `time` feature is enabled.

ROMemories: Indeed, that works

src/ariel-os/Cargo.toml
kaspar030: > > Maybe prefix with `timer-`, so the features are called `timer-generic-queue-N`?
> 
> I don't think we should have `generic` in the name, that's an implementation detail. The name should mention _what_ the feature is for (timers, and it's some sort of queue so the queue needs sizing), but the _nature_ of the queue is an implementation detail (if the nature of the queue is relevant for users to pick a sizing, the properties that are relevant should be explained, but not the implementation itself).

I think we should at best prefix with `timer-` so a) there's an idea what queue is meant, b) the original name is still in so grepping finds it.

People "in the know" might be better informed if the "qeneric-queue" is still visible.

ROMemories: Fine, but in general 

> People "in the know" might be better informed if the "qeneric-queue" is still visible.

would just mean our documentation would be lacking.

src/ariel-os/Cargo.toml
```diff
@@ -165,6 +160,21 @@ ble-peripheral = ["ble", "ariel-os-embassy/ble-peripheral"]
 # Enables BLE to act as a central. Can be used in conjunction with `ble-peripheral`.
 ble-central = ["ble", "ariel-os-embassy/ble-central"]
 
+#! ## Timer queue size configuration
+#! Should only be enabled when threading is enabled, too.
+#! Default is 64 entries for regular boards and 8 entries for those with little ram.
+#! When more than this number of concurrent timers are used, the system will loop (and not sleep/idle anymore).
+## Enables support for up to 8 concurrent timers. (default on `ram-tiny` boards)
```
ROMemories, line 167: Also users have no idea what `ram-tiny` boards are (not a blocker for merging though).

PR #1559, opened 2025-11-24
refactor(esp32): mark that MCU as disabled until we fix linking

ROMemories:
# Description

Temporarily disables building for thr (plain) ESP32 as linking of `isr_stack` is broken on it.

This is not a breaking change as it is not currently working.

## Testing

Checked with the following command that nothing gets built:

```sh
laze build --global -s 'context::esp32'
```

## Issues/PRs References

- Unblocks #1454 

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -390,6 +390,7 @@ contexts:
     parent: esp
     selects:
       - xtensa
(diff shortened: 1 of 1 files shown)


kaspar030:
Linking is fixed in #1328, will probably split that out.

PR #1565, opened 2025-11-25
refactor(stack-painting)!: revisit `Stack`'s API for clarity

ROMemories:
# Description

This renames most of `Stack`'s methods with the goal of making their meaning clearer.

## How to review this PR

The docs can be seen in #1538 (along with a few other minor changes from that PR).

## Issues/PRs References

- `Stack` was introduced in #977

## Change Checklist

### DIFF
--- a/src/ariel-os-rt/src/stack.rs
+++ b/src/ariel-os-rt/src/stack.rs
@@ -115,14 +115,14 @@ impl Stack {
     /// Returns the amount of currently free stack space.
     #[must_use]
     #[inline(always)]
-    pub fn free(&self) -> usize {
-        self.size() - self.used()
+    pub fn current_free_space(&self) -> usize {
+        self.size() - self.current_usage()
     }
 
     /// Returns the amount of currently used stack space.
     #[must_use]
     #[inline(always)]
-    pub fn used(&self) -> usize {
+    pub fn current_usage(&self) -> usize {
         self.highest - (sp() - STACK_PAINT_IGNORE)
     }
 
@@ -145,12 +145,12 @@ impl Stack {
         free
     }
 
-    /// Returns the maximum stack space used since last repaint.
+    /// Returns the peak amount of stack used since last repaint.
     ///
     /// This re-calculates and thus runs in `O(n)`!
     #[must_use]
     #[inline(always)]
-    pub fn used_max(&self) -> usize {
+    pub fn peak_usage(&self) -> usize {
         self.size() - self.free_min()
     }
 
@@ -180,7 +180,7 @@ impl Stack {
             // current stack frame's stack pointer.
             // This does not prevent this from being interrupted by an ISR, in which case
             // the stack is dirtied again, but that doesn't cause any unsafety and just
-            // makes any following `used_max()` call include whatever the ISR wrote on this stack.
+            // makes any following `peak_usage()` call include whatever the ISR wrote on this stack.
             unsafe {
                 write_volatile(pos as *mut u8, STACK_PAINT_COLOR);
             }


src/ariel-os-rt/src/stack.rs
```diff
@@ -94,27 +94,27 @@ impl Stack {
...
+    pub fn current_usage(&self) -> usize {
         self.highest - sp()
     }
 
     /// Returns the minimum free stack space since last repaint.
     ///
     /// This re-calculates and thus runs in `O(n)`!
     #[must_use]
-    pub fn free_min(&self) -> usize {
+    fn free_min(&self) -> usize {
```
ROMemories, line 133: It's hard to find a good name for this method and I don't see when users could need this, so this makes this method private. Users can still use `total_size() - peak_usage()` if needed.

kaspar030: IMO the names all made sense in tandem. `free_min()` + `used_max()`, and with `free()` and `used()`.

kaspar030:
@nponsard can you take a look? I think the API was fine and gets less consistent with this PR. We need more eyes here.

src/ariel-os-rt/src/stack.rs
```diff
@@ -94,27 +94,27 @@ impl Stack {
 
     /// Returns the total size of the current stack.
     #[must_use]
-    pub fn size(&self) -> usize {
+    pub fn total_size(&self) -> usize {
```
kaspar030: this one I think makes sense

kaspar030: But on the other hand there's only a total size. `stack.size()`.

nponsard: Would naming it `stack.capacity()` make sense here ? Since we're talking about usage it's an usage relative to the allocated capacity.

src/ariel-os-rt/src/stack.rs
nponsard: Maybe `lowest_free_space()` could make sense ?

src/ariel-os-rt/src/stack.rs
ROMemories: > Would naming it `stack.capacity()` make sense here ?

I like that, plus that mimics `Vec::capacity()`.

kaspar030: But the term is "stack size", not "stack capacity".

src/ariel-os-rt/src/stack.rs
```diff
@@ -94,27 +94,27 @@ impl Stack {
...
         self.highest - self.lowest
     }
 
     /// Returns the amount of currently free stack space.
     #[must_use]
-    pub fn free(&self) -> usize {
-        self.size() - self.used()
+    pub fn current_free_space(&self) -> usize {
```
nponsard, line 118: I do agree here that `stack.free()` could be misleading, it could be interpreted as a function that frees the memory on the stack or clears it.

kaspar030: `unused()`?

src/ariel-os-rt/src/stack.rs
bergzand: If we go with `capacity()` below, this could be named `excess_capacity()` or `unused_capacity()` or along those line

src/ariel-os-rt/src/stack.rs
ROMemories: `unused` has the original issue of being unclear about whether that refers to the *currently* unused stack space or to the space where the paint is completely fresh. That's the main issue I'm trying to address with this rename.

src/ariel-os-rt/src/stack.rs
bergzand: Preference for `capacity` from my side.

kaspar030: Yeah, but a stack has a size, not a capacity.

[Rust](https://doc.rust-lang.org/std/thread/struct.Builder.html#method.stack_size)
[pthreads](https://man7.org/linux/man-pages/man3/pthread_attr_setstacksize.3.html)
[Zephyr](https://docs.zephyrproject.org/latest/doxygen/html/group__thread__apis.html#gad5b0bff3102f1656089f5875d999a367)
[Windows](https://learn.microsoft.com/en-us/windows/win32/procthread/thread-stack-size)

src/ariel-os-rt/src/stack.rs
kaspar030: Then go with `unused_currently()` and `unused_since_last_repaint()`.

src/ariel-os-rt/src/stack.rs
kaspar030: Last thing we should do is create new terminology.

src/ariel-os-rt/src/stack.rs
ROMemories, line 133: I've found an other argument for keeping this method private: when playing with the test from #1577, I've found it was way too easy to get results where `current_usage() + free_min() != total_size()`, even when called within the same surrounding function (not passing the `Stack` instance around between functions/stack frames). This is likely because the call to `free_min()` influences the stack usage itself (after the first call to `current_usage()`).

As I still don't see a use case for this method (while I do totally see how `peak_usage()` can be used to size `executor_stack_required` for instance), I think it's best to just keep the method private.

src/ariel-os-rt/src/stack.rs
ROMemories: I've dropped the commit renaming `size()` to `total_size()`.

src/ariel-os-rt/src/stack.rs
kaspar030: I don't care enough

src/ariel-os-rt/src/stack.rs
```diff
@@ -136,14 +136,14 @@ impl Stack {
         free
     }
 
-    /// Returns the maximum stack space used since last repaint.
+    /// Returns the peak amount of stack used since last repaint.
     ///
     /// Equivalent to `size() - free_min()`.
```
kaspar030: I think this needs adjusting, with `free_min() now being private.

ROMemories: Indeed! I've simply removed the line.

PR #1567, opened 2025-11-25
refactor(threads)!: move `blocker` into ariel-os-threads

kaspar030:
# Description

This moves the `block_on()` implementation into `ariel-os-threads`, making it available to more internal crates (e.g., hals).

Within `ariel_os`, it is now in `ariel_os::thread::block_on()` (moved from `ariel_os::asynch::blocker::block_on`), making this a breaking change.

## Changelog entry

The `ariel_os::asynch::blocker::block_on()` function has been moved into `ariel_os::thread` and is now `ariel_os::thread::block_on()`.

## Change Checklist

### DIFF
--- a/src/ariel-os-threads/src/blocker.rs
+++ b/src/ariel-os-threads/src/blocker.rs
@@ -2,9 +2,10 @@
 
 #![expect(unsafe_code)]
 
-use ariel_os_threads::{ThreadId, current_tid, flags, flags::ThreadFlags};
 use core::pin::Pin;
 use core::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
+
+use crate::{ThreadId, current_tid, flags, flags::ThreadFlags};
 
 const THREAD_FLAG_WAKER: ThreadFlags = 1; // TODO: find more appropriate value
 
--- a/src/ariel-os-threads/src/lib.rs
+++ b/src/ariel-os-threads/src/lib.rs
@@ -36,6 +36,7 @@
 
 mod arch;
 mod autostart_thread;
+mod blocker;
 mod ensure_once;
 mod thread;
 mod threadlist;
@@ -62,6 +63,7 @@ pub mod events {
 }
 
 pub use ariel_os_runqueue::{RunqueueId, ThreadId};
+pub use blocker::block_on;
 pub use thread_flags as flags;
 
 #[cfg(feature = "core-affinity")]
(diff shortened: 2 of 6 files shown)


src/ariel-os-threads/src/lib.rs
```diff
@@ -43,6 +43,7 @@ mod threadlist;
 #[cfg(feature = "multi-core")]
 mod smp;
 
+pub mod blocker;
```
ROMemories: This makes the `blocker` module also visible in the `ariel_os::thread` module; I don't think that's what was intended.

kaspar030: It has to be exported somehow. you mean it should be doc hidden?

kaspar030: As is `ariel-os-embassy` imports the module into `asynch`, so the re-exports stays the same.

ROMemories: > you mean it should be doc hidden?

Yes or, even better, refactored in a way similar to the `api` module of `ariel-os-embassy` if that makes sense in that case.

kaspar030: this feels messy. `blocker` has no deps other than core, and is only needed when using threads. IMO the clean way is to just move it into `threads` and expose it as such.

src/ariel-os-threads/src/lib.rs
```diff
@@ -43,6 +43,8 @@ mod threadlist;
 #[cfg(feature = "multi-core")]
 mod smp;
 
+mod blocker;
```
ROMemories: optional: this would be better grouped with the other private modules above

src/ariel-os-threads/src/blocker.rs
```diff
@@ -2,7 +2,7 @@
 
 #![expect(unsafe_code)]
 
-use ariel_os_threads::{ThreadId, current_tid, flags, flags::ThreadFlags};
+use crate::{ThreadId, current_tid, flags, flags::ThreadFlags};
```
ROMemories: optional: `crate` imports would be better separated in a different import group after `core`/`std` imports (and after other crate imports, but there is non of these here)

src/ariel-os-threads/src/lib.rs
ROMemories: I agree it is better this way, even if that means having a breaking change.

ROMemories (approved):

```sh
laze -C examples/thread-async-interop/ build -b st-steval-mkboxpro run
```

Checked that there were no remaining occurrences of `blocker` in the codebase, except for the very definition of the module.

ROMemories (approved):

PR #1568, opened 2025-11-25
chore(sensors): make `cargo-manifest-check` happy

ROMemories:
# Description

This makes `cargo-manifest-check` happy with `ariel-os-sensors` and adds the check to the CI.

Using `ci-build:skip` is fine as the Clippy check checks that the manifest is still valid.

## Change Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -411,7 +411,7 @@ jobs:
           tool: cargo-manifest-check@0.2
 
       - name: cargo-manifest-check
-        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -411,7 +411,7 @@ jobs:
           tool: cargo-manifest-check@0.2
 
       - name: cargo-manifest-check
-        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
+        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-sensors src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
```
jkr-ariel, line 414: Maybe we could put each crate on a single line as we add more for readability.

ROMemories: This should only be temporary, we should be able to remove the entire list after #1328 lands and we make the remaining crates successfully. If this drags on, we'll consider reformatting.

PR #1570, opened 2025-11-26
docs(book): clarify how to disable laze modules through configuration

ROMemories:
# Description

This slightly clarifies how to disable laze modules through laze configuration files, following https://github.com/ariel-os/ariel-os/pull/1377#discussion_r2564538859.

## Change Checklist

### DIFF
--- a/book/src/build-system.md
+++ b/book/src/build-system.md
@@ -97,9 +97,9 @@ apps:
 The specified modules will be enabled for the application, some of which may enable associated Cargo features (as individually documented for each laze module).
 If a module is not available on a target—e.g., because networking is not available on the target, or not yet supported by Ariel OS—laze will prevent the application to be compiled for that target.
 
-### Disabling/forbidding laze modules for an application
+### Disabling laze modules for an application
 
-Conversely, to forbid laze modules through the configuration file, the [`conflicts` array][laze-module-conflicts-book] is used:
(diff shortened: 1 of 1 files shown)


book/src/build-system.md
```diff
@@ -97,7 +97,7 @@ apps:
 The specified modules will be enabled for the application, some of which may enable associated Cargo features (as individually documented for each laze module).
 If a module is not available on a target—e.g., because networking is not available on the target, or not yet supported by Ariel OS—laze will prevent the application to be compiled for that target.
 
-### Forbidding laze modules for an application
+### Disabling/forbidding laze modules for an application
```
ROMemories: It seems nothing was linking to that section (grepped for `forbidding-laze`), so no links are broken by this title change.

kaspar030:
```suggestion
### Disabling laze modules for an application
```

Let's drop the "forbid" wording. It is 'enable' or 'disable' a laze module.

ROMemories: Fine; the reason why I favored "forbid" is that this sounded more "declarative" (as I think laze configuration should be) while "disable" (and "enable") sounds more "imperative."

kaspar030: BTW, laze accepts both "conflicts" and "disables" in laze files. Only "-d"/"--disables" on the CLI though.

kaspar030 (approved):

PR #1574, opened 2025-11-27
docs: document support for boards with multiple MCUs and non-SMP cores

### LINKED ISSUE #1506: Document support for boards with multiple MCUs and for non-SMP cores
Currently there isn't a way for us to document support for laze builders such as `nrf5340dk-net`, or `nordic-thingy-91-x-nrf5340` and `nordic-thingy-91-x-nrf5340-net` (which don't exist yet), or AMP cores like on the STM32H755 (currently we only support the Cortex-M7).

We need to be able to document support for these laze builders in the book. It is unclear how the support pages for these should be structured: we could either have one page per laze builder or one page per actual piece of hardware, which would list the different laze builders. I personally think the latter would be clearer, but it might be a bit more difficult to implement.

Even if we don't yet support all of these variants, it seems useful to already start working on the docs, as that can give us some insights about how best to expose these, or even reconsider the laze builders we already have.

jkr-ariel:
# Description

This PR aims to document support for boards that contain more than 1 MCU (like the Nordic Thingy91:X with a nRF5340 and nRF9151) and MCUs having non symetric cores (like the STM32H755 with a Cortex-M7 and Cortex-M4).

## Issues/PRs References

- Closes #1506
- Depends on #1619
- Depends on #1658.

## Changelog entry

The hardware support documentation has been revamped: chips and boards now have dedicated pages, and board pages list laze builders that can be used.

## Change Checklist


jkr-ariel:
In order to document support for boards that have multiple MCUs, the `boards` map in `doc/support_matrix.yml` has been renamed to `builders` and a new key `boards` has been added, which is a list of objects that have a `name`, `url`, and `builders` which is a list of string corresponding to the laze builder(s) used by the board.

This will allow us to split the support matrix for a board between its laze builders.

jkr-ariel:
The board pages have been updated with a section per laze builder containing the support matrix for that builder.
The board index page now lists all the boards in a single section (previously the list was grouped by tier, but it is unclear wether we want tiers to be applied to builders or boards, see my comment on [#1506](https://github.com/ariel-os/ariel-os/issues/1506#issuecomment-3632875358) for more context).
The chip pages are unchanged.
The "Overview" page is currently in a broken state until we decide what to do (how to render boards vs builders + tiers).

doc/support_matrix.yml
```diff
@@ -817,58 +808,53 @@ boards:
...
-    url: https://web.archive.org/web/20250115224621/https://www.nordicsemi.com/Products/Development-hardware/nrf5340-dk
     chip: nrf5340
     tier: "1"
     support:
       user_usb: supported
       ethernet_over_usb: supported
+      ble: not_available
+
+  nrf5340dk-net:
+    chip: nrf5340
```
ROMemories: This needs a dedicated "chip", it particular we do support the HWRNG on it:

https://github.com/ariel-os/ariel-os/blob/7c95fb8ecda53730581d66a6d22450f3643ac0c2/src/ariel-os-nrf/src/hwrng.rs#L12

ROMemories: This begs the question of what even *is* a "chip" in this support info in the case of this two-MCUs-in-one-package chip: the net core does have an RNG that we currently support for instance. I need to think about this for a bit.

book/templates/board-page.md.tmpl
```diff
@@ -33,8 +40,15 @@ dt, dd {
   display: inline;
 }
 </style>
-{% for functionality in board_info.functionalities %}
-  {%- for comment in functionality.comments %}
+
+## References
```
ROMemories: Can we keep these references above the list of laze builders? I think the manufacturer link should be easy to get to for instance.

doc/gen_book.rs
```diff
@@ -107,16 +110,17 @@ impl SubCommandMatrix {
         // TODO: read the order from the YAML file instead?
         board_info.sort_unstable_by_key(|b| b.name.to_lowercase());
 
-        let boards = board_info
-            .into_iter()
-            .filter(|board_support| board_support.tier == self.tier.to_string())
-            .collect::<Vec<_>>();
+        // FIXME: deal with how to render the overview tables later on
```
ROMemories: Adding a comment here so we don't forget.

doc/schema.rs
```diff
@@ -37,15 +38,21 @@ pub struct ChipInfo {
     pub support: HashMap<String, SupportInfo>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct BuilderInfo {
+    pub chip: String,
+    pub tier: String,
```
ROMemories, line 45: I'm not *sure* tiers are an attribute of builders, I'd say they are an attribute of *boards*: you can't own one of the builders of a board but not the others, builders are only a logical concept. Regarding testing, I suppose the difference in testing that builders from a board could have is not too different from the different levels of testing that different pieces of functionalities could undergo even if part of the same builder.

Would considering tiers an attribute of builders make something I haven't considered simpler?

jkr-ariel: The main reason for having tiers in builders for now is that I was thinking of a case where a single board has multiple builders with different level of support e.g. an imaginary board with a nRF52840 and a nRF54L15 which is not currently supported, but users should be able to target and use the nrf52840 builder regardless.

ROMemories: Good point, fine to keep it as-is then.

doc/support_matrix.yml
jkr-ariel: In the case of MCUs like the nRF5340, it makes sense to me to have separate laze context for each core (so `nrf5340` or `nrf5340-app` and `nrf5340-net`) that the builders can just inherit from.

ROMemories: `nrf5340-net` is [already a laze "chip"](https://github.com/ariel-os/ariel-os/blob/54a3a4d236df6a6bfeba2b7dee130391ec04b259/boards/ariel-chips.yaml#L28) actually. Can you add the support info for it?

doc/gen_book.rs
ROMemories: I guess we either render support for laze builders instead of "boards", or we add a new column for laze builders (with duplicated rows for laze builders of the same board). No strong opinion here. Maybe we can also consider removing the chip column to make some room, as that info is now given in each board page.

doc/support_matrix.yml
ROMemories: Done in #1658.

doc/gen_book.rs
jkr-ariel: I think replacing the chips column with laze builders can work, I'll try this and we'll see how it looks.

jkr-ariel: Can now see how it looks: https://ariel-os-docs-preview-pr1574.surge.sh/dev/docs/book/hardware-functionality-support.html.
Boards that have more than one laze builder have sub-rows for the support info of each builder (currently only nRF5340-DK)

doc/support_matrix.yml
```diff
@@ -836,58 +824,53 @@ boards:
...
```
ROMemories: This can use the right chip as #1658 is in:
```suggestion
    chip: nrf5340-net
```

book/src/hardware-functionality-support.md
```diff
@@ -5,28 +5,4 @@
 > [!NOTE]
 > The tables below indicate whether we support using the piece of functionality in a portable manner, through an abstraction layer and platform-aware configuration.
```
ROMemories:
```suggestion
> The table below indicates whether we support using the piece of functionality in a portable manner, through an abstraction layer and platform-aware configuration.
```

book/src/hardware-functionality-support.md
```diff
@@ -5,28 +5,4 @@
...
 
-Hardware support is organized into tiers, each with their own testing policy:
-
-1. Tier 1 hardware gets regularly tested, either automatically or manually.
-2. Tier 2 hardware only gets tested infrequently, but Ariel OS maintainers do have access to the hardware.
-3. Tier 3 hardware is build-tested only, as Ariel OS maintainers do not have access to the hardware.
-
-Tiers therefore are not related to the functionality coverage of each piece of hardware, and only says something about how much testing they undergo.
```
ROMemories, line 14: I think we need to keep this, otherwise there's no definitions of the tiers anywhere anymore.

book/templates/support-matrix.html.tmpl
```diff
@@ -2,31 +2,32 @@
 <table class="support-matrix">
   <thead>
     <tr>
-      <th colspan="2">Chip</th>
       <th colspan="2">Testing Board</th>
```
ROMemories: Could we have an extra column with the tier level for each builder (between "Testing Board" and "Functionality")? Otherwise users have to go through each board page individually to find it. We could later consider sorting the table based on the tier level as well, but this is better done in a follow-up.

PR #1576, opened 2025-11-27
fix(stack-painting): inline reporting methods to not bias the result

ROMemories:
# Description

This makes a best effort attempt at inlining the `Stack` `pub` methods that report on stack usage, as their execution otherwise influence the results.

## Testing

The change of behavior is observable using the test from #1577.

## Change Checklist

### DIFF
--- a/src/ariel-os-rt/src/stack.rs
+++ b/src/ariel-os-rt/src/stack.rs
@@ -1,6 +1,10 @@
 //! Stack usage helpers.
 
 #![expect(unsafe_code)]
+#![expect(
+    clippy::inline_always,
+    reason = "otherwise usage reporting functions influence the result"
+)]
 
 use core::{marker::PhantomData, ptr::write_volatile};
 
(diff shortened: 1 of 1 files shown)


src/ariel-os-rt/src/stack.rs
```diff
@@ -94,49 +95,48 @@ impl Stack {
 
     /// Returns the total size of the current stack.
     #[must_use]
-    pub fn size(&self) -> usize {
+    #[inline(always)]
```
ROMemories, line 101: I'm aware that even `inline(always)` doesn't *guarantee* that they are actually inlined by the compiler, it's only a strong hit for it, but I guess that's the best we can easily do.

src/ariel-os-rt/src/stack.rs
```diff
@@ -1,6 +1,10 @@
 //! Stack usage helpers.
 
 #![expect(unsafe_code)]
+#![expect(
+    clippy::inline_always,
```
kaspar030, line 5: I'd rather only put this on the functions that matter

ROMemories: Do you really want this duplicated on the four methods? The lint description says:

> While there are valid uses of this annotation (and once you know when to use it, by all means `allow` this lint), it’s a common newbie-mistake to pepper one’s code with it.

To me it's better to have this lint `expect`ed once in that small, focused module, while giving a `reason` specific to the business logic, than simply `expect`ing this on the four methods, likely losing the reason in the process.

ROMemories:
I've unstacked this from #1565.

src/ariel-os-rt/src/stack.rs
kaspar030: ok

PR #1571, opened 2025-11-26
feat(net): remove default values of IPv6 device and gateway addresses

ROMemories:
# Description

Following https://github.com/ariel-os/ariel-os/pull/1377#discussion_r2565438614, this removes the default values of IPv6 device and gateway addresses that were added in #915.

This functionally reverts https://github.com/ariel-os/ariel-os/pull/915/commits/151fed5e352bbd788293e30462b673ed6833b28f.

## Testing

```sh
laze -C examples/udp-echo build -s ipv6 -b nrf52840dk
```
now errors out again, as expected.

## Change Checklist

- [x] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1578, opened 2025-11-27
feat(net): remove default values of IPv4 static configuration

chrysn:
# Description

Those values are no more suitable as static defaults than any other, and the original reason to have them (easy examples) went away when DHCP was enabled.

Docs are updated to list those values as examples rather than as defaults.

The utils macros are updated to also have a non-defaulting option for numerics. This makes sense because whether or not a default will be provided (or, otherwise, builds fail) is up to the using site and not the macro definition, *particularly* for numerics that could be used for any purpose

## Testing

This should not have an impact on any default binary configurations, as `network-config-ipv4-static` is not selected anywhere by default.

Tests I did:

* `laze -C examples/http-server/ build -b particle-xenon -s network-config-ipv4-static run` fails; first error: `error[E0080]: evaluation panicked: `CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN` environment variable was expected to provide the static IPv4 CIDR prefix length`
* `CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN=25 CONFIG_NET_IPV4_STATIC_GATEWAY_ADDRESS=10.42.0.1 CONFIG_NET_IPV4_STATIC_ADDRESS=10.42.0.61 laze -C examples/http-server/ build -b particle-xenon -s network-config-ipv4-static run` responds to `curl http://10.42.0.61/`
* Same for coap-server (with aiocoap-client and coap://)

## Issues/PRs References

Follow-up for: https://github.com/ariel-os/ariel-os/pull/1571

This came out of discussion in https://github.com/ariel-os/ariel-os/pull/1377#discussion_r2565438614. Out of the options of having IPv4 with defaults but IPv6 without, or having a ULA IPv6 default, or having no IPv4 default, this option was preferred.

Sorry #1377 authors for making the rebase harder, but with #1571 in, this sequence seems to make most sense for consistency.

## Change Checklist

### DIFF
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -34,9 +34,9 @@ DHCPv4 is used by default for network configuration, including for IP address al
 This is enabled by the `network-config-ipv4-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
 
 In order to provide a static configuration, select the `network-config-ipv4-static` [laze module](./build-system.md#laze-modules), which will take precedence.
-The configuration can be customized with the following environment variables:
+The configuration needs to be customized with the following environment variables:
 
-| Variable                                 | Default      |
+| Variable                                 | Example      |
 | --                                       | --           |
 | `CONFIG_NET_IPV4_STATIC_ADDRESS`         | `10.42.0.61` |
 | `CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN` | `24`         |
--- a/src/ariel-os-embassy/src/net.rs
+++ b/src/ariel-os-embassy/src/net.rs
@@ -178,23 +178,20 @@ fn __ariel_os_network_config() -> embassy_net::Config {
 
     cfg_if::cfg_if! {
         if #[cfg(feature = "network-config-ipv4-static")] {
-            use ariel_os_utils::{ipv4_addr_from_env_or, u8_from_env_or};
+            use ariel_os_utils::{ipv4_addr_from_env, u8_from_env};
 
-            let ipaddr = ipv4_addr_from_env_or!(
+            let ipaddr = ipv4_addr_from_env!(
                 "CONFIG_NET_IPV4_STATIC_ADDRESS",
-                "10.42.0.61",
                 "static IPv4 address",
             );
 
-            let gw_addr = ipv4_addr_from_env_or!(
+            let gw_addr = ipv4_addr_from_env!(
                 "CONFIG_NET_IPV4_STATIC_GATEWAY_ADDRESS",
-                "10.42.0.1",
                 "static IPv4 gateway address",
             );
 
-            const PREFIX_LEN: u8 = u8_from_env_or!(
+            const PREFIX_LEN: u8 = u8_from_env!(
                 "CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN",
-                24,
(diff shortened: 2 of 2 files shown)


src/ariel-os-embassy/src/net.rs
```diff
@@ -178,23 +178,20 @@ fn __ariel_os_network_config() -> embassy_net::Config {
...
+            let gw_addr = ipv4_addr_from_env!(
                 "CONFIG_NET_IPV4_STATIC_GATEWAY_ADDRESS",
-                "10.42.0.1",
                 "static IPv4 gateway address",
             );
 
-            const PREFIX_LEN: u8 = u8_from_env_or!(
+            const PREFIX_LEN: u8 = u8_from_env!(
                 "CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN",
-                24,
```
ROMemories, line 197: I'm ok with removing the default addresses, but I think we should keep this default for the subnet size. I think this is a reasonable default for many use cases, and if users need a different size, they are aware of it and the docs for that are right in the same table that they had to read to configure the addresses anyway.

book/src/networking.md
```diff
@@ -34,9 +34,9 @@ DHCPv4 is used by default for network configuration, including for IP address al
 This is enabled by the `network-config-ipv4-dhcp` [laze module](./build-system.md#laze-modules), selected by default.
 
 In order to provide a static configuration, select the `network-config-ipv4-static` [laze module](./build-system.md#laze-modules), which will take precedence.
-The configuration can be customized with the following environment variables:
+The configuration needs to be customized with the following environment variables:
 
-| Variable                                 | Default      |
+| Variable                                 | Example      |
 | --                                       | --           |
 | `CONFIG_NET_IPV4_STATIC_ADDRESS`         | `10.42.0.61` |
```
ROMemories, line 41: Grepping for `10.42.0` finds many other occurrences in the codebase, that we need to find a solution for if we want to go with this.

src/ariel-os-embassy/src/net.rs
chrysn: That's not something that is in any way default in IPv4.

If we want user convenience here, we should parse CIDR notation.

book/src/networking.md
chrysn: I did that grep and checked them all.

They are all in tests and examples, and that code has all long been switched to using DHCP, where it's at best a 1-in-254 chance that that actual address turns out being used. All those values only apply if a) auto-configuration spits them out by chance, or b) if the user already makes a config change (enabling static IPv4), and since nothing says that the user needs to enable static-IPv4 to get those exact values, that nothing doesn't need to be amended to say they also need to put in the example values.

src/ariel-os-embassy/src/net.rs
bergzand: I'm fine with either option here. Parsing CIDR notation would be great to have (good first issue etc.), but pretty low on my priority list now.

kaspar030: > That's not something that is in any way default in IPv4.

That might be true, but if we can find the most common subnet size users will encounter and save them from adding `CONFIG_NET_IPV4_STATIC_CIDR_PREFIX_LEN` to their command lines, IMO we should.

PR #1431, opened 2025-10-08
feat(builtin-sensors): introduce an LIS2DU12 driver

ROMemories:
# Description

This introduces a manual implementation of driver for the LIS2DU12 low-power 3-axis accelerometer.
Configuration is currently non-existent and some specific points need to be investigated further.

## Tested

Can be tested on an STEVAL-MKBOXPRO using the following:

```sh
laze -C examples/sensors-debug/ build -b st-steval-mkboxpro run
```

## Issues/PRs references

Depends on #1412.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1581, opened 2025-11-28
refactor(sensors): improve the `From<[...]>` docs of `ReadingChannels`

ROMemories:
# Description

This improves the docs of the `From` implementations required to instantiate `ReadingChannels`, as suggested by @chrysn in https://github.com/ariel-os/ariel-os/pull/1325#issuecomment-3306467576.

~~I hope to do something similar for `Samples`, but the API is a bit different and this is left for another PR.~~ It's done in #1582.

## How to review this PR

## Testing

~~This needs some more testing by using it for sensor drivers, so keeping as draft for now.~~ #1431 has landed, so this is now easy to review.

The size of the `text` and the `bss` between main and this PR should be compared (they must be unaffected):

```sh
laze -C examples/sensors-debug/ build -b st-steval-mkboxpro size
```

and that same example can be tested as still working in hardware.

## Issues/PRs References

- Depends on #1431 (because that PR makes it easier to review)

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -68,6 +68,7 @@ jobs:
               -p ariel-os \
               -p coapcore \
               --features "
+                  ariel-os-sensors/max-sample-min-count-12,
                   bench,
                   coap,
                   core-affinity,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,7 @@ jobs:
             --verbose
             --locked
             --features "
+                ariel-os-sensors/max-sample-min-count-12,
                 ble,
                 coap,
                 coap-transport-udp,
@@ -301,6 +302,7 @@ jobs:
                 -p ariel-os \
                 -p coapcore \
                 --features "
+                    ariel-os-sensors/max-sample-min-count-12,
                     bench,
                     ble,
                     coap,


src/ariel-os-macros/Cargo.toml
```diff
@@ -31,14 +31,14 @@ trybuild = "=1.0.111"
...
-max-sample-min-count-8 = []
-max-sample-min-count-9 = []
-max-sample-min-count-12 = []
+max-sample-min-count-3 = ["max-sample-min-count-2"]
+max-sample-min-count-4 = ["max-sample-min-count-3"]
+max-sample-min-count-5 = ["max-sample-min-count-4"]
+max-sample-min-count-6 = ["max-sample-min-count-5"]
+max-sample-min-count-7 = ["max-sample-min-count-6"]
+max-sample-min-count-8 = ["max-sample-min-count-7"]
+max-sample-min-count-9 = ["max-sample-min-count-8"]
```
ROMemories: We may want to add `max-sample-min-count-10` and `max-sample-min-count-11` later, but this is out of scope of this PR.

ROMemories: Done in #1606.

src/ariel-os-macros/src/define_count_adjusted_sensor_enums.rs
```diff
@@ -46,9 +46,12 @@ pub fn define_count_adjusted_sensor_enums(_item: TokenStream) -> TokenStream {
         }
     });
 
-    let reading_channels_from_impls = (1..=count).map(|i| {
+    // Starting at 2 as the first one is not feature-gated and manually written.
+    let reading_channels_from_impls = (2..=count).map(|i| {
```
chrysn, line 50: Could we do that always up to the maximum? This would not only save us the other commit, it would ensure that whenever a user has (say) 3 enabled and builds their local docs for their local build, they will see that more are possible if more features are active. Rather than having `get_allocation_size`, this would be using a constant. (That constant would need to be kept in sync with the features that are defined in the Cargo.toml, but that's just easier than keeping the list in `get_allocation_sizes` in sync).

ROMemories: > This would not only save us the other commit, it would ensure that whenever a user has (say) 3 enabled and builds their local docs for their local build, they will see that more are possible if more features are active.

I don't think so unfortunately: `doc_cfg` still requires the Cargo feature to be enabled for the items to appear in the docs, its only effect is to add a flair for items which are `cfg`-gated.

chrysn: That's sad, but the code generated by the macro won't just be seen by rustdoc. I can't test it quickly for this case, but I know that generally, attempting to access an implementation that is cfg'd out nowadays gives an error that says "exists but is not cfg'd", and those might profit from having them generated in a way that tools see through it as well.

ROMemories: > attempting to access an implementation that is cfg'd out nowadays gives an error that says "exists but is not cfg'd"

Indeed, I'll see what I can do.

ROMemories: I tried hard-coding `count` to `12`, and changing `from_3` to `from_6` in the LIS2DU12 driver:

https://github.com/ariel-os/ariel-os/blob/45d7fc946ddb2ae646834ca973968d780d95975b/src/sensors/ariel-os-sensor-lis2du12/src/i2c.rs#L183

but I can't get anything better than:

```sh
error[E0599]: no function or associated item named `from_6` found for struct `Samples` in the current scope
   --> src/sensors/ariel-os-sensor-lis2du12/src/i2c.rs:183:32
    |
183 |         let samples = Samples::from_6(
    |                                ^^^^^^ function or associated item not found in `Samples`
    |
note: if you're trying to build a new `Samples` consider using one of the following associated functions:
      Samples::from_1
      Samples::from_2
      Samples::from_3
   --> [redacted]/ariel-os/src/ariel-os-sensors/src/sensor.rs:15:1
    |
 15 | ariel_os_macros::define_count_adjusted_sensor_enums!();
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the macro `ariel_os_macros::define_count_adjusted_sensor_enums` (in Nightly builds, run with -Z macro-backtrace for more info)
help: there is an associated function `from_1` with a similar name
    |
183 -         let samples = Samples::from_6(
183 +         let samples = Samples::from_1(
    |
```

so I'd rather keep it as-is for now, to not have to keep that constant in sync. I do however plan to completely drop this codegen and write this manually in `ariel-os-sensors` to avoid the dependency on `ariel-os-macros`, the proc-macro, and the potential MSRV issues between `ariel-os-sensors` and `ariel-os-macros`, so we'll have the opportunity to revisit that when doing so.

src/ariel-os-sensors/Cargo.toml
```diff
@@ -25,16 +25,46 @@ defmt = ["dep:defmt"]
...
+  "max-sample-min-count-9",
+]
+max-sample-min-count-11 = [
+  "ariel-os-macros/max-sample-min-count-11",
+  "max-sample-min-count-10",
+]
+max-sample-min-count-12 = [
+  "ariel-os-macros/max-sample-min-count-12",
+  "max-sample-min-count-11",
+]
```
nponsard, line 67: Took me some time to understand what was happening here.
Maybe a comment indicating that the highest feature enables the previous ones could help.

ROMemories:
> A bit unfortunate that the features 10-12 are appearing first in the docs but I don't think we can do anything about that.

Agreed, I suppose upstream would consider sorting these with version sort instead of plain lexicographical order, since [rustfmt did switch to version sort in edition 2024](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/rustfmt-version-sorting.html).
Edit: this doesn't make sense, the `impl` blocks would need to be sorted by `T` in `impl From<T> for …`

src/ariel-os-sensors/Cargo.toml
ROMemories: Would it be ok to move this comment into #1325 (which is `ci-build:skip`)? CI is already green on this `ci-build:stm32` PR.

ROMemories: Also these should become simpler when we remove the codegen and the dependency on `ariel-os-macros` (as mentioned in https://github.com/ariel-os/ariel-os/pull/1581#discussion_r2630253852).

ROMemories:
Set to draft again after realizing that the `text` size does increase significantly when testing for `st-steval-mkboxpro`, which contains a 3-channel sensor (unlike the `stm32u083c-dk` which only contains a 1-channel sensor).

Edit: goes from 36488 to 36356 bytes.
Edit: actually everything is fine, it's the other way around: the size on *main* is now 36356 bytes, because #1637 has landed in the meantime.

PR #1588, opened 2025-12-02
feat(sensors): add custom reading signal

nponsard:
# Description

This is a custom Signal type to provide better API stability.

## Open Questions

Currently if `receive_reading` is called multiple times without `send_reading` being called in between, multiple `Futures` will be handed out but only one will per `send_reading` call. This could lead to code waiting forever for a signal, should an error be returned by `receive_reading` when a `Future` has already been handed out and not resolved ?

## Change Checklist

### DIFF
--- a/src/ariel-os-sensors/src/signal.rs
+++ b/src/ariel-os-sensors/src/signal.rs
@@ -1,0 +1,183 @@
+//! This module contains a custom [`Signal`] struct meant to be used in the [`ariel-os-sensors`][ariel-os-sensors] ecosystem
+
+use core::{
+    cell::Cell,
+    pin::Pin,
+    task::{Context, Poll, Waker},
+};
+
+use embassy_sync::blocking_mutex::{Mutex, raw::CriticalSectionRawMutex};
+
+#[derive(Debug, Default)]
+enum SignalState<T> {
+    #[default]
+    None,
+    Waiting(Waker),
+    Ready(T),
+}
+
+/// Custom signal struct inspired by [`embassy_sync::signal::Signal`] and [`embassy_sync::channel::Channel`]
+///
+/// This is meant for single-producer and single-consumer signaling.
+///
+/// This struct has been created for the [`ariel-os-sensors`][ariel-os-sensors] ecosystem.
+///
+/// [ariel-os-sensors]: crate
+// This struct exists for multiple reasons:
+// - Get a lightweight [`Future`] that can be easily stored in a struct like [`ReadingWaiter`][crate::sensor::ReadingWaiter]
+// - Keep a stable API that doesn't change with [`embassy_sync`] versions
+pub struct Signal<T> {
+    inner: Mutex<CriticalSectionRawMutex, Cell<SignalState<T>>>,
+}
+
+impl<T> Default for Signal<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T> Signal<T> {
+    /// Create a new empty [`Signal`]
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
+            inner: Mutex::new(Cell::new(SignalState::<T>::None)),
+        }
+    }
+
+    /// Signal that a new value is available and will replace the previous value if it wasn't read.
+    pub fn signal(&self, new: T) {
+        self.inner.lock(|cell| {
+            let state = cell.take();
+            match state {
+                SignalState::None => {
+                    cell.set(SignalState::Ready(new));
+                }
+                SignalState::Ready(_prev) => {
+                    cell.set(SignalState::Ready(new));
+                }
+                SignalState::Waiting(read_waker) => {
+                    cell.set(SignalState::Ready(new));
+                    read_waker.wake();
+                }
+            }
+        });
+    }
+
+    /// Returns a future that will return once a value is available.
+    ///
+    /// This is not meant to have multiple tasks waiting for a signal. If multiple tasks are waiting
+    /// then a signal sent with [`Self::signal`] will reach only one task at random.
+    pub fn wait(&'static self) -> ReceiveFuture<'static, T> {
+        ReceiveFuture { signaling: self }
+    }
+
+    fn poll_wait(&self, cx: &mut Context<'_>) -> Poll<T> {
+        self.inner.lock(|cell| {
+            let state = cell.take();
+            match state {
+                SignalState::None => {
+                    cell.set(SignalState::Waiting(cx.waker().clone()));
+                    Poll::Pending
+                }
+
+                // Multiple tasks waiting for a reading, this shouldn't happen
+                SignalState::Waiting(prev_waker) => {
+                    if prev_waker.will_wake(cx.waker()) {
+                        cell.set(SignalState::Waiting(prev_waker));
+                    } else {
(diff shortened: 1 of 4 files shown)


src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub const fn new() -> Self {
+        Self {
+            inner: Mutex::new(Cell::new(SignalState::<T>::None)),
+        }
+    }
+
+    /// Signal that a new value is available and will replace the previous value if it wasn't read.
+    ///
+    /// Returns [`Some`] if an unread value was already there, returns [`None`] otherwise.
+    pub fn signal(&self, new: T) -> Option<T> {
```
ROMemories: Do we really need to return an `Option<T>`? `embassy_sync::Signal` doesn't do that and I don't see how it can be useful to sensor drivers.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Custom signal struct inspired by [`embassy_sync::signal::Signal`] and [`embassy_sync::channel::Channel`]
+///
+/// This is meant for single-producer and single-consumer signaling.
+///
+/// This struct has been created for and is to be used only in the [`ariel-os-sensors`][ariel-os-sensors] ecosystem.
+///
+/// This struct exists for multiple reasons:
+/// - Get a lightweight [`Future`] that can be easily stored in a struct like [`ReadingWaiter`][ReadingWaiter]
+/// - Keep a stable API that doesn't change with [`embassy_sync`] versions
```
ROMemories: That's definitely a good thing to mention the reasons for creating this, but I don't think this has to go in the docs, could be just a regular comment.

src/ariel-os-sensors/src/signal.rs
nponsard: For our use case it could be useful to know if we replaced an unread value (since there should be one read per trigger), the content wouldn't matter though.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,123 @@
...
+    None,
+    Waiting(Waker),
+    Ready(T),
+}
+
+/// Custom signal struct inspired by [`embassy_sync::signal::Signal`] and [`embassy_sync::channel::Channel`]
+///
+/// This is meant for single-producer and single-consumer signaling.
+///
+/// This struct has been created for and is to be used only in the [`ariel-os-sensors`][ariel-os-sensors] ecosystem.
```
ROMemories: This seems unnecessarily scary, the primitive is actually quite generic. Is this to allow for future evolution? I don't think much could be done that wouldn't involve changing the interface, which we couldn't really do in this crate anyway.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,123 @@
...
+    signaling: &'ch Signal<T>,
+}
+
+impl<T> Future for SignalReceiveFuture<'_, T> {
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<T> {
+        self.signaling.poll_wait(cx)
+    }
+}
```
ROMemories, line 116: This also needs some unit tests. `embassy_futures::block_on()` and `join::join()` might come in handy.

src/ariel-os-sensors/src/signal.rs
ROMemories: I'm not convinced there is really a use case for it, but ok.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,123 @@
...
+                }
+                SignalState::Ready(res) => Poll::Ready(res),
+            }
+        })
+    }
+}
+
+/// A future that will resolve once a signal is sent.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct SignalReceiveFuture<'ch, T> {
```
ROMemories: Missed that one in the first pass: I think this should be named `ReceiveFuture`, similarly to [`embassy_sync::channel::ReceiveFuture`](https://docs.rs/embassy-sync/latest/embassy_sync/channel/struct.ReceiveFuture.html) since it's already in the `signal` module.

src/ariel-os-sensors/src/signal.rs
nponsard: I felt like it needed a reason to be in `ariel-os-sensors` instead of another crate like `ariel-os-utils` or `ariel-os-embassy-common`

ROMemories: Technically *where* the type comes from is an implementation detail; here one of the main reason to roll our own is API stability.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        assert!(
+            executor
+                .spawn(async move {
+                    signal.signal(wanted);
+                })
+                .is_ok()
+        );
+
+        assert_eq!(future.await, wanted);
```
ROMemories: If this is equivalent enough, this would avoid the extra dependencies:

```suggestion
        embassy_futures::block_on(async {
            embassy_futures::join::join(
                async {
                    signal.signal(wanted);
                },
                async {
                    assert_eq!(future.await, wanted);
                },
            )
            .await;
        });
```
Sure we don't get the syntactic sugar but that doesn't seem worth it.

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,172 @@
...
+impl<T> Future for ReceiveFuture<'_, T> {
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<T> {
+        self.signaling.poll_wait(cx)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ROMemories, line 119: Could one of the test try to send a second, different value, just to test that the `Signal` can be used multiple times?

src/ariel-os-sensors/src/signal.rs
```diff
@@ -0,0 +1,172 @@
...
+    use super::*;
+    use futures_executor::ThreadPool;
+    use futures_util::task::SpawnExt;
+    use static_cell::StaticCell;
+
+    #[futures_test::test]
+    async fn future_returns() {
+        let executor = ThreadPool::new().unwrap();
+
+        static SIGNAL: StaticCell<Signal<u8>> = StaticCell::new();
```
ROMemories, line 126: The position of that `static` item (not being placed at the top of the scope) made me realize we weren't passing `--tests` to Clippy. This should eventually be addressed in #1590.

ROMemories:
Codeberg, where the binaries of `cargo-manifest-check` are hosted, is [currently down](https://status.codeberg.org/status/codeberg), that's why the CI check is failing.

Edit: it is now back up

src/ariel-os-sensors/src/signal.rs
ROMemories: Manually checked Clippy would not raise concerns with the following:

```sh
cargo +nightly clippy -p ariel-os-sensors --tests
```

PR #1591, opened 2025-12-03
feat(sensors): add `clear()` function to `Signal`

nponsard:
# Description

This adds a `clear()` function to `ariel-os-sensors`' `Signal`, allowing to clear any pending value.

## Issues/PRs References

- Depends on #1588 

## Change Checklist

### DIFF
--- a/src/ariel-os-sensors/src/signal.rs
+++ b/src/ariel-os-sensors/src/signal.rs
@@ -99,6 +99,18 @@ impl<T> Signal<T> {
             }
         })
     }
+
+    /// Removes any pending value from the signal
+    pub fn clear(&self) {
+        self.inner.lock(|cell| {
+            let state = cell.take();
+            match state {
+                // Do nothing, `cell.take()` already set the state to `SignalState::None`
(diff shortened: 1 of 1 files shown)


src/ariel-os-sensors/src/signal.rs
```diff
@@ -45,6 +45,18 @@ impl<T> Signal<T> {
         }
     }
 
+    /// Removes any pending value from the signal
+    pub fn clear(&self) {
```
ROMemories: `embassy::signal::Signal` calls that `reset()`; honestly I like `clear()`, sounds less scary for something that is normal to do and more consistent with `Vec::clear()`.

ROMemories: Can you move this method below `signal()` or maybe even below `wait()`? The order is reflected in the docs and I think it's best to follow the expected lifecycle (`signal`ed and then either `clear` or a`wait`ed).

ROMemories: This needs some unit tests.

PR #1592, opened 2025-12-03
chore(coap): update dependencies

chrysn:
# Description

serde_cbor tripped things up in https://github.com/ariel-os/ariel-os/pull/1590 -- sure could fix this, but actually, that old crate just has to go. Out. Now.

Our local upgrading was painless (the technically breaking coap-handler-implementations 0.5 to 0.6 did not affect us), but it did affect our upstream riot-module-examples/riot-coap-handler-demos which backs our Blinky. (Mainly because it can do many more things than just make LEDs bink).

## Testing

The most affected test to run is tests/coap-blinky. I did this on a particle-xenon, with the CLI invocation from its README.

## Issues/PRs References

- Blocked by #1593

## Open Questions

* [x] Of the 12 updated dependencies, how many will I actually review?
* [x] How many allow-listed (grandfathered) libraries can I remove from that list?

## Change Checklist

- [x] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- n/a I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1590, opened 2025-12-03
ci(clippy): run Clippy on tests too

ROMemories:
# Description

This makes Clippy check tests too.

## Issues/PRs References

- Depends on #1592

## Change Checklist

- [x] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1593, opened 2025-12-03
chore: update stable Rust to 1.90

chrysn:
# Description

Our fixed Rust version is getting old, to the point where dependencies don't even work with that any more if we update them.

## Testing

We're building on nightly all the time, I don't see what can go wrong here.

## Issues/PRs References

Blocks: https://github.com/ariel-os/ariel-os/pull/1592 and thus transitively https://github.com/ariel-os/ariel-os/pull/1590

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -197,7 +197,7 @@ jobs:
       - name: Install stable toolchain
         if: steps.should-skip.outputs.result != 'true'
         # configure MSRV here:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```suggestion
          version: "1.90.0.0"
```

chrysn: Replaced 1.85.0.0 with 1.90.0.0. Not sure why GH didn't allow me to just apply it (or why my grep didn't catch it in the first place).

PR #1204, opened 2025-08-18
ci: free up more space

ROMemories:
# Description

In the spirit of #947, and in an attempt at fixing the nightly/full build that is failing because of lack of space, this deletes more unused software that is provided in the Actions' image.
Software already installed in the image is [documented here](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md).

This is a stopgap solution, I think what we should do is prune the build directory as we go instead.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1594, opened 2025-12-03
ci: free up even more space

ROMemories:
# Description

This attempts to free even more space on the GH image to avoid the nightly build failure. In a way this is a follow-up to #1204.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,6 +110,7 @@ jobs:
             /opt/ghc \
             /opt/hostedtoolcache/CodeQL \
             /usr/lib/jvm \
+            /usr/local/.ghcup \
             /usr/local/julia* \
             /usr/local/lib/android \
             /usr/local/share/chromium \


.github/workflows/build.yml
```diff
@@ -110,6 +110,7 @@ jobs:
             /opt/ghc \
             /opt/hostedtoolcache/CodeQL \
             /usr/lib/jvm \
+            /usr/local/.ghcup \
```
ROMemories, line 113: This path comes from the [same post](https://dev.to/mathio/squeezing-disk-space-from-github-actions-runners-an-engineers-guide-3pjg) #1204 was based on. I have no idea why I didn't also put that one here the first time.

Edit: according to the linked post, this is supposed to clear up 3 GB.

ROMemories:
Going with `ci-build:rp` for now, just to get the Build workflow running (and thus check that the path at least *exists* before attempting to delete it).

PR #1596, opened 2025-12-03
refactor(coap): prepare multiple transports

chrysn:
# Description

Just a small refactor base on which I can explore multiple back-ends with less divergence from main.

This is ground work for

* CoAP-over-GATT (BLE)
* Slipmux
* CoAP-over-TCP

## Testing

Build tests suffice, but I did run this with the coap-blinky test on particle-xenon.

## Change Checklist


ROMemories:
To me that's a bit too big for a self-approval; I wouldn't mind too much if it were only moving code around, but this also adds a bit of documentation in an `ariel-os*` crate. In this case that means I'll have to update #1544.

chrysn:
Ack; updating my calibration. Sorry for trouble.

Note that the added documentation is on a private item, and was really a late fix-up (because our lints are set so strict that they even require private functions to talk about their panic behavior, which in this case makes little sense because the whole module is run-once code).

PR #1600, opened 2025-12-04
feat(uart): forward BufRead implementation

chrysn:
# Description

The embedded-io-async Read trait is not zero-copy, but our UARTs are buffered. Applications that consume data on the fly (such as upcoming slipmux integration) still need to allocate their receive buffer, even if it's just 1 byte long, and keep reading into that.

The embedded-io-async family offers a better trait, BufRead, that suspends until the producer has collected any number of bytes in its internal buffer (it returns when there is one, but with fast interfaces or under load it can be more), offers the user a view into the buffer, with explicit consumption of the buffer. (In other contexts this might be called peeking into the buffer, just for searchability).

As our UART setup involves buffers in all current implementations (not by a trait, and that's something we should fix, but by the drop-in-replacement-safe `::new()` method of the UART types), it's safe to say that not just the current but also future types have some level of buffering. (And it doesn't matter if it's just 4 byte: Even then, the API is no worse than reading into a local buffer, but it automatically enhances on larger-buffered devices).

## Testing

I've done some testing with nRF on a to-be-created PR around slipmux, but this is best checked by the compiler and code review, because in the end it will exclusively depend on any BufRead implementations of the backing HALs to be correct.

## Change Checklist

  * This would be documented if we had traits for the things we drop-in-replace; I'd like to change that at some point to make things like this actually documented, but either way, it is visible automatically in traits implemented by a type.

### DIFF
--- a/src/ariel-os-embassy-common/src/uart.rs
+++ b/src/ariel-os-embassy-common/src/uart.rs
@@ -117,6 +117,21 @@ macro_rules! impl_async_uart_for_driver_enum {
         }
 
 
+        impl embedded_io_async::BufRead for $driver_enum<'_> {
+            async fn fill_buf(&mut self) -> Result<&[u8], Self::Error> {
+                match self {
+                    $( Self::$peripheral(uart) => embedded_io_async::BufRead::fill_buf(&mut uart.uart).await, )*
+                }
+            }
+
+            fn consume(&mut self, amt: usize) {
+                match self {
+                    $( Self::$peripheral(uart) => embedded_io_async::BufRead::consume(&mut uart.uart, amt), )*
+                }
+            }
+        }
+
+
         impl embedded_io_async::ReadReady for $driver_enum<'_> {
             fn read_ready(&mut self) -> Result<bool, Self::Error> {
                 match self {


bergzand:
The issue with providing `BufRead` for our uart implementation is that the esp-hal doesn't allow for this. ESP uart peripherals have hardware FIFOs for the uart that provide the guarantees required for the buffered UART traits, but the downside is that (at least by esp-hal) they are unable to provide a `BufRead` implementation.

There's already a macro (`impl_async_uart_bufread_for_driver_enum`) to add `BufRead` implementations, that looks identical to what you provide here. It's not yet used anywhere though. We can do that for the chips that support it, but that means having diverging support between ESP and the other vendors.

chrysn:
@bergzand pointed me at an unused existing implementation, and trouble due to the ESP UART actually ignoring the buffer we provide and running on its own internal FIFO.

Putting this on hold as a PR, but leaving it open to track that actually we *should* have ReadBuf provided.

chrysn:
Notes from talking this through with danielb [on chat](https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$EbhNoeGRNJZRlrfnLDPPPKDT249IiTA_l5pjK2KJdec?via=matrix.org&via=fehler-in-der-matrix.de&via=utwente.io): ESP hardware has a non-memory-mapped internal buffer. (Looking at ESP-IDF is a red herring: They do have [code that memcpy's on read](https://github.com/espressif/esp-idf/blob/683ddf8a6eeaf9189679273ba37b3fc82f515df0/components/esp_driver_uart/src/uart.c#L1723C5-L1723C20), but they have an extra layer of buffering they read from).

Sensible path forward might be to largely disregard the ESP's internal buffer, spool in user provided memory as we do for all platforms (ESP users are generally not so memory constrained), and provide a consistent BufRead'able view from there. Documentation can accompany this that nudges people towards using BufRead when they can.

(Exclusive Read'ing users, eg. those who need to hold on what they've read for a long time, may then still get the option to detect that there is a hidden buffer, set the buffers up such that there is no own buffer and BufRead panics because the buffer is 0 long -- but that might easily be a premature optimization.)

PR #1601, opened 2025-12-04
feat(macros): enable using `config!` in internal crates

chrysn:
# Description

This enables using the `config!` macro in crates that can not include ariel-os due to circular dependencies. If any internal module that makes large changes to the USB config needs to set override-usb-config, it now can.

Sure, on the long run I'd like to use descriptive controls to build up the USB config. But until that is in place, this generalizes the existing mechanism by a simple extension.

## Testing

I tried this in an upcoming PR for updating the USB configuration (yet to be linked back) -- but this should be an "if it builds, it is fine" situation.

## Issues/PRs References

This follows a pattern established in <https://github.com/ariel-os/ariel-os/pull/759>.

## Change Checklist

### DIFF
--- a/src/ariel-os-macros/src/config.rs
+++ b/src/ariel-os-macros/src/config.rs
@@ -39,8 +39,8 @@
 ///
 /// # Panics
 ///
-/// This macro panics when the `ariel-os` crate cannot be found as a dependency of the crate where
-/// this macro is used.
+/// The `ariel-os` crate (or `ariel-os-embassy`, for use in crates which `ariel-os` depends on)
+/// must be a dependency of the crate where this macro is used, otherwise this macro panics.
 #[proc_macro_attribute]
 pub fn config(args: TokenStream, item: TokenStream) -> TokenStream {
     #[allow(clippy::wildcard_imports)]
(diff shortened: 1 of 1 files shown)


src/ariel-os-macros/src/config.rs
```diff
@@ -39,8 +39,8 @@
 ///
 /// # Panics
 ///
-/// This macro panics when the `ariel-os` crate cannot be found as a dependency of the crate where
-/// this macro is used.
+/// This macro panics when the `ariel-os` crate (or `ariel-os-embassy`, for use in crates which
+/// `ariel-os` depends on) cannot be found as a dependency of the crate where this macro is used.
```
ROMemories: Usually I would rather not surface these internal details in the docs, but because this could lead to a surprising *non* panic I guess it's fine in this case.

I'm not sure about the logical articulation though: this reads to me as if it would panic when *either* of them cannot be found.

chrysn: > logical articulation

It searches for one crate or the other. But the precise logic there is *really* a thing people can look at the implementation.

As for documentation of internal use, see general discusson on chat, but I don't think that was a blocking aspect.

ROMemories: > It searches for one crate or the other.

Sure, but the doc comment is currently phrased it terms of panics: it says it panics when `ariel-os` *or* `ariel-os-embassy` cannot be found, while actually it panics when *neither* of them can be.

chrysn: Then please make a suggestion, or apply the minimum context sensitivity and cooperativity that a reader would have who reads the full sentence including "for use in".

ROMemories: (Needs to be reflowed properly:)
```suggestion
/// The `ariel-os` crate (or `ariel-os-embassy`, for use in crates which
/// `ariel-os` depends on) must be a dependency of the crate where this macro is used, otherwise this macro panics.
```

ROMemories (approved):
Please squash (and optionally reflow the doc comment)!

PR #1607, opened 2025-12-05
fix(esp): fix xtensa isr stack linking

kaspar030:
# Description

Previously we tried to move the xtensa ISR stack to the beginning of RAM, to get overflow protection. Turns out there was some other data mapped. Also, in #1328, this caused more linking errors.

So, just go with the (default) end-of-ram stack. This PR also does add a linker assertion checking whether there's enough space for the configured size.

Re-enables esp32, as in my testing it works again now.

## Issues/PRs References

Split out of #1328.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -392,7 +392,6 @@ contexts:
     parent: esp
     selects:
       - xtensa
-      - currently-broken # `isr_stack` is not placed at the right location
     env:
       RUSTC_TARGET: xtensa-esp32-none-elf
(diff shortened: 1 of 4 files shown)


kaspar030:
> I can't test on a plain ESP32.

I tested the http-client example one the plain esp32 devkit. Works fine with default settings, link-fails with `-s executor-thread` because somehow there's not enough memory, and the new linker check correctly fails on that. #1328 will make use of otherwise unused memory for the esp heap, there it compiles fine.

PR #1537, opened 2025-11-19
docs(book): document the `network-config-override` laze module

ROMemories:
# Description

This documents the `network-config-override` laze module in the book.

## Issues/PRs References

- Closes #740
- Depends on #1377
- Depends on #1612

## Change Checklist

- [x] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1612, opened 2025-12-08
fix(net): introduce a laze module to use the network config override

ROMemories:
# Description

This addresses the issue reported in https://github.com/ariel-os/ariel-os/pull/1537#issuecomment-3552064963, where it was actually not possible to use the `network-config-override` Cargo feature because the laze configuration would not allow it.

This introduces a `network-config-override` *laze module*, that enables the (now hidden) Cargo feature of the same name, and which provides both `network-config-ipv4` and `network-config-ipv6`.

## Testing

Enable the `network-config-override` laze module for the `udp-echo` example (without changing anything else); the linker will raise an error about a missing symbol.
Then, add [the example from `#[ariel_os::config]`](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/attr.config.html#examples) to that example (you will also need to add `heapless` as a dependency of the example):

```rust
#[ariel_os::config(network)]
const NETWORK_CONFIG: embassy_net::Config = {
    use embassy_net::Ipv4Address;

    embassy_net::Config::ipv4_static(embassy_net::StaticConfigV4 {
        address: embassy_net::Ipv4Cidr::new(Ipv4Address::new(10, 42, 0, 61), 24),
        dns_servers: heapless::Vec::new(),
        gateway: Some(Ipv4Address::new(10, 42, 0, 1)),
    })
};
```

The example should work when tested with the provided IP address. You can also try to change the IP to check it is indeed the one being used; the example prints the selected IP at startup.

## Issues/PRs References

- ~~The documentation for this will be updated in #1537.~~ Actually I've moved the docs in this PR, as it's hard to keep links consistent otherwise.

## Changelog Entry

The `network-config-override` Cargo feature has been removed from the documentation, as it was not possible to actually use it, and has been replaced by the laze module of the same name, which must be used instead.

## Change Checklist

### DIFF
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -65,6 +65,11 @@ The configuration must be customized with the following environment variables:
 > [!NOTE]
 > Non-static IPv6 address allocation will be supported in the future.
 
+#### Custom Configuration Provider
+
+Instead of using DHCP or passing static configuration through environment variables it is also possible to use a custom configuration provider if needed.
+To do that, the `network-config-override` [laze module](./build-system.md#laze-modules) needs to be enabled, and the [`#[ariel_os::config]` attribute macro][config-attr-macro-rustdoc] can be used to provide the configuration.
+
 ### Support for Network Protocols
 
 Support for various network protocols can be enabled through [Cargo features listed in the documentation][rustdoc-homepage].


src/ariel-os/Cargo.toml
```diff
@@ -125,7 +125,7 @@ usb-hid = ["ariel-os-embassy/usb-hid"]
 #! specific system functionality.
 #! The features below need to be enabled so that the provided custom
 #! configuration is taken into account.
-## Enables custom network configuration.
+# Enables custom network configuration.
```
ROMemories, line 127: Actually we can't simply hide this Cargo feature here, because it's referenced from the docs of `#[ariel_os::config]`.

src/ariel-os-macros/src/config.rs
```diff
@@ -41,6 +41,8 @@
 ///
 /// The `ariel-os` crate (or `ariel-os-embassy`, for use in crates which `ariel-os` depends on)
 /// must be a dependency of the crate where this macro is used, otherwise this macro panics.
+///
+/// [network-config-override-book]: https://ariel-os.github.io/ariel-os/dev/docs/book/networking.html#custom-configuration-provider
```
ROMemories, line 45: There is already a couple of links from the rustdoc docs to the book, we'll have to find a solution to link to the right version when we do host them. However I think this is out-of-scope of this PR, and it seems to me that we do need do need to link to the book to not reference laze modules in the rustdoc docs directly.

src/ariel-os/Cargo.toml
ROMemories: The docs of `config` are now updated accordingly.

src/ariel-os-macros/src/config.rs
```diff
@@ -9,10 +9,10 @@
 ///
 /// - The name of the driver the constant provides configuration for.
 ///
-/// | Driver    | Expected type                      | Cargo feature to enable   |
-/// | --------- | ---------------------------------- | ------------------------- |
-/// | `network` | `embassy_net::Config`              | `network-config-override` |
-/// | `usb`     | `embassy_usb::Config`              | `override-usb-config`     |
+/// | Driver    | Expected type                      | How to enable                                         |
+/// | --------- | ---------------------------------- | ----------------------------------------------------- |
+/// | `network` | `embassy_net::Config`              | See the [book section][network-config-override-book]. |
```
bergzand, line 14: Can't we just state that the `network-config-override` laze module needs to be enabled, and save users an indirection? Possibly adding an extra column to the table with a link to the book.

```suggestion
/// | `network` | `embassy_net::Config`              | Enable the `network-config-override` laze module. |
```

book/src/networking.md
```diff
@@ -65,6 +65,11 @@ The configuration must be customized with the following environment variables:
 > [!NOTE]
 > Non-static IPv6 address allocation will be supported in the future.
 
+#### Custom Configuration Provider
+
+Instead of using DHCP or passing static configuration through environment variables it is also possible to use a custom configuration provider if needed.
+To do that, the `network-config-override` [laze module](./build-system.md#laze-modules) needs to be enabled, and the [`#[ariel_os::config]` attribute macro][config-attr-macro-rustdoc] can be used to provide the configuration.
```
bergzand, line 71: When appending the `network-config-override` Laze module to the list of `selects` in the example, the `udp-echo` example doesn't build because laze has an ordering issue between the existing `network` module and the new `network-config-override` module in the laze file. The `network-config-override` module needs to precede the `network` module. Simply following the instructions here is not sufficient to get laze to compile the example. 
I've seen this more often with config-related network laze modules.

I think it's fair to add a note here, or on this page in general that config modules need to precede the generic `network` module.

src/ariel-os-macros/src/config.rs
ROMemories: The goal here was to not have a direct reference to any laze construct in the API docs: we don't have any for now, and users need the context of the book to understand that.

book/src/networking.md
ROMemories: Indeed! I suppose the habit made me totally miss that :thinking: Seems out-of-scope of this PR though, I'll open a separate PR, as indeed I can reproduce with `network-config-ipv4-static` as well.

ROMemories:
> > you will also need to add heapless as a dependency of the example
> 
> Can we add this note to the example in the docs?

I don't think this is needed, the compiler already calls out the missing dependency, and we don't really do the same in some other cases.

book/src/networking.md
kaspar030: > I think it's fair to add a note here, or on this page in general that config modules need to precede the generic `network` module.

... which we should fix more generally.

ROMemories: Opened https://github.com/ariel-os/ariel-os/pull/1796 to clarify this.

src/ariel-os-macros/src/config.rs
bergzand: > The goal here was to not have a direct reference to any laze construct in the API docs: we don't have any for now, and users need the context of the book to understand that.

I don't follow your train of thought here.
There can still be a link to the book here to provide users with the instructions *how* a laze module should be enabled. But requiring users to always follow the link, even when they are already familiar with laze and just need to know *which* laze to select to make the example work is just increasing the barrier for use.

bergzand:
> I don't think this is needed, the compiler already calls out the missing dependency

Examples should just work. A compiler error caused by an example provided by us just adds a barrier to usage. A single line below the example noting the extra dependency can save a user a bit of work trying to figure out what the issue is, even if it is *just* a missing dependency.

> and we don't really do the same in some other cases.

I'm fine with adding a note about extra dependencies in the other cases.

src/ariel-os-macros/src/config.rs
ROMemories: > I don't follow your train of thought here.

My point is that I think rustdoc docs should only refer to Rust concepts. Imagine this being published on docs.rs, I think referencing laze modules would look odd.

If you really want this, we can add the laze module in a `ci-build:skip` follow-up.

ROMemories:
> Examples should just work.

I had actually missed that you were referencing the already existing example. Given this snippet has already been there for a long time, somewhat improving this example seems orthogonal to this PR and can be done in a `ci-build:skip` follow-up.

Generally I would like to move on with this PR so it gets out of the way of #1389, and possibly address the comments in a doc-only follow-up (because these would be purely additive changes anyway).

src/ariel-os-macros/src/config.rs
bergzand: > My point is that I think rustdoc docs should only refer to Rust concepts. Imagine this being published on docs.rs, I think referencing laze modules would look odd.

Now I see what you mean here.

At this point I think trying to avoid mentioning anything Laze in the rustdoc docs makes it more awkward to use for our users. I would revisit this if Laze at some point is no longer mandatory to use with Ariel OS, but at this point any developer of Ariel OS has to deal with Laze. 

> If you really want this, we can add the laze module in a `ci-build:skip` follow-up.

Fine with me.

bergzand:
> I had actually missed that you were referencing the already existing example. Given this snippet has already been there for a long time, somewhat improving this example seems orthogonal to this PR and can be done in a `ci-build:skip` follow-up.

It was part of the testing description of this PR, hence I tested it and ran into the issue. :shrug:

> Generally I would like to move on with this PR so it gets out of the way of #1389, and possibly address the comments in a doc-only follow-up (because these would be purely additive changes anyway).

Fine with me!

PR #1613, opened 2025-12-09
feat: add slipmux CoAP support

chrysn:
# Description

slipmux is a means by which IP, direct CoAP and debug messages can be sent over a serial line (based on SLIP). It is mostly aimed at devices that can not use a more structured form of communication with the host such as USB Ethernet, or where that structured form is inaccessible to the user (eg. because they are not authorized on their PC to set up Ethernet connections).

For example, it enables devices with hard-wired serial adapters such as older ESP32, the microbit-v2 or nRF9151-DK, to serve as network border routers (once we have routing functionality), and allows CoAP traffic into them even without any network setup whatsoever.

The present implementation discards any actual network messages and only deals in CoAP; this can be extended later. It runs either over actual UART or over USB serial; <del>the latter is vastly incomplete</del>. (It is of somewhat questionable utility, but there are cases when it does make sense, eg. when using a WebSerial connection into a webapp, which then enables end-to-end ACE/EDHOC/OSCORE security from the Internet into that device).

The main contribution of this PR is initially not the slipmux integration itself, but setting sbd up in such a way that a selectable module can require there to be a suitable UART in the first place; this is acting in tandem with https://github.com/ariel-os/sbd/pull/68. Consequentially, this avoids the need for any `pin.rs` module with a large ifdef sequence, and runs from descriptive data instead.

## Testing

Send CoAP requests via slipmux, eg. using Jelly.

Current invocations are:
* Real UART:
  * `laze build -b bbc-microbit-v2 -C examples/coap-server -s coap-transport-slipmux run`
  * `laze build -b nrf9151-dk -C examples/coap-server -s coap-transport-slipmux -s coap-server-config-unprotected run` (disabling CoAP security so that it can run on a board without an RNG)
* USB:
  * `laze build -b nrf9151-dk -particle-xenon -C tests/coap -D LOG=debug -s coap-transport-slipmux -s coap-server-config-unprotected run`

In all cases, it's best used with the [Jelly crate](https://github.com/Teufelchen1/jelly) provided by Teufelchen, who also wrote the slipmux implementation used for this port. Launch `Jelly /dev/ttyACM0` (or whichever port it is connected to), and observe that a green (successful) response comes back to the /.well-known/core request on the right side.

[edit, added]: Note that if you send raw commands (like, entering foobar / return in the user input), that shows in the debug logs, as we don't have a shell or a virtual UART.

## Issues/PRs References

- Builds on https://github.com/ariel-os/sbd/pull/68.
- Depends on #2181

Over time it may make sense to split the slip and UART aspects.

## Change Checklist

### DIFF
--- a/src/ariel-os-coap/src/transport_slipmux.rs
+++ b/src/ariel-os-coap/src/transport_slipmux.rs
@@ -1,0 +1,292 @@
+// FIXME: This does not populate coap_client; probably it even could (but only with weird IP
+// address semantics).
+
+use ariel_os_debug::log::{Hex, debug, warn};
+use embassy_sync::once_lock::OnceLock;
+
+mod over_uart {
+    use ariel_os_embassy::cell::SameExecutorCell;
+    use static_cell::ConstStaticCell;
+
+    use super::*;
+
+    type UartAssignment = ariel_os_boards::pins::HOST_FACING_UART;
+
+    // FIXME There has to be a better idiom for a function deep down to take peripherals that does
+    // not involve spawning a task that terminates at startup?
+    //
+    // (Therefore, not bothering with optimizing any of its access)
+    static TAKE_THIS: OnceLock<SameExecutorCell<core::cell::RefCell<Option<UartAssignment>>>> =
+        OnceLock::new();
+
+    #[ariel_os_macros::task(autostart, peripherals)]
+    async fn pass_on_uart(peripherals: UartAssignment) {
+        TAKE_THIS
+            .init(SameExecutorCell::new_async(Some(peripherals).into()).await)
+            .ok()
+            .expect("Autostart is only called once");
+    }
+
+    // FIXME: Move to BufRead once <https://github.com/ariel-os/ariel-os/pull/1600> is resolved.
+    pub(super) async fn take() -> impl embedded_io_async::Read + embedded_io_async::Write {
+        let peripherals: UartAssignment = TAKE_THIS
+            .try_get()
+            .unwrap()
+            .get_async()
+            .await
+            .unwrap()
+            .take()
+            .unwrap();
+
+        static RX_BUF: ConstStaticCell<[u8; 16]> = ConstStaticCell::new([0; 16]);
+        static TX_BUF: ConstStaticCell<[u8; 16]> = ConstStaticCell::new([0; 16]);
+        let rx_buf = RX_BUF.take();
+        let tx_buf = TX_BUF.take();
+
+        let mut config = ariel_os_hal::hal::uart::Config::default();
+        config.baudrate = ariel_os_hal::uart::Baudrate::_115200;
+
+        use ariel_os_hal::uart::Assignment;
+        let (tx, rx) = peripherals.into_pins();
+
+        <UartAssignment as Assignment>::Device::new(rx, tx, rx_buf, tx_buf, config)
+            .expect("Invalid UART configuration")
+    }
+}
+
+pub(crate) async fn coap_run_slipmux(mut handler: impl coap_handler::Handler) -> ! {
+    use embedded_io_async::{Read, Write};
+    use slipmux::DecodeStatus;
+
+    // For the time being we play the simple game, where we just have a server, and can afford to
+    // read until something has arrived, then write out what has been written, and continue.
+    //
+    // This will not be sustainable with the next generation coap-handler, and then we may need to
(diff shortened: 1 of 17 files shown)


src/ariel-os-coap/src/transport_slipmux.rs
```diff
@@ -0,0 +1,337 @@
...
+    //
+    // This will not be sustainable with the next generation coap-handler, and then we may need to
+    // split the UARTs.
+
+    //let uart = usb_serial::take().await;
+    let mut uart = over_uart::take().await;
+
+    let mut decoder = slipmux::Decoder::new();
+    // Small -- just enough so we can show some bytes for diagnostics. (We can do larger if
+    // ReferencedLatestFrame learns to use a single buffer).
```
Teufelchen1: You can also always implement your own handler. No need to use one of the pre-builts if they don't suit your usecase. :) 

```Rust
pub trait FrameHandler {
    // Required methods
    fn begin_frame(&mut self, frame_type: FrameType);
    fn write_byte(&mut self, byte: u8);
    fn end_frame(&mut self, error: Option<Error>);
}
```

src/ariel-os-coap/src/transport_slipmux.rs
```diff
@@ -0,0 +1,337 @@
...
+                //BR break;
+            }
+        }
+
+        //BR }
+
+        //BR uart.consume(consume);
+
+        if coap_ready {
+            // Compensate for checksum -- FIXME: where should this best be done?
```
Teufelchen1: If you come up with a good idea, let me know! I would prefer for the `slipmux` crate to handel this but it is not as straight forward as one thinks: You can only know that you deserialezed the fcs after the frame completed, at which point the bytes have already been forwarded to the handler which, well, is in the users hands. And here we are >.<

src/ariel-os-coap/src/transport_slipmux.rs
chrysn: Trying that out in fb754dad8604c502687d71ac038d95d14c7c1303. Turns out all I need to store is a heapless vec and a byte for overflow, because while I don't get the info on what kind of frame ended in `end_frame`, I don't need that there, and I get it from where I feed the byte in too.

chrysn:
This is now ready for an *architectural* review, to check directions. (No style, comments, typos yet unless doing that is *less* effort for you as a reviewer).

Architecture has shifted a bit to use less "initialize some global periphs from an autostart task" and more "the autostart task that gets the periphs also runs the slipmux loop, which then hands out its buffer to CoAP". This is mostly because at some point, I'd like to support actual SLIP too -- at which point (!) I'd move the implementation over to a more general place.

That also made it easy to complete the USB implementation; that's still a separate commit because if there's any trouble in review, we can postpone it.

chrysn:
Hm, having the option to do slipmux (which, for reasons beyond this scope, doesn't need an RNG in its implementation, unlike the UDP version), breaks a hack I've had around for selecting the security mode in laze (which, incidentally, also came up around safety checks in https://github.com/ariel-os/ariel-os/pull/1783 / https://github.com/ariel-os/sbd/issues/110):

All applications so far selected `?coap-server-config-demokeys`, which succeeded in running them with demokeys, but if a user said `-s coap-server-config-storage`, that worked just as well. (The available server-configs are -storage, -unprotected and -demokeys, all of which provides_unique coap-server-config).

My intention with that line was to say that there is a default server-config that is not coap-server-config-unprotected (which, as in https://github.com/ariel-os/ariel-os/pull/1783, should never be selected unintentionally and always need explicit user action, either from the command line or from laze.yml). This worked so long as there was never a case where coap-server-demokeys would not work, but the rest would. Now that there are boards where coap-server can be used without HWRNG, all of a sudden I run into CI errors at build time where boards don't select any coap-server-config. (And coap-server does not depend in laze on coap-server-config because then it'd pull in the available coap-server-config-unprotected -- if I added that dependency, it would be used automatically in this case).

So in a sense, things worked as intended, but by failing at build time rather than with a laze "I need this but can't have it so I won't try", CI complains.

@kaspar030, do you have any better idea on this than adding a Laze feature by which a module can be in "never auto-select just because it provides something, only when it is named"?

chrysn:
With the upcoming rebase, this is 

Blocked-by: #1930

ROMemories:
Could you split the UART part out of this?

PR #1615, opened 2025-12-09
ci(clippy): lint `ariel-os-native`

ROMemories:
# Description

This fixes the couple of Clippy errors in `ariel-os-native` and adds the crate to the list checked by Clippy in CI.

## Testing

```sh
laze -C examples/random/ build -b native run 
```

## Issues/PRs References

- Depends on #1614

## Change Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -151,6 +151,7 @@ jobs:
             -p ariel-os-hal
             -p ariel-os-identity
             -p ariel-os-macros
+            -p ariel-os-native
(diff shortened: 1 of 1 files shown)


ROMemories:
> Should we do something about the warnings coming from `rustc` about unused imports/functions ?

That's indeed orthogonal, but yes suppressing those warnings for native would be good; they stem from the fact that native is reusing the dummy HAL that was originally only intended to satisfy HAL-agnostic tools.

PR #1618, opened 2025-12-10
refactor(coap): split out UDP transport

chrysn:
# Description

CoAP can eventually have multiple transports: over UDP (present), TCP, Bluetooth, or serial.

This PR threads the split prepared in #1596 out to a sparate module, laze and features. While this enables no new functionality (and thus the laze module and features may seem moot), it makes sense to make the split here, as it separates refactoring from later additions.

In theory, the the split might enable running CoAP without an RNG (because it's not CoAP-as-a-whole that needs one, just the UDP implementation for token randomization, and the cryptography parts) when using a more deterministic transport and no crypto, no such transport is present at the moment.

## Testing

* CoAP example still runs on any board where it did before.
* The split out of the HWRNG (where now `-s coap-server-config-unprotected` would work) can not be practically tested for whether things work alone, because so far there is only one transport implementation, and that too needs an RNG. I tested building and running in #1613 where there *is* another transport that is not RNG dependent.

## Issues/PRs References

Split out off #1613

## Open questions

The separation into commits is a bit arbitrary in that some modules could be made conditional earlier or later. If that is a concern to reviewers, I'd rather squash the 3 commits than try which move works where (which'd need extra testing compared to the constantly tested single commits produced linearly).

## Change Checklist

### DIFF
--- a/src/ariel-os-coap/src/lib.rs
+++ b/src/ariel-os-coap/src/lib.rs
@@ -19,6 +19,7 @@ mod stored;
 mod transport_udp;
 
 use ariel_os_embassy::cell::SameExecutorCell;
+#[cfg(feature = "coap-server")]
 use coap_handler_implementations::ReportingHandlerBuilder as _;
 use embassy_sync::watch::Watch;
 
@@ -124,11 +125,12 @@ async fn coap_run_impl(handler: impl coap_handler::Handler + coap_handler::Repor
         } else if #[cfg(feature = "coap-server-config-demokeys")] {
             let security_config = demo_setup::build_demo_ssc();
         } else if #[cfg(feature = "coap-server-config-unprotected")] {
-            let security_config = coapcore::seccfg::AllowAll;
+            // Not setting the config to `coapcore::seccfg::AllowAll` because it won't be taken up
+            // by a security context anyway (handler is not wrapped in an OscoreEdhocHandler).
         } else {
-            // We could pick another policy too to get 4.04 errors, but "there may be something but
-            // I won't tell you" is just as good an answer, and may prune some more branches even.
-            let security_config = coapcore::seccfg::DenyAll;
+            // Not setting the config to `coapcore::seccfg::DenyAll` because it won't be taken up
+            // by a security context anyway -- and as there is no sever configured, there is no
+            // resource to which the policy would be applied.
 
             #[cfg(all(feature = "coap-server", not(feature = "doc")))]
             compile_error!("No CoAP server configuration chosen out of the coap-server-config-* features.");
@@ -137,8 +139,12 @@ async fn coap_run_impl(handler: impl coap_handler::Handler + coap_handler::Repor
 
     // FIXME: Should we allow users to override that? After all, this is just convenience and may
     // be limiting in special applications.
+    #[cfg(feature = "coap-server")]
     let handler = handler.with_wkc();
-    #[cfg(not(feature = "coap-server-config-unprotected"))]
+    #[cfg(any(
+        feature = "coap-server-config-storage",
+        feature = "coap-server-config-demokeys"
+    ))]
     let handler = coapcore::OscoreEdhocHandler::new(
         handler,
         security_config,


src/ariel-os-coap/Cargo.toml
```diff
@@ -65,9 +65,11 @@ serde_yaml = "0.9.34"
 # laze's name for this (where coap-server makes more sense).
 coap-server = []
 
-coap-server-config-storage = ["dep:ariel-os-storage"]
+coap-server-config-storage = ["dep:ariel-os-storage", "dep:ariel-os-random"]
```
ROMemories:
```suggestion
coap-server-config-storage = ["dep:ariel-os-random", "dep:ariel-os-storage"]
```

src/ariel-os-coap/Cargo.toml
```diff
@@ -65,9 +65,11 @@ serde_yaml = "0.9.34"
...
 coap-server-config-unprotected = []
-coap-server-config-demokeys = []
+coap-server-config-demokeys = ["dep:ariel-os-random"]
+
+coap-transport-udp = ["dep:embassy-net", "ariel-os-embassy/net", "dep:embedded-nal-async", "dep:ariel-os-random"]
```
ROMemories:
```suggestion
coap-transport-udp = ["dep:ariel-os-random", "dep:embassy-net", "dep:embedded-nal-async", "ariel-os-embassy/net"]
```

ROMemories (commented):
That's what [cargo-manifest-check](https://codeberg.org/AudaciousAxiom/cargo-manifest-check) is expecting (haven't checked my suggestions locally however): implied features should be sorted lexicographically, with `dep:` features grouped first.

chrysn:
Thanks, all applied. GitHub appears to have operational difficulties or suffer from its own anti-download protection (curious thing for a hosting platform), so build tests can't currently run, but the lints finally passed.

It's all just a setting-up-features and moving-code-into-a-module change, so I think the biggest review question is if you're happy with the 3 separate steps (where it's hard to check whether any of the dependencies-becoming-optional could maybe have been done a commit earlier too, IMO that doesn't matter) or prefer to have it squashed.

laze-project.yml
```diff
@@ -1457,6 +1456,18 @@ modules:
         executor_stacksize_required:
           - "32768"
 
+  - name: coap-transport-udp
+    help: The transport of CoAP that uses the CoAP-over-UDP transport on the network stack.
+    # On the long run, this should not be unique, but currently it is.
```
```suggestion
    # In the long run, this should not be unique, but currently it is.
```

src/ariel-os-coap/src/transport_udp.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    server
+        .run(
+            &mut unconnected,
+            &mut handler,
+            &mut ariel_os_random::fast_rng(),
+        )
+        .await
+        .expect("UDP error");
+    unreachable!("embassy-net's sockets do not get closed (but embedded-nal-coap can't know that)");
```
ROMemories, line 76: This is all exactly moved from `src/lib.rs` 👍

src/ariel-os-coap/src/transport_udp.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use super::udp_nal;
+use super::{CLIENT_READY, CONCURRENT_REQUESTS};
+
+/// Runs the CoAP handler on CoAP-over-UDP indefinitely.
+///
+/// # Panics
+///
+/// This can only be run once, as it sets up a system wide CoAP handler.
+pub(crate) async fn coap_run_udp(mut handler: impl coap_handler::Handler) -> ! {
```
ROMemories, line 19: optional: is there really a reason to make it `pub(crate)`? The module is private anyway.

```suggestion
pub async fn coap_run_udp(mut handler: impl coap_handler::Handler) -> ! {
```

src/ariel-os-coap/src/lib.rs
ROMemories, line 152: `ariel-os-random` is not enabled as a dependency if no `coap-server-config-*` feature are enabled. I'm not sure what the right fix is here: maybe just move this into the `cfg_if` block and make it a noop if none of them are enabled?

laze-project.yml
```diff
@@ -1481,6 +1492,7 @@ modules:
     selects:
       - coap
       - sw/storage
+      - random
```
ROMemories: Because the ordering of these has meaning for laze, I would rather stick to the lexicographical order unless there is a reason not to:

```suggestion
      - random
      - sw/storage
```

laze-project.yml
```diff
@@ -1438,8 +1438,7 @@ modules:
       by coap-client and/or coap-server as per the needs of other modules and
       the application.
     selects:
-      - random
-      - network
+      - coap-transport
```
ROMemories, line 1441: Checked that laze refuses to build if we `--disable coap-transport-udp` (as it should).

src/ariel-os-coap/src/transport_udp.rs
chrysn: I prefer the clarity of not needing to know if the module I'm in is pub (or if anyone pub-exports it) to understand that this is not a pub function and I can therefore change it w/o being a publicly breaking change.

src/ariel-os-coap/src/lib.rs
chrysn: The coap-server-config-unprotected and coap-server-config-storage laze modules select the laze "random" module. That in term enables ariel-os/random.

ROMemories: Sure but when trying to check the individual crate with Clippy:

```sh
cargo +nightly clippy -p ariel-os-coap --features ariel-os-embassy/executor-none
```

You'll get the `compile_error` about having to select a transport (as expected), but *also* errors about the `ariel_os_random` dependency being missing (because it's an optional dependency). That seems easy enough to fix by changing the condition under which this statement is compiled in.

chrysn: Fixed in 41891dece915192d4f53622fee11d8c60a46e6cb -- commit message has the details. The fix is not *quite* as trivial as expected, but also ensures that we set variables and build components more precisely when we need them.

ROMemories (approved):

PR #1617, opened 2025-12-10
docs(book): introduce a book page on native

ROMemories:
# Description

This introduces a book page about native, building on #1616 which introduces its support info.

## Issues/PRs References

- Depends on #1616
- Depends on #1619
- Depends on #1640

## Open Questions

- Are there other things that need to go in that page for now?

## Changelog entry

A native target has been introduced: it allows running an application as a Linux process. This is especially useful for experimenting without a physical board, testing applications, and for simulation purposes. A subset of features is currently supported, and will be expanded in the future.

## Change Checklist

- [x] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1619, opened 2025-12-10
docs(book): introduce the concept of laze builders explicitly

ROMemories:
# Description

This explicitly introduces the concept of laze builders, which was not documented as such before (deliberately, because it wasn't clear whether introducing the concept was necessary wrt to boards). #1506 does now call for that.

~~*This is still a WIP as there may be other occurrences of "board" to replace (but this needs to be done with care, they cannot just all be substituted with "laze builder").*~~

## How to review this PR

## Testing

Remaining occurrences of "board" can be found with the following:

```sh
rg --iglob '!book/src/boards/**' --iglob '!**/*.svg' --iglob '!book/src/support_matrix_tier*.html' -i board book
```

## Issues/PRs References

- Needed for #1506
- Needed for #1617

## Open Questions

- We might want to rename the `boards` *key* in SBDs to `builders` (but that might be too laze-specific for SBDs?).

## Changelog entry

The concept of laze builder is now explained in the documentation.

## Change Checklist

### DIFF
--- a/book/src/testing.md
+++ b/book/src/testing.md
@@ -8,7 +8,7 @@ Please refer to the [`embedded-test` documentation][embedded-test-docs] for
 more info.
 
 The build system of Ariel OS integrates the `embedded-test`-based testing so that
-once set up, tests can be run by issuing `laze build -b <board> test`.
+once set up, tests can be run by issuing `laze build -b <builder> test`.
 `embedded-tests` can be used for any target that has `probe-rs` support (which currently means all targets).
 Both async and non-async code can be tested.
 
@@ -123,7 +123,7 @@ more information.
 To run a test, execute from within the crate's directory:
 
 ```shell
-laze build -b <board> test
+laze build -b <builder> test
 ```
 
 [embedded-test-docs]: https://docs.rs/embedded-test/latest/embedded_test/
--- a/book/src/vscode-configuration.md
+++ b/book/src/vscode-configuration.md
@@ -12,7 +12,7 @@ It is also recommended to use the the `Even Better TOML` exentsion to have TOML 
 
 ## Configuration for developing Ariel OS apps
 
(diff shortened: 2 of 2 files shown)


book/src/build-system.md
```diff
@@ -51,13 +51,24 @@ To enable or disable laze modules for an out-of-tree application, see [below](#e
...
+## laze builders
+
+laze *builders* are top-level [laze contexts](#laze-contexts) that can be passed on the command line to select which piece of hardware to compile for, using the following:
+
+```sh
+laze build -b <builder>
+```
+
+In simple cases, e.g., boards with only one single-core MCU, laze builders directly correspond to boards.
+However a board—and even a chip—can also have multiple laze builders, e.g., when a board comprises multiple MCUs, or when an MCU contains multiple cores with separate flash memories.
```
bergzand: I know what the message here is. I'm stumbling over the 'single-core' here as that case is not necessarily on the single core aspect, but more whether the cores allow smp and whether there's only one (Ariel OS) or multiple separate application running on the mcu.

I think the current text is good enough and conveys the idea, but could be improved. What do you think?

ROMemories: I agree that "single-core" itself is not exactly the right property, and this may be misleading that early in the docs (even if it is clearly phrased as just an example).

I've rephrased a bit to talk about "compilation targets" instead; if you think that "compilation target" is too ill-defined or a bit too cyclic a definition, I suppose we can just drop that part of the sentence altogether, and only "define" simple cases in opposition to other cases which are easier to hint to.

bergzand: Thank you for rephrasing. I think this clarifies the message. We can refine it later if we still get questions around this.

bergzand (approved):
LGTM

PR #1622, opened 2025-12-15
feat(boards): add Adafruit Feather nRF52840

chrysn:
# Description

This adds two widely available boards.

## Testing

The blinky test work just fine, and the GPIO example too. The USB serial example tests the USB port, and the CoAP example confirms that USB Ethernet works too. (IDK how that could even not work on the same chip which it is regularly used on.)

## Open Questions

* [x] Does it work? I didn't test yet, because I want to extract the bootloader before I flash my board, so I can later work on making Ariel support that bootloader directly.

  I tested the GPIO and blinky (also modified to use led1) examples on the -express.

  I don't have the -sense at hand, so it's tier 3, but they are similar enough.
* [x] Does anyone have the -sense version so we can make it tier 2? (It should really make not much of a difference, and I see no way how anything USB should not work there when it works on the -express, but that's the way our test reports are currently structured)

  [edit, added]: Asked around -- answer is no, so it stays tier 3.

* [x] @ROMemories, do you want to take any finer tuning to my rather blunt disabling of the yamllint test for comment indentation? The current board files would flag a lint even though I see them as correct, and there's an upstream issue about false positives there.

## Changelog entry

The Adafruit Feather nRF52840 Express and Sense boards are now supported.

## Change Checklist

### DIFF
--- a/boards/adafruit-feather-nrf52840-express.yaml
+++ b/boards/adafruit-feather-nrf52840-express.yaml
@@ -26,6 +26,9 @@ boards:
           - LED2 # bus name in schematic
           - D4 # LED part designator in schematic
           - Conn # name on silkscreen
+
+          # Workaround-for: https://github.com/adrienverge/yamllint/issues/384
+          # yamllint disable-line rule:comments-indentation
       # To be mapped:
       # * neopixel style RGB LED (WS2812B) on P0_16
       # Only to be mapped if we start expressing aspects that are not user controllable:
@@ -42,5 +45,8 @@ boards:
           - UserSw # silkscreen label
           - SW2 # button part designator in schematic
           - SWITCH # bus name in schematic
+
+          # Workaround-for: https://github.com/adrienverge/yamllint/issues/384
+          # yamllint disable-line rule:comments-indentation
       # Only to be mapped if we start expressing aspects that are not user controllable:
       # * SW1 is a reset pin at P0_18/RESET
--- a/boards/adafruit-feather-nrf52840-sense.yaml
+++ b/boards/adafruit-feather-nrf52840-sense.yaml
@@ -21,6 +21,9 @@ boards:
           - D4 # LED part designator in schematic
           - LED2 # bus name in schematic
           - Conn # name on silkscreen
+
+          # Workaround-for: https://github.com/adrienverge/yamllint/issues/384
+          # yamllint disable-line rule:comments-indentation
       # To be mapped:
       # * neopixel style RGB LED (WS2812B) on P0_16
       # Only to be mapped if we start expressing aspects that are not user controllable:
@@ -37,5 +40,8 @@ boards:
           - UserSw # silkscreen label
           - SW2 # button part designator in schematic
           - SWITCH # bus name in schematic
+
+          # Workaround-for: https://github.com/adrienverge/yamllint/issues/384
+          # yamllint disable-line rule:comments-indentation
       # Only to be mapped if we start expressing aspects that are not user controllable:
       # * SW1 is a reset pin at P0_18/RESET


chrysn:
My tests passed for -express (reflected now in the PR main post); -sense is not claimed to be tested as a tier 3 device (but I've used them and they're quite reasonably documented).

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - D3 # silkscreen
+      led1:
+        color: blue
+        pin: P1_10
+        active: high
+        aliases:
+          - LED2 # bus name in schematic
+          - D4 # LED part designator in schematic
+          - Conn # name on silkscreen
+      # To be mapped:
```
ROMemories, line 32: Regarding your open question, I would much rather only disable the check for this comment only (the `disable-line` comment has to be indented this way, otherwise the lint triggers on it):
```suggestion
          # yamllint disable-line rule:comments-indentation
      # To be mapped:
```

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -0,0 +1,44 @@
...
+    buttons:
+      # FIXME: "button0" name is purely synethetic to work with the current
+      # sbd::ariel convention of naming them sequentially.
+      button0:
+        pin: P1_02
+        active: low
+        aliases:
+          - UserSw # silkscreen label
+          - SW2 # button part designator in schematic
+      # Only to be mapped if we start expressing aspects that are not user controllable:
```
ROMemories, line 51:
```suggestion
          # yamllint disable-line rule:comments-indentation
      # Only to be mapped if we start expressing aspects that are not user controllable:
```

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -0,0 +1,44 @@
...
+  # different names in use: "Adafruit Feather nRF52840 Express" on shop as of
+  # 2025-12-14, whereas schematics dating from 2025-05-05 label it as "Adafruit
+  # nRF52840 Bluefruit Feather Express". The board itself does not have any
+  # "express" text on it at all.
+  adafruit-feather-nrf52840-express:
+    chip: nrf52840
+    flags:
+      - has_usb_device_port
+    leds:
+      # FIXME: led0/led1 names are purely synethetic to work with the current
```
ROMemories:
```suggestion
      # FIXME: led0/led1 names are purely synthetic to work with the current
```
Added as a suggestion to `typos` upstream.

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -0,0 +1,44 @@
+boards:
+  # different names in use: "Adafruit Feather nRF52840 Express" on shop as of
+  # 2025-12-14, whereas schematics dating from 2025-05-05 label it as "Adafruit
+  # nRF52840 Bluefruit Feather Express". The board itself does not have any
+  # "express" text on it at all.
+  adafruit-feather-nrf52840-express:
+    chip: nrf52840
+    flags:
+      - has_usb_device_port
```
ROMemories, line 9: If I understand correctly, the USB port is only user USB, and you're expected to use an external debug probe with the SWD connector, or use a DFU bootloader (or similar)?

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - D2 # LED part designator in schematic
+          - D3 # silkscreen
+      led1:
+        color: blue
+        pin: P1_10
+        active: high
+        aliases:
+          - LED2 # bus name in schematic
+          - D4 # LED part designator in schematic
+          - Conn # name on silkscreen
```
ROMemories, line 28: Checked these with the schematic and the pictures from Adafruit's website.

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -0,0 +1,44 @@
...
+      # * neopixel style RGB LED (WS2812B) on P0_16
+      # Only to be mapped if we start expressing aspects that are not user controllable:
+      # * orange D3 is hardwired to LiPo charge port STAT
+
+    buttons:
+      # FIXME: "button0" name is purely synethetic to work with the current
+      # sbd::ariel convention of naming them sequentially.
+      button0:
+        pin: P1_02
+        active: low
```
ROMemories, line 43: Based on the schematic, it seems this requires an *internal* pull-up to not have the input floating. That should work with the `gpio` example because there is a hard-coded pull-up there, but it could be useful to at least have a TODO comment here until we can encode that in SBD files.

boards/adafruit-feather-nrf52840-sense.yaml
```diff
@@ -0,0 +1,39 @@
+boards:
+  adafruit-feather-nrf52840-sense:
+    chip: nrf52840
+    flags:
+      - has_usb_device_port
```
ROMemories, line 5: Based on the pictures, it seems there is no SWD connector on this variant, and you have to use a pogo jig on the pads on the back of the PCB (or use a USB bootloader)?

boards/adafruit-feather-nrf52840-sense.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+    buttons:
+      # FIXME: "button0" name is purely synethetic to work with the current
+      # sbd::ariel convention of naming them sequentially.
+      button0:
+        pin: P1_02
+        active: low
+        aliases:
+          - UserSw # silkscreen label
+          - SW2 # button part designator in schematic
```
ROMemories, line 41: Technically if we have `LED2` which is only a net label, I guess we should have `SWITCH` as well (same for the Express)?

boards/adafruit-feather-nrf52840-express.yaml
chrysn: There is a USB bootloader, which gets removed when flashing Ariel OS, because we don't really support general bootloaders yet. Once that changes, this USB port will also be bootloader USB.

(This is the same situation as with particle-xenon).

boards/adafruit-feather-nrf52840-express.yaml
chrysn: I understood that this would need further modelling and then be refined everywhere, but I added a comment right away to make things easier for then (now that all schematics are still open).

boards/adafruit-feather-nrf52840-sense.yaml
chrysn: The predominant way to use this now is indeed to either use pogo pins, or to solder wires onto the test points.

Life will be easier if we support bootloaders, but that's not on the current priority list.

boards/adafruit-feather-nrf52840-sense.yaml
chrysn: Yes, added. I started with the LED bus names because they were the only designators in the "LED{number}" format, but then it's good for consistency to have the bus name in.

boards/adafruit-feather-nrf52840-express.yaml
chrysn: That's 4 occurrences in 2 files, and 2 lines each (because I don't override lints without reason, and the 2nd line links to the upstream bug).

I think that this hampers readability and merely serves to justify an arbitrary broken linter rule that we cling to for no good reason, but that's for general discussion on our linter strictness and not in here, I want to get things done.

boards/adafruit-feather-nrf52840-express.yaml
chrysn: Addressed / see other comment.

chrysn:
Thanks for the review and the quick turnaround times!

PR #1623, opened 2025-12-15
feat: choose correct gdb for each architecture

nponsard:
# Description

This PR updates the `gdb` task to use the gdb executable that corresponds to the MCU's architecture.

`probe-rs` gdb server is not yet available for Xtensa

## Testing

Can be tested with `probe-rs` compatible MCUs: 

- first flash an example, cancel the execution when the firmware started:
   ```
   laze -C examples/blinky build -b <board> -s probe-rs run
   ```
- Start the gdb server: 
   ```
   laze -C examples/blinky build -b <board> -s probe-rs debug
   ```
- In another terminal, connect to the gdb server: 
   ```
   laze -C examples/blinky build -b <board> -s probe-rs gdb
   ```
 

## Open Questions

Does this need to be more documented ? 

I chose the name `GDB_EXEC` to be more explicit than just "`GDB`", is this the right variable name ?

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -159,7 +159,9 @@ contexts:
 
       gdb:
         cmd:
-          - arm-none-eabi-gdb -ex "target extended-remote localhost:1337" ${out} $@
+          # On debian, install gdb-multiarch and do
+          # alias gdb=gdb-multiarch
+          - gdb -ex "target extended-remote localhost:1337" ${out} $@
         build: false
         ignore_ctrl_c: true
 


kaspar030:
I think we should go through a `GCC_PREFIX` variable (which would contain `xtensa-esp32s2-elf-`, `arm-none-eabi-`, ...), and then define `GDB_EXEC` once as `${GCC_PREFIX}gdb`.

laze-project.yml
```diff
@@ -452,9 +453,7 @@ contexts:
       CARGO_TARGET_PREFIX: CARGO_TARGET_XTENSA_ESP32S2_NONE_ELF
       CC: xtensa-esp32s2-elf-gcc
```
kaspar030, line 455:
```suggestion
      CC: ${GCC_PREFIX}gcc
```

Should we then move this into the ariel-os context?

laze-project.yml
```diff
@@ -486,6 +485,7 @@ contexts:
       CARGO_ENV:
         - TARGET_CC="${CC}"
         - TARGET_CFLAGS="${CFLAGS}"
```
kaspar030, line 490: Also move into ariel-os context?

nponsard: I'm not sure I undersand the effects of moving that to the ariel-os context. This would need some testing.

kaspar030: So the main idea would be that the ariel os context contains:

```
contexts:
  artiel-os:
    env:
      CC: ${GCC_PREFIX}gcc
      GDB: ${GCC_PREFIX}gdb
      CARGO_ENV:
        - TARGET_CC="${CC}"
        - TARGET_CFLAGS="${CFLAGS}"
```

... and child contexts then specify `$GCC_PREFIX`, and can also override `CC` and `CFLAGS`.

nponsard:
Tested against the reproducer of #1697 to be sure it's still handled.

laze-project.yml
```diff
@@ -907,6 +905,7 @@ modules:
           - -Clink-arg=-Tisr_stack.x
           - -Clink-arg=-Tlinkme-region-alias.x
           - -Clink-arg=-Tlinkme.x
+        GCC_PREFIX: riscv64-elf-
```
nponsard: Seems like `riscv64-elf-gcc` can build riscv32 binaries, I couldn't find `riscv32-elf-gcc` on my system yet coap examples are building for esp32c6 (on main and on this PR).

Tested with
```
laze -C examples/coap-server/ build -b seeedstudio-xiao-esp32c6 -s probe-rs run
```

Still working correctly and I'm able do debug it.

nponsard:
I'm cancelling the refactor, we cannot set a gnu toolchain prefix that would work for multiple linux distributions (tried debian and arch).

laze-project.yml
```diff
@@ -159,7 +159,9 @@ contexts:
 
       gdb:
         cmd:
-          - arm-none-eabi-gdb -ex "target extended-remote localhost:1337" ${out} $@
+          # On debian, install gdb-multiarch and do
+          # alias gdb=gdb-multiarch
+          - gdb -ex "target extended-remote localhost:1337" ${out} $@
```
nponsard, line 164: I discovered `gdb-multiarch`, it's enabled by default on Arch linux but for debian you need to install `apt install gdb-multiarch`

Which should be the default ? 

`gdb` or `gdb-multiarch`?


I tested gdb for `nrf9160` (arm), `esp32c6` (RiscV32), there is gdb support for xtensa too but probe-rs debug support isn't there yet.

kaspar030: well, I guess `gdb-multiarch` hits more architectures than plain `gdb`.
... now it bothers me that we have no pattern for for figuring out at least some things from the host system ...

kaspar030: > it's enabled by default on Arch linux

That means my (Arch) `gdb` is multi-arch capable?

nponsard: you can check it by entering the command `set architecture` in you gdb, it should output all the available architectures.

nponsard: > That means my (Arch) gdb is multi-arch capable?

That's what I mean yes, and that's probably why there is no equivalent to `gdb-multiarch` on Arch's repos

PR #1624, opened 2025-12-15
refactor(threads): make idle threads a feature

kaspar030:
# Description

This refactors idle-threads to be a cargo feature of `ariel-os-threads`, and selects it unconditionally if `multi-core` is enabled. It is also available for single core MCUs if needed.

We need this for e.g., esp, to avoid nested interrupts there.

## Issues/PRs References

Split out of #1328.

## Change Checklist

### DIFF
--- a/src/ariel-os-threads/src/arch/mod.rs
+++ b/src/ariel-os-threads/src/arch/mod.rs
@@ -4,6 +4,10 @@ use crate::Thread;
 pub trait Arch {
     const DEFAULT_THREAD_DATA: Self::ThreadData;
 
+    /// Stack size for the idle threads.
+    #[cfg(feature = "idle-threads")]
+    const IDLE_THREAD_STACK_SIZE: usize = 256;
+
     type ThreadData;
 
     /// Sets up the stack for newly created threads and returns the sp.
--- a/src/ariel-os-threads/src/lib.rs
+++ b/src/ariel-os-threads/src/lib.rs
@@ -81,7 +81,7 @@ use thread::{Thread, ThreadState};
 
 #[cfg(feature = "multi-core")]
 use smp::{Multicore, schedule_on_core};
-#[cfg(feature = "multi-core")]
+#[cfg(any(feature = "multi-core", feature = "idle-threads"))]
 use static_cell::ConstStaticCell;
 
 /// Dummy type that is needed because [`CoreAffinity`] is part of the general API.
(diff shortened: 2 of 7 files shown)


ROMemories:
How do I test this?

kaspar030:
> How do I test this?

So with `-DLOG=debug`, this should now show a line at boot logging how many idle threads were started, and what stacksize is used. So e.g. on `esp32s3`, this should report `2` and `2048` by default. On rp, it is `2` and `256`.

Adding `-DFEATURES+=ariel-os/idle-threads` should enable an idle thread where there usually is none, e.g., on any single-core cortex-m. Enabling one if it is not needed is wasteful but shouldn't break anything (only maybe change thread ids). Currently (on main) idle threads are automatically enabled when `multi-core` is enabled (through the former `ariel-os-threads` feature activating the latter).

src/ariel-os-threads/src/arch/mod.rs
```diff
@@ -4,6 +4,10 @@ use crate::Thread;
 pub trait Arch {
     const DEFAULT_THREAD_DATA: Self::ThreadData;
 
+    /// Stack size for the idle threads.
+    #[cfg(feature = "idle-threads")]
```
ROMemories, line 8: Do these need to be feature-gated? They cost nothing anyway, but it might still make sense to feature-gate them?

src/ariel-os-threads/src/lib.rs
```diff
@@ -81,7 +81,7 @@ use thread::{Thread, ThreadState};
 
 #[cfg(feature = "multi-core")]
 use smp::{Multicore, schedule_on_core};
-#[cfg(feature = "multi-core")]
+#[cfg(any(feature = "multi-core", feature = "idle-threads"))]
 use static_cell::ConstStaticCell;
```
ROMemories, line 85: optional: alternatively the import could be moved into `setup_idle_threads` and the multicore block of `start_threading()` to avoid the more complex `cfg` condition.

ROMemories (approved):
Successfully tested the following on an rpi-pico-w (which has a support multicore MCU):

```sh
laze -C examples/log build -s sw/threading -DLOG=debug -b rpi-pico-w run
```

It does print the following as expected:

```
DEBUG ariel-os-threads: setting up 2 idle thread(s) with stack size 256
```

I've also tried making the `multi-core` feature *not* enable `idle-threads` and it panics (as expected), for lack of idle threads.

I've also tried needlessly enabling the `idle-threads`, and it indeed does not seem to cause any issues:

```sh
laze -C examples/log build -s sw/threading -DLOG=debug -DEFEATURES+=ariel-os/idle-threads -b stm32u083c-dk run
```

and

```sh
laze -C examples/log build -s sw/threading -DLOG=debug -DEFEATURES+=ariel-os/idle-threads -d multi-core -b rpi-pico-w run
```

src/ariel-os-threads/src/arch/mod.rs
kaspar030: I feature gated them as the alternative is to `#[allow(unused, reason="...")]`. IMO this encodes its intended use perfectly.

src/ariel-os-threads/src/lib.rs
kaspar030: I prefer to keep as is.

PR #1628, opened 2025-12-16
refactor(sensors): move the registry into a separate crate

ROMemories:
# Description

This moves the sensor registry into a separate crate. This has two main benefits:

- This leaves really only the `Sensor` trait (and everything necessary for it) in the `ariel-os-sensors`, thus ever minimizing the necessity of breaking changes, which could fragment the sensor driver ecosystem.
- This makes the `ariel-os-sensors` not depend on [`linkme`](https://docs.rs/linkme/latest/linkme/), whose implementation is very platform dependent and may not be supported on every platform (esp. Wasm), thus making sure we don't run into incompatibility issues later.

## How to review this PR

## Testing

- This PR makes sure that the docs stay exactly the same, including every linked item, except for the `Registry` link from the `Sensor::trigger_measurement()` method.
- Successfully tested the following:

```sh
```

## Change Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -156,6 +156,7 @@ jobs:
             -p ariel-os-random
             -p ariel-os-rt
             -p ariel-os-sensors
+            -p ariel-os-sensors-registry
             -p ariel-os-sensors-utils
             -p ariel-os-storage
             -p ariel-os-threads
@@ -435,7 +436,7 @@ jobs:
           tool: cargo-manifest-check@0.2
 
       - name: cargo-manifest-check
-        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-sensors src/ariel-os-sensors-utils src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
+        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-sensors src/ariel-os-sensors-registry src/ariel-os-sensors-utils src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
 
   lint-ruff:
     runs-on: ubuntu-latest


src/ariel-os-sensors/src/lib.rs
```diff
@@ -25,13 +25,6 @@
 //! - Making it easy and as transparent as possible to substitute a specific sensor device by a
 //!   similar one from the same category.
 //!
-//! # Accessing sensor driver instances
-//!
-//! Registered sensor driver instances can be accessed using
-//! [`REGISTRY::sensors()`](registry::Registry::sensors).
-//! Sensor drivers implement the [`Sensor`] trait, which allows to trigger measurements and obtain
-//! the resulting readings.
-//!
```
ROMemories, line 34: Removed this as this is only describing the registry.

This documentation actually doesn't appear when looking at the docs of the `ariel-os` crate (only when rendering the docs of `ariel-os-sensors`).

src/ariel-os/src/sensors.rs
```diff
@@ -0,0 +1,64 @@
...
+//! To avoid handling floats, [`Sample`] returned by [`Sensor::wait_for_reading()`]
+//! are integers, and a fixed scaling value is provided in
+//! [`ReadingChannel`](sensor::ReadingChannel), for each [`Sample`] returned.
+//! See [`Sample`] for more details.
+//!
+//! # For implementors
+//!
+//! Sensor drivers must implement the [`Sensor`] trait.
+//!
+//! [`Sample`]: ariel_os_sensors::sensor::Sample
```
ROMemories, line 59: This is just the copy-paste of the existing documentation from `ariel-os-sensors`. I suppose we could factor it out later by providing docs in macros from `ariel-os-sensors` using `#[doc = …]`.

src/ariel-os-sensors/src/sensor.rs
```diff
@@ -22,9 +22,9 @@ pub trait Sensor: Send + Sync {
     /// Clears the previous reading.
     ///
     /// To obtain readings from every sensor drivers this method can be called in a loop over all
-    /// sensors returned by [`Registry::sensors()`](crate::registry::Registry::sensors), before
-    /// obtaining the readings with [`Self::wait_for_reading()`] in a second loop, so that the
-    /// measurements happen concurrently.
+    /// sensors returned by `Registry::sensors()`, before obtaining the readings with
+    /// [`Self::wait_for_reading()`] in a second loop, so that the measurements happen
+    /// concurrently.
```
anlavandier, line 27: I'm not a fan of having documentation that references a crate that this crate is not depending on. I think that this piece of information should really only be in the main ariel-os documentation, which is already the case as of this PR.

ROMemories: > I'm not a fan of having documentation that references a crate that this crate is not depending on.

Actually I think that's fine in principle, and that it is [only a limitation of rustdoc](https://github.com/rust-lang/rust/issues/74481) (even though indeed we lose the automated link checking). `ariel-os` can't override this piece of docs in the trait. I agree however that how and where we document this could be revisited (and mostly moved to the registry docs), but in this PR I really just wanted to keep things as-is as much as possible, and the revisit the docs in a separate, `ci-build:skip` PR (especially as the `ci-build:stm32` CI has already run now).

anlavandier: Does it make sense to have usage tips in the documentation of a trait crate ? I personally don't think so but I understand that this is out of scope for this PR

anlavandier:
I'm not sure what I need to do to unblock the merging

ROMemories:
> I'm not sure what I need to do to unblock the merging

Just marking all comment threads as resolved.

PR #1637, opened 2025-12-17
fix(stts22h-sensor): fix `wait_for_reading()`'s cancel safety

ROMemories:
# Description

Following https://github.com/ariel-os/ariel-os/pull/1431#discussion_r2626628049, this adds a test checking the cancel safety of `wait_for_reading()`.

## Testing

CI will do it for the test.

```sh
```

## Change Checklist

### DIFF
--- a/src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
+++ b/src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
@@ -176,10 +176,10 @@ impl<I2C: I2c + Send> Stts22h<I2C> {
 
 impl<I2C: Send> Sensor for Stts22h<I2C> {
     fn trigger_measurement(&self) -> Result<(), TriggerMeasurementError> {
+        self.reading.clear();
+
         match self.state.get() {
-            State::Measuring => {
-                self.reading.clear();
-            }
+            State::Measuring => {}
             State::Enabled => {
                 self.state.set(State::Measuring);
             }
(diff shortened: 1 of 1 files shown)


src/sensors/ariel-os-sensor-stts22h/src/i2c.rs
```diff
@@ -432,6 +432,36 @@ mod tests {
...
+                let waiter = STTS22H.wait_for_reading();
+                // Cancel the Future.
+                drop(waiter);
+
+                STTS22H.trigger_measurement().unwrap();
+
+                let reading = STTS22H.wait_for_reading().await.unwrap();
+                let (_channel, sample) = reading.sample();
+
+                assert_eq!(sample.value(), Ok(2500));
```
ROMemories: Based on https://github.com/ariel-os/ariel-os/pull/1431#discussion_r2626628049, we might want this be the following instead (i.e., the second reading instead of the first one), by always `clear()`ing the signal in `trigger_measurement()`. I need to think about this for a bit:

```suggestion
                assert_eq!(sample.value(), Ok(1800));
```

@bergzand Can you confirm that this is indeed the issue you mentioned?

bergzand: Yes, this is indeed the issue. I would expect the driver to return a new measurement when the old `reading` is dropped and a new measurement is triggered. In case of the mock device here `1800`.

From a user perspective, I think it is currently not possible to:
1. Trigger a measurement
2. Decide, based on other factors, that the measurement is no longer useful
3. Trigger a new measurement.
4. Immediately wait on the value of the second measurement.

Step 4 *might* return the second measurement if scheduling happens to align.

ROMemories: Agreed, changed the test value to `1800` and made the driver always `clear()` in `trigger_measurement()` as you suggested.

PR #1641, opened 2025-12-19
refactor(sensors): remove the dependency on `ariel-os-macros`

ROMemories:
# Description

This refactors how `ReadingChannels` and `Samples` are defined to not rely on a proc-macro. This has multiple benefits:

- This removes dependencies required for proc-macros (e.g., `syn`, `quote`) for sensor driver crates, reducing maintenance and improving compile times (for driver implementors).
- Because proc-macros have to be in a `proc-macro` crate, the proc-macro had to live in `ariel-os-macros`. This removes the dependency on `ariel-os-macros`, simplifying MSRV management, as `ariel-os-sensors`'s MSRV is decoupled from `ariel-os`'s.
- This arguably makes the implementation simpler.

## How to review this PR

This refactor should be exactly equivalent: in particular the docs must be exactly the same and there mustn't be new exported items.

## Testing

```sh
```

```sh
```

CI also helps a lot in the case of this refactor. However Cargo features `max-sample-min-count-$i` with `$i > 3` cannot be easily tested right now because no sensor driver use them.

## Issues/PRs References

- This is the refactor mentioned in https://github.com/ariel-os/ariel-os/pull/1581#discussion_r2630253852
- Depends on #1642
- Depends on #1643
- Depends on #1645

## Open Questions

- Should *declarative* macros be used instead, to make it more obvious that copy-pasted items (actually generated with Vim macros) are the same?

## Change Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -559,7 +559,6 @@ dependencies = [
 name = "ariel-os-sensors"
 version = "0.1.0"
 dependencies = [
- "ariel-os-macros",
  "critical-section",
  "defmt 1.0.1",
  "embassy-futures",
--- a/src/ariel-os-sensors/Cargo.toml
+++ b/src/ariel-os-sensors/Cargo.toml
@@ -10,7 +10,6 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
-ariel-os-macros = { workspace = true }
 defmt = { workspace = true, optional = true }
 embassy-sync = { workspace = true }
 pin-project-lite = { workspace = true }


src/ariel-os-sensors/src/sensor/reading_channels.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+}
+
+impl ReadingChannels {
+    /// Returns an iterator over the underlying [`ReadingChannel`] items.
+    ///
+    /// For a given sensor driver, the number and order of items match the one of
+    /// [`Samples`].
+    #[must_use]
```
ROMemories, line 134: These `must_use` attributes are the only *additions* in this refactor.

Edit: not exactly true anymore, see newer comments below

src/ariel-os-sensors/src/sensor/reading_channels.rs
```diff
@@ -0,0 +1,202 @@
...
+/// # For implementors
+///
+/// [`ReadingChannels`] can be instantiated using [`From`] implementations.
+///
+/// Sensor driver crates must enable the appropriate `max-sample-min-count-*` Cargo feature
+/// on this crate.
+/// For instance, a 3-axis accelerometer driver crate must enable `max-sample-min-count-3`
+/// to be able to expose 3 [`ReadingChannel`]s.
+///
+/// [`Sensor::wait_for_reading()`]: super::Sensor::wait_for_reading()
```
ROMemories, line 14: This link reference definition is needed because the `ReadingChannels` type was previously defined in the same module as `Sensor`.

src/ariel-os-sensors/src/sensor/reading_channels.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+}
+
+impl ReadingChannels {
+    /// Returns an iterator over the underlying [`ReadingChannel`] items.
+    ///
+    /// For a given sensor driver, the number and order of items match the one of
+    /// [`Samples`].
+    ///
+    /// [`Samples`]: super::Samples
```
ROMemories, line 133: This link reference definition is needed because the `ReadingChannels` type was previously defined in the same module as `Samples`.

src/ariel-os-sensors/src/lib.rs
```diff
@@ -51,7 +51,6 @@
 //!
 #![no_std]
 #![cfg_attr(nightly, feature(doc_cfg))]
-#![deny(clippy::pedantic)]
```
ROMemories, line 54: The workspace lints (which do include `clippy::pedantic`) are still inherited: removing this here is needed otherwise the `clippy::too_many_lines` lint triggers, while it was disabled at workspace level in #1643.

PR #1643, opened 2025-12-19
chore(clippy): allow `too_many_lines`

ROMemories:
# Description

This allows `clippy::too_many_lines` at workspace level, which is part of `clippy::pedantic`. My main motivation is that this lint can make otherwise-orthogonal refactors non-commutative when `expect`ed on a function.

We're already aware that it is generally a good thing to keep functions short, but the limit is arbitrary (100 lines by default, even though of course configurable) and readability is better enforced by code review imo anyway.

## Issues/PRs References

- Makes #1641 slightly easier.

## Change Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -187,6 +187,7 @@ indexing_slicing = "deny"
 missing_safety_doc = { level = "warn" }
 pedantic = { level = "warn", priority = -1 }
 too_long_first_doc_paragraph = "warn"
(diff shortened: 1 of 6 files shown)


chrysn (approved):
The argument about workflow is particularly convincing: Concurrent work easily makes this trigger, and I agree that a reviewer is the better judge of what is too long.

PR #1646, opened 2025-12-19
refactor(sensors): replace `pin-project` with `pin-project-lite`

ROMemories:
# Description

This replaces `pin-project` with `pin-project-lite`, for the one usage in the `ariel-os-sensors` crate. Both crates are from the same author, but `pin-project-lite` only uses declarative macros instead of a proc-macro, participating to reducing the dependencies of sensor driver crates (as does #1641). See [the docs](https://docs.rs/pin-project-lite/latest/pin_project_lite/) for more about the trade-offs of `pin-project` vs `pin-project-lite` (I'm not aware of any downsides in our case, except it needed a tiny refactor to be usable).

`pin-project-lite` is also a dependency of `futures-util` which is in many places of our tree, so this isn't a new dependency (and it's already vetted). 

## How to review this PR

Hiding whitespace in the diff might also be useful.

## Testing

```sh
```

## Change Checklist

### DIFF
--- a/src/ariel-os-sensors/src/sensor.rs
+++ b/src/ariel-os-sensors/src/sensor.rs
@@ -138,17 +138,19 @@ impl Future for ReadingWaiter {
     }
 }
 
-#[must_use = "futures do nothing unless you `.await` or poll them"]
-#[pin_project::pin_project(project = ReadingWaiterInnerProj)]
-enum ReadingWaiterInner {
-    Waiter {
-        #[pin]
-        waiter: signal::ReceiveFuture<'static, ReadingResult<Samples>>,
-    },
-    Resolved,
-    Err {
-        err: ReadingError,
-    },
+pin_project_lite::pin_project! {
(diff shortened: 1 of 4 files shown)


src/ariel-os-sensors/src/sensor.rs
```diff
@@ -138,15 +138,19 @@ impl Future for ReadingWaiter {
...
+    #[must_use = "futures do nothing unless you `.await` or poll them"]
+    #[project = ReadingWaiterInnerProj]
+    enum ReadingWaiterInner {
+        Waiter {
+            #[pin]
+            waiter: signal::ReceiveFuture<'static, ReadingResult<Samples>>,
+        },
+        Err {
+            err: ReadingError,
+        },
```
ROMemories, line 151: `pin-project-lite` doesn't currently support tuple variants, so a tiny refactor is needed (not a problem as this enum is private).

nponsard (approved):
Looks good to me, tested on the `stm32u083c-dk` as well.

PR #1652, opened 2026-01-06
feat(sensors): add variants required for pressure sensors

ROMemories:
# Description

This adds a couple a variants required for pressure sensors to the sensor API.

## Change Checklist

### DIFF
--- a/src/ariel-os-sensors/src/category.rs
+++ b/src/ariel-os-sensors/src/category.rs
@@ -49,6 +49,8 @@ pub enum Category {
     Ph,
     /// Pressure sensor.
     Pressure,
+    /// Pressure & temperature sensor.
+    PressureTemperature,
     /// Push button.
     PushButton,
     /// Temperature sensor.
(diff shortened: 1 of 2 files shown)


src/ariel-os-sensors/src/category.rs
```diff
@@ -49,6 +49,8 @@ pub enum Category {
     Ph,
     /// Pressure sensor.
     Pressure,
+    /// Pressure & temperature sensor.
+    PressureTemperature,
```
ROMemories, line 53: Pressure sensors (like many MEMS sensors) often also have a temperature sensing element, even though it might not be very accurate. Pressure sensor *drivers* don't *have* to expose such readings but it still sometimes makes sense.

bergzand: In general it would be great to have a comment around this in the documentation. (but out of scope here)

ROMemories: Agreed, now that we've given this some more thoughts I'll add actual guidelines in the docs.

PR #1653, opened 2026-01-07
feat(network)!: bump `embassy-net` to v0.8.0

ROMemories:
# Description

This bumps `embassy-net` to v0.8.0. ~~This uses an unreleased version of picoserve for now as the latest released is not compatible with `embassy-net` v0.8.~~

BREAKING CHANGE: the crate is re-exported in the `reexports` module.

Also bumps reqwless and picoserve to embassy-net 0.8.0 compatible versions.

## Testing

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/http-client/ build -b espressif-esp32-c6-devkitc-1 run
```

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/http-server/ build -b rpi-pico-w run
```

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/http-client/ build -b rpi-pico-w run
```

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/tcp-echo/ build -b rpi-pico-w run
```

```sh
laze -C examples/tcp-echo/ build -d network-config-dhcp -b nrf52840dk run
```

```sh
laze -C examples/udp-echo/ build -d network-config-dhcp -b nrf52840dk run
```

```sh
laze -C examples/tcp-echo/ build -d network-config-dhcp -b st-nucleo-wb55 run
```

## Issues/PRs References

- Depends on #1720

## Changelog entry

`embassy-net` has been updated to v0.8.0: the [network stack](https://ariel-os.github.io/ariel-os/0.4.0/docs/api/ariel_os/net/type.NetworkStack.html) is now provided by that version, which is a breaking change. The crate is re-exported from `ariel_os::reexports`.

## Change Checklist

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -234,7 +234,7 @@ version = "0.1.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.buffered-io]]
-version = "0.5.4"
+version = "0.6.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.bytemuck]]
@@ -329,6 +329,10 @@ criteria = "safe-to-deploy"
 version = "0.9.6"
 criteria = "safe-to-deploy"
 
+[[exemptions.const-oid]]
+version = "0.10.2"
+criteria = "safe-to-deploy"
+
 [[exemptions.const-sha1]]
 version = "0.3.0"
 criteria = "safe-to-deploy"
@@ -391,6 +395,14 @@ criteria = "safe-to-deploy"
 
 [[exemptions.ctr]]
 version = "0.9.2"
+criteria = "safe-to-deploy"
+
+[[exemptions.curve25519-dalek]]
+version = "4.1.3"
+criteria = "safe-to-deploy"
+
+[[exemptions.curve25519-dalek-derive]]
+version = "0.1.1"
 criteria = "safe-to-deploy"
 
 [[exemptions.cyw43]]
@@ -469,6 +481,18 @@ criteria = "safe-to-deploy"
 version = "0.13.3"
 criteria = "safe-to-deploy"
 
+[[exemptions.der]]
+version = "0.8.0"
+criteria = "safe-to-deploy"
+
+[[exemptions.der_derive]]
+version = "0.8.0"
+criteria = "safe-to-deploy"
+
+[[exemptions.deranged]]
+version = "0.5.8"
+criteria = "safe-to-deploy"
+
 [[exemptions.digest]]
 version = "0.10.7"
 criteria = "safe-to-deploy"
@@ -479,6 +503,14 @@ criteria = "safe-to-deploy"
 
 [[exemptions.ecdsa]]
 version = "0.16.9"
+criteria = "safe-to-deploy"
+
+[[exemptions.ed25519]]
+version = "2.2.3"
+criteria = "safe-to-deploy"
+
+[[exemptions.ed25519-dalek]]
+version = "2.2.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.elliptic-curve]]
@@ -510,7 +542,7 @@ version = "0.3.0@git:0feffeb3c90748112a64c6570a39e1251311accb"
 criteria = "safe-to-deploy"
 
 [[exemptions.embassy-net]]
-version = "0.7.1@git:f88320aea64cf24026dd95ce3405ac23253b1117"
+version = "0.8.0@git:3b185bf1b3852867e4b932456600621f86f78bce"
 criteria = "safe-to-deploy"
 
 [[exemptions.embassy-net-driver]]
@@ -609,6 +641,10 @@ criteria = "safe-to-deploy"
 version = "0.8.0"
 criteria = "safe-to-deploy"
 
+[[exemptions.embedded-nal-async]]
+version = "0.9.0"
+criteria = "safe-to-deploy"
+
 [[exemptions.embedded-nal-coap]]
 version = "0.1.0-alpha.5"
 criteria = "safe-to-deploy"
@@ -634,7 +670,7 @@ version = "0.7.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.embedded-tls]]
-version = "0.17.0"
+version = "0.18.0"
 criteria = "safe-to-deploy"
 
 [[exemptions.ena]]
@@ -961,10 +997,6 @@ criteria = "safe-to-deploy"
 version = "0.2.9"
 criteria = "safe-to-deploy"
 
-[[exemptions.lhash]]
-version = "1.1.0"
-criteria = "safe-to-deploy"
-
 [[exemptions.libloading]]
 version = "0.8.7"
 criteria = "safe-to-deploy"
@@ -1073,6 +1105,10 @@ criteria = "safe-to-deploy"
 version = "0.13.2"
 criteria = "safe-to-deploy"
 
+[[exemptions.p384]]
+version = "0.13.1"
+criteria = "safe-to-deploy"
+
 [[exemptions.panic-semihosting]]
 version = "0.6.0"
 criteria = "safe-to-deploy"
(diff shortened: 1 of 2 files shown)


src/ariel-os-coap/src/udp_nal/util.rs
```diff
@@ -61,6 +61,7 @@ impl embedded_io_async::Error for Error {
             Self::SendError(udp::SendError::NoRoute) | Self::BindError(udp::BindError::NoRoute) => {
                 embedded_io_async::ErrorKind::AddrNotAvailable
             }
+            Self::SendError(udp::SendError::PacketTooLarge) => embedded_io_async::ErrorKind::OutOfMemory,
```
ROMemories, line 64: @chrysn Do you think this is the right variant to return? It's not a dynamic allocation error, but it's still a "not enough buffer space/memory available" error.

examples/http-server/src/main.rs
```diff
@@ -4,7 +4,7 @@
 
 mod routes;
 
-use ariel_os::{asynch::Spawner, cell::StaticCell, net, time::Duration};
+use ariel_os::{asynch::Spawner, cell::StaticCell, net};
```
ROMemories: This shouldn't have been using the `ariel_os::time` types anyway since they are passed to `picoserve`, as specifically called out in the docs of that module.

examples/tcp-echo/src/main.rs
```diff
@@ -1,8 +1,9 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::{debug::log::*, net, reexports::embassy_net, time::Duration};
+use ariel_os::{debug::log::*, net, reexports::embassy_net};
```
ROMemories: Same as for the `http-server` example.

examples/http-server/Cargo.toml
```diff
@@ -8,6 +8,7 @@ publish = false
 ariel-os = { path = "../../src/ariel-os" }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 embassy-sync = { workspace = true }
+embassy-time = "0.5.0"
```
ROMemories: `embassy-net` v0.7 expects `embassy-time` v0.5's types.

supply-chain/config.toml
ROMemories, line 1: I've added exemptions for these new versions as well, as unfortunately it didn't seem very practical to audit these for this PR.

src/ariel-os-coap/src/udp_nal/util.rs
chrysn: I think it's the best match so far; we might still open a PR at embedded-io-async to add a "too large to send".

ROMemories:
A new version of `reqwless` has been released yesterday which should make it compatible, but this is likely better done after #1328.

ROMemories:
Seems like `reqwless` v0.14 requires Rust 1.91; we don't necessarily need to bump our MSRV because reqwless is only part of an example, but the stable release we use is that of our MSRV:

https://github.com/ariel-os/ariel-os/blob/5974f22ac112b6b2465f9e9415eb3ba35fc4a141/.github/workflows/build.yml#L201

examples/http-server/Cargo.toml
```diff
@@ -8,8 +8,9 @@ publish = false
 ariel-os = { path = "../../src/ariel-os" }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 embassy-sync = { workspace = true }
-picoserve = { version = "0.14.1", default-features = false, features = [
+picoserve = { version = "0.18.0", default-features = false, features = [
```
kaspar030, line 11: This technically requires Rust 1.93, but also +nightly, so we get away with Ariel's MSRV not increasing.

examples/http-server/laze.yml
```diff
@@ -3,7 +3,7 @@ apps:
     env:
       global:
         executor_stacksize_required:
-          - "16384"
+          - "32768"
```
kaspar030, line 6: This crashed on the Xtensa s3 devkit with the previous `16384`. Tried `20480`, then just doubled.

examples/http-server/Cargo.toml
kaspar030: Actually xtensa still breaks here.

kaspar030: Fixed with #1830.

nponsard:
Successfully tested example `http-server` on `rpi-pico` (usb-ethernet), `nrf52840dk`, `seeedstudio-xiao-esp32c6` and `espressif-esp32-s3-devkitc-1`.

kaspar030:
> It seems like `http-client` needs more stacksize on `esp32` and `esp32s3`. With `executor_stacksize_required` at 32768 it crashes, setting it to 65536 fixes it.

Wow that means this needs a lot more.

> Tested on `esp32c6` with `executor_stacksize_required` at 32768 with no issues.

... and not the same on all platforms. :(

I hope this is caused by `reqwless` now supporting certificate verification with `embedded-tls`, but enabling too many cyphers - git upstream feature-gates some.

I pushed the change to 65536. Let's go with this, wait the next reqwless version, then see and possibly make this MCU dependent.

nponsard:
This won't solve #1827 
I tried with 
```
laze -C examples/tcp-echo/ build -d network-config-dhcp -b stm32u083c-dk run
```
And I'm still getting the error 

```
error[E0308]: mismatched types
   --> src/ariel-os-embassy/src/net.rs:215:30
    |
215 |                 dns_servers: heapless::Vec::new(),
    |                              ^^^^^^^^^^^^^^^^^^^^ expected `Vec<Ipv4Address, 3>`, found `VecInner<_, _, VecStorageInner<[MaybeUninit<_>; _]>>`
    |
    = note: expected struct `heapless::vec::Vec<Ipv4Addr, 3>`
               found struct `VecInner<_, _, vec::storage::VecStorageInner<[MaybeUninit<_>; _]>>`
```

ROMemories:
I should clarify that, on top of the issues raised above, we've initially delayed merging this to wait for the ecosystem to catch up on the release of `embedded-nal-async` v0.9, to not inflict the pain of dependency hell on our users. A few weeks/months later, we can now consider this again.

examples/http-client/Cargo.toml
```diff
@@ -13,8 +13,13 @@ ariel-os = { path = "../../src/ariel-os", features = [
   "time",
 ] }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
+
+# Needed with embedded-tls 0.18.0, which reqwless 0.14.0 depends on.
```
ROMemories: I was actually confused by this: this is actually a bug in `embedded-tls` v0.18 (which is the one that should enable that feature on `der`), and is [now fixed upstream](https://www.github.com/drogue-iot/embedded-tls/pull/182). `embedded-tls` seems to be about to release a v0.19 (the changelog is ready).

nponsard:
`ariel-os-coap` is still using `embedded-nal-async` v0.8.0, this doesn't seem to break anything, but something we may want to monitor.

examples/http-client/Cargo.toml
kaspar030: Yup, it's been ready for some days. Then, `reqwless` needs a release that uses that. (IMO, no need to wait.)

examples/http-client/Cargo.toml
```diff
@@ -13,8 +13,14 @@ ariel-os = { path = "../../src/ariel-os", features = [
   "time",
 ] }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
+
+# A bug in embedded-tls 0.18.0, which reqwless 0.14.0 depends on, makes this
+# temporarily needed.
+# The `heapless` feature is needed since 0.8.0-rc.2.
```
ROMemories, line 19: I've rephrased this comment a bit.

kaspar030:
I think this just needs squashing, right?

ROMemories:
I've updated the changelog entry to explain that the network stack type that Ariel OS exposes is affected by this. Unfortunately we can't make this an implementation detail because it is expected to be passed *by users* to other crates that need to be able to accept it. We might want to look into better documenting these, both for us and for users (as was hinted to by #1229).

PR #1472, opened 2025-10-27
ci: check more crates with `cargo-manifest-check`

jkr-ariel:
# Description

This PR adds `cargo-manifest-check` CI checks for more crates inside `src/`. The crates chosen were thoses which are updated infrequently (based on their last commit date) to avoid confilcts as much as possible.

## Issues/PRs references

Part of #1217.
Follow up to #1235.

## Change checklist

- [x] I have cleaned up my commit history and squashed fixup commits.
- [x] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

PR #1656, opened 2026-01-07
chore(ci): check all crates with `cargo-manifest-check`

### LINKED ISSUE #1217: ci: introducing `cargo-manifest-check` piece by piece

jkr-ariel:
# Description

This PR extends the `Cargo.toml` verification to all crates in the `src/`, and the workspace `Cargo.toml`. Previously, only some crate were checked.

## Issues/PRs References

- Closes #1217.
- Follow up to #1235 and #1472.
- Depend on #1328

## Change Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -446,7 +446,7 @@ jobs:
           tool: cargo-manifest-check@0.3
 
       - name: cargo-manifest-check
-        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-sensors src/ariel-os-sensors-registry src/ariel-os-sensors-utils src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
+        run: cargo manifest-check
 
   lint-ruff:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


jkr-ariel:
We should probably wait for #1328 to be merged before merging it to reduce friction.

.github/workflows/main.yml
```diff
@@ -433,7 +433,7 @@ jobs:
           tool: cargo-manifest-check@0.3
 
       - name: cargo-manifest-check
-        run: cargo manifest-check src/ariel-os-identity src/ariel-os-buildinfo src/ariel-os-debug-log src/ariel-os-coap src/ariel-os-macros src/ariel-os-power src/ariel-os-runqueue src/ariel-os-sensors src/ariel-os-sensors-registry src/ariel-os-sensors-utils src/ariel-os-stm32-mapping src/ariel-os-utils tests/ examples/
+        run: cargo manifest-check src/ tests/ examples/
```
ROMemories: There's also the root `Cargo.toml` (this could be done as a follow-up but I think this is better done in this PR if possible):

```suggestion
        run: cargo manifest-check
```

PR #1658, opened 2026-01-08
docs(support-matrix): document `nrf5340-net` chip

jkr-ariel:
# Description

This PR documents support for the `nrf5340-net` laze context.

## Changelog entry

(nRF5340) Both the application core and the network core of this MCU are now supported and their usage is documented.

## Change Checklist


doc/support_matrix.yml
```diff
@@ -174,7 +174,24 @@ chips:
...
+  nrf5340-net:
+    name: nRF5340 network core
+    manufacturer: Nordic Semiconductor
+    support:
+      gpio: supported
+      debug_output: supported
+      hwrng: supported
+      i2c_controller: supported
+      spi_main: supported
+      uart: supported
```
ROMemories: These are not implemented yet:
```suggestion
      i2c_controller: not_currently_supported
      spi_main: not_currently_supported
      uart: not_currently_supported
```

doc/support_matrix.yml
```diff
@@ -174,7 +174,24 @@ chips:
...
+      logging: supported
+      storage: not_available
```
ROMemories: Persistent storage [is actually supported](https://github.com/ariel-os/ariel-os/blob/54a3a4d236df6a6bfeba2b7dee130391ec04b259/src/ariel-os-storage/build.rs#L14):

```suggestion
      storage: supported
```

doc/support_matrix.yml
```diff
@@ -174,7 +174,24 @@ chips:
       logging: supported
       storage: supported
       wifi: not_available
+      ble: not_available
+
+  nrf5340-net:
+    name: nRF5340 network core
+    manufacturer: Nordic Semiconductor
+    support:
+      gpio: supported
```
ROMemories: @nponsard There is a GPIO peripheral, but isn't it the case that control of individual pins need to be transferred to the net core explicitly? Should we mark this as `not_currently_supported` for now?

doc/support_matrix.yml
```diff
@@ -174,7 +174,24 @@ chips:
       logging: supported
       storage: supported
       wifi: not_available
+      ble: not_available
+
+  nrf5340-net:
```
ROMemories, line 179: @nponsard As you've used that core quite a bit, can you check this new support info?

doc/support_matrix.yml
nponsard: This is right, the datasheet states
> By default, all pins are assigned to the application core

`supported_with_caveats` could also be a good label, with the description saying that pins need to be assigned to the network core from the application core.

doc/support_matrix.yml
jkr-ariel: The [storage](https://github.com/ariel-os/ariel-os/tree/main/examples/storage) example cannot be built for the `nrf5340dk-net` builder currently. Should we change that?

nponsard: storage should be `not_currently_supported` for `nrf5340-net`, in theory it should be easy to add support but this is outside the scope of this PR.

doc/support_matrix.yml
nponsard: I agree with this, these are supported for the application core (`nrf5340`) but not the network core (`nrf5340-net`)

doc/support_matrix.yml
ROMemories: Agreed, I had only checked the build script:
```suggestion
      storage: not_currently_supported
```

doc/support_matrix.yml
ROMemories: I agree that `supported_with_caveats` would be the right one.

doc/support_matrix.yml
```diff
@@ -174,7 +174,24 @@ chips:
       logging: supported
       storage: supported
       wifi: not_available
+      ble: not_available
+
```
nponsard, line 178: `ethernet_over_usb` and `user_usb` can be marked as supported for `nrf5340`, I successfully tested them.

ROMemories: Actually I think `user_usb` (and therefore `ethernet_over_usb`) are only marked as supported *on boards*, when there is a user USB port (the only exception currently is on `esp32s3`, which I should have added on the ESP32-S3-based boards in #1561; will fix that). So I think there isn't anything to do here.

nponsard: Ahh I see, its the fact that it's marked as `not_available` on chips that confused me

doc/support_matrix.yml
nponsard: Oh I didn't spot it in the middle of the stm32 chips in the build script

doc/support_matrix.yml
```diff
@@ -174,7 +174,27 @@ chips:
...
+  nrf5340-net:
+    name: nRF5340 network core
+    manufacturer: Nordic Semiconductor
+    support:
+      gpio:
+        status: supported
```
nponsard:
```suggestion
        status: supported_with_caveats
```
https://github.com/ariel-os/ariel-os/pull/1658#discussion_r2672182989

ROMemories:
Could you rebase #1574 on top of this to be able to easily check the output? Other than that that's likely fine by me.

ROMemories (approved):
I've checked the output of #1574 after rebasing.

PR #1659, opened 2026-01-08
docs(sensors): clarify the meaning of compound sensor driver categories

ROMemories:
# Description

This clarifies the meaning of compound sensor driver categories and whether being part of such a compound category implies also being part of each "subcategory" of it.

## Change Checklist

### DIFF
--- a/src/ariel-os-sensors/src/category.rs
+++ b/src/ariel-os-sensors/src/category.rs
@@ -1,14 +1,26 @@
 /// Categories a sensor driver can be part of.
 ///
-/// A sensor driver can be part of multiple categories.
+/// A sensor driver may be part of multiple categories.
 ///
 /// # For sensor driver implementors
 ///
-/// Many mechanical sensor devices (e.g., accelerometers) include a temperature sensor as
-/// temperature may slightly affect the measurement results.
+/// Many MEMS sensor devices (e.g., accelerometers) include a temperature sensing element in
+/// addition to their main sensing element, as temperature may slightly affect the measurement
+/// results.
+/// Sensor *drivers* are not under the obligation of exposing such temperature readings, even if
+/// they are exposed by the sensor device.
+/// They may however still be fetched by the sensor driver internally, especially to dynamically
+/// compute the accuracy of the main reading returned by the sensor driver.
 /// If temperature readings are not exposed by the sensor driver, the sensor driver must not be
 /// considered part of a category that includes temperature ([`Category::Temperature`] or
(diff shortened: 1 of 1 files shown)


src/ariel-os-sensors/src/category.rs
```diff
@@ -1,14 +1,24 @@
...
+/// Sensor *drivers* are not under the obligation of exposing such temperature readings, even if
+/// they are exposed by the sensor device.
+/// They may however still be fetched by the sensor driver internally, especially to dynamically
+/// compute the accuracy of the main reading returned by the sensor driver.
 /// If temperature readings are not exposed by the sensor driver, the sensor driver must not be
 /// considered part of a category that includes temperature ([`Category::Temperature`] or
-/// [`Category::AccelerometerTemperature`] in the case of an accelerometer).
+/// [`Category::AccelerometerTemperature`] in the case of an accelerometer), even if the sensor
+/// *device* does expose them.
+///
```
bergzand, line 20: Adding the why or when to *not* expose the extra sensor in these devices helps for implementors. As always, feel free to rephrase or mangle into the above paragraph.

```suggestion
///
/// One reason to *not* expose the extra sensor readings is a lack of precision or accuracy with the extra sensor.
///
```

src/ariel-os-sensors/src/category.rs
```diff
@@ -1,14 +1,24 @@
...
+/// Sensor drivers may be part of multiple categories and should then list all of them: e.g., being
+/// part of the [`Category::AccelerometerTemperature`] does *not* imply also being part of the
+/// [`Category::Accelerometer`] category, and the sensor should list both of them.
```
bergzand: Should this be a 'must'?

```suggestion
/// [`Category::Accelerometer`] category, and the sensor must list both of them.
```

bergzand (changes requested):

src/ariel-os-sensors/src/category.rs
ROMemories: Yes of course! Replaced "sensor" with "sensing element" and merged it with the paragraph right before this.

src/ariel-os-sensors/src/category.rs
ROMemories: I initially did write "should" on purpose as my reasoning was that the definitions of categories were technically disjoint (even though semantically related), but I agree using "must" is just simpler (even though maybe that means that `AccelerometerTemperature` *does* imply semantically `Accelerometer`?).

I've also replaced "sensor" with "sensor driver" for clarity.

PR #1660, opened 2026-01-08
docs(esp32s3): fix user USB support info on ESP32-S3-based boards

ROMemories:
# Description

Following #1561 and https://github.com/ariel-os/ariel-os/pull/1658#discussion_r2672375175, this fixes the support info about user USB on ESP32-S3 chips and boards: `user_usb` requires a user USB *port* and therefore can only be marked as supported on boards.

## Change Checklist

### DIFF
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -462,7 +462,6 @@ chips:
         status: not_currently_supported
         comments:
           - requires partitioning support
-      user_usb: supported
       ethernet_over_usb: not_currently_supported
       wifi: supported
       ble: not_currently_supported
@@ -483,7 +482,7 @@ chips:
         status: not_currently_supported
         comments:
           - requires partitioning support
-      user_usb: supported
+      user_usb: needs_testing # This should be moved to a board when one is added.
       ethernet_over_usb: not_currently_supported
       wifi: supported
       ble: not_currently_supported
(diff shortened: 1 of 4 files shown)


doc/support_matrix.yml
```diff
@@ -938,13 +937,15 @@ boards:
...
     support:
+      user_usb: supported
 
   heltec-wifi-lora-32-v3:
     name: Heltec WiFi LoRa 32 V3
     url: https://web.archive.org/web/20250807184214/https://heltec.org/project/wifi-lora-32-v3/
     chip: esp32s3
     tier: "3"
     support:
+      user_usb: not_available
```
ROMemories, line 948: From my reading of the board's schematic, the USB port is only connected to the USB-to-UART chip, not to the USB pins of the ESP32-S3 chip.

doc/support_matrix.yml
```diff
@@ -483,7 +482,7 @@ chips:
         status: not_currently_supported
         comments:
           - requires partitioning support
-      user_usb: supported
+      user_usb: needs_testing # This should be moved to a board when one is added.
```
ROMemories, line 485: The generation script seems to always require a `user_usb` key to be present, as that chip is not present on any supported boards currently.

doc/support_matrix.yml
nponsard: That's what I see too, GPIO19 and GPIO20 (D- and D+) of the esp chip doesn't seem to be connected to anything.

ROMemories: Exactly, technically they are connected to the male headers, but that's not exactly a USB port.

PR #1661, opened 2026-01-09
fix(nrf): use non-lfxo pins for nrf5340

kaspar030:
# Description

`P0_00` and `P0_01` are gated behind lxfo. just use others.

## Issues/PRs References

Split out of #1328.

## Change Checklist

### DIFF
--- a/examples/sensors-debug/src/pins.rs
+++ b/examples/sensors-debug/src/pins.rs
@@ -14,7 +14,11 @@ pub type SensorI2c = i2c::controller::TWISPI0;
 pub type SensorI2c = i2c::controller::SERIAL0;
 #[cfg(all(
     context = "nrf",
-    not(any(context = "bbc-microbit-v2", context = "nordic-thingy-91-x-nrf9151"))
+    not(any(
+        context = "bbc-microbit-v2",
+        context = "nordic-thingy-91-x-nrf9151",
+        context = "nrf5340"
+    ))
 ))]
 ariel_os::hal::define_peripherals!(Peripherals {
     i2c_sda: P0_00,
@@ -24,6 +28,11 @@ ariel_os::hal::define_peripherals!(Peripherals {
 ariel_os::hal::define_peripherals!(Peripherals {
     i2c_sda: P0_16,
     i2c_scl: P0_08,
+});
+#[cfg(context = "nrf5340")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    i2c_sda: P0_20,
+    i2c_scl: P0_22,
 });
 #[cfg(context = "nordic-thingy-91-x-nrf9151")]
 ariel_os::hal::define_peripherals!(Peripherals {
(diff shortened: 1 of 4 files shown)


examples/sensors-debug/src/pins.rs
```diff
@@ -25,6 +29,11 @@ ariel_os::hal::define_peripherals!(Peripherals {
     i2c_sda: P0_16,
     i2c_scl: P0_08,
 });
+#[cfg(context = "nrf5340")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    i2c_sda: P0_02,
+    i2c_scl: P0_03,
```
ROMemories: P0.02 and P0.03 are the pins dedicated to NFC, this only works because we have `nfc-pins-as-gpio` enabled (and actually I don't think we should be enabling that, but that's a separate issue), I don't think these tests and examples should be using such specific pins. Using P0.04 and P0.05 instead should be fine though.

tests/gpio-interrupt-nrf/src/main.rs
```diff
@@ -35,8 +35,8 @@ ariel_os::hal::define_peripherals!(ButtonPeripherals {
 
 #[cfg(context = "nrf5340")]
 ariel_os::hal::define_peripherals!(ButtonPeripherals {
-    btn_0: P0_00,
-    btn_1: P0_01,
+    btn_0: P0_02,
+    btn_1: P0_03,
```
ROMemories: As long as we do have `nfc-pins-as-gpio` enabled, we should still keep P0.02 and P0.03 here however.

Edit: actually they could also be changed to P0.20 and P0.22, but keeping P0.02 and P0.03 here is fine too.

tests/gpio/src/pins.rs
```diff
@@ -18,8 +18,8 @@ ariel_os::hal::define_peripherals!(Peripherals {
 
 #[cfg(context = "nrf5340")]
 ariel_os::hal::define_peripherals!(Peripherals {
-    pin_0: P0_00,
-    pin_1: P0_01,
+    pin_0: P0_02,
+    pin_1: P0_03,
```
ROMemories: These could be using P0.20 and P0.22 (as P0.04 and P0.05 are already used).

ROMemories:
Switched the build label to `ci-build:small` as `nrf5340dk` is part of it and this still needs to check that it doesn't break unrelated builders.

kaspar030:
So, P0.20 and P0.22 everywhere?

ROMemories:
> So, P0.20 and P0.22 everywhere?

Yes, that would work.

tests/i2c-controller/src/pins.rs
```diff
@@ -30,6 +34,11 @@ ariel_os::hal::define_peripherals!(Peripherals {
     i2c_sda: P0_09,
     i2c_scl: P0_08,
 });
+#[cfg(context = "nrf5340")]
+ariel_os::hal::define_peripherals!(Peripherals {
+    i2c_sda: P0_20,
+    i2c_scl: P0_22,
```
ROMemories: Actually these don't work for me when running this test in hardware. The following does work though, and these pins are the ones compatible with Arduino-shields on the nRF5340-DK:

```suggestion
    i2c_sda: P1_02,
    i2c_scl: P1_03,
```

ROMemories (commented):
I've successfully tested the other tests and examples however, so this will be ready to merge after changing the pins in the `i2c-controller` test.

PR #1674, opened 2026-01-12
docs(gen-book): sort laze builders by tier then alpha

jkr-ariel:
# Description

This PR sorts the support matrix in the "Overview" page of the book by laze builder tier, then alpha order.
For board that have more than one builder, only the builder with the lowest tier (meaning the highest level of testing) is considered.
Boards with multiple builders have their builders sorted by tier.

## Issues/PRs References

Follow up to #1506.

## Change Checklist

### DIFF
--- a/book/src/support_matrix.html
+++ b/book/src/support_matrix.html
@@ -26,69 +26,6 @@
   <tbody>
 	<tbody class="odd">
       <tr>
-	    <td rowspan="2"><a href="boards/adafruit-feather-nrf52840-express.html">Adafruit Feather nRF52840 Express</a></td>
-	  </tr>
-	  <tr>
-	    <td><code>adafruit-feather-nrf52840-express</code></td>
-		<td style="text-align: center;">2</td>
-		  <td class="support-cell" title="supported">✅</td>
-		  <td class="support-cell" title="supported">✅</td>
-		  <td class="support-cell" title="supported">✅</td>
-		  <td class="support-cell" title="supported">✅</td>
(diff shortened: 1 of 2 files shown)


ROMemories (commented):
I think we should also move the Tier column to be in first position, otherwise it's not so clear on which keys the table gets sorted. Can you do this in another PR stacked on this one, and also remove "Testing" from the "Testing Board" column at the same time? This made sense when we named it that but not so much anymore since we have chip pages now.

Then I'll approve both PRs at the same time.

jkr-ariel:
> I think we should also move the Tier column to be in first position, otherwise it's not so clear on which keys the table gets sorted. 

I think the name of the board needs to be in the first column otherwise it would look weird with boards that have more than one builder since the row is split.
Unless you mean put the tier just before the builder name?

book/src/support_matrix.html
ROMemories, line 1: @jkr-ariel By the way, could you update the `.gitattributes` following #1574 so this file shows up as auto-generated in the diff view?

jkr-ariel: I noticed that I also forgot to change the column name from "Testing Boards" to "Boards", I'll do it in the same follow-up.

ROMemories: > I noticed that I also forgot to change the column name from "Testing Boards" to "Boards"

I think that's better kept for another PR, we might also consider reshuffling the functionality columns at the same time, for instance to get "Debug Output" and "Logging" closer to each other.

PR #1682, opened 2026-01-13
feat(boards): merge Nordic Thingy 91:X builders into one board

jkr-ariel:
# Description

This PR updates the `nordic-thingy-91-x` board (previously called `nordicy-thingy-91-x-nrf9151`) by merging it with the `nordic-thingy-91-x-nrf5340-net` laze builder.

## Change Checklist

### DIFF
--- a/book/src/boards/nordic-thingy-91-x.md
+++ b/book/src/boards/nordic-thingy-91-x.md
@@ -35,6 +35,34 @@ laze build -b nordic-thingy-91-x-nrf9151
 |Bluetooth Low Energy|<span title="not available on this piece of hardware">–</span>|
 |Hardware Random Number Generator|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>[^only-available-through-the-cryptocell]|
 |Persistent Storage|<span title="supported">✅</span>|
+### `nordic-thingy-91-x-nrf5340-net`
+
+- **Tier:** 2
+- **Chip:** [nRF5340 network core](../chips/nrf5340-net.md)
+- **Chip Ariel OS Name:** `nrf5340-net`
+
+To target this laze builder, run the following command in the root of your Ariel OS app:
+
+```bash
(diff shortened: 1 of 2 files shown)


nponsard:
The pin assignments look good, shouldn't the keys for the nrf5340 begin with 0 too ? like `led0` and `button0`

boards/nordic-thingy-91-x.yaml
```diff
@@ -0,0 +1,21 @@
...
+  nordic-thingy-91-x-nrf9151:
+    chip: nrf9151
+    leds:
+      led0:
+        pin: P0_29
+    buttons:
+      button0:
+        pin: P0_26
+
+  nordic-thingy-91-x-nrf5340:
```
ROMemories: As I said, I think we should keep this one for another PR, after we rename `nrf5340` to `nrf5340-app`; or instead do the rename first and name this `-app` directly.

jkr-ariel: Ok, let's wait after we rename `nrf5340`.

doc/support_matrix.yml
```diff
@@ -826,6 +826,13 @@ builders:
           - requires supporting the onboard nRF7002 chip
       ble: not_available
 
+  nordic-thingy-91-x-nrf5340-net:
+    chip: nrf5340-net
+    tier: "2"
+    support:
+      wifi: not_available
```
ROMemories: I think that one is actually available: Figure 2 of the Thingy:91 X User Guide shows the nRF7002 Wi-Fi chip is connected over SPI to the nRF5340 chip, the net core does have one SPIM peripheral, and the nRF chips are able to route peripheral outputs to most pins, so I suppose in theory it would be usable (it's unclear whether the pretty small flash+RAM memories of the net core are actually enough however). So I think this should copy the comment from the `nordic-thingy-91-x-nrf9151` support info about `wifi`.

ROMemories (approved):

PR #1683, opened 2026-01-13
feat(sensors): introduce a GNSS time extension

nponsard:
# Description

This PR adds the `GnssTimeExt` extension, defining a way to send a timestamp with nanosecond precision over sensor samples.

## Testing

Testing happens in #1330 

## Changelog Entry

A sensor extension has been added, `GnssTimeExt`, that allows to read the time and date from a GNSS sensor.

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -109,6 +109,8 @@ jobs:
             -p embedded-hal@1.0.0 \
             -p embedded-hal-async \
             -p embedded-io-async@0.6.1
+
+          cargo doc --no-deps -p ariel-os-sensors-gnss-time-ext
           RUSTDOCFLAGS='-D warnings --cfg context="esp32c6" --cfg nightly' cargo doc \
               --target=riscv32imac-unknown-none-elf \
               --no-deps \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -163,6 +163,7 @@ jobs:
             -p ariel-os-random
             -p ariel-os-rt
             -p ariel-os-sensors
+            -p ariel-os-sensors-gnss-time-ext
             -p ariel-os-sensors-registry
             -p ariel-os-sensors-utils
             -p ariel-os-storage
@@ -316,6 +317,7 @@ jobs:
                 --no-deps \
                 -p ariel-os \
                 -p ariel-os-sensors \
+                -p ariel-os-sensors-gnss-time-ext \
                 -p coapcore \
                 --features "
                     ariel-os-sensors/max-sample-min-count-12,
--- a/src/ariel-os-sensors-gnss-time-ext/Cargo.toml
+++ b/src/ariel-os-sensors-gnss-time-ext/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "ariel-os-sensors-gnss-time-ext"
+# Versioned separately from the rest of Ariel OS.
+version = "0.1.0"
+edition.workspace = true
+# This crate's MSRV is decoupled from Ariel OS's, but should be no higher than `ariel-os-sensors`'s.
+rust-version = "1.90"
+repository.workspace = true
+license.workspace = true
+
+[dependencies]
+ariel-os-sensors = { path = "../ariel-os-sensors" }
+defmt = { workspace = true, optional = true }
+
+[dev-dependencies]
+ariel-os-sensors = { path = "../ariel-os-sensors", features = [
+  "max-sample-min-count-4",
+] }
+static_cell = { workspace = true }
+
+[features]
+defmt = ["dep:defmt", "ariel-os-sensors/defmt"]
+
+_test = []
+
+[lints]
+workspace = true
--- a/src/ariel-os-sensors-gnss-time-ext/src/lib.rs
+++ b/src/ariel-os-sensors-gnss-time-ext/src/lib.rs
@@ -1,0 +1,548 @@
+//! # GNSS Time Extension
+//!
+//! This extension to [ariel-os-sensors] allows you to read the time information from [`Samples`] produced
+//! by a compatible GNSS driver.
+//!
+//! # For users
+//!
+//! Import the trait [`GnssTimeExt`], you then access the time information on [`Samples`] generated
+//! by a compatible driver by using those functions:
+//!
+//! - [`GnssTimeExt::time_of_fix_timestamp()`]: UTC time in whole seconds since the UNIX epoch.
+//! - [`GnssTimeExt::time_of_fix_timestamp_nanos()`]: UTC time in nanoseconds since the UNIX epoch.
+//!   Some sensors may have worse than nanosecond precision but this function
+//!   will always return nanoseconds.
+//!
+//!
+//! # For implementors
+//!
+//! Get parts with [`convert_datetime_to_parts()`].
+//! You need to return a [`ReadingChannel`] with the label [`Label::OpaqueGnssTime`] containing the first
+//! part *directly followed* by a [`ReadingChannel`] with the label [`Label::Opaque`] containing the second part.
+//!
+//! [ariel-os-sensors]: ariel_os_sensors
+#![cfg_attr(not(any(test, context = "native")), no_std)]
+#![deny(missing_docs)]
+
+use ariel_os_sensors::{
+    Label, Reading as _,
+    sensor::{ReadingChannel, Sample, SampleError, Samples},
+};
+
+// 2025-01-01T00:00Z in seconds
+// Obtained using `date --date "2025-01-01 0:00Z" -u +%s`
+const ARIEL_EPOCH: i64 = 1_735_689_600;
+
+/// Error returned when trying to access the time information on [`Samples`] coming from a GNSS sensor.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssTimeExtError {
+    /// Error reading value from sample, see inner error.
+    Reading(SampleError),
+    /// Sensor does not provide the correct [`ReadingChannels`](ariel_os_sensors::sensor::ReadingChannels).
+    InvalidSensor,
+}
+
+impl core::fmt::Display for GnssTimeExtError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            GnssTimeExtError::Reading(e) => write!(f, "reading error: {e}"),
+            GnssTimeExtError::InvalidSensor => write!(f, "invalid sensor"),
+        }
+    }
+}
+
+impl core::error::Error for GnssTimeExtError {}
+
+/// Error returned when trying to convert the time to two parts.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssTimeExtConvertError {
+    /// Timestamp is either too far in the past or too far in the future.
+    Overflow,
+}
+
+impl core::fmt::Display for GnssTimeExtConvertError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            GnssTimeExtConvertError::Overflow => write!(f, "time conversion overflowed"),
+        }
+    }
+}
+
+impl core::error::Error for GnssTimeExtConvertError {}
+
+// Only implement for Samples
+mod private {
+    use ariel_os_sensors::sensor::Samples;
+
+    pub trait Sealed {}
+    impl Sealed for Samples {}
+}
+
+/// Trait to access time information from [`Samples`] coming from a GNSS sensor.
+pub trait GnssTimeExt: private::Sealed {
+    /// Returns the UTC time in whole seconds since the UNIX epoch.
+    ///
+    /// # Errors
+    ///
+    /// Returns errors if the reading is not compatible with this extension
+    /// or one [`Sample`] returned an error.
+    fn time_of_fix_timestamp(&self) -> Result<i64, GnssTimeExtError>;
+
+    /// Returns the time of the last fix in nanoseconds since the UNIX epoch.
+    ///
+    /// # Errors
+    ///
+    /// Returns errors if the reading is not compatible with this extension
+    /// or one [`Sample`] returned an error.
+    fn time_of_fix_timestamp_nanos(&self) -> Result<i128, GnssTimeExtError> {
+        Ok(
+            i128::from(self.time_of_fix_timestamp()?) * 1_000_000_000i128
+                + i128::from(self.time_of_fix_subsec_nanos()?),
+        )
+    }
+    /// Returns the nanoseconds part of the time of the last fix.
+    ///
+    /// # Errors
+    ///
+    /// Returns errors if the reading is not compatible with this extension
+    /// or one [`Sample`] returned an error.
+    fn time_of_fix_subsec_nanos(&self) -> Result<i64, GnssTimeExtError>;
+}
+
+impl GnssTimeExt for Samples {
+    fn time_of_fix_timestamp(&self) -> Result<i64, GnssTimeExtError> {
+        let sample = get_element_after_marker(self.samples(), Label::OpaqueGnssTime, 0)
+            .ok_or(GnssTimeExtError::InvalidSensor)?;
+
+        let since_ariel_epoch: i64 = sample.1.value().map_err(GnssTimeExtError::Reading)?.into();
+        Ok(ARIEL_EPOCH + since_ariel_epoch)
+    }
+    fn time_of_fix_subsec_nanos(&self) -> Result<i64, GnssTimeExtError> {
+        let sample = get_element_after_marker(self.samples(), Label::OpaqueGnssTime, 1)
+            .ok_or(GnssTimeExtError::InvalidSensor)?;
+
+        Ok(sample.1.value().map_err(GnssTimeExtError::Reading)?.into())
+    }
+}
+
+/// Converts time to parts to be put into [`Samples`].
+///
+/// `timestamp_nanos` is an UTC timestamp as nanoseconds since the UNIX epoch.
+///
+/// The first element of the tuple is to be inserted in a [`ReadingChannel`] that has the label
+/// [`Label::OpaqueGnssTime`].
+/// The second element must be inserted in a [`ReadingChannel`] right after the [`ReadingChannel`] of the first element and
+/// that has the label [`Label::Opaque`].
+///
+/// You would return [`Samples`] like this:
+///
+///     # use ariel_os_sensors_gnss_time_ext::mock_sensor::MockSensor;
+///     # static SENSOR: static_cell::StaticCell<MockSensor> = static_cell::StaticCell::new();
+///     # use ariel_os_sensors_gnss_time_ext::{convert_datetime_to_parts, GnssTimeExtConvertError};
+///     # use ariel_os_sensors::{sensor::{Samples, Sample, ReadingChannel, ReadingChannels}, Label};
+///     // For demonstration purposes, In a normal implementation the Samples should be returned by
+///     // `wait_for_reading()` implemented by the Sensor.
+///     let sensor = SENSOR.init(MockSensor::new(ReadingChannels::from([
+///         ReadingChannel::new(
+///             Label::OpaqueGnssTime,
+///             0,
+///             ariel_os_sensors::MeasurementUnit::Second,
+///         ),
+///         ReadingChannel::new(Label::Opaque, -9, ariel_os_sensors::MeasurementUnit::Second),
+///     ])));
+///
+///     let timestamp_ns: i128 = 982107720621000000;
+///     let (first, second) = convert_datetime_to_parts(timestamp_ns)?;
+///     let samples = Samples::from_2(
+///         sensor,
+///         [
+///             Sample::new(
+///                 first,
+///                 ariel_os_sensors::sensor::SampleMetadata::UnknownAccuracy,
+///             ),
+///             Sample::new(
+///                 second,
+///                 ariel_os_sensors::sensor::SampleMetadata::UnknownAccuracy,
+///             ),
+///         ],
+///     );
+///     # Ok::<(), GnssTimeExtConvertError>(())
+///
+/// # Errors
+/// Returns an error when `timestamp_nanos` is either too far in the past or too far in the future.
+//
+// # Implementation details
+//
+// The first part is the seconds difference with ARIEL_EPOCH, the second part is the sub-second
+// part of timestamp_nanos, in nanoseconds.
+//
+// This is done so we can have the correct second and sub-second part without having to read both.
+// Otherwise the seconds part will be off by 1 second when the timestamp is between the POSIX epoch and
+// the Ariel epoch, handling this rounding adds complexity.
+pub fn convert_datetime_to_parts(
+    timestamp_nanos: i128,
+) -> Result<(i32, i32), GnssTimeExtConvertError> {
+    let timestamp_seconds = i64::try_from(timestamp_nanos / 1_000_000_000)
+        .map_err(|_| GnssTimeExtConvertError::Overflow)?;
+
+    let seconds_since_epoch = i32::try_from(
+        timestamp_seconds
+            .checked_sub(ARIEL_EPOCH)
+            .ok_or(GnssTimeExtConvertError::Overflow)?,
+    )
+    .map_err(|_| GnssTimeExtConvertError::Overflow)?;
+
+    let nanoseconds: i32 = (timestamp_nanos % 1_000_000_000)
+        .try_into()
+        .map_err(|_| GnssTimeExtConvertError::Overflow)?;
+
+    Ok((seconds_since_epoch, nanoseconds))
+}
+
+fn get_element_after_marker(
+    iter: impl Iterator<Item = (ReadingChannel, Sample)>,
+    marker: Label,
+    position: usize,
+) -> Option<(ReadingChannel, Sample)> {
+    let mut peekable = iter.peekable();
+    while let Some((c, _)) = peekable.peek() {
+        if c.label() == marker {
+            break;
+        }
+        peekable.next();
+    }
+    let result = peekable.nth(position);
+
+    // if it's not an opaque channel we're doing something wrong
+    if let Some((c, _)) = result
+        && !(c.label() == Label::Opaque || c.label() == marker)
+    {
+        None
+    } else {
+        result
+    }
+}
+
+/// Mock sensor module for tests.
+/// Has to be public for docs tests.
+#[cfg(any(test, feature = "_test"))]
+pub mod mock_sensor {
+    use ariel_os_sensors::{Sensor, sensor::ReadingChannels};
+
+    /// Mock sensor for tests.
+    pub struct MockSensor {
+        reading_channels: ariel_os_sensors::sensor::ReadingChannels,
+    }
+
+    impl MockSensor {
+        /// Create a new mock sensor.
+        pub const fn new(reading_channels: ReadingChannels) -> Self {
+            Self { reading_channels }
+        }
+    }
+
+    impl Sensor for MockSensor {
+        fn categories(&self) -> &'static [ariel_os_sensors::Category] {
+            unimplemented!()
+        }
+        fn display_name(&self) -> Option<&'static str> {
+            unimplemented!()
+        }
+
+        fn trigger_measurement(
+            &self,
+        ) -> Result<(), ariel_os_sensors::sensor::TriggerMeasurementError> {
+            unimplemented!()
+        }
+
+        fn wait_for_reading(&'static self) -> ariel_os_sensors::sensor::ReadingWaiter {
+            unimplemented!()
+        }
+
+        fn reading_channels(&self) -> ariel_os_sensors::sensor::ReadingChannels {
+            self.reading_channels
+        }
+
+        fn set_mode(
+            &self,
+            _mode: ariel_os_sensors::sensor::Mode,
+        ) -> Result<ariel_os_sensors::sensor::State, ariel_os_sensors::sensor::SetModeError>
+        {
+            unimplemented!()
+        }
+
+        fn state(&self) -> ariel_os_sensors::sensor::State {
+            unimplemented!()
+        }
+
+        fn label(&self) -> Option<&'static str> {
+            unimplemented!()
+        }
+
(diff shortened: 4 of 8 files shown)


src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//!   will always return nanoseconds.
+//!
+//!
+//! # For implementors
+//!
+//! Get parts with [`convert_datetime_to_parts()`].
+//! You need to return a channel with the label [`Label::OpaqueGnssTime`] containing the first
+//! part followed by a channel with the label [`Label::Opaque`] containing the second part.
+#![cfg_attr(not(context = "native"), no_std)]
+
```
ROMemories:
```suggestion
#![deny(missing_docs)]

```

src/ariel-os-gnss-time-extension/Cargo.toml
ROMemories, line 1: Can you make sure `cargo manifest-check` is happy about this manifest? We may not need to add the crate to that CI check if we land #1656 first (and to avoid unneeded conflicts).

src/ariel-os-gnss-time-extension/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition.workspace = true
+# This crate's MSRV is decoupled from Ariel OS's.
+rust-version = "1.90"
+repository.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+ariel-os-debug = { path = "../ariel-os-debug" }
```
ROMemories: I don't think this dependency is used:

```suggestion
```

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+use ariel_os_sensors::{
+    Label, Reading as _,
+    sensor::{ReadingChannel, Sample, SampleError, Samples},
+};
+use time::{UtcDateTime, macros::utc_datetime};
+
+const ARIEL_EPOCH: UtcDateTime = utc_datetime!(2024-01-01 0:00);
```
ROMemories: I suppose we could make it 2025 now?

```suggestion
const ARIEL_EPOCH: UtcDateTime = utc_datetime!(2025-01-01 0:00);
```

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//!   Some sensors may have worse than nanosecond precision but this function
+//!   will always return nanoseconds.
+//!
+//!
+//! # For implementors
+//!
+//! Get parts with [`convert_datetime_to_parts()`].
+//! You need to return a channel with the label [`Label::OpaqueGnssTime`] containing the first
+//! part followed by a channel with the label [`Label::Opaque`] containing the second part.
+#![cfg_attr(not(context = "native"), no_std)]
```
ROMemories: This could help with unit testing:

```suggestion
#![cfg_attr(not(any(test, context = "native")), no_std)]
```

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    Label, Reading as _,
+    sensor::{ReadingChannel, Sample, SampleError, Samples},
+};
+use time::{UtcDateTime, macros::utc_datetime};
+
+const ARIEL_EPOCH: UtcDateTime = utc_datetime!(2024-01-01 0:00);
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum GnssTimeExtError {
```
ROMemories: This error type should implement `core::error::Error` but that was likely not possible because `SampleError` was missing a `core::error::Error` implementation. I've opened #1684 to fix that.

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//! - [`GnssTimeExt::time_of_fix_timestamp_nanos()`]: UTC time in nanoseconds since UNIX epoch.
+//!   Some sensors may have worse than nanosecond precision but this function
+//!   will always return nanoseconds.
+//!
+//!
+//! # For implementors
+//!
+//! Get parts with [`convert_datetime_to_parts()`].
+//! You need to return a channel with the label [`Label::OpaqueGnssTime`] containing the first
+//! part followed by a channel with the label [`Label::Opaque`] containing the second part.
```
ROMemories: I think it's important to insist on this as it's not obvious from the docs alone why is needed:

```suggestion
//! part *directly followed* by a channel with the label [`Label::Opaque`] containing the second part.
```

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        Ok(sample.1.value().map_err(GnssTimeExtError::Reading)?.into())
+    }
+}
+
+/// Convert time to parts to be put into Samples.
+///
+/// # Panics
+/// Panics if the seconds since Ariel's epoch overflow.
+#[must_use]
+pub fn convert_datetime_to_parts(utc_datetime: UtcDateTime) -> (i32, i32) {
```
ROMemories: I have quite some qualms about having a type from `time` in the public API of such a foundational crate:

- Time is complicated and time libraries tend to go extinct: there is already a couple of them in the ecosystem, including the newest `jiff` by BurnSushi.
- `time` isn't even in 1.0 and updating the dependency would be a very significant breaking change in this crate.

I unfortunately don't have an obvious fix in mind. Maybe we can accept exposing the epoch for sensor drivers (while still making sure *users* of the API don't need to work with it)?

src/ariel-os-gnss-time-extension/src/lib.rs
ROMemories: Is there another way of creating this that doesn't rely on a proc-macro? I'd rather avoid these in this type of crate.

src/ariel-os-gnss-time-extension/src/lib.rs
nponsard: A solution could be to request from the sensor a timestamp in nanoseconds since UNIX epoch, subtracting the timestamp corresponding to the start of the Ariel EPOCH would work, and would also avoid using the `utc_datetime!()` proc macro.

nponsard: implemented in commit 2dd3b7498cb945aa87882c1f924e871ab5e0965f

src/ariel-os-gnss-time-extension/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "ariel-os-gnss-time-extension"
+# Versioned separately from the rest of Ariel OS.
+version = "0.1.0"
+edition.workspace = true
+rust-version = "1.90"
+repository.workspace = true
+license.workspace = true
+# This crate's MSRV is decoupled from Ariel OS's.
```
ROMemories: This comment should go back on the `rust-version` key.

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+/// # Panics
+/// Panics if the seconds since Ariel's epoch overflow or the nanoseconds in a second overflow,
+/// the first panic would happen if the timestamp is too far in the future or too far in the past.
+///
+/// The Ariel epoch needs to be changed before this happens.
+///
+/// For the second panic, it should never happen unless something went very wrong,
+/// as the value range should be between 1,000,000 and -1,000,000.
+#[must_use]
+pub fn convert_datetime_to_parts(utc_datetime: i64) -> (i32, i32) {
```
ROMemories: Could we have some unit tests of this function, with well-chosen corner cases?

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// Convert time to parts to be put into Samples.
+///
+/// `utc_datetime` is an UTC timestamp as nanoseconds since UNIX epoch.
+///
+/// # Panics
+/// Panics if the seconds since Ariel's epoch overflow or the nanoseconds in a second overflow,
+/// the first panic would happen if the timestamp is too far in the future or too far in the past.
+///
+/// The Ariel epoch needs to be changed before this happens.
```
ROMemories: Seeing this it looks to me this is somewhat similar to [`SystemTimer::elapsed()`](https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html#method.elapsed) and I think this should actually return an `Option` or even an `Error`. Maybe having a dedicated error would actually allow us to keep the epoch an implementation detail (because this current text wouldn't allow me as a user to actually know when this function would or would not panic).

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Returns the nanoseconds part of the time of the last fix.
+    ///
+    /// # Errors
+    ///
+    /// Returns errors if the reading is not compatible with this extension
+    /// or one [`Sample`] returned an error.
+    fn time_of_fix_delta_ns(&self) -> Result<i64, GnssTimeExtError>;
+}
+
+impl GnssTimeExt for Samples {
```
ROMemories: As this crate does provide an implementation for the foreign `Samples` type, could we actually seal the `GnssTimeExt` trait? I don't think there is any other type it would make sense to implement it for, is it?

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+/// `utc_datetime` is an UTC timestamp as nanoseconds since UNIX epoch.
+///
+/// # Panics
+/// Panics if the seconds since Ariel's epoch overflow or the nanoseconds in a second overflow,
+/// the first panic would happen if the timestamp is too far in the future or too far in the past.
+///
+/// The Ariel epoch needs to be changed before this happens.
+///
+/// For the second panic, it should never happen unless something went very wrong,
+/// as the value range should be between 1,000,000 and -1,000,000.
```
ROMemories: This would just be an actual bug in the implementation, right? I don't think this needs to be documented (unless this can be triggered by the function inputs).

You might however want to consider using the `saturing_`/`wrapping_`/`checked_` variants of the subtraction/multiplication operations, while keeping an eye on binary size.

src/ariel-os-gnss-time-extension/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+#![cfg_attr(not(any(test, context = "native")), no_std)]
+#![deny(missing_docs)]
+
+use ariel_os_sensors::{
+    Label, Reading as _,
+    sensor::{ReadingChannel, Sample, SampleError, Samples},
+};
+
+// 2025-01-01T00:00:00.000Z in nanoseconds
+const ARIEL_EPOCH: i64 = 1_735_689_600_000_000;
```
ROMemories: How did you obtain this? (This could then go into a comment.)

src/ariel-os-gnss-time-extension/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "ariel-os-gnss-time-extension"
```
ROMemories: Maybe we should name this crate that? It only makes sense with the sensor API, and `-ext` seems to be a relatively common prefix in the ecosystem already.

```suggestion
name = "ariel-os-sensors-gnss-time-ext"
```

src/ariel-os-gnss-time-extension/src/lib.rs
nponsard: I used `Deno`
The one liner is 
```
echo 'new Date("2025-01-01 0:00Z").getTime() * 1000' | deno
```

Should this be in the comment ?

ROMemories: If there's an easy way of using coreutils's `date` for that instead, that could go in the comment.

src/ariel-os-gnss-time-extension/src/lib.rs
nponsard: Yeah that would be an implementation issue.

nponsard: Using `checked_sub` twice adds 24B of `text` and 124B of `bss`

nponsard:
Added unit tests, renamed the crate and fixed a few issues found with unit tests.

src/ariel-os-sensors-gnss-time-ext/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "ariel-os-sensors-gnss-time-ext"
+# Versioned separately from the rest of Ariel OS.
+version = "0.1.0"
+edition.workspace = true
+# This crate's MSRV is decoupled from Ariel OS's.
```
ROMemories:
```suggestion
# This crate's MSRV is decoupled from Ariel OS's, but should be no higher than `ariel-os-sensors`'s.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
+//! # GNSS Time Extension
+//!
+//! This extension allows you to read the time information from [`Samples`] produced
```
ROMemories: This is otherwise not mentioned anywhere:
```suggestion
//! This extension to [`ariel-os-sensors`] allows you to read the time information from [`Samples`] produced
```

.github/workflows/build-deploy-docs.yml
```diff
@@ -66,6 +66,7 @@ jobs:
           RUSTDOCFLAGS='--cfg nightly -Z unstable-options --extern-html-root-url trouble_host=https://docs.embassy.dev/trouble-host/0.5.0 ' cargo doc \
               --no-deps \
               -p ariel-os \
+              -p ariel-os-sensors-gnss-time-ext \
```
ROMemories: That's definitely a good idea and I've opened #1849 to do the same for `ariel-os-sensors`. We'll merge #1849 first if that makes more sense for referencing the items from `ariel-os-sensors` from the docs from this new crate.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
ROMemories: Links to `ariel-os-sensors`'s items are broken here because they are not in scope. You can specify the paths explicitly using `[`Samples`]: ariel_os_sensors::sensor::Samples`.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+//! by a compatible GNSS driver.
+//!
+//! # For users
+//!
+//! Import the trait [`GnssTimeExt`], you then access the time information on [`Samples`] generated
+//! by a compatible driver by using those functions:
+//!
+//! - [`GnssTimeExt::time_of_fix_timestamp()`]: UTC time in seconds since UNIX epoch.
+//! - [`GnssTimeExt::time_of_fix_timestamp_nanos()`]: UTC time in nanoseconds since UNIX epoch.
```
ROMemories:
```suggestion
//! - [`GnssTimeExt::time_of_fix_timestamp()`]: UTC time in whole seconds since the UNIX epoch.
//! - [`GnssTimeExt::time_of_fix_timestamp_nanos()`]: UTC time in nanoseconds since the UNIX epoch.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+mod private {
+    use ariel_os_sensors::sensor::Samples;
+
+    pub trait Sealed {}
+    impl Sealed for Samples {}
+}
+
+/// Trait to use to access time information on [`Samples`] coming from a GNSS sensor.
+pub trait GnssTimeExt: private::Sealed {
+    /// Returns the UTC time in seconds since UNIX epoch.
```
ROMemories:
```suggestion
    /// Returns the UTC time in whole seconds since the UNIX epoch.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns errors if the reading is not compatible with this extension
+    /// or one [`Sample`] returned an error.
+    fn time_of_fix_timestamp(&self) -> Result<i64, GnssTimeExtError>;
+
+    /// Returns the time of the last fix in nanoseconds since UNIX epoch.
```
ROMemories:
```suggestion
    /// Returns the time of the last fix in nanoseconds since the UNIX epoch.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+        let sample = get_element_after_marker(self.samples(), Label::OpaqueGnssTime, 1)
+            .ok_or(GnssTimeExtError::InvalidSensor)?;
+
+        Ok(sample.1.value().map_err(GnssTimeExtError::Reading)?.into())
+    }
+}
+
+/// Convert time to parts to be put into Samples.
+///
+/// `timestamp_nanos` is an UTC timestamp as nanoseconds since UNIX epoch.
```
ROMemories:
```suggestion
/// `timestamp_nanos` is an UTC timestamp as nanoseconds since the UNIX epoch.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+        match self {
+            GnssTimeExtError::Reading(e) => write!(f, "reading error: {e}"),
+            GnssTimeExtError::InvalidSensor => write!(f, "invalid sensor"),
+        }
+    }
+}
+
+impl core::error::Error for GnssTimeExtError {}
+
+/// Error returned when trying convert the time to two parts.
```
ROMemories:
```suggestion
/// Error returned when trying to convert the time to two parts.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+}
+
+/// Convert time to parts to be put into Samples.
+///
+/// `timestamp_nanos` is an UTC timestamp as nanoseconds since UNIX epoch.
+///
+/// The first element of the tuple is to be inserted in a Channel that has the label
+/// [`Label::OpaqueGnssTime`].
+/// The second element must be inserted in a Channel right after the channel of the first element and
```
ROMemories: I think these "Channel"s should be `ReadingChannel`s?

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+    fn time_of_fix_delta_nanos(&self) -> Result<i64, GnssTimeExtError> {
+        let sample = get_element_after_marker(self.samples(), Label::OpaqueGnssTime, 1)
+            .ok_or(GnssTimeExtError::InvalidSensor)?;
+
+        Ok(sample.1.value().map_err(GnssTimeExtError::Reading)?.into())
+    }
+}
+
+/// Convert time to parts to be put into Samples.
```
ROMemories:
```suggestion
/// Convert time to parts to be put into [`Samples`].
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+
+// Only implement for Samples
+mod private {
+    use ariel_os_sensors::sensor::Samples;
+
+    pub trait Sealed {}
+    impl Sealed for Samples {}
+}
+
+/// Trait to use to access time information on [`Samples`] coming from a GNSS sensor.
```
ROMemories:
```suggestion
/// Trait to access time information from [`Samples`] coming from a GNSS sensor.
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+                + i128::from(self.time_of_fix_delta_nanos()?),
+        )
+    }
+    /// Returns the nanoseconds part of the time of the last fix.
+    ///
+    /// # Errors
+    ///
+    /// Returns errors if the reading is not compatible with this extension
+    /// or one [`Sample`] returned an error.
+    fn time_of_fix_delta_nanos(&self) -> Result<i64, GnssTimeExtError>;
```
ROMemories: Seems like this could be [aligned with the std](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.subsec_nanos); don't know how I missed that existing name before.
```suggestion
    fn time_of_fix_subsec_nanos(&self) -> Result<i64, GnssTimeExtError>;
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+    let mut peekable = iter.peekable();
+    while let Some((c, _)) = peekable.peek() {
+        if c.label() == marker {
+            break;
+        }
+        peekable.next();
+    }
+    let result = peekable.nth(position);
+
+    // if its not an opaque channel we're doing something wrong
```
ROMemories:
```suggestion
    // if it's not an opaque channel we're doing something wrong
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+
+/// Convert time to parts to be put into Samples.
+///
+/// `timestamp_nanos` is an UTC timestamp as nanoseconds since UNIX epoch.
+///
+/// The first element of the tuple is to be inserted in a Channel that has the label
+/// [`Label::OpaqueGnssTime`].
+/// The second element must be inserted in a Channel right after the channel of the first element and
+/// that has the label [`Label::Opaque`].
+///
```
ROMemories, line 138: An example about using the parts wouldn't hurt, but it is likely better added after landing the first driver that uses this.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
nponsard: I think the link wasn't made because `ariel_os_sensors` wasn't built.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
nponsard: Indeed, I changed it.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
nponsard: I added an simple example

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
ROMemories: Ah they are indeed in scope, it might be because `ariel_os_sensors` is not built *prior* to generating the docs of `ariel_os_sensors_gnss_time_ext` then.

nponsard: I'm struggling to debug this, on my machine doing 

```
cargo +nightly doc -p ariel-os-sensors -p ariel-os-sensors-gnss-time-ext --no-deps --open        
```

Has the correct links, even after deleting the `target` directory.

nponsard:
I can't figure out what's going wrong with the links between `ariel-os-sensors` and `ariel-os-sensors-gnss-time-ext`. I can't reproduce the issue outside of the CI.

I tried this: 

```
docker run -it debian:13
apt update
apt install rustup gcc clang git
rustup toolchain install nightly-2026-01-20 --target riscv32imac-unknown-none-elf --component rust-src --profile minimal --no-self-update
git clone https://github.com/ariel-os/ariel-os

cd ariel-os

git fetch --no-tags --force --update-head-ok 'https://github.com/ariel-os/ariel-os.git' '+647ad06fe0d6ce53351d726e4c2ebe4115058257:refs/commit/647ad06fe0d6ce53351d726e4c2ebe4115058257'

git checkout 647ad06fe0d6ce53351d726e4c2ebe4115058257

RUSTDOCFLAGS='--cfg nightly -Z unstable-options --extern-html-root-url trouble_host=https://docs.embassy.dev/trouble-host/0.5.0' cargo doc --no-deps -p ariel-os -p ariel-os-sensors -p ariel-os-sensors-gnss-time-ext -p coapcore --features "ariel-os-sensors/max-sample-min-count-12,bench, ble, coap, core-affinity, csprng, defmt, dns, executor-thread, external-interrupts, hwrng, i2c, mdns, multicast, net, no-boards, random, ariel-os-coap/doc, sensors, spi, storage, tcp, threading, uart, udp, usb, usb-hid, coapcore/_nightly_docs"


cat target/doc/ariel_os_sensors_gnss_time_ext/fn.convert_datetime_to_parts.html
```

Spot `<p>You would return <a href="../ariel_os_sensors/sensor/samples/struct.Samples.html"`, that means the link worked.

nponsard:
The issues I'm having seem to happen at random when testing locally.

nponsard:
Solved by invoking `cargo doc` another time

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+/// [`Label::OpaqueGnssTime`].
+/// The second element must be inserted in a [`ReadingChannel`] right after the [`ReadingChannel`] of the first element and
+/// that has the label [`Label::Opaque`].
+///
+/// You would return [`Samples`] like this:
+///
+///     // For demonstration purposes, In a normal implementation the Samples should be returned by
+///     // `wait_for_reading()` implemented by the Sensor.
+///     let sensor = SENSOR.init(MockSensor::new(ReadingChannels::from([
+///         ReadingChannel::new(Label::Altitude, 0, ariel_os_sensors::MeasurementUnit::Meter),
```
ROMemories: Am I missing something or is the `Opaque` `ReadingChannel` missing (and this `Altitude` channel extraneous)?

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+    }
+    fn time_of_fix_subsec_nanos(&self) -> Result<i64, GnssTimeExtError> {
+        let sample = get_element_after_marker(self.samples(), Label::OpaqueGnssTime, 1)
+            .ok_or(GnssTimeExtError::InvalidSensor)?;
+
+        Ok(sample.1.value().map_err(GnssTimeExtError::Reading)?.into())
+    }
+}
+
+/// Convert time to parts to be put into [`Samples`].
```
ROMemories: (Can be ignored and done later if I am indeed missing something below):
```suggestion
/// Converts time to parts to be put into [`Samples`].
```

.github/workflows/main.yml
```diff
@@ -316,6 +317,7 @@ jobs:
                 --no-deps \
                 -p ariel-os \
                 -p ariel-os-sensors \
+                -p ariel-os-sensors-gnss-time-ext \
```
ROMemories, line 320: It would be better to keep this aligned with `build-deploy-docs.yml`, but this is better done in a `ci-build:skip` follow-up, in which we also split out the rendering of `ariel-os-sensors` for clarity.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
nponsard: Ah, I copied from the the wrong test... 
I fixed it.

nponsard:
@ROMemories squashed, rebased #1330 on it and it work :tada: 

```
[INFO ] Time of fix: Ok(1772640243) (gnss_sensor_reader gnss-sensor-reader/src/main.rs:63)
[INFO ] Time of fix ns : Ok(1772640243965000000) (gnss_sensor_reader gnss-sensor-reader/src/main.rs:64)
```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,533 @@
...
+}
+
+// To obtain a timestamp with nanoseconds (works only for positive timestamps)
+//  `date --date "2038-01-19T04:04:04.040404Z" -u +%s%N`
+//
+// For a negative timestamp, get the seconds timestamp with `date --date "1956-12-13T20:45:54.1343212Z" -u +%s`,
+// if the sub-second part is non-zero, add 1. Then multiply the seconds part by 1_000_000_000, and add
+// `date --date "1956-12-13T20:45:54.1343212Z" -u +%N` - 1_000_000_000.
+#[cfg(test)]
+mod tests {
```
nponsard: Tests aren't run in CI, working on it.

nponsard: Fixed.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    use ariel_os_sensors::sensor::ReadingChannels;
```
ROMemories: Would fix the fmt error:

```suggestion
    use ariel_os_sensors::sensor::ReadingChannels;

```

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
ROMemories: They do indeed show in the CI logs.

src/ariel-os-sensors-gnss-time-ext/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+    {
+        None
+    } else {
+        result
+    }
+}
+
+#[cfg(any(test, feature = "_test"))]
+/// Mock sensor module for tests.
+/// Has to be public for docs tests.
```
ROMemories: I'd rather keep the `#[cfg]` attribute closer to the item, unless the attribute is about the doc tests themselves (like [`#[expect(clippy::doc_markdown)]`](https://rust-lang.github.io/rust-clippy/master/index.html?search=doc_mar#doc_markdown) would be):
```suggestion
/// Mock sensor module for tests.
/// Has to be public for docs tests.
#[cfg(any(test, feature = "_test"))]
```

ROMemories (commented):
Good to go after these.

PR #1684, opened 2026-01-14
feat(sensors): implement `core::error::Error` on every `*Error` type

ROMemories:
# Description

This makes sure `core::error::Error` is implemented on every type in `ariel-os-sensors` whose name ends with `Error` (and therefore is an error type). That trait was actually only missing on `SampleError`.

## Testing

Tested the following (which doesn't actually rely on the `Error` trait):

```sh
```

More testing should happen in #1683.

## Issues/PRs References

Found when looking at #1683.

## Change Checklist

### DIFF
--- a/src/ariel-os-sensors/src/sample.rs
+++ b/src/ariel-os-sensors/src/sample.rs
@@ -11,6 +11,17 @@ pub enum SampleError {
     ChannelDisabled,
 }
 
+impl core::fmt::Display for SampleError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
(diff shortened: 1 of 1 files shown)


chrysn:
Brief drive-by remark: We might want to consider thiserror for this kind of code. Yeah it's just 9 lines, but with the boilerplate removal afforded by thiserror, it'd be down to 2 lines plus one more derive, with identical results.

(ed: and thiserror is already in our cargo tree).

ROMemories:
> We might want to consider thiserror for this kind of code.

I do myself like `thiserror` quite a lot, but I don't think the proc-macro is worth the cost in this specific crate.

PR #1689, opened 2026-01-14
ci(sensors): run Clippy on the sensor drivers

ROMemories:
# Description

This makes Clippy run in CI on the sensor drivers and fixes Clippy's lints on them.

## How to review this PR

## Testing

Successfully tested the `sensors-debug` example on the st-steval-mkboxpro (which does have the two sensors onboard).

## Change Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -270,6 +270,17 @@ jobs:
             --
             --deny warnings
 
+      # Lint sensor drivers.
+      - name: Clippy on sensor drivers
+        uses: clechasseur/rs-clippy-check@286fd9545eb133adad54b4af34bcb642e266ddd1 # v5
+        with:
+          args: |
+            --locked
+            -p ariel-os-sensor-lis2du12
+            -p ariel-os-sensor-stts22h
+            --
+            --deny warnings
+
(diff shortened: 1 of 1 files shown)


src/sensors/ariel-os-sensor-stts22h/src/lib.rs
```diff
@@ -16,13 +16,13 @@ const PART_NUMBER: &str = "STTS22H";
 #[expect(dead_code)]
 #[derive(Copy, Clone, PartialEq, Eq)]
 enum Register {
-    WhoAmIRegAddr = 0x01,
+    Whoami = 0x01,
```
ROMemories, line 19: In the datasheet of this sensor, the register is named `WHOAMI` (with no underscores), while on the LIS2dU12, it's called `WHO_AM_I`, so the casing follows the datasheets as closely as possible.

anlavandier: I think that there is an argument to be made about how being internally consistent is more important than following datasheets religiously but seeing as other sensors don't even have WHOAMI registers (e.g. the SHT3x), it doesn't really matter.

PR #1690, opened 2026-01-15
docs(tcp): add tcp-client example

nponsard:
# Description

This PR adds a simple TCP client example allowing to test connectivity without needing `hwrng`. 

## Testing

Ran on `rpi-pico-w` as wifi is broken on ESP: 

```
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<password> laze -C examples/tcp-client build -b rpi-pico-w run
```

## Changelog entry

A `tcp-client` example is now available: it makes it easy to test Internet connectivity without requiring a HWRNG.

## Change Checklist

### DIFF
--- a/examples/tcp-client/README.md
+++ b/examples/tcp-client/README.md
@@ -1,0 +1,20 @@
+# tcp-client
+
+## About
+
+This application is testing basic
+[Embassy](https://github.com/embassy-rs/embassy) _networking_ usage with Ariel OS.
+
+## How to run
+
+In this directory, run
+
+    laze build -b rpi-pico-w run
+
+The application will try to connect to [tcpbin.com](https://tcpbin.com/), a simple echo server using TCP.
+
+Look [here](../README.md#networking) for more information about network configuration.
+
+If everything goes well, you should see the server's response:
+
+    [INFO ] txd: Hello world!
--- a/examples/tcp-client/src/main.rs
+++ b/examples/tcp-client/src/main.rs
@@ -1,0 +1,57 @@
+// modification of https://github.com/embassy-rs/embassy/blob/main/examples/nrf9160/src/bin/modem_tcp_client.rs
+
+#![no_main]
+#![no_std]
+
+use core::str::FromStr as _;
+
+use embedded_io_async::Write as _;
+
+use ariel_os::time::Timer;
+use ariel_os::{
+    debug::log::{info, warn},
+    net,
+    reexports::embassy_net,
+    time::Duration,
+};
+
+#[ariel_os::task(autostart)]
+async fn tcp_echo() {
+    let stack = net::network_stack().await.unwrap();
+
+    // Increase the buffer size if you want to send bigger packets.
+    let mut rx_buffer = [0; 256];
+    let mut tx_buffer = [0; 256];
+
+    info!("waiting for interface to come up...");
+    stack.wait_config_up().await;
+
+    loop {
+        let mut socket = embassy_net::tcp::TcpSocket::new(stack, &mut rx_buffer, &mut tx_buffer);
+        socket.set_timeout(Some(Duration::from_secs(10)));
+
+        info!("Connecting...");
+        // Connect to https://tcpbin.com/ without using DNS
+        let host_addr = embassy_net::Ipv4Address::from_str("45.79.112.203").unwrap();
+        if let Err(e) = socket.connect((host_addr, 4242)).await {
+            warn!("connect error: {:?}", e);
+            Timer::after_secs(10).await;
+            continue;
+        }
+        info!("Connected to {:?}", socket.remote_endpoint());
+
+        let msg = b"Hello world!\n";
+        for _ in 0..10 {
+            #[allow(unused_variables, reason = "log macro sometimes doesn't use this")]
+            if let Err(e) = socket.write_all(msg).await {
+                warn!("write error: {:?}", e);
+                break;
+            }
+            // We put trust in the server to return a valid utf8 string
+            info!("txd: {}", core::str::from_utf8(msg).unwrap());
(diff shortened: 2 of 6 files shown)


ROMemories:
This isn't really specific to nRF, is it?

examples/tcp-client/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## How to run
+
+In this directory, run
+
+    laze build -b rpi-pico-w run
+
+The application will try to connect to [tcpbin.com](https://tcpbin.com/), a simple echo server using TCP.
+
+Look [here](../README.md#networking) or more information about network configuration.
```
ROMemories: I do see this comes from the other networking examples:

```suggestion
Look [here](../README.md#networking) for more information about network configuration.
```

examples/tcp-client/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    info!("waiting for interface to come up...");
+    stack.wait_config_up().await;
+
+    loop {
+        let mut socket = embassy_net::tcp::TcpSocket::new(stack, &mut rx_buffer, &mut tx_buffer);
+        socket.set_timeout(Some(Duration::from_secs(10)));
+
+        info!("Connecting...");
+        // Connect to https://tcpbin.com/ without using DNS
```
ROMemories, line 34: Thanks for adding this comment and the explanation in the README, seeing that random address was weird at first.

examples/tcp-client/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+};
+
+#[ariel_os::task(autostart)]
+async fn tcp_echo() {
+    let stack = net::network_stack().await.unwrap();
+
+    let mut rx_buffer = [0; 4096];
+    let mut tx_buffer = [0; 4096];
+
+    info!("waiting for interface to come up...");
```
ROMemories, line 26: Everything below this is indeed copy-pasted from the upstream Embassy example (as it should).

nponsard:
> This isn't really specific to nRF, is it?

This is not, let me rectify that

examples/tcp-client/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+    reexports::embassy_net,
+    time::Duration,
+};
+
+#[ariel_os::task(autostart)]
+async fn tcp_echo() {
+    let stack = net::network_stack().await.unwrap();
+
+    let mut rx_buffer = [0; 4096];
+    let mut tx_buffer = [0; 4096];
```
ROMemories: These buffers can likely be sized down as we've done in #1176 considering how short the byte string we're sending is (this example is for instance extremely close to not linking on the stm32u083c-dk). Something like 256 bytes is likely to be enough (but this really need testing).

examples/tcp-client/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+        info!("Connected to {:?}", socket.remote_endpoint());
+
+        let msg = b"Hello world!\n";
+        for _ in 0..10 {
+            if let Err(e) = socket.write_all(msg).await {
+                warn!("write error: {:?}", e);
+                break;
+            }
+            info!("txd: {}", core::str::from_utf8(msg).unwrap());
```
ROMemories, line 51: Technically that's a server-controllable DoS, but this doesn't matter much for this example.

ROMemories:
Successfully tested on the rpi-pico-w: tcpbin does echo back the string.

examples/tcp-client/README.md
nponsard: Oh, the typo comes from `examples/tcp-echo` XD

examples/tcp-client/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+        if let Err(e) = socket.connect((host_addr, 4242)).await {
+            warn!("connect error: {:?}", e);
+            Timer::after_secs(10).await;
+            continue;
+        }
+        info!("Connected to {:?}", socket.remote_endpoint());
+
+        let msg = b"Hello world!\n";
+        for _ in 0..10 {
+            if let Err(e) = socket.write_all(msg).await {
```
ROMemories, line 46: There's an `unused` warning on `e` because it's only used in a log macro (there's a couple of other occurrences in the codebase already).

nponsard: Does this need to be fixed or is it fine to leave it like that ?

examples/tcp-client/src/main.rs
nponsard: Successfully tested with a buffer of 256.

examples/tcp-client/src/main.rs
ROMemories: Even if that makes it diverge from upstream, I think we shouldn't have warnings, even in examples. Here's another example in the codebase:

https://github.com/ariel-os/ariel-os/blob/8cd2c14af738a78610041261a405d13b1bb987f4/src/ariel-os-threads/src/sync/wait_queue.rs#L80

examples/tcp-client/src/main.rs
```diff
@@ -0,0 +1,57 @@
+// modification of https://github.com/embassy-rs/embassy/blob/main/examples/nrf9160/src/bin/modem_tcp_client.rs
```
ROMemories, line 1: For reference, that's <https://github.com/embassy-rs/embassy/blob/08e89c21b0097777f12bba45d2be9150d268fc75/examples/nrf9160/src/bin/modem_tcp_client.rs> at the moment

nponsard:
Squashed commits, ready to merge if there are no further comments.

PR #1691, opened 2026-01-15
fix(esp): don't force frame pointers on xtensa

kaspar030:
# Description

As suggested by @MabezDev, don't enable force-frame-pointers on xtensa.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -384,8 +384,6 @@ contexts:
       RUSTFLAGS:
         # linkall first
         - -Clink-arg=-Tlinkall.x
-        # this might be needed for backtraces. it is needed for probe-rs.
-        - -C force-frame-pointers
 
   - name: esp32
     parent: esp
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -866,6 +864,8 @@ modules:
         # riscv needs a bit more isr stack.
         isr_stacksize_required_default: "4096"
         RUSTFLAGS:
+          # this might be needed for backtraces. it is needed for probe-rs.
+          - -C force-frame-pointers
```
ROMemories, line 868: Is this required only on ESP's RISC-V-based MCU or on all RISC-V MCUs?

kaspar030: If needed at all, probably on all RISC_V based MCUs. Anyhow, we only support the esp riscv currently. Once we add another kind, we'll have to re-evaluate all those settings.

MabezDev (approved):

PR #1694, opened 2026-01-15
fix(rt): disable esp watchdog in init()

kaspar030:
# Description

It seems like with some output and the initial stack painting, the esp wdt (enabled by the bootloader IIUC) triggers before `esp_hal::init()` disables it. So, disable it early.

## Change Checklist


kaspar030:
not needed actually.

Issue #1195, opened 2025-07-31
esp32s3: setting log level lower than info hangs during init


nponsard:
### Summary

On main, running 
```
laze -C examples/blinky build  -DLOG=trace -b espressif-esp32-s3-devkitc-1 run            
```

The MCU seems to hang, displaying
```
DEBUG ariel_os_rt::startup()
DEBUG ariel-os-rt: ISR stacksize: 32768
Error:   × Broken pipe
```

This doesn't happen on tag v0.2.1

### Hardware/ Setup

Testing on a xiao esp32s3 sense board
Xtensa Rust 1.88.0.0
Ariel OS Commit ce5d056d924d1963006df1fbe3ff48809bdfba14 

### Would you like to work on fixing this bug?

Yes

nponsard:
Using git bisect I found that it should come from one of these commits: 

67dcfd61cf9e620d8589df10c672e223c687edb3
a42af036f283636e7adfa59cb0059d549290cb40
f8ce22648c80de04e6aa0c343d95bbda18598cd3
59d57a1383c80efc98715ee9bd76caa78001628d
631e94cc51b04e88ca63d710acd7214b0d25ae3c
1cf3a4cc2b8b75cfeb961f0118475d708761d106
5c1a2b81e38e5c27df2434700f4164b8a9c51ef6
534ea02491e8c43f8d48ed1f760315b7cda561be

There is some build errors when trying to build between 5c1a2b81e38e5c27df2434700f4164b8a9c51ef6 and 534ea02491e8c43f8d48ed1f760315b7cda561be

kaspar030:
Could you try increasing the ISR stacksize?

nponsard:
> Could you try increasing the ISR stacksize?

I just tried with an ISR stacksize of 40960, 65536 and 131072. I observed the same behavior.

nponsard:
77721c127280292107e48f39107d1b33b85dcfa1 is the last commit where the initialization works with -DLOG=debug

kaspar030:
I think this is related to https://github.com/ariel-os/ariel-os/issues/1438.

nponsard:
> I think this is related to [#1438](https://github.com/ariel-os/ariel-os/issues/1438).

Commenting the line with `stack.repaint();` does indeed solve my issue.

kaspar030:
[...]

PR #1695, opened 2026-01-15
fix(rt): consider Xtensa register spill area when stack painting

kaspar030:
# Description

Xtensa needs a four word register spill area below SP, which previously we carelessly overwrote when painting the stack.

image

Fixed in this PR by not painting those four words.

This really nice catch was debugged and found by @bugadani, thanks a bunch!

## Issues/PRs References

Fixes https://github.com/ariel-os/ariel-os/issues/1195.

## Change Checklist

### DIFF
--- a/src/ariel-os-rt/src/stack.rs
+++ b/src/ariel-os-rt/src/stack.rs
@@ -13,6 +13,15 @@ use crate::arch::sp;
 /// Bytes that's used to pain stacks.
 const STACK_PAINT_COLOR: u8 = 0xCC;
 
+/// Number of bytes to ignore when painting / measuring.
+/// Will only paint from `stack_lowest` to `sp()-STACK_PAINT_IGNORE`.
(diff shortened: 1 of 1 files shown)


nponsard:
This does fix my issue in #1195 !

src/ariel-os-rt/src/stack.rs
nponsard, line 127: Should `STACK_PAINT_IGNORE` also be taken into account here ?

kaspar030: Yeah, let's do it.

kaspar030:
I have also updated the safety comment in `repaint()`.

bugadani:
I don't think this fix needs to be +14kloc 👀

kaspar030:
> I don't think this fix needs to be +14kloc 👀

ow, that was close. :) Thanks!

PR #1697, opened 2026-01-16
fix: set TARGET_CC and TARGET_CFLAGS for our Xtensa targets

anlavandier:
# Description

This PR set `TARGET_CC` and `TARGET_CFLAGS` explicitly for our 3 xtensa targets (`esp32`, `esp32s2` and `esp32s3`).
This means that applications that use C dependencies compiled with the `cc` or `cmake` crates will now correctly compile.
## Testing

Trying to compile the following code (by copying it in any example) should fail on main and work with this PR for all three above targets
```rust
#![no_main]
#![no_std]

use ariel_os::debug::{ExitCode, exit, log::*};
use core::ffi::c_char;
// This is needed for the C code to be compiled. Otherwise the build will fail at the linking stage 
// because the symbol `snprintf` is never defined
extern crate tinyrlibc;

unsafe extern "C" {
    fn snprintf(buf: *mut c_char, len: usize, fmt: *const c_char, ...) -> i32;
}

#[ariel_os::task(autostart)]
async fn main() {
    info!("Hello World!");

    let mut buf: [c_char; 256] = [0; 256];
    let fmt: c_char = 61;

    unsafe {
        snprintf(buf.as_mut_ptr(), 256, &fmt);
    }
    exit(ExitCode::SUCCESS);
}
```
`tinyrlibc = "*"` needs to be added to the `Cargo.toml`. 

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -484,8 +484,8 @@ contexts:
       CARGO_TARGET_PREFIX: CARGO_TARGET_XTENSA_ESP32S3_NONE_ELF
       CC: xtensa-esp32s3-elf-gcc
       CARGO_ENV:
-      - TARGET_CC="${CC}"
-      - TARGET_CFLAGS="${CFLAGS}"
+        - TARGET_CC="${CC}"
+        - TARGET_CFLAGS="${CFLAGS}"
 
   - name: esp32s3fx8
     parent: esp32s3


kaspar030:
So I was a bit quiet on this PR thinking if we could just always export `TARGET_CC`/`TARGET_CFLAGS`, e.g., in the `ariel-os` context. Alas, we don't have that info, and this fixes a real issue, so let's go with it.

kaspar030:
confirmed the mentioned issue still persists, and is still fixed, when rebasing on `main`.

Issue #1678, opened 2026-01-13
bug: esp WiFi issues


kaspar030:
#1328 bumps esp-hal to 1.0.0, which in turn requires much deeper integration with our scheduler.

Somehow, WiFi doesn't work correctly. Often, it just doesn't connect:

```
...
     Finished in 6.68s                                                                                                                                         
[INFO ] IPv4: DOWN (embassy_net embassy-net/src/lib.rs:750)                                                                                                    
[WARN ] esp_wifi_internal_tx(interface, ptr, len) returned an error: 12290 (12290) (esp_radio src/wifi/mod.rs:1785)
[WARN ] esp_wifi_internal_tx(interface, ptr, len) returned an error: 12294 (12294) (esp_radio src/wifi/mod.rs:1785)
[WARN ] esp_wifi_internal_tx(interface, ptr, len) returned an error: 12294 (12294) (esp_radio src/wifi/mod.rs:1785)
[WARN ] esp_wifi_internal_tx(interface, ptr, len) returned an error: 12294 (12294) (esp_radio src/wifi/mod.rs:1785)
[WARN ] esp_wifi_internal_tx(interface, ptr, len) returned an error: 12294 (12294) (esp_radio src/wifi/mod.rs:1785)
...
```

Note the errors:

- '12290' also happens when WiFi is working, it means embassy-net tried to send before the stack is up and is unrelated.
- '12294' though means 'wifi stack is in invalid state', after that, wifi is non-functional.

Hints so far:
- the timer thread starves at priority 2 if the Ariel embassy thread loops
- the Wifi channel of the used AP makes a difference, channel 1 is more likely to connect than a higher one

I suspect this is actually multiple issues.

nponsard:
from the comment on `esp_wifi_internal_tx` in `esp-wifi-sys`:
12290 is `ESP_ERR_WIFI_NOT_STARTED` "WiFi is not started"
12294 is `ESP_ERR_WIFI_STATE` "WiFi internal state is not ready, e.g. WiFi is not started"

But the documentation for the `ESP_ERR_WIFI_STATE` constant is "WiFi internal state error".

kaspar030:
Also, at least on the s3, when using espflash, the initial stack painting needs to be disabled. This is quite a smell. 🙈 

*edit*: fixed in #1695.

nponsard:
This was already an issue before the dependency version bump #1195

kaspar030:
[...]

PR #1698, opened 2026-01-16
fix(esp): pull in and adapt upstream `CompatSemaphore` impl

kaspar030:
# Description

"The upstream Semaphore::take_with_deadline() calls our WaitQueue::wait_until() holding a critical section (through NonReentrantRawMutex), IIUC. The WaitQueue logic doesn't expect that, causing the call to never reschedule the calling thread."

This practically caused the esp-radio-rtos-driver timer thread to never trigger a timer.

Here's my branch where I'm trying to fix this.

I've copied over the upstream semaphore implementation and changed it to directly use our `WaitQueue` instead of going through the wrapper. I've then added a new `timeout::with_deadline()` variant that allows calling custom functions, which can be used to encode the "get semaphore with timeout" logic.

This PR should make WiFi work on Xtensa esp32(-s2, -s3). The RISC-V have some other issue still.

## Testing

My testing rig currently is `CONFIG_WIFI_NETWORK=foo CONFIG_WIFI_NETWORK=bar laze -Cexamples/http-client -b <some-esp-board> run`.

## Issues/PRs References

#1678

## Open Questions

- [x] This seems to work fine on esp32 and esp32s3, but not at all on the c6. investigating. => turns out on esp risc-v there is another issue.

## Change Checklist

### DIFF
--- a/src/ariel-os-esp/src/wifi/esp_wifi/semaphore.rs
+++ b/src/ariel-os-esp/src/wifi/esp_wifi/semaphore.rs
@@ -1,0 +1,402 @@
+// taken from upstream esp_radio_rtos_driver, modified to use Ariel's `WaitQueue` directly.
+// Source: https://github.com/esp-rs/esp-hal/blob/f5a5a408bf117e116fd16485706477352824aa30/esp-radio-rtos-driver/src/semaphore.rs#L450
+// (Apache 2.0/MIT)
+//
+// This is mostly changing `CompatSemaphore::take_with_deadline()` to use Ariel's
+// `wait_until_with_check()`.
+// Upstream version was holding a critical section while initiating
+// `WaitQueue::with_timeout()`, which doesn't work with Ariel's `WaitQueue` implementation.
+// The actual change to `CompatSemaphore::take_with_deadline()` is pretty small. It required moving
+// the waitqueue handle (`waiting`) out of `inner` into the main `CompatSemaphore` struct.
+
+// Unfortunately upstream doesn't provide safety comments.
+#![expect(unsafe_code)]
+
+use alloc::boxed::Box;
+use core::ptr::NonNull;
+
+use ariel_os_debug::log::{debug, trace};
+use ariel_os_threads::sync::WaitQueue;
+use esp_radio_rtos_driver::{
+    ThreadPtr, current_task, now,
+    semaphore::{SemaphoreImplementation, SemaphoreKind, SemaphorePtr},
+};
+use esp_sync::NonReentrantMutex;
+
+// SAFETY: These must match the upstream interface.
+unsafe extern "Rust" {
+    fn esp_rtos_task_priority(task: ThreadPtr) -> u32;
+    fn esp_rtos_set_task_priority(task: ThreadPtr, priority: u32);
+}
+
+/// # Safety
+///
+/// The task pointer must be valid and point to a task that was created using
+/// [`task_create`].
+#[inline]
+unsafe fn task_priority(task: ThreadPtr) -> u32 {
+    // Safety: function defined above must match upstream definition.
+    unsafe { esp_rtos_task_priority(task) }
+}
+
+/// # Safety
+///
+/// The task pointer must be valid and point to a task that was created using
+/// [`task_create`].
+#[inline]
+unsafe fn set_task_priority(task: ThreadPtr, priority: u32) {
+    // Safety: function defined above must match upstream definition.
+    unsafe { esp_rtos_set_task_priority(task, priority) }
+}
+
+enum SemaphoreInner {
+    Counting {
+        current: u32,
+        max: u32,
+    },
+    Mutex {
+        recursive: bool,
+        owner: Option<ThreadPtr>,
+        original_priority: u32,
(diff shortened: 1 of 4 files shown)


src/ariel-os-esp/src/wifi/esp_wifi/semaphore.rs
```diff
@@ -0,0 +1,391 @@
...
+use esp_sync::NonReentrantMutex;
+
+unsafe extern "Rust" {
+    fn esp_rtos_task_priority(task: ThreadPtr) -> u32;
+    fn esp_rtos_set_task_priority(task: ThreadPtr, priority: u32);
+}
+
+#[inline]
+unsafe fn task_priority(task: ThreadPtr) -> u32 {
+    unsafe { esp_rtos_task_priority(task) }
```
ROMemories, line 39:
```suggestion
    // SAFETY: this simply propagates the arguments; this function is just an alias.
    unsafe { esp_rtos_task_priority(task) }
```

src/ariel-os-esp/src/wifi/esp_wifi/semaphore.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+#[inline]
+unsafe fn set_task_priority(task: ThreadPtr, priority: u32) {
+    unsafe { esp_rtos_set_task_priority(task, priority) }
```
ROMemories, line 49:
```suggestion
    // SAFETY: this simply propagates the arguments; this function is just an alias.
    unsafe { esp_rtos_set_task_priority(task, priority) }
```

src/ariel-os-esp/src/wifi/esp_wifi/semaphore.rs
ROMemories, line 1: I've diffed this file against the upstream's version and only reviewed the differences.

src/ariel-os-threads/src/sync/wait_queue.rs
```diff
@@ -66,6 +66,42 @@ impl WaitQueue {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if this is called outside of a thread context.
+    pub fn wait_until_with_check(
+        &self,
+        deadline: embassy_time::Instant,
+        check: impl Fn(CriticalSection<'_>) -> bool,
+    ) -> bool {
+        ariel_os_debug::log::trace!("WaitQueue::wait_until()");
```
ROMemories:
```suggestion
        ariel_os_debug::log::trace!("WaitQueue::wait_until_with_check()");
```

src/ariel-os-threads/src/timeout.rs
```diff
@@ -145,6 +145,66 @@ pub(crate) unsafe fn with_deadline(
...
+    deadline: embassy_time::Instant,
+    check: impl Fn(CriticalSection<'_>) -> bool,
+    wait: impl FnOnce(CriticalSection<'_>),
+    on_timeout: impl FnOnce(CriticalSection<'_>),
+) -> bool {
+    let deadline = deadline.as_ticks();
+    let initial = critical_section::with(|cs| {
+        if check(cs) {
+            ariel_os_debug::log::debug!(
+                "with_deadline: initial check succeeded, early out for deadline={}",
```
ROMemories: Same for the other occurrences below:
```suggestion
                "with_deadline_check: initial check succeeded, early out for deadline={}",
```

src/ariel-os-threads/src/sync/wait_queue.rs
```diff
@@ -66,6 +66,42 @@ impl WaitQueue {
         }
     }
 
+    /// Waits for this [`WaitQueue`] to be notified, with deadline and check fn (blocking).
+    ///
+    /// This function:
+    /// 1. calls `check()`
+    /// 2. if `check()` has returned true, return true
+    /// 3. else, wait on `Self` until `deadline` (or notification)
+    /// 4. call `check()` again, return its result
```
ROMemories:
```suggestion
    /// 1. calls `check()`
    /// 2. if `check()` has returned true, returns true
    /// 3. else, waits on `Self` until `deadline` (or notification)
    /// 4. calls `check()` again, returns its result
```

src/ariel-os-threads/src/sync/wait_queue.rs
```diff
@@ -66,6 +66,42 @@ impl WaitQueue {
...
+            crate::timeout::with_deadline_check(
+                deadline,
+                check,
+                |cs| {
+                    self.wait_cs(cs);
+                },
+                |cs| {
+                    ariel_os_debug::log::trace!("WaitQueue::wait_until_with_check() timeout");
+                    #[expect(unused_unsafe)]
+                    let waiters = unsafe { &mut *self.waiters.get() };
```
ROMemories, line 99: This `unsafe` block could still use a SAFETY comment referring to the critical section.

src/ariel-os-esp/src/wifi/esp_wifi/semaphore.rs
kaspar030: I've added safety comments (copied from upstream) to the functions itself.

PR #1699, opened 2026-01-19
refactor(chips)!: rename `nrf5340` to `nrf5340-app`

jkr-ariel:
# Description

This PR renames the `nrf5340` chip laze context to `nrf5340-app`, to better differentiate between the `nrf5340-net` context.

## Changelog entry

The laze contexts and builders targeting the application core of the nRF5340 chip have been renamed. The chip laze context is now named `nrf5340-app` and the laze builder of the nRF5340-DK board targeting the application core is now `nrf5340dk-app`. This is a breaking change for applications relying on the laze context for feature-gating, and when targeting this specific development kit.

## Change Checklist

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -57,7 +57,7 @@
   - [nRF52832](./chips/nrf52832.md)
   - [nRF52833](./chips/nrf52833.md)
   - [nRF52840](./chips/nrf52840.md)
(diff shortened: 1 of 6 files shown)


ROMemories (commented):
The builder used by the `ci-build:small` label also needs to be renamed. Other than that LGTM, we should be able to merge after that.

Issue #1696, opened 2026-01-16
feat: initial support for nrf54LM20-DK


chrysn:
### Description

Development kits for the nRF54LM20A are now available through Nordic's Early Access Program.

It'd be nice to have support for them in Ariel OS.

### Motivation

As that chip family promises to be more powerful[^1] and cheaper[^2] than the popular nRF52840, I think it has a good chance of becoming a successor.

[^1]: Some BLE 6.0 features, twice the memory, twice the clock rate, high speed rather than just full speed USB, more I2S-ish channels, tamper and glitch detection; mixed power consumption (enhanced 0.9µA vs. 1.5µA for RTC-wakeup-but-no-RAM-retention sleep, but worse 0.7µA vs. 0.4µA off-mode power)
[^2]: Mouser lists them at 2.29/pc @ 1k, over 3.04 for the nRF52840.

Also, this chip is under consideration for the development kit being developed in https://github.com/ariel-os/hophop-open-hardware

### Requirements

This issue is about having basic support -- basically, hello world running, but let's see what else gets 

1. The Hello World should run.
2. GPIOs, I2C and UARTs could run.
3. Let's see whether they changed anything in USB or whether that'll just work.
4. Access to the CRACEN crypto accelerator would be awesome.

### Are you planning to do it yourself in a pull request?

Yes

chrysn:
Exploratory notes, while there's no PR:

Trying probe-rs (main; something around nRF54 was added after 0.30) recognizes the debugger with this rule:

```udev
SUBSYSTEMS=="usb", ATTRS{idVendor}=="1366", ATTRS{idProduct}=="1069", GROUP="plugdev", MODE="0660"
```

probe-rs recognizes the chip well (beware that it won't tell you this in `probe-rs info`!), and has flashing, but [there's something wonky about how it interacts with the on-board debugger](https://github.com/probe-rs/probe-rs/issues/3775) (and by wonky I mean that if you don't set `--verify` you'll be off to wild goose chases all over the place, and we don't have a non-invasive workaround yet, just "patch the source to not do double buffering").

Then, all the tests/examples that I've tried that are plausible did work out (not a lot as I didn't wire up any peripherals), with the notable exception of examples/alloc that trips over an Exception in memset somewhere deep in embedded-alloc's critical section.

Branch is up at https://github.com/chrysn-pull-requests/ariel-os/tree/nrf54, no PR yet.

PR #1701, opened 2026-01-19
chore: update embassy HALs

chrysn:
# Description

For #1696, I need the latest version of embassy-nrf (possibly even main, but at least 0.9). This would also be nice because it'll eventually pull in @nponsard's work on the RNGs for nRF91 and nRF53.

This is currently troublesome, as the only HAL that depends on hal-internal 0.4.0 has been released by embassy is stm32. (So we might actually consider whether we *can* get away with having different (patched) versions of that crate).

## Branches created for this

In our embassy clone at https://github.com/ariel-os/embassy:

* embassy-hal-internal-v0.4.0+ariel-os
* embassy-stm32-v0.5.0+ariel-os
* embassy-nrf-v0.9.0+ariel-os (actually a bit off topic b/c only post 0.9 the hal-internal 0.4 was used; still, good to have it, as cherry-picking the OptionalPeripherals commit needed manual merge conflict work)

## Testing

So far, this doesn't even build -- but at least STM32 gets past the point of not complaining about mismatched dependencies, "just" about no chip feture being selected, probably b/c the stm32 HAL update changed something there.

## Issues/PRs References

Contributes-to: #1696

## Open Questions

* [ ] Will it blend?

## Change Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1987,7 +1987,7 @@ dependencies = [
 [[package]]
 name = "embassy-hal-internal"
 version = "0.3.0"
-source = "git+https://github.com/ariel-os/embassy?rev=0feffeb3c90748112a64c6570a39e1251311accb#0feffeb3c90748112a64c6570a39e1251311accb"
+source = "git+https://github.com/ariel-os/embassy?rev=b292e1e2eaf54804c95e8cdc8bbcb11e03452538#b292e1e2eaf54804c95e8cdc8bbcb11e03452538"
 dependencies = [
  "cortex-m",
  "critical-section",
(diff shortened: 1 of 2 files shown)


chrysn:
It's probably best to put this on hold until the -nrf and -rp HALs (which are the other two that depend on e-hal-internal) have released.

For #1696 (nrF54), there *were* changes after the last e-hal-internal 0.3.0 version (0d60ab316fe3916e44668ca4c98565cf339c6679), but they only affect the SAADC burst mode and the docs.

kaspar030:
I'm wondering if we can patch our forks of -nrf and -rp to use (our fork of) `embassy-hal-internal` 0.4.0? The changes are minimal and mostly additions.

kaspar030:
I wrote some tooling for automating the branch & patch juggling and will try with the recent embassy crate releases.

PR #1703, opened 2026-01-20
docs: do reference release blockers in release checklist

chrysn:
# Description

As discussed in todays' weekly, checking for release critical issues is part of the release process.

We did also discuss how to treat them when "just fix them" is not an issue (rougly "if we can't fix it but still want to release, make a note in the book where that feature is described and downgrade the severity"), but I don't think that those are mature enough to go into the docs.

## Change Checklist

- n/a I have tested and performed a self-review of my own code.


ROMemories:
@chrysn Could you make the CI happy by changing the commit title (and PR title)? I think we need to merge this.

Issue #1686, opened 2026-01-14
Removing `syn` v1 from our tree


ROMemories:
`syn` v2 was released in 2023 and now only a few of our dependencies bring in the v1. Because of the position of `syn` in compilation trees, it should be worthwhile to avoid compiling two versions of it, to improve compilation times.

Therefore we should aim to remove the v1 from our tree: this may involve bumping our dependencies, going upstream to bump transitive dependencies, or switching crates when crates we rely on are unmaintained.

Below is the list of crates that depend on `syn` v1:

- [x] `cbindgen` v0.24.5, imported through `liboscore` v0.2.4: need to go upstream to bump `cbindgen`; liboscore 0.2.6 has updated syn to v2 #1893
- [x] `data-encoding-macro-internal` v0.1.16, imported through `data-encoding-macro` v0.1.18, through `cbor-edn` v0.0.9; done in #1706 by bumping `data-encoding-macro-internal`'s dependencies in our lockfile
- [ ] `num_enum_derive` v0.5.11, imported through `num_enum` v0.5.11, through `cyw43` v0.5.0: need to go upstream to bump `num_enum`; https://github.com/embassy-rs/embassy/pull/5263, need to bump to the next release of `cyw43`
- [x] `proc-macro-error` v1.0.4, imported through `hax-lib-macros` (Git dependency), through `hax-lib` (Git dependency); need to update `hax-lib` in `ariel-os-runqueue`
- [ ] `svgbobdoc` v0.3.0, imported through `rlsf` v0.2.1 through two other crates: need to ask upstream to cut a new release (in progress upstream)
- [x] `usbd-hid-macros` v0.8.2, imported through `usbd-hid`: need to go upstream to bump `syn` (done upstream, and released as `usbd-hid-macros` 0.10.0), and make Embassy HAL crates bump `usbd-hid` to 0.10.0 as well (done, waiting for next release) #1932

ROMemories:
All dependencies have now updated syn to v2 upstream, we are now only waiting for new releases.

PR #1706, opened 2026-01-20
chore: bump `hax-lib` to v0.3.6

ROMemories:
# Description

This bumps `hax-lib` (used only in `ariel-os-runqueue`) from a Git dependency to v0.3.6 from crates.io. The main goal of this is actually to remove the transitive dependency on [the unmaintained `proc-macro-error`](https://rustsec.org/advisories/RUSTSEC-2024-0370), which itself brought in `syn` v1, which #1686 is trying to remove from our dependencies.

After this is in, I'll add a `cargo-deny` ban on `proc-macro-error` so it doesn't come back in our tree (as a transitive dependency).

## Testing

CI should be enough, we're not really using hax currently anyway.

## Issues/PRs References

- Works towards #1686

## Change Checklist

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -106,15 +106,6 @@ audit-as-crates-io = true
 [policy.esp-wifi-sys-esp32s3]
 audit-as-crates-io = true
 
-[policy.hax-lib]
-audit-as-crates-io = true
-
-[policy.hax-lib-macros]
-audit-as-crates-io = true
-
-[policy.hax-lib-macros-types]
-audit-as-crates-io = true
-
 [policy.nrf-mpsl]
 audit-as-crates-io = true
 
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1735,7 +1735,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d162beedaa69905488a8da94f5ac3edb4dd4788b732fadb7bd120b2625c1976"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.114",
```
ROMemories, line 1738: I've just realized that this bumped `syn` to v2 for `data-encoding-macro-internal` (which worked towards #1686), because `data-encoding-macro-internal` allows *both* syn v1 and syn v2 as a dependency.

PR #1707, opened 2026-01-20
feat(boards): add nRF5340 app core builder to `nordic-thingy-91-x` board

jkr-ariel:
# Description

This PR adds the `nordic-thingy-91-x-nrf5340-app` builder to the `nordic-thingy-91-x` board, and updates the support matrix accordingly.

## Issues/PRs References

Follow up to #1699.

## Change Checklist

### DIFF
--- a/book/src/boards/nordic-thingy-91-x.md
+++ b/book/src/boards/nordic-thingy-91-x.md
@@ -35,6 +35,34 @@ laze build -b nordic-thingy-91-x-nrf9151
 |Bluetooth Low Energy|<span title="not available on this piece of hardware">–</span>|
 |Hardware Random Number Generator|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>[^only-available-through-the-cryptocell]|
 |Persistent Storage|<span title="supported">✅</span>|
+### `nordic-thingy-91-x-nrf5340-app`
+
+- **Tier:** 2
+- **Chip:** [nRF5340 application core](../chips/nrf5340-app.md)
+- **Chip Ariel OS Name:** `nrf5340-app`
+
+To target this laze builder, run the following command in the root of your Ariel OS app:
+
+```bash
+laze build -b nordic-thingy-91-x-nrf5340-app
+```
+
+#### Support Matrix
+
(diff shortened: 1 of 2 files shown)


boards/nordic-thingy-91-x.yaml
```diff
@@ -8,6 +8,15 @@ boards:
       button0:
         pin: P0_26
 
+  nordic-thingy-91-x-nrf5340-app:
+    chip: nrf5340-app
+    leds:
+      led0:
+        pin: P0_14
```
jkr-ariel, line 15: According to the [documentation](https://docs.nordicsemi.com/bundle/ug_thingy91x/page/UG/thingy91x/hw_description/leds.html), the pin P0.14 corresponds to the red channel for this LED, we can maybe add the other pins for the blue and green channels in a follow up (this is also true for the nRF9151 LED).

doc/support_matrix.yml
```diff
@@ -826,6 +826,18 @@ builders:
...
 
+  nordic-thingy-91-x-nrf5340-app:
+    chip: nrf5340-app
+    tier: "2"
+    support:
+      wifi:
+        status: not_currently_supported
+        comments:
+          - requires supporting the onboard nRF7002 chip
+      ble: not_available
```
ROMemories: I don't think this is necessary, as it will be inherited correctly from `nrf5340-app`:

```suggestion
```

boards/nordic-thingy-91-x.yaml
ROMemories: Agreed, and thanks for pointing to the docs!

boards/nordic-thingy-91-x.yaml
```diff
@@ -8,6 +8,15 @@ boards:
       button0:
         pin: P0_26
 
+  nordic-thingy-91-x-nrf5340-app:
```
ROMemories, line 11: @nponsard Can you try out these pins in hardware?

ROMemories (commented):
LGTM after removing the extraneous line and testing the pins in hardware.

PR #1712, opened 2026-01-22
feat(boards): add green and blue LED channels to `nordic-thingy-91-x`

jkr-ariel:
# Description

This PR adds the pin information for the green and blue LED channels on the Nordic Thingy 91:X according to the [documentation](https://docs.nordicsemi.com/bundle/ug_thingy91x/page/UG/thingy91x/hw_description/leds.html). Currently, only the red channel is specified.

## Issues/PRs References

Follow to #1707.

## Change Checklist

### DIFF
--- a/boards/nordic-thingy-91-x.yaml
+++ b/boards/nordic-thingy-91-x.yaml
@@ -4,6 +4,13 @@ boards:
     leds:
       led0:
         pin: P0_29
+        color: red
+      led1:
+        pin: P0_31
+        color: green
+      led2:
+        pin: P0_30
+        color: blue
     buttons:
       button0:
         pin: P0_26
@@ -13,6 +20,13 @@ boards:
     leds:
       led0:
         pin: P0_14
+        color: red
+      led1:
+        pin: P0_26
+        color: green
+      led2:
+        pin: P0_15
+        color: blue
     buttons:
       button0:
         pin: P0_24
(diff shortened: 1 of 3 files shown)


boards/nordic-thingy-91-x.yaml
```diff
@@ -4,6 +4,13 @@ boards:
     leds:
       led0:
         pin: P0_29
+        color: red
+      led1:
+        pin: P0_30
+        color: blue
+      led2:
+        pin: P0_31
+        color: green
```
ROMemories: I would prefer to follow the RGB order (especially as that's also the one used in the manufacturer's docs) instead of the pin names, but we don't really have guidelines about that:
```suggestion
        pin: P0_31
        color: green
      led2:
        pin: P0_30
        color: blue
```

boards/nordic-thingy-91-x.yaml
```diff
@@ -13,6 +20,13 @@ boards:
     leds:
       led0:
         pin: P0_14
+        color: red
+      led1:
+        pin: P0_15
+        color: blue
+      led2:
+        pin: P0_26
+        color: green
```
ROMemories:
```suggestion
        pin: P0_26
        color: green
      led2:
        pin: P0_15
        color: blue
```

nponsard:
By modifying the blinky example to change which led (`peripherals.led0`, `peripherals.led1`, `peripherals.led2`) is used and flashing to `nordic-thingy-91-x-nrf5340-app` and `nordic-thingy-91-x-nrf9151` I can confirm that the pins are correct and the color description are correct.

nponsard:
I second @ROMemories' comment about changing the order to red-green-blue .

PR #1713, opened 2026-01-22
fix(esp): riscv task switching

### LINKED ISSUE #1711: bug: esp risc-v context switching broken
#1328 brought in our own context switching for the RISC-V esp32. Unfortunately sth seems broken.

- broken since #1328
- related:
  - #1678 
  - #1127

nponsard:
# Description

This PR rewrites the task switching on risc-v to better handle having a timer interrupt with the same priority.
This aims to fix the unreliable wifi on esp32c6.

## Testing

Tested with the fixes of #1698: 
```
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<password> laze -C examples/http-client build -b espressif-esp32-c6-devkitc-1 -DLOG=info run
```
and
```
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<password> laze -C examples/tcp-client build -b espressif-esp32-c6-devkitc-1 -DLOG=info run
```

On multiple access points with different 2.4Ghz channels.

Tested on esp32-c3 by @ROMemories 

## Issues/PRs References

- Fixes #1711 

## Change Checklist

### DIFF
--- a/src/ariel-os-threads/src/arch/riscv.rs
+++ b/src/ariel-os-threads/src/arch/riscv.rs
@@ -14,6 +14,9 @@ use portable_atomic::Ordering;
 
 use crate::{Arch, SCHEDULER, Thread, cleanup};
 
+const CONFIG_ISR_STACKSIZE: usize =
+    ariel_os_utils::usize_from_env_or!("CONFIG_ISR_STACKSIZE", 2048, "ISR stack size (in bytes)");
+
 static _CURRENT_CTX_PTR: portable_atomic::AtomicPtr<ThreadData> =
     portable_atomic::AtomicPtr::new(core::ptr::null_mut());
 
@@ -63,6 +66,12 @@ pub struct ThreadData {
 impl Arch for Cpu {
     type ThreadData = ThreadData;
     const DEFAULT_THREAD_DATA: Self::ThreadData = default_trap_frame();
+
+    /// Stack size for the idle threads.
+    /// On RISC-V (esp-hal), interrupts don't automatically change which stack they use,
+    /// idle thread stack needs to be sized accordingly.
+    #[cfg(feature = "idle-threads")]
+    const IDLE_THREAD_STACK_SIZE: usize = CONFIG_ISR_STACKSIZE;
 
     /// Triggers software interrupt for the context switch.
     fn schedule() {


nponsard:
With `-C force-frame-pointers`, depending on the toolchain, log level and flasher I get different outcomes. Sometimes it works, sometimes it crashes or hangs:

```
probe-rs 0.31.0 (git commit: 3de1cae)
espflash 4.3.0
```

Tested on `examples/http-client`:

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<password> laze -C examples/http-client/ build -b espressif-esp32-c6-devkitc-1 -DLOG=<log level> [-s probe-rs] [-s nightly] run
```

| toolchain version  | log level | flasher program | outcome                                                                                                                                                                                 |
| ------------------ | --------- | --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1.92.0             | info      | espflash        | works                                                                                                                                                                                   |
| 1.92.0             | debug     | espflash        | works                                                                                                                                                                                   |
| 1.92.0             | trace     | espflash        | output stops (probably crash), last line is always `[DEBUG] with_deadline: initial check failed, calling wait() for deadline=18446744073709551615`                                      |
| 1.92.0             | info      | probe-rs        | works                                                                                                                                                                                   |
| 1.92.0             | debug     | probe-rs        | output stops, last line is always `[DEBUG] ariel-os-embassy::init_task() done`                                                                                                          |
| 1.92.0             | trace     | probe-rs        | output stops, last line is always `[DEBUG] ariel-os-embassy::init_task() done`                                                                                                          |
| nightly-2026-01-20 | info      | espflash        | panic: `panicked at src/ariel-os-threads/src/ensure_once.rs:42:33: RefCell already borrowed`                                                                                            |
| nightly-2026-01-20 | debug     | espflash        | works                                                                                                                                                                                   |
| nightly-2026-01-20 | trace     | espflash        | output stops (probably crash), last line is always `[DEBUG] with_deadline: initial check failed, calling wait() for deadline=18446744073709551615`                                      |
| nightly-2026-01-20 | info      | probe-rs        | works                                                                                                                                                                                   |
| nightly-2026-01-20 | debug     | probe-rs        | `panicked at /home/nponsard/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/defmt-rtt-1.1.0/src/channel.rs:38:31: range start index 1082312328 out of range for slice of length 1` |
| nightly-2026-01-20 | trace     | probe-rs        | crash, probe-rs doesn't give more info                                                                                                                                                  |


This doesn't happen when disabling `-C force-frame-pointers`.

kaspar030:
> With `-C force-frame-pointers`, depending on the toolchain, log level and flasher I get different outcomes. Sometimes it works, sometimes it crashes or hangs:

Those failures do look like memory corruption. It makes some sense that enabling forced frame pointers break things, it does put some more expectations on stack frame handling, which the context switching code obviously meddles with ...

nponsard:
I tried adding something to store the stack frame as expected but it didn't help

nponsard:
Found it ! IDLE_THREAD_STACK_SIZE was too small and leaking into other stuff when interrupts were triggered.

src/ariel-os-threads/src/arch/riscv.rs
```diff
@@ -147,118 +158,76 @@ const fn default_trap_frame() -> ThreadData {
...
-unsafe extern "C" {
-    fn FROM_CPU_INTR0();
-}
-
 global_asm!(
     r#"
 
-    .section .trap.rust, "ax"          // FIXME: is this right ?
-    .globl FROM_CPU_INTR0
+    .section .trap, "ax"          // FIXME: is this right ?
```
nponsard, line 169: Changed to .trap to match what upstream does

ROMemories:
Successfully tested the `http-client` and `udp-echo` (not `tcp-echo`) examples by merging this with #1698 on espressif-esp32-c3-lcdkit and espressif-esp32-c6-devkitc-1. I'll do the rest of the review later.

~~In the meantime, could you add pointers to references/specifications you used (as code comments) to make reviewing easier?~~ (edit: Nevermind, I hadn't seen how close it was upstream; even though of course that wouldn't hurt.)

src/ariel-os-threads/src/arch/mod.rs
```diff
@@ -1,12 +1,16 @@
...
+
 /// Arch-specific implementations for the scheduler.
 pub trait Arch {
     const DEFAULT_THREAD_DATA: Self::ThreadData;
 
-    /// Stack size for the idle threads.
+    /// Stack size for the idle threads. The interrupts will execute in this context.
     #[cfg(feature = "idle-threads")]
-    const IDLE_THREAD_STACK_SIZE: usize = 256;
+    const IDLE_THREAD_STACK_SIZE: usize = CONFIG_ISR_STACKSIZE;
```
ROMemories: I'm not sure I understand this, aren't the idle threads supposed to be doing *nothing*?

nponsard: I should add a feature gate for risc-v here.
On risc-v with how interrupts are handled in esp-hal the stack pointer is not changed when an interrupt interrupts the execution of the code.
The idle thread is explicitly waiting for interrupts to interrupt it.

kaspar030: > The idle thread is explicitly waiting for interrupts to interrupt it.

Well, it is not opting out. The "normal" state (always unless within a critical section) is to allow ISRs to interrupt.

nponsard: I meant that it executes `wfi` that is literally "wait for interrupts"

src/ariel-os-threads/src/arch/mod.rs
```diff
@@ -1,13 +1,22 @@
...
 
     /// Stack size for the idle threads.
-    #[cfg(feature = "idle-threads")]
+    #[cfg(all(feature = "idle-threads", not(context = "riscv")))]
     const IDLE_THREAD_STACK_SIZE: usize = 256;
 
+    /// Stack size for the idle threads.
+    /// On RISC-V interrupts don't automatically change which stack they use and may use this one.
+    #[cfg(all(feature = "idle-threads", context = "riscv"))]
+    const IDLE_THREAD_STACK_SIZE: usize = CONFIG_ISR_STACKSIZE;
```
kaspar030: This can be moved to the `arch/riscv.rs` `impl Arch for Cpu`.

nponsard: Didn't realize this was in the trait, thanks :+1:

src/ariel-os-threads/src/arch/riscv.rs
```diff
@@ -148,117 +160,79 @@ const fn default_trap_frame() -> ThreadData {
...
-
-
-        call {sched}
-
-        // if a1 is null, we need to return to the previous task
-        beqz    a1, restore_stack
+    sys_switch:
+
+        // save some registers that will be used below
+        addi sp, sp, -0x10
```
ROMemories, line 175: Can you add a comment explaining how to compute that value? I suppose this is pretty much the number of `sw` instructions that writes to the stack times four (but then are we not over-reserving one word)?

src/ariel-os-threads/src/arch/riscv.rs
ROMemories, line 30: Can you quickly explain why it's ok for the order of these to be different from upstream's? Does the order not matter as long as we save/restore in a consistent way?

src/ariel-os-threads/src/arch/riscv.rs
```diff
@@ -85,23 +103,17 @@ impl Arch for Cpu {
 
     /// Enable and trigger the appropriate software interrupt.
     fn start_threading() {
```
ROMemories, line 105: For reference (no need to address this in this PR), this method (and maybe the trait itself) should be made `unsafe`, requiring at least the caller to only call this once to start threading.

src/ariel-os-threads/src/arch/riscv.rs
nponsard: The order does not matter, I made the choice to store the registers ordered by their generic id (x14 is a4 for example). There is a neat table showing the name aliases here: https://projectf.io/posts/riscv-cheat-sheet/#rv32-abi-registers

src/ariel-os-threads/src/arch/riscv.rs
nponsard: It was hard to find a good source for it, the stack pointer has to be 16-bytes aligned. 

> The stack grows downwards (towards lower addresses) and the stack pointer shall be aligned to a 128-bit boundary upon procedure entry. 

https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc

ROMemories: I mean the source is the spec itself; but that makes sense, thanks!

PR #1714, opened 2026-01-22
docs(book): remove the pin on `espflash` in the Getting Started

ROMemories:
# Description

The appropriate descriptors are now provided since #1328, this removes the pin on espflash as it is now unnecessary:

https://github.com/ariel-os/ariel-os/blob/2cab2507aa0784e916a131641d9031f703880df1/src/ariel-os-esp/src/lib.rs#L10

## Change Checklist

### DIFF
--- a/book/src/getting-started.md
+++ b/book/src/getting-started.md
@@ -60,7 +60,7 @@ It explains how to compile and run the `hello-word` example to verify your setup
    ```sh
    cargo install espup --locked
    espup install
-   cargo install espflash@3.3.0 --locked
+   cargo install espflash --locked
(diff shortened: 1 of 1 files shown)


ROMemories:
We might want to wait a bit before merging (edit: so keeping as draft for now) to not throw off people following the dev version of the book instead of that of the version they're actually using (as we don't currently explicitly document how to that). I've added the `release-blocker` label to not forget about it.

nponsard (approved):
Ready to be merged when we cut a new release or we have a clear way for users to see the book for the version they want.

Issue #1169, opened 2025-07-21
Metrics monitoring in CI


ROMemories:
There is a few metrics that would be very beneficial to keep track of in CI builds:

- [ ] The binary size of at least a subset of examples (e.g., with the `size`/`size -A` laze task)
- [ ] The BSS size of at least a subset of examples
- [ ] Compilation times, using [`--timings`](https://doc.rust-lang.org/cargo/reference/timings.html)

These would need to be monitored across the supported HALs, and on both stable and nightly.

At least at first these would be reported as PR comments by a bot, ideally with warnings when significant changes occur, similarly to [what it is done for the Rust compiler for build times](https://github.com/rust-lang/rust), with an explicit confirmation required to move forward with the PR.

Being able to compute the changes requires being able to access the previous values, which need to be stored somewhere and then fetched when needed (GitHub actions do not provide facilities for this).

kaspar030:
This probably has multiple parts:

1. Collection of data

We have some experience with this in RIOT OS. There, each "app for board" pair is one build, and we parse the whole output for lines starting with "{", and if those lines parse as JSON, we merge it into a metrics json file for that build.
The metrics of all builds then get combined into a single large JSON file that can be downloaded.
E.g., https://ci.riot-os.org/results/2dc3563362a145039b49c3b76fb3e386/metrics.json.

Collection any single-line JSON output is very convenient, e.g., the test scripts that run on hardware can also just output e.g., `{"executor_thread_stack_max_usage": 12344}`, and that becomes available for metrics.

TBH I'm not sure how to do this with laze. For tasks, laze could watch the called processes stdout itself, but for commands run by ninja, that's not trivial. Maybe a helper utility that calls a process, passes through signals and stdio, and parses out single line json metrics would help. That could implemented in a way that allows us to plug in the guts into laze, and in addition use it in our laze rules.

3. Storing data

Github has "artifacts", maybe we can use those for storing metrics. We could also host a redis or any k/v on tatooine or fly and give access to our github actions.

4. Handling the data

Assuming we have collected a bunch of JSONs or metrics in a db, we need to make sense of them.
Some (Rust) utility looking for regressions etc. needs to be written.

5. Reporting

Here I think I'd prefer a solution that does not run in a GHA workflow, but rather implement a simple bot (in Rust of course) that listens on github web hooks and does the (bot) logic in a proper language. We do have a server to host this, and probably the smallest fly.io machine is also sufficient.
[...]

PR #1715, opened 2026-01-23
ci: introduce job to monitor binary size

jkr-ariel:
# Description

This PR adds a CI job in the "Build" workflow to track binary size from examples by posting a comment on PRs with the output of the `llvm-size` command. As this is more of a proof of concept to test waters, only the `blinky` example with the `nrf52840dk` builder is checked.

## Issues/PRs References

Linked to #1169.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -363,6 +363,69 @@ jobs:
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
           # Build the udp-echo example with static IPv6
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6 -s network-config-ipv6-static
+
+      # Arbitrarily choose an example binary to check its size in the next step.
+      # It is not possible to choose the same binary across multiple builders
+      # because not everything is present in the 1/12 partition.
+      - name: Select ELF to be used for size info
+        if: steps.should-skip.outputs.result != 'true' && github.event_name != 'schedule' && matrix.partition == '1/12'
+        run: |
+          case "${LABEL}" in
+            'ci-build:esp')
+              echo "EXAMPLE_PATH=build/bin/espressif-esp32-s3-devkitc-1/cargo/xtensa-esp32s3-none-elf/release/tcp-echo" >> "$GITHUB_ENV"
+              ;;
+            'ci-build:full')
+              echo "EXAMPLE_PATH=build/bin/nrf52840dk/cargo/thumbv7em-none-eabihf/release/blinky" >> "$GITHUB_ENV"
+              ;;
+            'ci-build:native')
+              echo "EXAMPLE_PATH=build/bin/native/cargo/x86_64-unknown-linux-gnu/release/example-random" >> "$GITHUB_ENV"
+              ;;
+            'ci-build:nrf')
+              echo "EXAMPLE_PATH=build/bin/nrf52840dk/cargo/thumbv7em-none-eabihf/release/blinky" >> "$GITHUB_ENV"
+              ;;
+            'ci-build:rp')
+              echo "EXAMPLE_PATH=build/bin/rpi-pico2/cargo/thumbv8m.main-none-eabihf/release/tcp-echo" >> "$GITHUB_ENV"
+              ;;
+            'ci-build:small')
+              echo "EXAMPLE_PATH=build/bin/nrf52840dk/cargo/thumbv7em-none-eabihf/release/blinky" >> "$GITHUB_ENV"
+              ;;
+            'ci-build:stm32')
+              echo "EXAMPLE_PATH=build/bin/st-nucleo-wb55/cargo/thumbv7em-none-eabihf/release/udp-echo" >> "$GITHUB_ENV"
+              ;;
+            *)
+              # Can happen if `check-labels` finds no labels but still completes
+              # successfully because of a bug.
+              exit 1
+              ;;
+          esac
+        env:
+          LABEL: ${{ needs.check-labels.outputs.label }}
+
+      - name: Get size of example binary
+        if: steps.should-skip.outputs.result != 'true' && github.event_name != 'schedule' && matrix.partition == '1/12'
+        id: size
+        run: |
+          echo "${EXAMPLE_PATH}" | awk -F/ '{print "EXAMPLE_BUILDER="$3}' >> "$GITHUB_ENV"
+          echo "${EXAMPLE_PATH}" | awk -F/ '{print "EXAMPLE_BINARY="$NF}' >> "$GITHUB_ENV"
+          sudo apt update
+          sudo apt install llvm -y
+          {
+            echo 'binsize<<EOF'
+            llvm-size -x "${EXAMPLE_PATH}"
+            echo 'EOF'
+          } | tee -a "$GITHUB_OUTPUT"
+
+      # - name: Comment binary size info on PR
+        # if: steps.should-skip.outputs.result != 'true' && github.event_name != 'schedule' && matrix.partition == '1/12'
+        # uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2
+        # with:
+          # message: |
+            # <!-- binary size -->
+            # Binary size of the `${{ env.EXAMPLE_BINARY }}` example on the `${{ env.EXAMPLE_BUILDER }}` builder:
+            # ```
+            # ${{ steps.size.outputs.binsize }}
+            # ```
+
       # cleanup previous build artifacts. do in background (no need to wait).
       - name: "cleanup stable build artifacts"
         if: github.event_name == 'schedule' && steps.should-skip.outputs.result != 'true'


.github/workflows/build.yml
```diff
@@ -219,7 +219,7 @@ jobs:
           targets: riscv32imc-unknown-none-elf,riscv32imac-unknown-none-elf,thumbv6m-none-eabi,thumbv7m-none-eabi,thumbv7em-none-eabi,thumbv7em-none-eabihf,thumbv8m.main-none-eabi,thumbv8m.main-none-eabihf
           # rust-src: Used for -Zbuild-std.
           # rustfmt: Used by bindgen for liboscore
-          components: rust-src, rustfmt
+          components: rust-src, rustfmt, llvm-tools-preview
```
ROMemories: This might be enough (and might help fix the CI failure)?
```suggestion
          components: rust-src, rustfmt, llvm-tools
```

ROMemories:
*Could* be a `PATH` issue.

jkr-ariel:
Asking for merge even though it is not completely fleshed out in order to be able to test the job easily once it is integrated on main.

nponsard:
You should be able to try the CI on your fork, set main with the changes you want, then open a pull request on your own fork, targeting the main branch of your fork.

nponsard:
Also I'm concerned that you chose to have this on the `build` CI file, those jobs take the longest to run and would run on a different environment when we enable the self hosted runners.

nponsard:
This will also run 20 times as it's using a matrix strategy with 20 variants, I don't think that what you want.

jkr-ariel:
> You should be able to try the CI on your fork, set main with the changes you want, then open a pull request on your own fork, targeting the main branch of your fork.

For some reason, Github ignores workflows that require write permissions for forks. Not sure why, even though I have the setting as read/write on my repo.

.github/workflows/binary-size.yml
```diff
@@ -0,0 +1,36 @@
...
+
+# Execute this workflow after the "Build" workflow has completed.
+on:
+  workflow_run:
+    workflows: ["Build"]
+    types: [completed]
+
+# commenting needs these
+permissions:
+  pull-request: write
```
nponsard: Seems like you're using the wrong key: https://github.com/jkr-ariel/ariel-os/actions/runs/21828713192

[it should be `pull-requests`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions)

```suggestion
  pull-request: write
  pull-requests: write
```

Try with this fix on your fork

jkr-ariel: After the fix, the error message about the unknown key disapeared, but the workflow is still ignored...

.github/workflows/binary-size.yml
```diff
@@ -0,0 +1,37 @@
+name: Check binary size
+
+# Execute this workflow after the "Build" workflow has completed.
+on:
+  workflow_run:
+    workflows: ["Build"]
+    types: [completed]
+
+# commenting needs these
+permissions:
```
ROMemories: [Can you move the `pull-requests: write`](https://docs.zizmor.sh/audits/#excessive-permissions) to the job(s) that require it (likely only the one commenting)?
```suggestion
permissions: {}
```

.github/workflows/binary-size.yml
```diff
@@ -0,0 +1,37 @@
...
+  pull-requests: write
+
+jobs:
+  build:
+    # Only run this job if the Build workflow completed successfully.
+    if: github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch'
```
ROMemories: I might be misunderstanding how this interacts, but I think this would require having `workflow_dispatch` as a trigger type?
```suggestion
    if: github.event.workflow_run.conclusion == 'success'
```

.github/workflows/build.yml
```diff
@@ -287,6 +287,19 @@ jobs:
         run: |
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --select nightly --disable xtensa
 
+      # Upload the binary used by the binary-size workflow to keep track of
+      # binary size in the PR.
+      # The condition below assures that we are only uploading the binary once.
+      # The `blinky` for nrf52840dk example was choosen arbitrarily (because
```
ROMemories: Seems like `typos` is actually not linting `.github`:
```suggestion
      # The `blinky` for nrf52840dk example was chosen arbitrarily (because
```

.github/workflows/build.yml
```diff
@@ -287,6 +287,19 @@ jobs:
...
+      # it is present in the 1/20 partition).
+      # In the future, we'll want to check multiple binaries with multiple builders.
+      - name: Upload example binary for size workflow
+        if: github.event_name != 'schedule' && matrix.partition == '1/20'
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        with:
+          name: example
+          path: build/bin/nrf52840dk/blinky/blinky.elf
```
ROMemories: I think the existence of this file depends on which `ci-build:*` label is attached, doesn't it? I suppose it's fine to only run this on `ci-build:full` or `ci-build:nrf` or `ci-build:small` for now.

jkr-ariel: Yes, I think we might need to add a bit of logic to chose a target binary depending on which build label is attached. But that can be done in a follow up.

ROMemories: It seems [the Action indeed does not fail when the path is not present](https://github.com/actions/upload-artifact?tab=readme-ov-file#customization-if-no-files-are-found) (which is what I was concerned of) so I think it's fine.

.github/workflows/binary-size.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Download example binary
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
+        with:
+          name: example
+          run-id: ${{ github.event.workflow_run.id }}
+
+      - name: Get size of example binary
+        id: size
+        run: echo binsize="$(llvm-size example)" | tee -a "$GITHUB_OUTPUT"
```
ROMemories: Good idea to also print it in the logs, should make debugging easier 👍

ROMemories (approved):

nponsard:
I did some tinkering, the workflows that have `workflow_run` as trigger only run the version from the main branch. You can test on your own fork by changing the workflow on the main branch of your fork. 

Example run: https://github.com/nponsard/ariel-os/actions/runs/21902129268/job/63232850131

.github/workflows/binary-size.yml
jkr-ariel: That's correct. It was just something I forgot to remove.

ROMemories:
@jkr-ariel It would indeed be useful you do a round of testing before actually merging this.

nponsard:
One observation is that if the run is cached, the artifact file isn't created so it's not uploaded and the size workflow cannot find it.

Also if you just re-run only the "Final Results" job, it will count as a new workflow run that doesn't have the artifact.

nponsard:
Have you considered building the binaries you want to analyze directly in the workflow? It may simplify things.

ROMemories:
> Have you considered building the binaries you want to analyze directly in the workflow? It may simplify things.

Thing is, this would likely eventually result in duplicating the Build workflow, as we later increase the number of applications/builders we monitor. Not exactly sure what the best option is here. I think we just need to start with *something*, that doesn't make Build fail on main, and iterate on that as we learn from it.

jkr-ariel:
The approach I wanted to use with the `upload-artifact` and `download-artifact` actions did not work, not sure why. Instead,  there are two new steps in the Build workflow that take the size of the `blinky` example on the nRF52840-DK and comment it on the PR. This feels more "hacky" that the `upload-artifact` option but at least it works...

[Here](https://github.com/jkr-ariel/ariel-os/pull/24#issuecomment-4134169990) is an example on my fork

.github/workflows/build.yml
```diff
@@ -355,6 +358,28 @@ jobs:
...
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }}
 
+      - name: Get size of example binary
+        if: steps.should-skip.outputs.result != 'true' && github.event_name != 'schedule' && matrix.partition == '1/12'
+        id: size
+        run: |
+          sudo apt install llvm
+          {
+            echo 'binsize<<EOF'
+            llvm-size -x build/bin/nrf52840dk/cargo/thumbv7em-none-eabihf/release/blinky
```
ROMemories: If the build label isn't `ci-build:nrf`, doesn't this file exist only because of caching? I'm concerned the job could start failing, or return outdated/inconsistent numbers.

jkr-ariel: I think this should also be built with the `ci-build:small` label. Ideally, we would choose a subset of binaries depending on the build label.

jkr-ariel: Confirmed by looking into the ninja file generated by the laze command used in the workflow (for the 1/12 partition):
```bash
$ CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:1/12 -G >/dev/null && awk '/^build/ {print $2}' build/build-global.ninja | grep blinky
[...]
build/bin/nrf52840dk/cargo/thumbv7em-none-eabihf/release/blinky:
[...]
```

Also, the `nrf52840dk` builder is present in the `ci-build:small` label so it should be fine.

ROMemories: Sure, but couldn't that break if the label were `ci-build:rp` and the ELF wasn't in the cache for whatever reason?

jkr-ariel: > Sure, but couldn't that break if the label were ci-build:rp and the ELF wasn't in the cache for whatever reason?

To avoid this I added a step to select an arbitrary ELF based on the build label.

.github/workflows/build.yml
```diff
@@ -363,6 +366,68 @@ jobs:
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
           # Build the udp-echo example with static IPv6
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6 -s network-config-ipv6-static
+
+      # Arbitrarily choose an exmaple binary to check its size in the next step.
```
```suggestion
      # Arbitrarily choose an example binary to check its size in the next step.
```

ROMemories:
Can you try to rebase on the current main? I think that would fix the CI failure, but I'm not sure why it's happening.

jkr-ariel:
I think the [CI failure](https://github.com/ariel-os/ariel-os/actions/runs/24555907650/job/71792307622?pr=1715) is expected? It should not appear once this is on main.

nponsard:
> I think the [CI failure](https://github.com/ariel-os/ariel-os/actions/runs/24555907650/job/71792307622?pr=1715) is expected? It should not appear once this is on main.

If this also fails in the merge queue we won't be able to merge it.

jkr-ariel:
Changed the `on` event to `pull_request_target` which should fix the CI failure.

ROMemories (commented):
> Changed the `on` event to `pull_request_target` which should fix the CI failure.

I don't think that should be necessary for that specific failure, I don't think we should be using `pull_request_target` for the Build workflow. It's still unclear to me why that failure happened.

.github/workflows/build.yml
```diff
@@ -363,6 +366,69 @@ jobs:
...
+          echo "${EXAMPLE_PATH}" | awk -F/ '{print "EXAMPLE_BINARY="$NF}' >> "$GITHUB_ENV"
+          sudo apt update
+          sudo apt install llvm -y
+          {
+            echo 'binsize<<EOF'
+            llvm-size -x "${EXAMPLE_PATH}"
+            echo 'EOF'
+          } | tee -a "$GITHUB_OUTPUT"
+
+      - name: Comment binary size info on PR
```
ROMemories: The job failed in CI for the 1/12 partition because `pull_request` is indeed not enough for posting comments on the PR. However, I also have qualms using `pull_request_target` on the whole Build workflow because [the security implications are complicated](https://docs.zizmor.sh/audits/#dangerous-triggers). So to unblock this I propose we remove the step (and the permission) that posts the comment, and keep only the size measurement and it being printed to the CI logs. That will already be useful to test those in production, and we'll reconsider how we display those later.

.github/workflows/build.yml
```diff
@@ -111,6 +111,9 @@ jobs:
     # Ignore the PR labels in nightly builds.
     if: ${{ github.event_name == 'schedule' || needs.check-labels.outputs.label != 'ci-build:skip' }}
 
+    permissions:
+      pull-requests: write
+
```
ROMemories: The extra permission is only needed to post the comment, right? We shouldn't have overly broad permissions:


```suggestion
```

ROMemories (commented):

ROMemories (approved):

Issue #1716, opened 2026-01-23
re-enable esp32 multi-core support


kaspar030:
The esp-hal 1.0.0 peripheral instances are `!Send` to model that they don't support being migrated between MCU cores.
Our abstraction requires peripherals to be Send.

As a workaround, we've disabled multi-core support on the Xtensa esp32 and marked the esp-hal peripherals `Send` in our fork since #1328.

This is unfortunate, we need to re-enable multi-core support.

PR #1717, opened 2026-01-23
docs(book): clarify current esp32-s3 multicore support status

kaspar030:
# Description

Represent #1716 in book.

## Change Checklist

### DIFF
--- a/book/src/multithreading.md
+++ b/book/src/multithreading.md
@@ -18,7 +18,6 @@ The maximum number of threads is defined by the [`THREAD_COUNT`][max-thread-coun
 ### Multicore Support
 
 Ariel OS currently supports symmetric multiprocessing (SMP) on the following MCUs:
-  - ESP32-S3
   - RP2040
   - RP235XA
 
@@ -27,6 +26,13 @@ To disable multicore, disable the `multi-core` [laze module][laze-modules-book].
 
 > [!NOTE]
 > Porting single-core applications to support multicore requires no changes to them.
+
+> [!WARNING]
+> Ariel OS used to support multicore on ESP32-S3.
+> This is currently disabled because peripheral instances from esp-hal have to be pinned
+> to the core they've been initialized on, which Ariel OS's abstractions currently cannot
+> model.
+> See [#1716](https://github.com/ariel-os/ariel-os/issues/1716).
 
 ### Priority Scheduling
 


book/src/multithreading.md
```diff
@@ -28,6 +27,9 @@ To disable multicore, disable the `multi-core` [laze module][laze-modules-book].
 > [!NOTE]
 > Porting single-core applications to support multicore requires no changes to them.
 
+> [!NOTE]
```
bergzand:
```suggestion
> [!WARNING]
```
This should be a `!WARNING` or maybe even a `!CAUTION`.
`!NOTE`s are for General information or additional context. `!WARNING`s for critical information that highlights a potential risk. `!CAUTION`s for information about potential issues that require caution.

book/src/multithreading.md
```diff
@@ -28,6 +27,9 @@ To disable multicore, disable the `multi-core` [laze module][laze-modules-book].
...
+> Ariel OS used to support multicore on ESP32-S3. This is currently disabled, see [#1716](https://github.com/ariel-os/ariel-os/issues/1716).
```
bergzand: Saves readers a click. Feel free to rephrase

```suggestion
> Ariel OS used to support multicore on ESP32-S3.
> This is currently disabled because peripherals from esp-hal cannot be used from multiple cores while Ariel OS requires this.
> See [#1716](https://github.com/ariel-os/ariel-os/issues/1716) for details and the current status.
```

kaspar030: I was intentionally vague here - this puts the blame on upstream, which it not necessarily is. If our abstraction would accept `!Send` peripherals and handle pinning, we could support esp-hal multicore.

bergzand: Makes sense, how about 
> This is currently disabled because Ariel OS is current unable to pin used peripherals to a single core, as required by the esp32-s3 hardware.

?

kaspar030: > This is currently disabled because Ariel OS is current unable to pin used peripherals to a single core, as required by the esp32-s3 hardware.

better, how about what I pushed? adds "cannot model this", because it is possible to pin threads in ariel, just not without breaking the whole api for the periph abstractions.

bergzand: > better, how about what I pushed? adds "cannot model this", because it is possible to pin threads in ariel, just not without breaking the whole api for the periph abstractions.

Perfect, thanks!

PR #1720, opened 2026-01-26
build: make the MSRV 1.91 and bump the toolchain

ROMemories:
# Description

This bumps the MSRV to 1.91.
This is in particular required for #1653 to be able to compile the latest `reqwest` version, even though it is only used in an example.

## Testing

- I've grepped for `1.90` in the codebase, the other occurences are from sensor-related crates whose MSRV we don't want to bump.
- Successfully tested the `log` example on nrf52840dk and on espressif-esp32-s3-devkitc-1.

## Issues/PRs References

- The previous MSRV bump was #1593 for reference.

## Changelog entry

Ariel OS's MSRV is now 1.91. `laze build install-toolchain` can be used to update the toolchains.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -214,7 +214,7 @@ jobs:
       - name: Install stable toolchain
         if: steps.should-skip.outputs.result != 'true'
         # configure MSRV here:
-        uses: dtolnay/rust-toolchain@1.90
+        uses: dtolnay/rust-toolchain@1.91.1
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -226,7 +226,7 @@ jobs:
         # No tag yet after https://github.com/esp-rs/xtensa-toolchain/pull/38 (will probably be @v1.5.5)
         uses: esp-rs/xtensa-toolchain@main
         with:
-          version: "1.90.0.0"
+          version: "1.91.1.0"
```
ROMemories, line 229: This release still being as a pre-release upstream I'm not sure this is going to work.

ROMemories: It actually doesn't seem to be an issue; the CI logs looks like that version is getting installed as expected:

```
[info]: Installing Xtensa Rust 1.91.1.0 toolchain
```

nponsard: 1.92 xtensa seems to be stable now too !

PR #1725, opened 2026-01-27
feat(gen-book): `update-book` laze tasks removes chip and board pages

jkr-ariel:
# Description

This PR adds a `rm` command to remove all files in `book/src/boards` and `book/src/chips` before generating the new files.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1857,6 +1857,7 @@ builders:
       update-book:
         build: false
         cmd:
+          - rm -f "${relroot}/book/src/boards/*" "${relroot}/book/src/chips/*"
           - doc/gen_book.rs summary doc/support_matrix.yml book/src/SUMMARY.md --template-path book/templates/SUMMARY.md.tmpl
           - doc/gen_book.rs board-pages doc/support_matrix.yml book/src/boards/ --template-path book/templates/board-page.md.tmpl
           - doc/gen_book.rs board-index doc/support_matrix.yml book/src/boards/index.md --template-path book/templates/board-index.md.tmpl


laze-project.yml
```diff
@@ -1857,6 +1857,7 @@ builders:
       update-book:
         build: false
         cmd:
+          - rm -f book/src/boards/* book/src/chips/*
```
ROMemories: This should somehow refer to the "root dir" explicitly to be sure what files we're deleting, as there is no guarantee that the task will be executed at the project's root.

jkr-ariel: Added `${relroot}` before the relative path. However when running `laze build update-book` inside a subfolder, the command is not executed.

```
$ cd book
$ laze -v build update-book
laze: jobserver not inherited
laze: project root: /home/jkr/src/ariel-os relpath: book project_file: laze-project.yml
laze: configured jobserver with limit 22
laze: building all for all
laze: reading cache: local paths don't match
laze: parsing 77 files took 4.359519ms
laze: stat'ing 77 files took 83.142µs
configured 0 builds (took 2.08416ms).
laze: writing cache took 83.282µs.
laze: error: no matching target for task "update-book" found.
```

ROMemories: @kaspar030 Is there another way of doing that? What's the difference between [`root`](https://kaspar030.github.io/laze/dev/reference/variables.html#root) and `relroot`?

laze-project.yml
```diff
@@ -1866,6 +1857,7 @@ builders:
       update-book:
         build: false
         cmd:
+          - rm -f ${relroot}/book/src/boards/* ${relroot}/book/src/chips/*
```
ROMemories: @kaspar030 Also, in which shell is this being executed? Anything to be aware of when using wildcards here (apart from exceeding the maximum number of arguments)?

ROMemories: These really should be quoted, `relroot` could contain spaces:

```suggestion
          - rm -f "${relroot}/book/src/boards/*" "${relroot}/book/src/chips/*"
```

laze-project.yml
kaspar030: `${root}` the relative to a possibly imported project root. contains just `.` for the main project, and e.g., `build/imports/ariel-os` when building from our templates, when `${root}` was used in an imported ariel-os build file.

`${relroot}` contains the relative path to the project root from the file `${relroot}` was used in. E.g.. if `crates/foo/laze.yml` contains a module that has a variable `mypath: ${relroot}/common/foo.c`, that evaluates to `../../common/foo.c`.

In this case, `${root}/book/src/boards/*` should do the right thing.
The task is not found unless `--global` is used. Even in that case, all commands get executed with the project root as cwd. But without the `${root}` prefix, when running `laze build -g update-book` in an oot laze project that imports ariel-os, the task would try to rm `book/...` relative to the oot root and not relative to the ariel-os import root. (whether that task should be exported in that case is a different question).

ROMemories:
Closing: we'll instead tackle this by deleting the contents of `book/src/chips` and `book/src/boards` using the Cargo script instead of using the easy-to-misuse `rm`.

PR #1730, opened 2026-01-28
refactor(esp): keep the app descriptor static private

ROMemories:
# Description

This moves the `esp_app_desc!()` macro call introduced by #1328 into a private module to keep the generated static private.

## How to review this PR

The `ESP_APP_DESC` item no longer shows up in the docs of `ariel_os_esp` (accessible through `ariel_os::hal`).

## Testing

Successfully tested on espressif-esp32-c6-devkitc-1 and espressif-esp32-s3-devkitc-1 by flashing the `log` example. Removing the macro call does show an error:

```
ESP-IDF App Descriptor […] missing in your`esp-hal` application.
```

## Change Checklist

### DIFF
--- a/src/ariel-os-esp/src/lib.rs
+++ b/src/ariel-os-esp/src/lib.rs
@@ -7,7 +7,9 @@
 #[cfg(feature = "wifi")]
 extern crate alloc;
 
-esp_bootloader_esp_idf::esp_app_desc!();
+mod app_desc {
+    esp_bootloader_esp_idf::esp_app_desc!();
+}
 
 pub mod gpio;
 


nponsard:
Should there be a comment explaining why we put the macro invocation in a module ?

ROMemories:
> Should there be a comment explaining why we put the macro invocation in a module ?

That wouldn't hurt but I think it's a somewhat common pattern for controlling visibility. Given this is a release blocker, I'm merging now, and we can add the comment when we revisit this (or in a `ci-build:skip` follow-up).

PR #1738, opened 2026-01-29
ci: add a CI security check workflow using zizmor

### LINKED ISSUE #1705: Integrating zizmor in CI
> `zizmor` is a static analysis tool for GitHub Actions.

Integrating [zizmor](https://github.com/zizmorcore/zizmor) in our CI would be useful to help find potential security issues with our CI configuration.

jkr-ariel:
# Description

This PR adds a CI workflow that checks for security issues in our CI using [zizmor](https://docs.zizmor.sh/), a static analysis tool for Github Actions.

## Issues/PRs References

Closes #1705.

## Changelog Entry

CI workflows are now checked for security issues with [zizmor](https://docs.zizmor.sh/), a static analysis tool for Github Actions. 

## Change Checklist

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -5,7 +5,7 @@ permissions:
   pull-requests: write
 
 on:
-  workflow_run:
+  workflow_run: # zizmor: ignore[dangerous-triggers]
     workflows: ["Deploy docs"]
     types:
       - completed
--- a/.github/workflows/dependencies-action.yml
+++ b/.github/workflows/dependencies-action.yml
@@ -1,7 +1,7 @@
 name: PR dependencies
 
 on:
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
     types: [opened, edited, closed, reopened]
   pull_request:
     types: [synchronize]
--- a/.github/workflows/matrix-bot.yml
+++ b/.github/workflows/matrix-bot.yml
@@ -1,6 +1,6 @@
 name: Matrix bot
 on:
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
     types: [opened, ready_for_review, closed]
 
 permissions: {}


ROMemories:
You could run zizmor locally and start fixing the issues it finds. Having one commit per lint would make it easier to review.

chrysn:
This adds a lot of explicit values.

Are there actual security concerns covered by those, or just blanket "but there are permissions associated with that token so we need to minimize" habits that the lint authors enforce independently of whether there's actual value in limiting them (say, because the token has no value in terms of harming the project, but merely serves to GitHub to bypass their own overzealous rate limiting)?

I'm fine with anything that protects project secrets, but my impression is that GH actions just have bad defaults, and when *then* all our tokens do is get read access on public parts (and all our inputs *are* public) and are only worth protecting to GitHub for their own resource protection … why put that on everyone who edits the already complex CI files.

.github/workflows/build.yml
```diff
@@ -214,7 +222,7 @@ jobs:
       - name: Install stable toolchain
         if: steps.should-skip.outputs.result != 'true'
         # configure MSRV here:
-        uses: dtolnay/rust-toolchain@1.91.1
+        uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master
```
kaspar030, line 226: this was not master before

kaspar030:
> This adds a lot of explicit values.

Do you mean the commit pinning or the assigned permissions?

chrysn:
I mean `persist-credentials: false`, esp. in contexts when all we hand out are read permissions.

kaspar030:
> I mean `persist-credentials: false`, esp. in contexts when all we hand out are read permissions.

I see. Does zizmor do suggestions?

jkr-ariel:
> > I mean `persist-credentials: false`, esp. in contexts when all we hand out are read permissions.
> 
> I see. Does zizmor do suggestions?

Each zizmor lint is documented, for instance [the one we are talking about](https://docs.zizmor.sh/audits/#artipacked).

In this particular case, I think we can safely ignore this lint becase we never use `upload-artifact` on the whole repo, only the rendered docs. Furthermore, starting at `actions/checkout@v6.0.0` the credentials are stored in a temporary directory outside the checked out repo which largely negates this issue.

.github/workflows/preview-docs.yml
```diff
@@ -5,15 +5,15 @@ permissions:
...
+  workflow_run: # zizmor: ignore[dangerous-triggers]
     workflows: ["Deploy docs"]
     types:
       - completed
 
 jobs:
   preview:
     runs-on: ubuntu-latest
-    if: >
+    if: >-
```
ROMemories, line 16: Wow, that [was always evaluated to true](https://docs.zizmor.sh/audits/#unsound-condition)‽

.github/workflows/build.yml
ROMemories: It wasn't, but for consistency it would likely be a good idea to switch to using the `toolchain` config key in this PR, which also has the benefit of being more explicit.

.github/workflows/preview-docs.yml
```diff
@@ -5,15 +5,15 @@ permissions:
   pull-requests: write
 
 on:
-  workflow_run:
+  workflow_run: # zizmor: ignore[dangerous-triggers]
```
ROMemories, line 8: I looked at these three `ignore`s (for `dangerous-triggers`), and I indeed don't think there is a straightforward replacement.

PR #1743, opened 2026-01-29
refactor(gpio): use anonymous type parameters in constructors

ROMemories:
# Description

This replaces the generics introduced in #1328 on the `Input` and `Output` constructors to use anonymous type parameters instead. This keeps these constructors in line with what we had before, is arguably easier to read, and makes these constructors consistent with those of I2C, SPI, and UART as done in #1737 and similar.

## Testing

```sh
laze -C examples/gpio/ build -b nrf52840dk run
```

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/gpio.rs
+++ b/src/ariel-os-hal/src/gpio.rs
@@ -66,10 +66,7 @@ pub struct Input {
 
 impl Input {
     /// Returns a configured [`Input`].
-    pub fn new<T: IntoPeripheral<'static, P>, P: HalInputPin + 'static>(
-        pin: T,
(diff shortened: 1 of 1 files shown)


src/ariel-os-hal/src/gpio.rs
ROMemories, line 75: Unfortunately we can't do the same here because the type actually has to be named in the return type, but it also [wasn't using an anonymous type parameter before #1328](https://ariel-os-docs-preview-pr1010.surge.sh/dev/docs/api/ariel_os/gpio/struct.Input.html#method.builder) for the same reason.

Issue #1648, opened 2026-01-05
"native" networking


kaspar030:
Wire up networking for the native target.

PR #1744, opened 2026-01-30
feat: add tun/tap support for native

chrysn:
# Description

As on the label: This adds networking support.

There's barely any code change, as embassy-net supports it neatly.

What is currently missing is:

* [x] Choosing *which* tap interface to pick

  [edirt, added: now allowing to override the default using an env  var]
* [x] Maybe rename because it seems we can only do tap (but maybe tun works too and I just held it wrong)

  [edit, added: leaving it, as embassy-net-tuntap may transparently improve]
* [x] Document how to use it, including how to set up a tap interface easily.

## Testing

With the current state and tap0 set up to serve DHCP, this already works:

```
$ laze build -b native -C examples/coap-server/ run
```

```
$ aiocoap-client coap://10.42.0.225/hello
Hello from Ariel OS
```

## Issues/PRs References

Closes: https://github.com/ariel-os/ariel-os/issues/1648

## Open question

* If anyone can provide the how-to for non-<del>Linux/</del>NetworkManager platforms, happy to add it; otherwise I think documentation for the primary dev platform will suffice for the moment.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -495,6 +495,8 @@ contexts:
       - has_device_identity
       - has_hwrng
       - sw/benchmark
+    provides_unique:
+      - c-function-abort
     disables:
       - semihosting
       - defmt


chrysn:
I just found (when doing reviews) that this is Linux-only. In https://github.com/embassy-rs/embassy/issues/5340, I' testing the waters for using tun-rs, which might enable more platforms. (But which platforms do we even aim for in native?)

chrysn:
There is https://github.com/embassy-rs/embassy/pull/5341 pending on settable hardware addresses, but I think that even with the single default address, this should be a good-enough start.

book/src/native-target.md
```diff
@@ -27,7 +27,45 @@ See [the support info of `native`][native-builder-support] for details.
 Native itself enables [multithreading][multithreading-book], and creates one "virtual core" per Ariel OS thread using host threads.
 This means that threads all run in *parallel* from the point of view of Ariel OS and of the application.
 
+## Networking
+
+Applications that set the `network` [laze module]
+will by select the `tuntap` module, which opens the `tap0` tap device
+(or any other name given in the `ARIEL_NATIVE_TUNTAP` environment variable)
```
kaspar030, line 34: note to self: we need conventions on variable names

book/src/native-target.md
```diff
@@ -27,7 +27,45 @@ See [the support info of `native`][native-builder-support] for details.
 Native itself enables [multithreading][multithreading-book], and creates one "virtual core" per Ariel OS thread using host threads.
 This means that threads all run in *parallel* from the point of view of Ariel OS and of the application.
 
+## Networking
+
+Applications that set the `network` [laze module]
+will by select the `tuntap` module, which opens the `tap0` tap device
```
ROMemories, line 33:
```suggestion
will be able to select the `tuntap` module, which opens the `tap0` tap device
```

book/src/native-target.md
```diff
@@ -27,7 +27,45 @@ See [the support info of `native`][native-builder-support] for details.
...
+(or any other name given in the `ARIEL_NATIVE_TUNTAP` environment variable)
+and exchanges traffic through there.
+
+If the device is has not been created, is in use or otherwise inaccessible,
```
ROMemories, line 37:
```suggestion
If the device has not yet been created, is in use or otherwise inaccessible,
```

src/ariel-os-native/src/lib.rs
```diff
@@ -16,3 +16,26 @@ pub fn init() -> OptionalPeripherals {
...
 
 pub struct SWI {}
+
+pub mod tuntap {
+    pub type NetworkDevice = embassy_net_tuntap::TunTapDevice;
+
+    /// Creates a TUN/TAP network device as configured in the environment.
+    ///
+    /// Unlike the hardware device initializers that are inline in `ariel_os_embassy::init_task`,
+    /// this doesn't work automatically from init'ing with peripherals ans a spawner, but goes
```
ROMemories, line 26:
```suggestion
    /// this doesn't work automatically from init'ing with peripherals and a spawner, but goes
```

src/ariel-os/Cargo.toml
```diff
@@ -154,6 +154,8 @@ wifi-cyw43 = ["ariel-os-embassy/wifi-cyw43"]
 wifi-esp = ["ariel-os-embassy/wifi-esp"]
 # Selects STM32 Ethernet
 eth-stm32 = ["ariel-os-embassy/eth-stm32"]
+## Selects the network backend that goes through tun/tap (native only)
```
ROMemories, line 157: Cargo features that users should only enable through laze modules and that don't appear as `doc_cfg` flairs should not be listed here (and currently this makes it appear in the "Multicore functionality" section):
```suggestion
# Selects the network backend that goes through tun/tap (native only).
```

book/src/native-target.md
```diff
@@ -27,7 +27,45 @@ See [the support info of `native`][native-builder-support] for details.
...
+Error opening interface tap0: Operation not permitted (os error 1)
+```
+
+Setting up a suitable interface depends on your platform and preferred configuration:
+
+* To instruct NetworkManager to create a connection that gets enabled automatically
+  and forward traffic from any uplink interface, run:
+
+  ```console
+  $ sudo nmcli connection add type tun mode tap user $(id -u) ifname tap0 con-name tap0 ivp6.method shared ipv4.method shared
```
ROMemories, line 51: Is this command usable as-is? I'm getting the following error when trying it out (with NM 1.52):

```sh
Error: invalid <setting>.<property> 'user'.
```

book/src/native-target.md
chrysn: I'm going with "will automatically" here, because "will be able to" seems to imply that this is now an option (but it is not optional).

(Commenting here b/c this is where these changes have a point-to-point presence; I'll link back when it's a PR)

book/src/native-target.md
```diff
@@ -27,7 +27,45 @@ See [the support info of `native`][native-builder-support] for details.
...
+  ```
+
+* To create a manually managed device that only persists until the next reboot, run:
+
+  ```console
+  $ sudo ip tuntap add dev tap0 user $(id -u) mode tap
```
ROMemories, line 57: I've tried running this command and then the following one:

```sh
laze -C examples/udp-echo/ build -b native run
```

I'm getting the following panic immediately:

```sh
[redacted]/embassy-net-tuntap-0.1.0/src/lib.rs:233:23:
transmit error: Os { code: 5, kind: Uncategorized, message: "Input/output error" }
```

Is there anything else I should have run? How can I investigate this?

kaspar030: Can you try upping the tap interface? `sudo ip link set up dev tap0`

book/src/native-target.md
chrysn: That's weird -- I did test this, but do get that same error; investigating.

chrysn: Indeed this contains two errors, and my shell history shows that I've made both those mistakes in mixed settings.

The one in #1751 I did test again, merely with the modification of s/0/123/ as there is already a tap0 that I don't want to break.

book/src/native-target.md
chrysn: How did you create the tap interface, the manual `ip tuntap` route or via a fixed NM command?

ROMemories: > Can you try upping the tap interface? `sudo ip link set up dev tap0`

Thanks, this does fix that error

kaspar030: > How did you create the tap interface, the manual `ip tuntap` route or via a fixed NM command?

I ran into this today on my workstation, using the manual `ip tuntap`. It needs the extra manual `ip link set up dev tapN`.

PR #1745, opened 2026-01-30
docs(book): revamp the glossary page

ROMemories:
# Description

This completely revamps the Glossary page of the book, which was really underdeveloped. I've focused for now on terms which can be confusing, seem to be often misunderstood, or which can refer to multiple things. I've tried to keep it balanced across manufacturers. The goal is not necessarily to completely define each term, but instead to emphasize how they relate to each other. I've added links to relevant Wikipedia pages and software projects, but have refrained from linking to manufacturer websites for now.

Missing terms can of course always be added in a later revision.

## Changelog Entry

The book's glossary has been corrected and expanded.

## Change Checklist

### DIFF
--- a/book/src/glossary.md
+++ b/book/src/glossary.md
@@ -1,10 +1,58 @@
 # Glossary
 
-## **Embassy HAL crates**
+- **[ARM (or Arm)][arm-arch-wikipedia]:** a family of instruction set architectures.
 
-Embassy crates that implement support for specific MCU families:
+- **[Arm (Arm Holdings)][arm-holdings]:** the company behind the ARM architecture family.
 
-- [embassy-nrf](https://docs.embassy.dev/embassy-nrf/)
-- [embassy-rp](https://docs.embassy.dev/embassy-rp/)
-- [embassy-stm32](https://docs.embassy.dev/embassy-stm32/)
-- [esp-hal](https://docs.esp-rs.org/esp-hal/) <!-- We cannot link to esp-hal specifically -->
+- **[Chip][integrated-circuit-wikipedia]:** an integrated circuit (IC).
+
+- **[Cortex-M][cortex-m-wikipedia]:** a family of 32-bit processor implementations from Arm.
+
+- **[ESP32][esp32-wikipedia]:** a family of 32-bit microcontrollers from Espressif.
+  Its older microcontrollers are based on the Xtensa architecture, while newer ones use RISC-V.
+  Can also specifically refer to the eponymous ESP32 microcontroller.
+
+- **[Embassy][embassy-homepage]:** a software project developing HALs for multiple microcontroller families (e.g., [`embassy-nrf`][embassy-nrf-embassydev], [`embassy-rp`][embassy-rp-embassydev], [`embassy-stm32`][embassy-stm32-embassydev]), along with other components (e.g., [`embassy-executor`][embassy-executor-docsrs], [`embassy-time`][embassy-time-docsrs]).
+
+- **[esp-hal][esp-hal-homepage]:** a HAL for ESP32 microcontrollers, developed by their manufacturer Espressif.
+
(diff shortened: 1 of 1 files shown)


book/src/application.md
```diff
@@ -14,7 +14,7 @@ Being ZSTs, they do not carry any data to the drivers, only their ownership is m
 > [!TIP]
 > If you are used to thinking about MCU peripherals as referenced by a base address (in the case of memory-mapped peripherals), you can think of these ZSTs as abstraction over these, with a zero-cost, statically-enforced lock ensuring exclusive access.
 
-These Embassy types are defined by [Embassy HAL crates][embassy-hal-crates] in the respective `peripherals` modules.
+These Embassy types are defined by HAL crates in the respective `peripherals` modules.
```
ROMemories, line 17: This section could likely be re-written to put a little less emphasis on Embassy, but this is out-of-scope of this PR.

book/src/glossary.md
```diff
@@ -1,10 +1,58 @@
...
+- **[esp-hal][esp-hal-homepage]:** a HAL for ESP32 microcontrollers, developed by their manufacturer Espressif.
+
+- **HAL (Hardware Abstraction Layer):** a software layer that makes specific pieces of hardware (e.g., microcontroller peripherals) easier to use, by hiding some of their details.
+
+- **[MCU][microcontroller-wikipedia]:** a microcontroller.
+  Contains a processor, memories, and peripherals to interact with the outside world.
+
+- **nRF:** a family of 32-bit microcontrollers developed by Nordic Semiconductor.
+
+- **Raspberry Pi:** can refer to the [family of single-board computers][raspberry-pi-sbc-wikipedia] or to [Raspberry Pi Holdings][raspberry-pi-holdings-wikipedia], the company that manufactures them as well as RP microcontrollers.
```
ROMemories, line 26: Technically there's also the [Raspberry Pi Foundation](https://en.wikipedia.org/wiki/Raspberry_Pi_Foundation) but I'm not sure how useful it is to refer to it.

ROMemories:
Keeping as draft for now in case we wouldn't want to land it between v0.3.0-rc.1 and v0.3.0.
Edit: marking as ready for review now that v0.3.0 has been tagged.

book/src/glossary.md
```diff
@@ -1,10 +1,58 @@
...
+- **[esp-hal][esp-hal-homepage]:** a HAL for ESP32 microcontrollers, developed by their manufacturer Espressif.
+
+- **HAL (Hardware Abstraction Layer):** a software layer that makes specific pieces of hardware (e.g., microcontroller peripherals) easier to use, by hiding some of their details.
+
+- **[MCU][microcontroller-wikipedia]:** a microcontroller.
+  Contains a processor, memory, and peripherals to interact with the outside world.
+
+- **nRF:** a family of 32-bit microcontrollers developed by Nordic Semiconductor.
+
+- **Raspberry Pi:** can refer to the [family of single-board computers][raspberry-pi-sbc-wikipedia] or to [Raspberry Pi Holdings][raspberry-pi-holdings-wikipedia], the company that manufactures them as well as RP microcontrollers.
```
ROMemories, line 26: Maybe "manufactures" isn't right, "designs" or "sells" may be better.

book/src/glossary.md
kaspar030: fine as is

book/src/glossary.md
kaspar030: I think it is ok

book/src/application.md
```diff
@@ -14,7 +14,7 @@ Being ZSTs, they do not carry any data to the drivers, only their ownership is m
 > [!TIP]
```
kaspar030, line 14: Do book improvements end up in the changelog?

ROMemories: It depends, but I think that one should (I didn't add an entry before because it was an indeterminate state wrt to the release/changelog).

PR #1754, opened 2026-02-02
feat(native): add storage support

chrysn:
# Description

Storage is missing on native so far; this adds it backed by a file.

The implementation tries to be rather KISS: The file is loaded in full at startup, and stored back in full on every write. It picks a sensible erase unit size (because the key-value storage won't spread some things across page boundaries).

This is useful for running CoAP tests and others.

## Testing

Run the storage example on native; its counter ticks up, and a storage file is crated.

## Open Questions

* [x] How much of the file do we set up from the build system?

  [edit:] Nothing at all; there is a constant default size, that should suffice for an MVP.

* [x] Where do we best configure the storage size and parameters such as the flash page?

   [edit:] In an initial implementation, both are just fixed (although size can be changed if the user provides a larger or smaller file)

* [ ] How much should this be KISS and how much should this be the most strict and most extremely configured storage backend (because then tests on native would show implicit assumptions early)?

* [ ] Should we fsync?
  * Pro: If the host system crashes, we don't lose any data. (But then again, this is mostly used in testing, and the host crashing is a rather unexpected event). In particular, we forward the guarantee that we claim in the flash system that writes are really there.
  * Con: can be quite slow, especially when writing often. The guarantee is wobbly anyway because we can't keep the user from copying the file and restoring it from a backup. (Especiially relevant for OSCORE, which may eventually rely on this property to hold lest it would break cryptographic promises).

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1279,7 +1279,7 @@ modules:
     help:
       Laze-only module that enables storage.x.
 
-      This is needed for all the boards where storage is memoyr mapped and
+      This is needed for all the boards where storage is memory mapped and
       visible to the linker, which is currently all except native.
     provides_unique:
       - storage-backend


chrysn:
Running on a small build for the early commits that alter the build system; I'd go back to build-on-native for testing later.

laze-project.yml
```diff
@@ -1258,12 +1259,40 @@ modules:
...
+    env:
+      global:
+        FEATURES:
+          - ariel-os/storage-backend-native
+
+  - name: storage-backend-linked
+    help:
+      Laze-only module that enables storage.x.
+
+      This is needed for all the boards where storage is memoyr mapped and
```
```suggestion
      This is needed for all the boards where storage is memory mapped and
```

laze-project.yml
```diff
@@ -1258,12 +1259,40 @@ modules:
...
     selects:
       - has_storage_support
+      - storage-backend
     env:
       global:
         FEATURES:
           - ariel-os/storage
+
+  - name: storage-backend-native
+    provides_unique:
```
kaspar030, line 1269: So the idea here is that `storage-backend` pulls in any of the following, but tries the `native` first b/c that is defined first?

chrysn: "Idea" would be taking too much credit; it's more "it happens to work this way".

I'd like to migrate to "It takes any storage-backend-* available, but realistically there will be just one", but currently `has_storage_support` just implies that we have a linker-mapped flash layer. Not sure how to best do this – can I just s/has_storage_support/storage-backend-linked in all the existing contexts?

kaspar030: Hm, I think this is fine, but maybe add a comment like "this needs to stay ordered before `storage-backend-linked`".

book/src/native-target.md
```diff
@@ -48,25 +48,34 @@ Setting up a suitable interface depends on your platform and preferred configura
...
-  Note that for provisioning addresses, you may need to run a DHCP server on that interface,
-  and depending on your application, you may need to set up routing manually.
-
 * For device-to-device communication between multiple native instances,
   you can create a bridge and attach one tap device per instance to the bridge;
   the setup for that is currently beyond the scope of this documentation.
 
 At the time of writing, the tap implementation is limited to Linux.
 
+## Storage
```
ROMemories, line 66: Drive-by comment (haven't looked at the rest at all): if we go with not fsyncing (which, after thinking about if for just 5 s, might make the most sense), I think there should be a warning about it in this section so it's not assumed that it is.

PR #1763, opened 2026-02-04
docs(ble): enable ble in API docs

nponsard:
# Description

Documents the ble module in the hosted api docs. This issue was found by @chrysn 

## Testing

Check the published docs for this PR and see that `ble` is available https://ariel-os-docs-preview-pr1763.surge.sh/dev/docs/api/ariel_os/

## Open Questions

Should trouble_host be reexported ?

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -63,7 +63,7 @@ jobs:
       # faster generation.
       - name: Build rustdoc docs
         run: |
-          RUSTDOCFLAGS='--cfg nightly' cargo doc \
+          RUSTDOCFLAGS='--cfg nightly -Z unstable-options --extern-html-root-url trouble_host=https://docs.embassy.dev/trouble-host/0.5.0 ' cargo doc \
               --no-deps \
               -p ariel-os \
               -p coapcore \
--- a/src/ariel-os-embassy/src/ble.rs
+++ b/src/ariel-os-embassy/src/ble.rs
@@ -1,3 +1,14 @@
+//! Provides control over the system BLE stack.
+//!
+//! All interactions happen through the [`trouble_host::Stack`] struct that can be obtained using
+//! [`ble_stack()`].
+//!
+//! The address of the device is randomly generated at boot and may be rotated during execution.
+//!
+//! # Current implementation
+//!
+//! The address is not currently rotated during execution; however this behavior may not be relied upon.
+
 use trouble_host::{
     Address,
     prelude::{AddrKind, BdAddr},
--- a/src/ariel-os-hal/src/hal/dummy/ble.rs
+++ b/src/ariel-os-hal/src/hal/dummy/ble.rs
@@ -13,6 +13,14 @@ impl Peripherals {
     }
 }
 
+/// Returns the system BLE stack, [`trouble_host::Stack`].
+///
+/// # Panics
+///
+/// For this function to succeed, it can only be called from the system executor, and only once:
+///
+/// - Panics if the stack was already taken.
+/// - Panics when not called from the system executor.
 pub async fn ble_stack() -> &'static Stack<'static, DummyController, DefaultPacketPool> {
     async { unimplemented!() }.await
 }


ROMemories:
We need documentation for the `ble` module and the `ble_stack()` function.

src/ariel-os-embassy/src/ble.rs
```diff
@@ -1,3 +1,10 @@
+//! Provides control over the BLE stack of the system.
+//!
+//! All interactions happens through the [`trouble_host::Stack`] struct that you can obtain using
+//! [`ble_stack()`].
+//!
+//! Currently the address of the device will be randomly generated at boot.
```
ROMemories: The address is *guaranteed* to be rotated at boot (this is important for privacy and security reasons, changing this *should* be a breaking change):
```suggestion
//! The address of the device is randomly generated at boot and may be rotated during execution.
//!
//! # Current implementation
//!
//! The address is not currently rotated during execution; however this behavior may not be relied upon.
```

src/ariel-os-embassy/src/lib.rs
```diff
@@ -93,6 +93,8 @@ pub mod reexports {
     pub use embassy_time;
     #[cfg(feature = "usb")]
     pub use embassy_usb;
+    #[cfg(feature = "ble")]
+    pub use trouble_host;
```
ROMemories: We shouldn't re-export `trouble_host`, as we also don't re-export `embedded-hal`, `rand_core`, and other similar crates, on purpose. Generally we don't currently re-export crates that only provide interfaces (at least so as to not easily end up in dependency hell).

src/ariel-os-hal/src/hal/dummy/ble.rs
```diff
@@ -13,6 +13,14 @@ impl Peripherals {
     }
 }
 
+/// Returns the BLE stack, [`trouble_host::Stack`], for this system.
```
ROMemories: I'm not sure what "this system" refers to, we don't use that phrasing in other parts of the docs:
```suggestion
/// Returns the system BLE stack, [`trouble_host::Stack`].
```

src/ariel-os-hal/src/hal/dummy/ble.rs
```diff
@@ -13,6 +13,14 @@ impl Peripherals {
...
+///
+/// Implementation of this function is dependent on the current system.
```
ROMemories: That should be transparent to users, and we don't say this for other items in `ariel_os::hal`:
```suggestion
```

src/ariel-os-hal/src/hal/dummy/ble.rs
```diff
@@ -13,6 +13,14 @@ impl Peripherals {
...
+/// This function must never be called more than once and must be called from the main executor.
+///
+/// # Panics
+/// - panics if the stack was already taken
+/// - panics when not called from the main executor
```
ROMemories: "must never" sounds too scary to me (it would make me wonder whether that could lead to undocumented UB), I think it can be moved into the Panics section (unless I misunderstood).

[The docs](https://ariel-os.github.io/ariel-os/dev/docs/book/async-support.html) call this the "system executor."

```suggestion
///
/// # Panics
///
/// This function should not be called more than once and must be called from the system executor:
///
/// - Panics if the stack was already taken.
/// - Panics when not called from the system executor.
```

src/ariel-os-embassy/src/ble.rs
```diff
@@ -1,3 +1,10 @@
+//! Provides control over the BLE stack of the system.
```
ROMemories:
```suggestion
//! Provides control over the system BLE stack.
```

src/ariel-os-embassy/src/ble.rs
```diff
@@ -1,3 +1,10 @@
...
+//!
+//! All interactions happens through the [`trouble_host::Stack`] struct that you can obtain using
```
ROMemories:
```suggestion
//! All interactions happen through the [`trouble_host::Stack`] struct that can be obtained using
```

src/ariel-os-embassy/src/lib.rs
nponsard: If I remove the export of `trouble_host`, the link in the module doc to `trouble_host::Stack` breaks, is there a way to still have a working link ?

chrysn: I agree that re-exporting is not for here (AFAIR we don't have precise policy, but I'd wing it at "we re-export if there's a self-contained subset of the other crate which we think may be more stable than that whole crate").

As for links, there has to be *some* mechanism, because docs.rs also manages to not build all crates in one go and still have links to dependencies go to the right place despite building with --no-dependencies, but I haven't found it (and think we should try harder to find it and link to docs.rs for everything that's not documented on our page).

As a stop-gap measure, there's always the option to link to the relevant docs.rs URI.

src/ariel-os-hal/src/hal/dummy/ble.rs
```diff
@@ -13,6 +13,14 @@ impl Peripherals {
     }
 }
 
+/// Returns the system BLE stack, [`trouble_host::Stack`].
+///
+/// # Panics
+///
+/// This function should not be called more than once and must be called from the system executor:
```
chrysn: Or, following up on https://github.com/ariel-os/ariel-os/pull/1763#discussion_r2763487248, to neither try to be hard or soft prescriptive and rather factual:

```suggestion
///
/// # Panics
///
/// For this function to succeed, it can only be called from the system executor, and only once:
```

(But up to you, the current phrasing works too.)

src/ariel-os-embassy/src/lib.rs
nponsard: I think I found an okay solution: https://doc.rust-lang.org/rustdoc/unstable-features.html#--extern-html-root-url-control-how-rustdoc-links-to-non-local-crates

.github/workflows/build-deploy-docs.yml
```diff
@@ -63,13 +63,14 @@ jobs:
       # faster generation.
       - name: Build rustdoc docs
         run: |
-          RUSTDOCFLAGS='--cfg nightly' cargo doc \
+          RUSTDOCFLAGS='--cfg nightly -Z unstable-options --extern-html-root-url trouble-host=https://docs.embassy.dev/trouble-host/0.5.0 ' cargo doc \
```
chrysn: It's a good fix for here, but we should follow up with generating this list for everything that is in our public dependencies.

src/ariel-os-embassy/src/lib.rs
chrysn: Awesome, thanks a lot!

ROMemories: > I think I found an okay solution

Does it work? The links still look broken to me in the docs preview:

Screenshot_20260204_133958

ROMemories: However I don't see these links being broken as a release blocker, so I think we should now merge this as-is and fix this later.

nponsard: argh, I wrote it first in the CI file, tested it on my machine, found that it needed to be `trouble_host` and not `trouble-host` but didn't fix it in the CI file.

nponsard: it works now

ROMemories: Using the crate name instead of the package name works; thanks a lot for looking into this!

PR #1764, opened 2026-02-04
fix(smp): remove an ignored `inline` on an externally exported function

ROMemories:
# Description

This removes an `#[inline]` attribute which was ignored by the compiler because it was on an externally exported function (a function with `no_mangle`).

This fixes the following compiler warning:

```sh
warning: `#[inline]` is ignored on externally exported functions
  --> src/ariel-os-threads/src/smp/rp.rs:90:1
   |
90 | #[inline]
   | ^^^^^^^^^
   |
   = help: externally exported functions are functions with `#[no_mangle]`, `#[export_name]`, or `#[linkage]`
   = note: `#[warn(unused_attributes)]` (part of `#[warn(unused)]`) on by default
```

This isn't forbidden by CI because we don't currently turn warnings from *the compiler* into errors (only from Clippy; but we should).

## Testing

Can be observed by running the following:

```sh
laze -C examples/threading-multicore/ build -b rpi-pico
```

## Change Checklist

### DIFF
--- a/src/ariel-os-threads/src/smp/rp.rs
+++ b/src/ariel-os-threads/src/smp/rp.rs
@@ -87,7 +87,6 @@ const SCHEDULE_TOKEN: u32 = 0x11;
 #[unsafe(no_mangle)]
 // SAFETY: this function is placed in RAM to improve execution latency.
 #[unsafe(link_section = ".data.ram_func")]
-#[inline]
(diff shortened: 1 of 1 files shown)


ROMemories:
Not marking this as draft as I think we can land this for the v0.3.0 (but no marking as `release-blocker` either because it's not critical at all).

PR #1771, opened 2026-02-04
feat(boards): add STM32WBA6x family and NUCLEO-WBA65RI board support

leftger:
# Description

This adds support for the STM32WBA6x microcontroller family (18 chip variants) and the ST NUCLEO-WBA65RI development board.

The WBA6x family is architecturally identical to the existing WBA5x family (Cortex-M33F with hardware RNG) but offers more memory (up to 2MB flash / 512KB RAM) and Bluetooth LE 6.0 support.

## Testing

Successfully built and tested the following examples on st-nucleo-wba65ri:
- hello-world - Basic chip support and UART console
- blinky - LED GPIO functionality
- gpio - GPIO/button support
- random - Hardware RNG functionality

All builds completed without errors.

## Issues/PRs References

- Depends on #1773

## Changelog entry

The STM32WBA65RI MCU and the ST NUCLEO-WBA65RI board are now supported.

## Change checklist


src/ariel-os-stm32-mapping/stm32_mapping.csv
```diff
@@ -1382,6 +1382,24 @@
...
+"stm32wba64cg"
+"stm32wba64ci"
+"stm32wba65cg"
+"stm32wba65ci"
+"stm32wba65mg"
+"stm32wba65mi"
+"stm32wba65pg"
+"stm32wba65pi"
+"stm32wba65rg"
+"stm32wba65ri"
```
ROMemories, line 1402: I've checked this list of new chips. 👍

src/ariel-os-stm32-mapping/Cargo.toml
ROMemories, line 1: This gets auto-generated by the build script as appropriate 👍

laze-project.yml
```diff
@@ -684,6 +684,18 @@ contexts:
       RUSTFLAGS:
         - --cfg capability=\"hw/stm32-rng\"
 
+  - name: stm32wba65ri
+    parent: stm32
+    selects:
+      - cortex-m33f
```
ROMemories, line 721: Does have a Cortex-M33 with an FPU 👍

laze-project.yml
```diff
@@ -684,6 +684,18 @@ contexts:
...
+    provides:
+      - has_hwrng
```
ROMemories, line 723: Does have a TRNG. 👍

laze-project.yml
```diff
@@ -684,6 +684,18 @@ contexts:
...
+      - has_storage_support
+    env:
+      PROBE_RS_CHIP: STM32WBA65RI
+      RUSTFLAGS:
+        - --cfg capability=\"hw/stm32-rng\"
```
ROMemories, line 728: That's indeed the proper one: the interrupt is `RNG`, and the `random` example would not compile otherwise 👍

boards/st-nucleo-wba65ri.yaml
```diff
@@ -0,0 +1,34 @@
+boards:
+  st-nucleo-wba65ri:
+    chip: stm32wba65ri
+    ariel:
+      swi: USART1
+    leds:
```
ROMemories, line 8: I've checked the pins of the LEDs and buttons against [the board's user manual](https://www.st.com/resource/en/user_manual/um3448-stm32wba-nucleo64-board-mb1801-and-mb2130-stmicroelectronics.pdf) 👍

boards/st-nucleo-wba65ri.yaml
```diff
@@ -0,0 +1,34 @@
+boards:
+  st-nucleo-wba65ri:
+    chip: stm32wba65ri
```
ROMemories, line 3: Based on the board's user manual, it seems that the board does have a user USB port (and the MCU does  have a USB peripheral, unlike the STM32WBA5 MCUs). However it has a USB IP that we didn't support yet.

Would you be able to rebase this PR onto #1773 (which should add support for this new IP, but is currently untested in hardware), and add the following here?

```suggestion
    chip: stm32wba65ri
    flags:
      - has_usb_device_port
```

To test it in hardware, would you be able to run the `usb-serial` example, or one of the networking example (e.g., `udp-echo`) [using Ethernet-over-USB](https://ariel-os.github.io/ariel-os/dev/docs/book/networking.html#network-link-selection)? Otherwise that's ok, we'll simply mark USB functionality as `needs_testing` for now.

nponsard:
Hello ! 
Thank you for your contribution, it seems like we forgot to document a step in the process to add a new MCU, the name needs to be added in `boards/ariel-chips.yaml`.

After that you can run `sbd-gen` to update all the generated files: 
```
sbd-gen generate-ariel boards -o src/ariel-os-boards --mode update
```

leftger:
Thank you for reviewing this PR. I believe I have addressed your request. Please let me know if there's anything else I need to do. 😄

boards/ariel-chips.yaml
```diff
@@ -42,4 +42,5 @@ ariel:
     - stm32u585ai
     - stm32wb55rg
     - stm32wba55cg
+    - stm32wba65ri
```
ROMemories, line 47: 👍

src/ariel-os-storage/build.rs
```diff
@@ -17,7 +17,7 @@ fn main() {
         (4 * KIBIBYTES, 2 * KIBIBYTES)
     } else if is_in_current_contexts(&["nrf52", "nrf5340-app", "nrf91", "rp", "stm32wb55rg"]) {
         (8 * KIBIBYTES, 4 * KIBIBYTES)
-    } else if is_in_current_contexts(&["stm32u585ai"]) {
+    } else if is_in_current_contexts(&["stm32u585ai", "stm32wba65ri"]) {
```
ROMemories, line 22: That MCU indeed has 8-KiB flash pages, and we use two pages for the persistent storage API. 👍

ROMemories, line 22: Could you run the `storage` example to check the persistent storage API works as expected?

laze-project.yml
ROMemories, line 728: This should work after rebasing on #1773:
```suggestion
        - --cfg capability=\"hw/stm32-rng\"
        - --cfg capability=\"hw/stm32-usb-synopsis-hs\"
```

kaspar030:
Does this MCU work without clock config in [this cursed function](https://github.com/ariel-os/ariel-os/blob/6adbd4682fb7cf846aca89af9816b24d8bf4170d/src/ariel-os-stm32/src/lib.rs#L104)?

leftger:
> Does this MCU work without clock config in [this cursed function](https://github.com/ariel-os/ariel-os/blob/6adbd4682fb7cf846aca89af9816b24d8bf4170d/src/ariel-os-stm32/src/lib.rs#L104)?

Could you elaborate on this? Do you want me to change the function? remove it? Any help is appreciated.

leftger:
Apologies for force-pushing and failing CI so many times. I am a bit tired (it's 5am where I'm at). I'll continue this when I wake up.

kaspar030:
> Could you elaborate on this? Do you want me to change the function? remove it? Any help is appreciated.

Most stm32 need some amount of clock configuration, usually when USB comes into the picture. If that is the case for this board, the linked function is currently the correct place for that. Might be that the defaults on this board are fine and you don't need to do anything! :)

src/ariel-os-storage/build.rs
leftger: It works

leftger:
Hi @kaspar030 I submitted this a while back. I was busy with life things and left this in the back burner. I think I have addressed all your requests. Please let me know if anything else is needed.

leftger:
I failed rustfmt. I force-pushed the change.

kaspar030:
The sbd-gen check failed, you probably need to re-generate `src/ariel-os-boards` with this:

`sbd-gen generate-ariel --mode update boards -o src/ariel-os-boards`

leftger:
OK done.

doc/support_matrix.yml
```diff
@@ -781,6 +781,26 @@ chips:
...
+      uart: not_currently_supported
+      logging: supported
+      storage:
+        status: supported_with_caveats
+        comments:
+          - removing items not supported
+      wifi: not_available
+      ble: not_currently_supported
+      user_usb: needs_testing
+      ethernet_over_usb: needs_testing
```
kaspar030, line 802:
```suggestion
      user_usb: supported
      ethernet_over_usb: supported
```

As you did add support for the USB IP, I'm assuming you tested these?

boards/st-nucleo-wba65ri.yaml
ROMemories: I've rebased and updated #1773 following #1840. As @kaspar030 mentioned, can you confirm this does work on your board?

doc/support_matrix.yml
kaspar030: Ah now I see the comment by @ROMemories and the dependency on #1773. :eyes:

kaspar030: ok, let's postpone testing.

kaspar030:
Thanks a lot @leftger for seeing this through!

PR #1773, opened 2026-02-05
feat(stm32): add support for the Synopsys USB OTG HS IP

ROMemories:
# Description

This adds support for the Synopsys USB OTG HS IP used in the STM32WBA65RI.

## Testing

Testing should happen in #1771.

## Changelog Entry

I don't think this needs an entry in itself, only support for the new MCU will be added to the changelog.

## Change Checklist


kaspar030:
I have tested on #1771 that this doesn't break the non-hs synopsys users.

PR #1774, opened 2026-02-05
build(doc): add laze task to generate documentation

bergzand:
# Description

This adds tasks to laze to generate the documentation with the same settings as done currently by the CI. This makes it easier to generate documentation locally, the magic incantation doesn't have to be grabbed from the CI files anymore. For example `laze build doc --open` builds and opens the api docs.

Next step is to use these tasks also in the CI, to guarantee the same process is used.

The esp documentation ends up in `./target/riscv…/doc`. I haven't found a good way to have it also end up in the same directory as the rest. `mv` would be a crude solution, but currently `laze build doc-esp --open` works, and adding the `mv` would break that

## Testing

Added tasks that need a look
 - `laze build doc`
 - `laze build doc-esp`
 - `laze build doc-nrf`
 - `laze build doc-rp`
 - `laze build doc-stm32`

## Issues/PRs References

None

## Open Questions

None

## Changelog Entry

Building documentation locally is now possible via `laze build doc`. Documentation for microcontroller families can now be built by appending the family name: `laze build doc-esp`.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -2020,14 +2020,14 @@ builders:
           - >-
             cargo doc
             --no-deps
+            -p ariel-os-stm32
             --features "
                 embassy-stm32/stm32wb55rg,
                 external-interrupts,
                 i2c,
                 spi,
                 uart,
-                "
-            -p ariel-os-stm32 $@
+                " $@
 
   - name: bbc-microbit
     parent: bbc-microbit-base


laze-project.yml
```diff
@@ -1900,6 +1906,128 @@ builders:
...
+            cargo doc
+            --no-deps
+            --features "
+            embassy-nrf/nrf52840,
+            external-interrupts,
+            i2c,
+            spi,
+            uart,
+            "
+            -p ariel-os-nrf $@
```
bergzand: In the ideal case, these chip family specific tasks would be a bit more generic and generate all the  documentation for a specific builder. But I don't see a clean way to do that here.

laze-project.yml
```diff
@@ -1900,6 +1906,128 @@ builders:
...
+            udp,
+            usb,
+            usb-hid,
+            coapcore/_nightly_docs
+            " $@
+
+      doc-esp:
+        build: false
+        export:
+          - CARGO_BUILD_TARGET: riscv32imac-unknown-none-elf
```
bergzand, line 1959: I haven't found a way to build the docs into `./target/doc` (without the `riscv32imac-unknown-none-elf` subdir in between)

nponsard:
Should these commands be used in the CI so we are sure they stay in sync ?

bergzand:
> Should these commands be used in the CI so we are sure they stay in sync ?

Yes! But I'm slightly afraid to touch that side and I'd rather leave that for a follow up. It also means that that a few more CI tasks need laze installed.

laze-project.yml
```diff
@@ -1900,6 +1906,128 @@ builders:
...
+            -p embedded-hal-async
+            -p embedded-io-async@0.6.1
+          - >-
+            cargo doc
+            --no-deps
+            -p ariel-os
+            -p coapcore
+            --features "
+            ariel-os-sensors/max-sample-min-count-12,
+            bench,
```
ROMemories: `ble` was added in #1763:
```suggestion
            bench,
            ble,
```

laze-project.yml
```diff
@@ -1900,6 +1906,128 @@ builders:
         cmd:
           - cargo fmt --all -- --config format_generated_files=false
 
+      doc:
```
ROMemories, line 1909: These only work if a nightly override is set up (which we don't document currently). Is there any way to make them work without one?
Edit: i.e., this should actually select the nightly toolchain

laze-project.yml
```diff
@@ -1900,6 +1906,128 @@ builders:
...
+            -p embedded-hal@1.0.0
+            -p embedded-hal-async
+            -p embedded-io-async@0.6.1
+          - >-
+            cargo doc
+            --no-deps
+            -p ariel-os
+            -p coapcore
+            --features "
+            ariel-os-sensors/max-sample-min-count-12,
```
ROMemories: optional: preserving the indent that we have in the workflows might make these easier to read.

laze-project.yml
```diff
@@ -1868,6 +1868,12 @@ builders:
       - name: LINK
         description: "HOST LINK (no-op)"
         cmd: "true"
+    env:
+      RUSTDOCFLAGS:
+        - -D warnings
+        - --cfg nightly
+        - -Z unstable-options
+        - --extern-html-root-url trouble_host=https://docs.embassy.dev/trouble-host/0.5.0'
```
ROMemories, line 1876: @kaspar030 Is there a way to keep these factored out between the doc tasks, without the env for the others?

laze-project.yml
```diff
@@ -1900,6 +1906,128 @@ builders:
...
+            spi,
+            storage,
+            tcp,
+            threading,
+            uart,
+            udp,
+            usb,
+            usb-hid,
+            coapcore/_nightly_docs
+            " $@
```
ROMemories: The `$@` does allow to pass `--open` in particular 👍

laze-project.yml
bergzand: I wish I could do that, but yaml doesn't strip out the newlines with the `>` marker if I add the proper indentation here. I already wasted an hour on wondering why my '>' didn't work :sob:

When I format it as:
```
      doc:
        build: false
        cmd:
          - >-
            cargo doc
              --no-deps
              -p ariel-os
              -p coapcore
```

It results in 
```
laze: executing `["sh", "-c", "cargo doc\n --no-deps\n -p ariel-os\n -p coapcore"]`
```

laze-project.yml
kaspar030: I think so, by creating a hierarchy for tasks. I have a branch for doing a similar thing for clippy, let me dig that out

laze-project.yml
ROMemories: That's right indeed, it seems we still can indent the list of features (possibly because it is quoted), which does still improve readability significantly imo.

bergzand: I already had a discussion with @kaspar030 whether Laze should always strip the newlines in a `cmd`, in turn allowing for preserving the full indentation here.

PR #1776, opened 2026-02-05
feat(ci): add clippy laze tasks

kaspar030:
# Description

This moves the clippy calls from the github workflow into laze tasks.

## Open Questions

- [ ] double check the feature/package lists are the same
- [ ] actually use the laze tasks in the workflow

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1900,6 +1900,172 @@ builders:
         cmd:
           - cargo fmt --all -- --config format_generated_files=false
 
+
+      clippy-all:
+        build: false
+        cmd:
+          - :clippy-host
+          - :clippy-coapcore
+          - :clippy-esp32
+          - :clippy-nrf52
+          - :clippy-nrf91
+          - :clippy-rp
+          - :clippy-stm32
+
+      clippy-run:
+        build: false
+
+        cmd:
+          - >-
+            cargo clippy
+            --locked
+            $CLIPPY_TARGET
+            --features "$CLIPPY_FEATURES"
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -1900,6 +1900,172 @@ builders:
...
+            --features "$CLIPPY_FEATURES"
+            $CLIPPY_PACKAGES
+            --
+            --deny warnings
+
+      clippy-esp32:
+        build: false
+        export:
+          - RUSTFLAGS: '--cfg context="esp32c6" --cfg getrandom_backend="custom"'
+          - CLIPPY_TARGET: --target=riscv32imac-unknown-none-elf
```
kaspar030, line 1932: There's no "proper" way of passing typed arguments from task to task, so I'm using exported env variables here.

laze-project.yml
```diff
@@ -1900,6 +1900,172 @@ builders:
...
+          - :clippy-stm32
+
+      clippy-run:
+        build: false
+
+        cmd:
+          - >-
+            cargo clippy
+            --locked
+            $CLIPPY_TARGET
```
kaspar030, line 1922: Using `$VAR` vs `${VAR}` makes the shell evaluate this and not laze.

bergzand: So `${VAR}` is evaluated by Laze and `$VAR` only works on Unix shells?

kaspar030: yes on the first statement, the second is a bit too open.

PR #1783, opened 2026-02-06
docs(CC): no arbitrary pins

chrysn:
# Description

I just ran the sensor example on a particle-xenon. That board is not in `pins.rs`, but there's a catch-all for the nRF chips. This means that it was by pure coincidence that there was not a button connected on those pins that, when pressed, would have shortened out when I2C was driven low.

This PR contains a policy change (or clarification – I for one would have expected this to be the case) that we don't drive arbitrary pins in examples, neither on unknown boards (if the example requires something present on some board that's not declared, it should just not run there) nor driving "user" pins without some checking.

## Testing

Docs and build system only; sensors-debug example does not build with particle-xenon any more.

## Open Questions

* [ ] Do we have other examples that do this?
* [ ] Do we need a conventional way to prompt the user for "yes I've attached the wiring from the README to the board"? Can this be a laze module that is required to be specified on the command line, or would we need to go through something that triggers later (like the user needs to set a define that if unset raises a compiler error)?

## Changelog Entry

Coding conventions to ensure safety against chip damage on out-of-tree boards; examples were adjusted accordingly.

## Change Checklist

### DIFF
--- a/book/src/coding-conventions.md
+++ b/book/src/coding-conventions.md
@@ -108,7 +108,7 @@ Practically, this comes in two stages:
 2. User-accessible general-purpose pins must only be used with some confirmation,
    either from automated testing or from manual input.
 
-   This means that it is safe to run any application on any bare board even if there are daughterboards
+   This means that it is safe to run any application on any board even if there are daughterboards
    (e.g., shields, wings, capes, clicks, or grove modules)
    connected.
 


book/src/coding-conventions.md
```diff
@@ -91,6 +91,24 @@ You can quickly format the code in the ariel-os repository using this command:
...
 
+## Applications (Tests and Examples)
+
+All our applications (currently, that is tests and examples)
+must be safe against damage on any board, including externally defined boards.
+They may alter the state of the system (eg. write to on-board memory),
+but must only drive GPIO pins based on board descriptions
+(or, until those are comprehensive enough, hard-coded knowledge of that board).
+
+They may expect that some configuration of pin headers in a board is im place,
```
kaspar030:
```suggestion
They may expect that some configuration of pin headers in a board is in place,
```

chrysn:
> even targeting boards/builders instead of MCUs doesn't really guarantee that a wire/button won't be attached between two pin

There are two levels of things we can do that I think we're mixing up here (mostly because I didn't keep them apart in the text):

1. If there is a board, and nothing is connected, any example/test you run should be safe (provided, as always, that you flash firmware built for the right board).
2. If there is anything connected to the board that is within what you can generally expect not to cause damage, our examples/tests should deal with that. ("generally expect" as in "if something is exclusively marked as an I2C pin but misused, it's not our fault").

Class 1 is what worries me most urgently, and that is what is fixed already in the code change here: In examples that talk to on-board peripherals, we configure pins, and doing that on a different board has some chance of hitting a button. So ideally I'd like to eradicate class 1 now, and fix 2 until the next release or so.

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+Practically, this comes in two stages:
+
+1. Driven pins must only be selected based on the board (and not, for example, on the MCU family).
+
+   This means that it is safe to run any application on any bare board (with no extra wiring / shields).
+
+2. User accessible general-purpose pins must only used be used with some confirmation,
+   either from automated testing or from manual input.
+
+   This means that it is safe to run any applycation on any bare board even if there are daughterboards
```
kaspar030:
```suggestion
   This means that it is safe to run any application on any bare board even if there are daughterboards
```

ROMemories:
As long as there are changes to applications in this PR, this should definitely not be `ci-build:skip`.

chrysn:
(Build was set to skip because all recent changes go to docs, and the one code change already passed.)

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
 laze build fmt
 ```
 
+## Applications (Tests and Examples)
+
+All our applications (currently, that is tests and examples)
+must be safe against damage on any board, including externally defined boards.
+They may alter the state of the system (eg. write to on-board memory),
```
ROMemories:
```suggestion
They may alter the state of the system (e.g., write to onboard memory),
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+but must only drive GPIO pins based on board descriptions
+(or, until those are comprehensive enough, hard-coded knowledge of that board).
+
+Practically, this comes in two stages:
+
+1. Driven pins must only be selected based on the board (and not, for example, on the MCU family).
+
+   This means that it is safe to run any application on any bare board (with no extra wiring / shields).
+
+2. User accessible general-purpose pins must only used be used with some confirmation,
```
ROMemories:
```suggestion
2. User-accessible general-purpose pins must only be used with some confirmation,
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+   either from automated testing or from manual input.
+
+   This means that it is safe to run any application on any bare board even if there are daughterboards
+   (eg. shields, wings, capes, clicks, grove modules)
```
ROMemories:
```suggestion
   (e.g., shields, wings, capes, clicks, or grove modules)
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+   are connected.
```
ROMemories: Not sure this is the correction you want:
```suggestion
   connected.
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+
+    > [!WARNING]
+    > This requirement has been recently introduced,
+    > and may not be upheld by all applications yet.
+
+    There is no sharp definition of a general purpose pin.
```
ROMemories:
```suggestion
    There is no sharp definition of a general-purpose pin.
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+    Some clear cases are the digital pins of an Arduino style shield (clearly general-purpose)
```
ROMemories:
```suggestion
    Some clear cases are the digital pins of an Arduino-style shield (clearly general-purpose)
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+2. User accessible general-purpose pins must only used be used with some confirmation,
+   either from automated testing or from manual input.
+
+   This means that it is safe to run any application on any bare board even if there are daughterboards
+   (eg. shields, wings, capes, clicks, grove modules)
+   are connected.
+
+    > [!WARNING]
+    > This requirement has been recently introduced,
+    > and may not be upheld by all applications yet.
```
ROMemories, line 117: I think this should be moved before "Practically," as this is also true for the first stage.

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+
+    There is no sharp definition of a general purpose pin.
+    Some clear cases are the digital pins of an Arduino style shield (clearly general-purpose)
+    and the I2C pins of a mikroBUS click (which are safe to drive in an I2C mode without extra checks).
+
+Applications in group 2 usually require some external components,
```
ROMemories, line 123: I'm not sure what "applications in group 2" refers to: is this the second item in the vertical list above? If so the list is introduced with the word "stage," and I'm not sure these two stages refer to applications.

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+sometimes just jumpers across some indicated pins,
+for example in peripheral loopback tests.
+Some means by which the application can adhere to this are:
+
+- using run-time indication (eg. reading a connected I2C EEPROM that confirms that some particular shield is connected),
```
ROMemories: Removing the hyphen makes it look less alien:
```suggestion
- using runtime indication (e.g., reading a connected I2C EEPROM that confirms that some particular shield is connected),
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+- using only low-strength driving modes (through internal pull-up/pull-down resistors), or
+- obtaining explicit user confirmation (e.g. by selecting a per-application module).
```
ROMemories:
```suggestion
- obtaining explicit user confirmation (e.g., by selecting a per-application module).
```

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+must be safe against damage on any board, including externally defined boards.
+They may alter the state of the system (eg. write to on-board memory),
+but must only drive GPIO pins based on board descriptions
+(or, until those are comprehensive enough, hard-coded knowledge of that board).
+
+Practically, this comes in two stages:
+
+1. Driven pins must only be selected based on the board (and not, for example, on the MCU family).
+
+   This means that it is safe to run any application on any bare board (with no extra wiring / shields).
```
kaspar030:
```suggestion
   This means that it is safe to run any application on any bare board (if no extra wiring / shields were added).
```

Make sure that the condition is about the board not *having* something added, and not about potentially *missing* something.

book/src/coding-conventions.md
kaspar030: I know what you are trying to convey here and I believe I agree, still I've been reading this a couple of times and I think I still couldn't reproduce it accurately. Let me try to explain in the next comment:

book/src/coding-conventions.md
chrysn: It is in this place because the one commit that you asked above to be removed (changing the application) makes sure that we do the easy thing (1), so that warning can be limited to the hard thing (2).

book/src/coding-conventions.md
```diff
@@ -91,6 +91,44 @@ You can quickly format the code in the ariel-os repository using this command:
...
+
+2. User accessible general-purpose pins must only used be used with some confirmation,
+   either from automated testing or from manual input.
+
+   This means that it is safe to run any application on any bare board even if there are daughterboards
```
kaspar030: So, is a board with a daughterboard attached still a bare board?

You mean that an application needs some confirmation (in whatever way) if it might do anything that might break anything if a daughterboard is connected?

chrysn:
```suggestion
   This means that it is safe to run any application on any board even if there are daughterboards
```

Sorry, that was bad wording.

Yes, I'd like applications that do things that become dangerous if anything is connected to the pins (be it jumpers, be it daughterboards) to require some form of confirmation.

The alternative is to say that for every test, always read the README if you have anything connected at all. But I think it's less of a hassle to try run the test, have it tell me to wire up these and those pins and then set `-s loopback-is-connected`, wire it up, and set that selection, *but only if it prompts me to*, and run all tests that don't need this in quiet, than to have to understand what every test is doing before I run it.

book/src/coding-conventions.md
ROMemories: > he one commit that you asked above to be removed (changing the application) makes sure that we do the easy thing

As I've said in my first review, other applications (`tests/i2c-controller`, `tests/spi-main`, `tests/spi-loopback`, and maybe others) would need to be modified for us to actually abide by (1), and I think that's better done in a separate PR, which could be done *after* landing this with the warning moved (so that at least we have the policy from now on).

PR #1784, opened 2026-02-08
feat(examples): Add USB ethernet option to the http-server example

bugadani:
This PR adds the `use-usb-ethernet` module to the http-server example. The module selects the usb-ethernet network interface and configures the device to use a static IP, and starts a DHCP server so the host side can get an IP.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,7 +292,7 @@ dependencies = [
  "embassy-nrf",
  "embassy-stm32",
  "embassy-sync 0.7.2",
- "embassy-time",
+ "embassy-time 0.5.0",
  "embassy-time-queue-utils",
  "embassy-usb",
  "embedded-hal 1.0.0",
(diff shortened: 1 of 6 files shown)


bugadani:
So the question is, is something like this desired? I have two paths in front of me:
 - fix up the issues of the PR
 - abandon it, now that the debugging has been done I personally no longer need this

kaspar030:
> So the question is, is something like this desired? I have two paths in front of me:
> 
>     * fix up the issues of the PR
> 
>     * abandon it, now that the debugging has been done I personally no longer need this

Sorry for the silence. :innocent:
 
Generally I think having a DHCP server is very desirable. In this PR it is added to just the http-server example, though. Would be great to have it alongside the other `network-config-FOO` options.

... there are some questions, like, probably this should conflict wifi (as we only support STA modes atm, and there is usually DHCP on the router already). It would make total sense with usb ethernet, and some sense with regular ethernet.

I wouldn't want to make you go through design iterations  here, so I suggest we close this here and we'll shamelessly copy the guts of it to some place more general.

bugadani:
> I wouldn't want to make you go through design iterations here, so I suggest we close this here and we'll shamelessly copy the guts of it to some place more general.

Sounds good, thanks :)

PR #1786, opened 2026-02-08
feat(esp): enable storage

bugadani:
# Description

This PR implements scanning the partition table for a `data/undefined` partition. The [ESP-IDF documentation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/partition-tables.html) doesn't give us a better predefined type, but we could even use our own if that's preferred.

The PR defines an alternate `sw/storage` context that omits the linker script. Instead, the storage implementation tries to look up the partition and prints a warning if it can't be found. The PR also modifies how the espflash command is built, and modifies the storage example to showcase how the partition table can be injected.

## Testing

I've ran the `storage` example on an ESP32-S3 devkit.

## Open Questions

The partition table required goes a bit against the portability aspects of Ariel OS. Is this acceptable?

## Changelog Entry

Storage support has been enabled for ESP32 devices.

## Change Checklist

### DIFF
--- a/book/src/boards/dfrobot-firebeetle-2-esp32-c6.md
+++ b/book/src/boards/dfrobot-firebeetle-2-esp32-c6.md
@@ -35,7 +35,7 @@ laze build -b dfrobot-firebeetle2-esp32-c6
 |Wi-Fi|<span title="supported">✅</span>|
 |Bluetooth Low Energy|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>|
 |Hardware Random Number Generator|<span title="supported">✅</span>|
-|Persistent Storage|<span title="available in hardware, but not currently supported by Ariel OS">❌</span>[^requires-partitioning-support]|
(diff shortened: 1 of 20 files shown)


kaspar030:
The PR looks great already!

> The partition table required goes a bit against the portability aspects of Ariel OS. Is this acceptable?

Hm. I guess there is no sensible default partition table.

Ideally we'd use some configuration to specify the size of the storage partition, then generate the partition table on-the-fly based on that and the actual flash size.
How much difference is there between different esp variants w.r.t. their requirements?
Can we break things, e.g., lose factory calibration or brick a device, by supplying a bad partition table?

bugadani:
> Hm. I guess there is no sensible default partition table.

There is a default partition table for each chip and both espflash and probe-rs program it unless you override it, but the default has no data partition.

> How much difference is there between different esp variants w.r.t. their requirements?

I'm not entirely sure, I think there are minor differences to where the bootloader and/or partition table need to be written, but not major ones. I'll need to check.

> Can we break things, e.g., lose factory calibration or brick a device, by supplying a bad partition table?

You can't brick these devices just by flashing a wrong partition table as long as you have physical access.

PR #1791, opened 2026-02-10
ci: autosquash before running Lint Commits

chrysn:
# Description

Currently, whenever fixups/squashes are present, Lint Commits is not helpful in two ways:

* It always fails, and users won't even look into it because "sure it fails because there are fixups".
* Even if they look in it, they'll need to look hard between all the fixup complaints to find if evern, and then which, text upset the check.

By rebasing before checking, fixups are removed -- that does introduce a source of errors, but then, if the fixups don't autosquash, that *is* an error to be addressed anyway.

## Testing

No clue; just hope and verify later by observing behavior?

[edit, added]: Tested in this PR, [see comment](https://github.com/ariel-os/ariel-os/pull/1791#issuecomment-3926626950).

## Open questions

Can we do that for DCO as well? I don't even find where that is in our CI steps.

[edit, added]: [No](https://github.com/ariel-os/ariel-os/pull/1791#issuecomment-3877435783), that'll just stay red as long as we use GitHub's tools for it.

## Issues/PRs References

Workarund-for: https://github.com/crate-ci/committed/issues/392

(which didn't really have any satisfactory conclusion, but then I also didn't read all the comments)

## Change Checklist

  self-review yes, I'd need help testing.

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -37,6 +37,7 @@ jobs:
           # Workaround-For: https://github.com/actions/checkout/issues/416
           git fetch origin "${GITHUB_BASE_REF}"
           git branch baseref origin/"${GITHUB_BASE_REF}"
+
           git config --global user.name "Committer of temporary autosquashes"
           git rebase baseref --autosquash --keep-base
 


.github/workflows/git.yml
```diff
@@ -21,6 +21,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-        with:
-          fetch-depth: 0
+      # If there are squash/fixup commits, the user will a) be aware of that
+      # and b) that's checked separately -- so we autosquash first. (If that
+      # doesn't work, *that* is an error of its own anyway).
+      - name: Rebase to avoid fixups/squashes from obscuring actual errors
+        run: git rebase ${{ github.base_ref }} --autosquash --keep-base
```
ROMemories: The benefit of doing it this way instead of [passing `--no-fixup` to committed](https://github.com/crate-ci/committed/blob/master/docs/reference.md#config-fields) is that it also checks that the fixup titles are valid, right?

ROMemories: (Generally I would prefer to have a blank line between CI steps, which makes the CI job easier to read at a glance for me.)

ROMemories: What values can `github.base_ref` take? If there's any way for a PR author, this could lead to [template injection](https://docs.zizmor.sh/audits/#template-injection).

ROMemories:
The DCO check is a [GitHub App](https://github.com/apps/dco).

.github/workflows/git.yml
chrysn: The benefit is that if you have a git history such as

```
commit be0454606b76586990d724facea4475488fda65c (HEAD -> main)
Author: chrysn <chrysn@fsfe.org>
Date:   Tue Feb 10 14:08:17 2026 +0100

    amend! bar
    
    Do bar properly

commit c86c61a0d114e22d7d23f476148ae781f61e8c0a
Author: chrysn <chrysn@fsfe.org>
Date:   Tue Feb 10 13:28:28 2026 +0100

    fixup! bar

commit cde2078ff11a10edaa9ae0aa88ef5f5b4043a5da
Author: chrysn <chrysn@fsfe.org>
Date:   Tue Feb 10 13:28:22 2026 +0100

    bar

commit f2ddf2a16a280c674914044ed35f5aaa7051f1ec
Author: chrysn <chrysn@fsfe.org>
Date:   Tue Feb 10 13:28:12 2026 +0100

    foo
```

then it will not check "bar" or "fixup! bar", but just:

```
commit 4513c26010b85aa13c8edeb194b0735e793d1584 (HEAD -> main)
Author: chrysn <chrysn@fsfe.org>
Date:   Tue Feb 10 13:28:22 2026 +0100

    Do bar properly

commit f2ddf2a16a280c674914044ed35f5aaa7051f1ec
Author: chrysn <chrysn@fsfe.org>
Date:   Tue Feb 10 13:28:12 2026 +0100

    foo
```

If this were provided in "committed", it'd be great, but as that issue was closed, I'd guess it's out of scope there.

chrysn: Not all PRs are against main. I *think* that if it's a PR in our repo, its base will always be a ref of ours. (But I don't know the GH CI security model well enough to tell, and frankly, don't trust CI with any secrets that do more than soft DoS.)

chrysn:
> The DCO check is a [GitHub App](https://github.com/apps/dco).

I'll just put that on my mental "things that are weird because GitHub" list then, and not bother with it any more.

.github/workflows/git.yml
ROMemories: Makes sense, I wasn't familiar with `amend!`, thanks!

ROMemories: Generally when in doubt I would err on the side of caution, but I guess in this case it's fine (and it *seems* that fortunately branch names can't contain spaces).

> don't trust CI with any secrets that do more than soft DoS

`GITHUB_TOKEN` is always present and [has too many permissions by default](https://docs.zizmor.sh/audits/#excessive-permissions). #1705 should address that.

.github/workflows/git.yml
```diff
@@ -21,6 +21,11 @@ jobs:
...
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-        with:
-          fetch-depth: 0
+
+      # If there are squash/fixup commits, the user will a) be aware of that
+      # and b) that's checked separately -- so we autosquash first. (If that
+      # doesn't work, *that* is an error of its own anyway).
+      - name: Rebase to avoid fixups/squashes from obscuring actual errors
+        run: git rebase ${{ github.base_ref }} --autosquash --keep-base
```
ROMemories: The check doesn't actually seem to be happy about the `base_ref` value:

```sh
Run git rebase main --autosquash --keep-base
fatal: invalid upstream 'main'
```

(While you're at it it would actually likely be better to use the [remediation recommended by zizmor](https://docs.zizmor.sh/audits/#remediation_25) to be on the safe side).

.github/workflows/git.yml
chrysn: In that case this definitely needs to be careful; going with the environment variable as zizmor suggests.

chrysn:
This is ready for another round of review after a lot of trial-and-error.

The current CI run also serves as a test: I've left one fixup in, "block fixup" still triggers, but comitted [sees](https://github.com/ariel-os/ariel-os/actions/runs/22179744903/job/64137653112?pr=1791#step:4:36) the 3 after squashing:

```

./committed
Linting commits:
* ff50ca77 ci: make Lint Commits act on the squashed commits
* 779a7176 ci: manually fetch base branch for commit message linting
* ddbb4c3a ci: autosquash before running Lint Commits

Against 'committed.toml':
```

.github/workflows/git.yml
```diff
@@ -22,5 +22,27 @@ jobs:
...
+          # not in GitHub's environment).
           fetch-depth: 0
+
+      # If there are squash/fixup commits, the user will a) be aware of that
+      # and b) that's checked separately -- so we autosquash first. (If that
+      # doesn't work, *that* is an error of its own anyway).
+      - name: Rebase to avoid fixups/squashes from obscuring actual errors
+        run: |
+          # Workaround-For: https://github.com/actions/checkout/issues/416
+          git fetch origin "${GITHUB_BASE_REF}"
```
ROMemories, line 38: Finding [the docs for `GITHUB_BASE_REF`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#default-environment-variables) wasn't that easy, but they seem to say that the variable wouldn't be set for the `merge_group` trigger, and this would therefore prevent the PR to actually land by failing in the queue. I suppose the step could be `if`-gated based on the event that triggered the workflow (and the `committed` step would need to be adjusted).

When looking for the docs, I've also stumbled upon [this StackOverflow question (from 2020)](https://stackoverflow.com/q/62097282) that says that this variable is/was only set for PRs from forks. I have no idea whether that's true, but if it is that would be a problem, at least for Renovate PRs, which are opened directly on the repo.

PR #1790, opened 2026-02-10
refactor(spi)!: alleviate the need for `SpiDevice` to be `'static`

ROMemories:
# Description

What the changelog entry says.

## Testing

Successfully tested this on the four boards:

```sh
laze -C tests/spi-main/ build -b espressif-esp32-c6-devkitc-1 -b nrf52840dk -b rpi-pico-w -b stm32u083c-dk
```

## Open Questions

- Companion to #1793

## Changelog Entry

A lifetime generic has been introduced on `ariel_os::spi::main::SpiDevice`. This avoids the need for storing it in a global `OnceCell` (or similar) inside applications. It is a breaking change but should not require changes to many existing applications in practice.

## Change Checklist

- [x] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1793, opened 2026-02-10
refactor(i2c)!: alleviate the need for `I2cDevice` to be `'static`

ROMemories:
# Description

What the changelog entry says.

## Testing

Tested by building with the following:

```sh
laze -C tests/i2c-controller/ build -b espressif-esp32-c6-devkitc-1 -b nrf52840dk -b rpi-pico-w -b st-steval-mkboxpro
```

## Issues/PRs References

- Companion to #1790

## Changelog Entry

A lifetime generic has been introduced on `ariel_os::i2c::controller::I2cDevice`. This avoids the need for storing it in a global `OnceCell` (or similar) inside applications. It is a breaking change but should not require changes to many existing applications in practice.

## Change Checklist

### DIFF
--- a/tests/i2c-controller/src/main.rs
+++ b/tests/i2c-controller/src/main.rs
@@ -54,10 +54,6 @@ cfg_if::cfg_if! {
     }
 }
 
-pub static I2C_BUS: once_cell::sync::OnceCell<
-    Mutex<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, hal::i2c::controller::I2c>,
-> = once_cell::sync::OnceCell::new();
-
 #[ariel_os::task(autostart, peripherals)]
 async fn main(peripherals: pins::Peripherals) {
     let mut i2c_config = hal::i2c::controller::Config::default();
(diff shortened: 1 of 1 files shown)


tests/i2c-controller/src/main.rs
```diff
@@ -54,21 +54,16 @@ cfg_if::cfg_if! {
     }
 }
 
-pub static I2C_BUS: once_cell::sync::OnceCell<
-    Mutex<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, hal::i2c::controller::I2c>,
-> = once_cell::sync::OnceCell::new();
```
ROMemories, line 59: A similar construct is used in the `sensors-debug` and `thermometer` examples. This PR doesn't change those because the bus is used for sensors and the architecture is a bit different. It's not clear whether it would be beneficial to refactor these as well, and I think it's better to wait until we have more visibility on how the sensor configuration/initialization is codegened.

ROMemories:
Ran the CI with `ci-build:stm32` and then `ci-build:small`.

PR #1797, opened 2026-02-11
feat(ci): also check if check-labels succeeded

nponsard:
# Description

The "Final Results" job of the Build workflow checks if the build job has succeeded or has been skipped. If the build job failed, "Final Results" fail. 
There is an issue where when there is no labels, the check-labels job fails, the build job is skipped and Final Results is considered a success. In my opinion the Final Results check should fail in that case. 

## Testing

Do a ci run without build labels, "Final Results" should fail.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -293,7 +293,8 @@ jobs:
     name: Final Results
     needs: [build, check-labels]
     steps:
-      - run: |
+      - name: Check if build jobs succeeded
+        run: |
           result="${{ needs.build.result }}"
           if [[ $result == "success" || $result == "skipped" ]]; then
             exit 0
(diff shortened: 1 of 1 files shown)


ROMemories:
> In my opinion the Final Results check should fail in that case.

I'm… not sure? Unless I'm missing something this can never result in the CI incorrectly passing because `check-labels` is also a required check, can it? That means that we would have two jobs failing while there is only one action required (which is something that #1791 is trying to avoid).

The Build *workflow* is already considered failing in this case, right? So this is not required for #1715, is it?

nponsard:
> The Build workflow is already considered failing in this case, right? So this is not required for https://github.com/ariel-os/ariel-os/pull/1715, is it?

I thought it would affect #1715, but I re-did some tests and the job is correctly skipped if check-labels fails even though Final Reults succeeds.

ROMemories:
> > The Build workflow is already considered failing in this case, right? So this is not required for #1715, is it?
> 
> I thought it would affect #1715, but I re-did some tests and the job is correctly skipped if check-labels fails even though Final Reults succeeds.

Thanks for testing, that was my understanding as well because the check-labels job failing should fail the whole workflow, which is what #1715 relies on.

If this is not functionally required, I'd rather keep only one check failing in case no labels are attached as explained above. Closing, we can revive this if we realize later that this is functionally needed (or if I've missed something).

PR #1799, opened 2026-02-11
docs(glossary): add serial-related terms to the glossary

ROMemories:
# Description

This clarifies the meaning and relationships of serial-related terms. As in #1745, this doesn't aim to fully define them, but instead emphasize their relationships.

## Changelog Entry

Not adding to the changelog, because #1745 is already there.

## Change Checklist


bergzand (approved):
I find it a bit confusing how this leaves out any hint of what it is (besides a peripheral), but that seems to be your goal here.

Issue #1782, opened 2026-02-06
Keeping online documentation for our different releases


ROMemories:
Currently we have two types of documentation: rustdoc docs, and the mdbook book. These are exposed at the respective URLs:

- <https://ariel-os.github.io/ariel-os/dev/docs/api/>
- <https://ariel-os.github.io/ariel-os/dev/docs/book/>

In these URLs `dev` refers to the development version, built from the `main` branch.

This means that the online docs of our previous releases gets overridden by the current state of `main`. It would be really useful to also keep online the docs of our previous releases, e.g., at the following URLs:

- <https://ariel-os.github.io/ariel-os/0.3.0/docs/api/>
- <https://ariel-os.github.io/ariel-os/0.3.0/docs/book/>

Note that, even though our tags follow the `v{version}` format, the URL paths should likely be just `/{version}/` instead, for consistency with docs.rs.

- [ ] To do this, after some out-of-band discussion, we found that the simplest and more robust option was likely to store the rustdoc docs and the mdbook docs in a dedicated repo in the `ariel-os` organization. When setting this up initially, these docs can be generated and pushed manually, for the four past releases we have.
- [ ] Then, the [CI workflow publishing the docs](https://github.com/ariel-os/ariel-os/blob/841322f9078f287291deb6e7c856c0f646906228/.github/workflows/build-deploy-docs.yml) on each push on `main` can be modified to assemble these already existing docs with the new docs from `main`, to have docs at `/dev/`, `/0.3.0/`, `/0.2.1/`, `/0.2.0/`, `/0.1.0/` all at the same time.
- [ ] When this is done, we'll need to find a solution for the already existing `/dev/` links to point at the version they should (it won't of course be possible to do this retroactively, but this should be doable for new versions).
- [ ] Later, we can consider automatically pushing the new docs to the docs repo, using a tag-based trigger.

ROMemories:
This should address #1114.

anlavandier:
> * [x] To do this, after some out-of-band discussion, we found that the simplest and more robust option was likely to store the rustdoc docs and the mdbook docs in a dedicated repo in the `ariel-os` organization. When setting this up initially, these docs can be generated and pushed manually, for the four past releases we have.

A [repo](https://github.com/ariel-os/docs) was opened with the previous versions of the docs.

ROMemories:
> A [repo](https://github.com/ariel-os/docs) was opened with the previous versions of the docs.

[...]

PR #1802, opened 2026-02-13
ci: Build documentation and book for older releases 

anlavandier:
# Description

This PR changes the deploy docs CI to also retrieve the previous version located in [ariel-os/docs](https://github.com/ariel-os/docs). 

## Issues/PRs References

Advances #1782 

## Open Questions

* How are users expected to select between versions ? Currently, this can only be done manually by changing the `/dev/` in the URL to the correct tag. 

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -168,6 +168,26 @@ jobs:
           mdbook build
           mv book ../_site/dev/docs
 
+      - name: Checkout older docs and book
+        if: github.event_name != 'pull_request'
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          repository: ariel-os/docs
+          ref: main
+          path: ./_older-docs
+
+      - name: Move older docs
+        if: github.event_name != 'pull_request'
+        run: |
+          for dir in ./_older-docs/tags/*/;
+          do
+            dir="${dir%*/}" # Remove trailing slash
+            tag="${dir##*/}" # Extract the release tag
+            mkdir -p "./_site/$tag/docs" # Create the directory for the versionned docs
+            mv "$dir/api" "./_site/$tag/docs/api" # Move the rustdoc API documentation
+            mv "$dir/book" "./_site/$tag/docs/book" # Move the book
+          done
+
       - name: Hash docs tree
         run: find ./_site -type f | xargs sha256sum | sha256sum > ./_hash && mv ./_hash ./_site/
 


anlavandier:
Once the PR is ready to land, the extra documentations versions can be removed from the docs preview by adding their respective dirs to a [`.surgeignore`](https://surge.sh/help/ignoring-files-and-directories) file.

ROMemories:
> the extra documentations versions can be removed from the docs preview by adding their respective dirs to a [`.surgeignore`](https://surge.sh/help/ignoring-files-and-directories) file

Instead, would it be possible to run the new job steps only when the workflow is triggered by a push (to main) event? That would avoid a little bit of work in the workflow and would avoid relying on a surge-specific feature.

anlavandier:
> > the extra documentations versions can be removed from the docs preview by adding their respective dirs to a [`.surgeignore`](https://surge.sh/help/ignoring-files-and-directories) file
> 
> Instead, would it be possible to run the new job steps only when the workflow is triggered by a push (to main) event? That would avoid a little bit of work in the workflow and would avoid relying on a surge-specific feature.

Done in [9644e4d](https://github.com/ariel-os/ariel-os/pull/1802/commits/9644e4d84dfef29d61d95793ceff46ea47da9dae)

anlavandier:
If the open question doesn't need to be answered right now then I think that this can be opened for reviews and subsequently merged

.github/workflows/build-deploy-docs.yml
```diff
@@ -168,6 +168,27 @@ jobs:
...
+      - name: Move older docs
+        if: github.event_name != 'pull_request'
+        run: |
+          for dir in ./_older-docs/tags/*/;
+          do
+            dir=${dir%*/}; # Remove trailing slash
+            tag=${dir##*/}; # Extract the release tag
+            mkdir -p ./_site/$tag/docs; # Create the directory for the versionned docs
+            mv $dir/api ./_site/$tag/docs/api; # Move the rustdoc API documentation
+            mv $dir/book ./_site/$tag/docs/book; # Move the book
```
ROMemories: This removes the semicolons as I don't think they are necessary, even in loops, and adds quoting. Generally [everything that *can* be quoted should be](https://github.com/koalaman/shellcheck/wiki/SC2046):
```suggestion
            dir="${dir%*/}" # Remove trailing slash
            tag="${dir##*/}" # Extract the release tag
            mkdir -p "./_site/$tag/docs" # Create the directory for the versionned docs
            mv "$dir/api" "./_site/$tag/docs/api" # Move the rustdoc API documentation
            mv "$dir/book" "./_site/$tag/docs/book" # Move the book
```

.github/workflows/build-deploy-docs.yml
```diff
@@ -168,6 +168,27 @@ jobs:
...
+          done;
+
```
ROMemories: No need for a semicolon in this position:
```suggestion
          done
```

ROMemories (commented):
It would likely be a good idea to comment out the `if`s to re-test this once before landing.

ROMemories:
> * How are users expected to select between versions ? Currently, this can only be done manually by changing the `/dev/` in the URL to the correct tag.

That's a problem for later, for now we just want them to be hosted, and be able to link to them.

anlavandier:
> It would likely be a good idea to comment out the `if`s to re-test this once before landing.

You can test the current state of the PR. I'm going to prepare a squashed and rebased commit with the `if`s uncommented to clean the history. Once you tell me that things look good on your end I'm pushing this commit and enabling auto-merge

ROMemories (approved):
Successfully tested the API docs and the book for the five versions: `dev`, `0.3.0`, `0.2.1`, `0.2.0`, `0.1.0`.

You can squash the first commits and remove the latest one as you said.

PR #1808, opened 2026-02-16
fix(rt): sleep instead of busy-looping

ROMemories:
# Description

This makes sure that the loop inside `ariel-os-rt` doesn't busy-loop when not using multithreading.

## Testing

Testing involves adding a log statement [inside the loop](https://github.com/ariel-os/ariel-os/blob/d5329fffc983517d2ff9bd2bbcf0965f26045bd1/src/ariel-os-rt/src/lib.rs#L191) (e.g., `ariel_os_debug::log::info!("looping");`) and checking that it gets printed way fewer times with this PR, for instance in the `minimal` example (be aware that testing this on main might crash your terminal given how many log lines will be printed). How often the printing happens depends on the HAL used (not just the processor architecture) because on STM32 the (enabled-by-default) time driver seems to be triggering more wake-ups for instance.

Testing this is more fun when mixing it with #1809.

Tested on:

- nrf52840dk
- stm32u083c-dk

I'm not sure how to test this on RISC-V and Xtensa as the only implementations we have are on ESP32s, on which we don't provide an SWI, which the `executor-interrupt` requires.

## Issues/PRs References

- Depends on #1809

## Changelog Entry

The internal runtime loop is now not busy-looping anymore when using the `executor-interrupt` flavor. This may bring significant power consumption improvements when using that executor flavor.

## Change Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ featurecomb = "0.2.0"
 portable-atomic = { version = "1.11.0", default-features = false, features = [
   "require-cas",
 ] }
+riscv = { version = "0.15.0", default-features = false }
 semihosting = { version = "0.1.24", default-features = false }
 
 embassy-embedded-hal = { version = "0.5.0", default-features = false }
--- a/src/ariel-os-rt/src/cortexm.rs
+++ b/src/ariel-os-rt/src/cortexm.rs
@@ -218,6 +218,11 @@ pub fn init() {
     }
 }
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn wfi() {
+    cortex_m::asm::wfi();
+}
+
 /// Returns a `Stack` handle for the currently active thread.
 pub(crate) fn stack() -> crate::stack::Stack {
     #[cfg(feature = "threading")]
--- a/src/ariel-os-rt/src/lib.rs
+++ b/src/ariel-os-rt/src/lib.rs
@@ -44,6 +44,7 @@ cfg_if::cfg_if! {
             use crate::stack::Stack;
 
             pub fn init() {}
+            pub fn wfi() {}
             pub fn sp() -> usize { 0 }
             pub fn stack() -> Stack { Stack::default() }
         }
@@ -187,8 +188,10 @@ fn startup() -> ! {
     {
         #[cfg(test)]
         test_main();
-        #[allow(clippy::empty_loop)]
-        loop {}
+
+        loop {
+            arch::wfi();
+        }
     }
 }
 
--- a/src/ariel-os-rt/src/riscv.rs
+++ b/src/ariel-os-rt/src/riscv.rs
@@ -9,6 +9,11 @@ fn main() -> ! {
 
 pub fn init() {}
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn wfi() {
+    riscv::asm::wfi();
+}
+
 /// Returns the current `SP` register value.
 pub(crate) fn sp() -> usize {
     let sp: usize;
--- a/src/ariel-os-rt/src/xtensa.rs
+++ b/src/ariel-os-rt/src/xtensa.rs
@@ -9,6 +9,16 @@ fn main() -> ! {
 
 pub fn init() {}
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn wfi() {
+    // The options are similar to those used for wfi on RISC-V and Cortex-M:
+    // the instruction does not modify memory or the stack, and does preserve flags.
+    // SAFETY: executing `waiti 0` is sound.
+    unsafe {
+        core::arch::asm!("waiti 0", options(nomem, nostack, preserves_flags));
+    }
+}
+
 /// Returns the current stack pointer register value
 pub(crate) fn sp() -> usize {
     let sp: usize;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -520,6 +520,7 @@ dependencies = [
  "ld-memory",
  "linkme",
  "portable-atomic",
+ "riscv",
 ]
 
 [[package]]
--- a/src/ariel-os-rt/Cargo.toml
+++ b/src/ariel-os-rt/Cargo.toml
@@ -40,6 +40,9 @@ portable-atomic = { workspace = true }
 [target.'cfg(context = "nrf")'.dependencies]
 cortex-m = { workspace = true, features = ["critical-section-single-core"] }
 
+[target.'cfg(context = "riscv")'.dependencies]
+riscv = { workspace = true }
+
 [target.'cfg(context = "stm32")'.dependencies]
 cortex-m = { workspace = true, features = ["critical-section-single-core"] }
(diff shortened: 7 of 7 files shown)


src/ariel-os-rt/src/cortexm.rs
```diff
@@ -218,6 +218,10 @@ pub fn init() {
     }
 }
 
+pub fn sleep_now() {
+    cortex_m::asm::wfe();
```
ROMemories: Would it be better to use `wfi` instead, in this crate?

src/ariel-os-rt/src/riscv.rs
```diff
@@ -9,6 +9,14 @@ fn main() -> ! {
 
 pub fn init() {}
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn sleep_now() {
+    // SAFETY: executing WFI is safe.
+    unsafe {
+        core::arch::asm!("wfi");
```
ROMemories: Still need to do more research on whether this needs memory barriers on this architecture.

src/ariel-os-rt/src/xtensa.rs
```diff
@@ -9,6 +9,14 @@ fn main() -> ! {
 
 pub fn init() {}
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn sleep_now() {
+    // SAFETY: executing WFI is safe.
+    unsafe {
+        core::arch::asm!("wfi");
```
ROMemories: Still need to do more research on whether this needs memory barriers on this architecture.

nponsard:
```suggestion
        core::arch::asm!("waiti 0");
```

According to `ariel-os-threads/src/arch/xtensa.rs`

ROMemories: Ah right, I've missed that because this actually never gets used anyway, so it was compiling fine; thanks!

src/ariel-os-rt/src/lib.rs
```diff
@@ -44,6 +44,7 @@ cfg_if::cfg_if! {
             use crate::stack::Stack;
 
             pub fn init() {}
+            pub fn sleep_now() {}
```
ROMemories: Is this name appropriate in this crate?

nponsard: Should this be named `wfi` like in `ariel-os-threads` ? Or maybe `wait_for_interrupt` ?

src/ariel-os-rt/src/cortexm.rs
kaspar030: yes, let's use `wfi`. Not sure about "better", but `wfi` has worked everywhere in the past, and we're also using that in the idle loops in `ariel-os-threads`.

src/ariel-os-rt/src/lib.rs
kaspar030: Or `cpu_idle()`?

kaspar030: or just `idle()`?

kaspar030 (approved):

Cargo.toml
```diff
@@ -73,6 +73,7 @@ featurecomb = "0.1.3"
 portable-atomic = { version = "1.11.0", default-features = false, features = [
   "require-cas",
 ] }
+riscv = { version = "0.15.0", default-features = false }
```
ROMemories, line 77: This technically introduces a new dependency, but when used with `ariel-os-esp`, `riscv` is already brought in by `esp-hal` anyway (that's why this doesn't use the latest, v0.16 version, even though this likely doesn't make a difference anyway as it's only used for `wfi()` currently, which is the same across these versions).

src/ariel-os-rt/src/riscv.rs
```diff
@@ -9,6 +9,11 @@ fn main() -> ! {
 
 pub fn init() {}
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn wfi() {
+    riscv::asm::wfi();
```
ROMemories, line 14: Using `riscv`'s implementation avoids an `unsafe` block, and avoids having to provide the right set of `asm!` options.

src/ariel-os-rt/src/xtensa.rs
```diff
@@ -9,6 +9,16 @@ fn main() -> ! {
 
 pub fn init() {}
 
+#[allow(dead_code, reason = "conditional compilation")]
+pub fn wfi() {
+    // The options are similar to those used for wfi on RISC-V and Cortex-M:
+    // the instruction does not modify memory or the stack, and does preserve flags.
```
ROMemories, line 15: [`esp-hal` doesn't actually specify these options](https://github.com/esp-rs/esp-hal/blob/c729564b17637d1e3e95ac0def95fbad734b5d90/esp-hal/src/interrupt/xtensa.rs#L275), but they were also [missing from `esp-hal-embassy` on RISC-V](https://github.com/esp-rs/esp-hal/blob/6b758d75b878d3ebf671a89c28de787551bdbdbc/esp-hal-embassy/src/executor/thread.rs#L231), so I suppose they're just missing upstream. I didn't not observe the slightest reduction in binary size when adding these options however.

It's still a bit unclear whether this needs a memory barrier, but the ISA documentation does not clearly state that it needs one. The documentation of `PS.INTLEVEL` does hint to using `RSYNC`, but I think that wouldn't make sense when modifying `PS.INTLEVEL` through `waiti`.

### References

- [Documentation of `asm!` options](https://doc.rust-lang.org/reference/inline-assembly.html#options).
- [Documentation of the Xtensa ISA](https://www.cadence.com/content/dam/cadence-www/global/en_US/documents/tools/silicon-solutions/compute-ip/isa-summary.pdf).

ROMemories:
I've done a new round of testing, it all seems to behave as expected.

Cargo.toml
nponsard: Another way would be to directly import esp-hal and use `esp_hal::riscv`.

kaspar030 (approved):

Cargo.toml
ROMemories: Yes, but that would only work because the only RISC-V MCUs we have right now happen to be ESPs.

src/ariel-os-rt/src/cortexm.rs
```diff
@@ -218,6 +218,10 @@ pub fn init() {
     }
 }
 
+pub fn wfi() {
+    cortex_m::asm::wfi();
+}
```
nponsard, line 224:
```suggestion
#[allow(dead_code, reason = "conditional compilation")]
pub fn wfi() {
    cortex_m::asm::wfi();
}
```

ROMemories: Indeed, otherwise it shows up with `laze -C examples/log/ build -s sw/threading -b nrf52840dk`.

kaspar030:
please squash. @nponsard I think your request is fixed?

PR #1809, opened 2026-02-16
refactor!: enable the time drivers only when `time` is enabled

ROMemories:
# Description

This makes sure time drivers provided by HALs are only enabled when the `time` Cargo feature is enabled. This has two benefits: when time is actually not needed it reduces unneeded wake-ups and reduces power consumption, and it would us to provide a different implementation of time drivers later if needed.

This is however a breaking change as users might have been relying on time drivers being always provided even if `time` was not enabled.

This can easily be split up if needed.

## Testing

Successfully tested the `blinky` example (which enables `time`)/the `i2c-controller` example (on boards without LEDs, by letting the example timeout) and the `log` example (which doesn't), on the following:

- espressif-esp32-c6-devkitc-1
- espressif-esp32-s3-devkitc-1
- rpi-pico-w
- stm32u083c-dk
- nrf52840dk

#1808 makes this easier to test, as it makes wake-ups more visible (at least on STM32).

## Changelog Entry

The `embassy-time` time driver is now *only* provided by HALs when the already-existing `time` Cargo feature is enabled. Existing applications that relied on it being always enabled will need to select that Cargo feature.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/Cargo.toml
+++ b/src/ariel-os-hal/Cargo.toml
@@ -74,6 +74,7 @@ spi = [
   "ariel-os-stm32/spi",
 ]
 
+## Enables a time driver, and functionality requiring `embassy-time`.
 time = [
   "ariel-os-esp/time",
   "ariel-os-nrf/time",


src/ariel-os-nrf/Cargo.toml
```diff
@@ -88,6 +86,8 @@ i2c = ["ariel-os-embassy-common/i2c"]
 ## Enables SPI support.
 spi = ["ariel-os-embassy-common/spi"]
 
+time = ["embassy-nrf/time", "embassy-nrf/time-driver-rtc1"]
```
ROMemories, line 90: Some documentation wouldn't hurt.

Edit: but this is tied to the guarantees we actually want on `ariel-os`, so I'll write it when that is settled.

src/ariel-os/Cargo.toml
```diff
@@ -56,6 +56,7 @@ threading = [
   "ariel-os-rt/threading",
 ]
 ## Enables the internal executor's timer queue, required for timer support.
+## Additionally enables the HAL-specific time driver.
```
ROMemories: We need to think carefully about what we want to guarantee here.

ROMemories: Do we want to specifically mention that this is Embassy's time driver? That would allow having an additive Cargo feature later (or even deprecate `time` if it gets to that) if we realize we need a separate one when doing power management (as `embassy-time` seems to be making the assumption that it actually counts ticks since startup).

src/ariel-os-hal/Cargo.toml
```diff
@@ -74,7 +74,12 @@ spi = [
   "ariel-os-stm32/spi",
 ]
 
-time = ["ariel-os-esp/time"]
+time = [
+  "ariel-os-esp/time",
```
ROMemories, line 79: There is nothing to do for ESP in this PR, because that was done as part of #1328.

src/ariel-os/Cargo.toml
kaspar030: I'm not sure we need to mention anything more here. Let's improve when we take more control or offer more options.

kaspar030: We're currently hard-selecting *some* timer base, e.g., rtc1 on nrf and "any" on stm32. This PR doesn't change that, and fixing that is orthogonal, I think.

Maybe make the first line here just say "Enables timing functionality." and keep the other too as informational.

src/ariel-os-nrf/Cargo.toml
ROMemories: I have added documentation, including in `ariel-os-hal`.

src/ariel-os/Cargo.toml
ROMemories: > Maybe make the first line here just say "Enables timing functionality." and keep the other too as informational.

Did that, hopefully that does the job.

> We're currently hard-selecting _some_ timer base, e.g., rtc1 on nrf and "any" on stm32.

We're also not yet appropriately selecting a tick rate, and the default one is likely too high on STM32 (but this is also orthogonal).

PR #1817, opened 2026-02-24
feat: add board seeedstudio-xiao-nrf52840-plus

chriscate:
# Description

Adding board support for Seeed Studio XIAO nRF52840 Plus.

## Testing

Built examples and also flashed a separate blinky to the board to verify it works.

## Changelog Entry

The Seeed Studio XIAO nRF52840 Plus board is now supported.

## Change Checklist

### DIFF
--- a/boards/seeedstudio-xiao-nrf52840-plus.yaml
+++ b/boards/seeedstudio-xiao-nrf52840-plus.yaml
@@ -1,0 +1,24 @@
+boards:
+  seeedstudio-xiao-nrf52840-plus:
+    chip: nrf52840
+    flags:
+      - has_usb_device_port
+    leds:
+      led0:
+        pin: P0_26
+        color: red
+        active: low
+        aliases:
+          - USER_LED_R
+      led1:
+        pin: P0_06
+        color: blue
+        active: low
+        aliases:
+          - USER_LED_B
+      led2:
+        pin: P0_30
+        color: green
+        active: low
+        aliases:
+          - USER_LED_G
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -980,6 +980,13 @@ builders:
     tier: "2"
     support:
 
+  seeedstudio-xiao-nrf52840-plus:
+    chip: nrf52840
+    tier: "3"
+    support:
+      user_usb: needs_testing
+      ethernet_over_usb: needs_testing
+
   seeedstudio-lora-e5-mini:
     chip: stm32wle5jc
     tier: "3"
@@ -1210,6 +1217,11 @@ boards:
     url: https://web.archive.org/web/20250827114042/https://wiki.seeedstudio.com/xiao_esp32c6_getting_started/
     builders:
       - seeedstudio-xiao-esp32c6
+
+  - name: Seeed Studio XIAO NRF52840 Plus
+    url: https://web.archive.org/web/20260101232815/https://wiki.seeedstudio.com/XIAO_BLE/
+    builders:
+      - seeedstudio-xiao-nrf52840-plus
 
   - name: Seeed Studio LoRa-E5 mini
     url: https://web.archive.org/web/20250802201959/https://wiki.seeedstudio.com/LoRa_E5_mini/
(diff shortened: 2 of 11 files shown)


nponsard:
I'm curious, how do you flash this board? Do you connect a j-link programmer to the pads under it?

boards/seeedstudio-xiao-nrf52840-plus.yaml
```diff
@@ -0,0 +1,22 @@
+boards:
+  seeedstudio-xiao-nrf52840-plus:
+    chip: nrf52840
+    leds:
+      led0:
```
kaspar030, line 7: confirmed LED pins, silkscreen names and `active: low` with upstream wiki. ✔️

kaspar030 (commented):
looks good already

chriscate:
> I'm curious, how do you flash this board? Do you connect a j-link programmer to the pads under it?

I personally use [this expansion board](https://wiki.seeedstudio.com/Seeeduino-XIAO-Expansion-Board/) just for flashing which breaks out the swd pads to male headers (using spring pins under the board).

nponsard:
> > I'm curious, how do you flash this board? Do you connect a j-link programmer to the pads under it?
> 
> I personally use [this expansion board](https://wiki.seeedstudio.com/Seeeduino-XIAO-Expansion-Board/) just for flashing which breaks out the swd pads to male headers (using spring pins under the board).

Thanks for the link !

boards/seeedstudio-xiao-nrf52840-plus.yaml
```diff
@@ -0,0 +1,22 @@
+boards:
+  seeedstudio-xiao-nrf52840-plus:
```
nponsard, line 2: For `user_usb` and `ethernet_over_usb` to be considered available on this board, you need to add 

```yaml
    flags:
      - has_usb_device_port
```

doc/support_matrix.yml
```diff
@@ -980,6 +980,13 @@ builders:
     tier: "2"
     support:
 
+  seeedstudio-xiao-nrf52840-plus:
+    chip: nrf52840
+    tier: "3"
+    support:
+      user_usb: supported
+      ethernet_over_usb: supported
```
nponsard: The USB looks correctly wired in the schematics.

Have you tested the examples `examples/udp-echo` and `examples/usb-serial` to confirm this works ? 

Otherwise these fields should be set to `needs_testing`.

boards/seeedstudio-xiao-nrf52840-plus.yaml
chriscate: added this flag

doc/support_matrix.yml
chriscate: swapped these to `needs_testing`

chriscate:
> Thank you for your contribution! Some small changes are needed before we merge this.

@nponsard I made the requested updates. One thing I noticed is now when I attempt to build the examples, I run into an issue with the http-server example. Previously I was able to build the examples although now it's attempting to build 45 and previously it was 35.

Here's the output & I did confirm I have the target and also tried running `cargo build -Zbuild-std` as it recommends, but it gave me an error saying that was for nightly only (I do have nightly it's just not my default so maybe that's why?).

> [11/45] CARGO build/bin/seeedstudio-xiao-nrf52840-plus/http-server/http-server.elf
   Compiling proc-macro2 v1.0.103
   Compiling defmt v1.0.1
   Compiling thiserror v2.0.18
   Compiling defmt-macros v1.0.1
   Compiling bitflags v1.3.2
   Compiling critical-section v1.2.0
   Compiling serde_core v1.0.228
error[E0463]: can't find crate for `core`
  |
  = note: the `thumbv7em-none-eabihf` target may not be installed
  = help: consider downloading the target with `rustup target add thumbv7em-none-eabihf`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`
For more information about this error, try `rustc --explain E0463`.
error: could not compile `bitflags` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `critical-section` (lib) due to 1 previous error
FAILED: build/bin/seeedstudio-xiao-nrf52840-plus/http-server/http-server.elf

kaspar030:
> I run into an issue with the http-server example.

That's probably because `http-server` needs the nightly toolchain (with the needed components and targets). Try if `laze build -s nightly install-toolchain` fixes this.

chriscate:
> > I run into an issue with the http-server example.
> 
> That's probably because `http-server` needs the nightly toolchain (with the needed components and targets). Try if `laze build -s nightly install-toolchain` fixes this.

Thanks, that resolved it. All of the examples built.

kaspar030:
Thanks @chriscate!

chriscate:
Awesome thanks for the guidance to get this in @kaspar030 and @nponsard !

PR #1823, opened 2026-02-25
chore(renovate): add a changelog:skip label by default on Renovate PRs

anlavandier:
# Description

This PR changes the renovate configuration to add a `changelog:skip`  label to renovate PRs by default.
This PR also changes the extension of the renovate configuration file from `.json` to `.json5` so that editors stop complaining about the presence of comments in the renovate configuration file. 

## Change Checklist

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -73,7 +73,8 @@
     }
   ],
   // Add a 'renovate' label to every PR for the matrix bot
-  "addLabels": ["renovate"]
+  // and add 'changelog:skip' label for the check-changelog CI job
+  "addLabels": ["renovate", "changelog:skip"]
   // TODO: uncomment once confident that renovates works properly
   // "patch": {
   //   "automerge": true


.github/renovate.json5
```diff
@@ -73,7 +73,8 @@
     }
   ],
   // Add a 'renovate' label to every PR for the matrix bot
-  "addLabels": ["renovate"]
+  // and add 'changelog:skip' label for the check-changelog CI job
+  "addLabels": ["renovate", "changelog:skip"]
```
ROMemories, line 77: I don't think we should be adding `changelog:skip` by default on every Renovate PR actually, at least because upgrading *some* of our crates is a breaking change, as we re-export a few of them. Could we change this to only apply the `changelog:skip` label to GitHub Actions? These are indeed unlikely to warrant a changelog entry, but I would rather keep the decision of whether a crate bump should go into the changelog an explicit one (and because the workflow is triggered on labeling, it really just takes two clicks).

anlavandier: Opened  #1842 to fix this

Issue #1827, opened 2026-02-26
Incompatible heapless version used to initialize embassy-net static configuration


nponsard:
### Summary

The version of embassy-net we use is using heapless v0.8.0, using heapless v0.9.1 breaks the type compatibility in `src/ariel-os-embassy/src/net.rs:249:26`. 

To reproduce the issue, you can run

```
laze -C examples/tcp-client/ build -b nrf52840dk -s ipv6   
```

### Relevant log output

```shell
   Compiling ariel-os-embassy v0.3.0 (/tmp/4133/ariel-os/src/ariel-os-embassy)
error[E0308]: mismatched types
   --> src/ariel-os-embassy/src/net.rs:249:26
    |
249 |             dns_servers: heapless::Vec::new(),
    |                          ^^^^^^^^^^^^^^^^^^^^ expected `Vec<Ipv6Address, 3>`, found `VecInner<_, _, ...>`
    |
    = note: expected struct `heapless::vec::Vec<Ipv6Addr, 3>`
               found struct `VecInner<_, _, vec::storage::VecStorageInner<[MaybeUninit<_>; _]>>`

```

### Would you like to work on fixing this bug?

Maybe

chrysn:
To test in https://github.com/ariel-os/ariel-os/pull/1197#issuecomment-3966250890, I successfully used (until unrelated trouble stopped the test):

```patch
diff --git a/src/ariel-os-embassy/src/net.rs b/src/ariel-os-embassy/src/net.rs
index 08b40b39..90b15c63 100644
--- a/src/ariel-os-embassy/src/net.rs
+++ b/src/ariel-os-embassy/src/net.rs
@@ -252 +252 @@ fn __ariel_os_network_config() -> embassy_net::Config {
-            dns_servers: heapless::Vec::new(),
+            dns_servers: Default::default(),
```

nponsard:
This also shows that the `__ariel_os_network_config()` function doesn't get built by the CI.

kaspar030:
We just bumped this in #1318.

anlavandier:
The question is do we downgrade back to v0.8.0 or do we apply chrysn patch ?

anlavandier:
```console
$ grep -r src/*  -e "heapless"
src/ariel-os-coap/src/stored.rs:fn generate_credpair() -> (heapless::Vec<u8, 60>, lakers::BytesP256ElemLen) {
[...]

PR #1828, opened 2026-02-26
ci: test all of our `network-config-*` options in CI

anlavandier:
# Description

This PR adds a explicit check for the modules `ipv6` and `network-config-ipv4-static`. Together they lead to the function `ariel_os_network_config` being checked in CI which it previously wasn't.

## Testing

~~With the current main, this should fail because of a mismatch in the version of `heapless` between what we use and what is required by `embassy-net`. This was introduced in #1318.~~

## Issues/PRs References
This attempts to have CI check for bugs like the one reported in #1827 which is now fixed.

## Open Questions

- [ ] Is this the best way to test this ? 
- [ ] Are there other functions that are never compiled and/or tested in the CI because of non default laze-modules ?

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -353,8 +353,16 @@ jobs:
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
           echo "LAZE_DISABLE=${LAZE_DISABLE}"
           laze build -mg info-boards | uniq
+          # This will test DHCP IPv4 as it is the default
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }}
 
+      - name: "static IPv4/IPv6 compilation tests (default toolchain)"
+        if: steps.should-skip.outputs.result != 'true'
+        run: |
+          # Build the udp-echo example with static IPv4
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
+          # Build the udp-echo example with static IPv6
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6 -s network-config-ipv6-static
       # cleanup previous build artifacts. do in background (no need to wait).
       - name: "cleanup stable build artifacts"
         if: github.event_name == 'schedule' && steps.should-skip.outputs.result != 'true'


anlavandier:
[This failed as expected](https://github.com/ariel-os/ariel-os/actions/runs/22444427898/job/64995001185?pr=1828)

anlavandier:
@ROMemories feel free to close this PR if you think that it is right. I'm putting it up for review now to avoid it being stuck in Draft Hell and forgotten

.github/workflows/build.yml
```diff
@@ -343,6 +343,11 @@ jobs:
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
           echo "LAZE_DISABLE=${LAZE_DISABLE}"
           laze build -mg info-boards | uniq
+          # Build the udp-echo example with Static IPV6
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='1::1' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='1::1' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6
```
kaspar030: please use `2001::db8::` prefix (IICU :-)) for compile testing.

ROMemories: [`2001:db8::`](https://en.wikipedia.org/wiki/IPv6_address#Special_addresses:~:text=2001%3Adb8%3A%3A%2F32)

anlavandier: Done in c5d3b1b0

.github/workflows/build.yml
```diff
@@ -343,6 +343,11 @@ jobs:
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
           echo "LAZE_DISABLE=${LAZE_DISABLE}"
           laze build -mg info-boards | uniq
+          # Build the udp-echo example with Static IPV6
```
```suggestion
          # Build the udp-echo example with static IPv6
```

.github/workflows/build.yml
```diff
@@ -343,6 +343,11 @@ jobs:
...
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6
+          # Build the udp-echo example with Static IPV4
```
```suggestion
          # Build the udp-echo example with static IPv4
```

.github/workflows/build.yml
```diff
@@ -343,6 +343,11 @@ jobs:
...
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
+          # This will test IPV4 DHCP as it is the default
```
```suggestion
          # This will test DHCP IPv4 as it is the default
```

.github/workflows/build.yml
```diff
@@ -343,6 +343,11 @@ jobs:
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
           echo "LAZE_DISABLE=${LAZE_DISABLE}"
           laze build -mg info-boards | uniq
+          # Build the udp-echo example with Static IPV6
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6
+          # Build the udp-echo example with Static IPV4
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
```
ROMemories: As discussed out of band, these should use a single builder instead of `--global` to minimize the CI cost, as this is only testing a software, hardware-independent difference from the full build. `rpi-pico-w` sounds like a reasonable choice for that builder.

.github/workflows/build.yml
anlavandier: Fixed in [6375dad](https://github.com/ariel-os/ariel-os/pull/1828/commits/6375dad337627697ed3a4794c72dfc84a3a5339d)

.github/workflows/build.yml
anlavandier: Fixed in [6375dad](https://github.com/ariel-os/ariel-os/pull/1828/commits/6375dad337627697ed3a4794c72dfc84a3a5339d)

.github/workflows/build.yml
anlavandier: Fixed in 6375dad

.github/workflows/build.yml
anlavandier: In my local tests I had to keep the `--global` flag but in any case this is fixed in [ea7db46](https://github.com/ariel-os/ariel-os/pull/1828/commits/ea7db46040ad1779f5cfd80e93c9771c86790501)

.github/workflows/build.yml
ROMemories: The "V" of "IPV6" is still uppercase, as is in the other occurrences.

anlavandier: Fixed the capitalization of the Vs in  [964b835](https://github.com/ariel-os/ariel-os/pull/1828/commits/964b8352b54eb08b5594bbc7576f22e0a36999c9)

.github/workflows/build.yml
```diff
@@ -343,6 +343,11 @@ jobs:
           echo "LAZE_BUILDERS=${LAZE_BUILDERS}"
           echo "LAZE_DISABLE=${LAZE_DISABLE}"
           laze build -mg info-boards | uniq
+          # Build the udp-echo example with static IPv6
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6
+          # Build the udp-echo example with static IPv4
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
```
ROMemories: Could you move this to a separate step? The `LAZE_*` environment variables are overridden by the CLI arguments anyway, and that would make observing the impact on CI time easier.

anlavandier: Fixed in e00cc62

.github/workflows/build.yml
```diff
@@ -343,8 +343,16 @@ jobs:
...
+          # This will test DHCP IPv4 as it is the default
           CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --partition hash:${{ matrix.partition }}
 
+      - name: "static IPv4/IPv6 compilation tests (default toolchain)"
+        if: steps.should-skip.outputs.result != 'true'
+        run: |
+          # Build the udp-echo example with static IPv4
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s network-config-ipv4-static
+          # Build the udp-echo example with static IPv6
+          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6
```
ROMemories: We'll later default to SLAAC instead:
```suggestion
          CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' CONFIG_NET_IPV6_STATIC_ADDRESS='2001:db8::0' CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS='2001:db8::1' laze build -DCARGO_ARGS+='--locked' --global --builders rpi-pico-w --partition hash:${{ matrix.partition }} --apps udp-echo -s ipv6 -s network-config-ipv6-static
```

PR #1829, opened 2026-02-26
refactor(net): decorrelate `network-config-override` and the ariel-os provided `network-config-*`

anlavandier:
# Description

This PR refactors the network configuration to decorrelate the `network-config-override` laze module and feature from the configuration presets of ariel OS (i.e. `network-config-ipv4-static`, `network-config-ipv6-static` and `network-config-ipv4-dhcp`).
This was done through:
- Removing `__ariel_os_network_config` which was declared here despite being supposed to be generated using the  `#[ariel_os::config(network)]` macro. 
- Changes the function `ariel-os-embassy::net::config` to choose with `cfg_if` between:
  - `__ariel_os_network_config` when `network-config-override` is used 
  - a new `ariel_os_network_config` when `context = "ariel-os"` that initializes the network configuration
  - `embassy_net::Config::default()` otherwise
- Remove the dependency on  `network-config-override` for  `network-config-ipv4-static` and `network-config-ipv6-static`
## Testing

`laze build -b nrf52840dk -s network-config-ipv4-static -C examples/udp-echo`
`laze build -b nrf52840dk -s ipv4 -s ipv6 -C examples/udp-echo`

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/src/net.rs
+++ b/src/ariel-os-embassy/src/net.rs
@@ -87,11 +87,8 @@ pub(crate) fn config() -> embassy_net::Config {
                 fn __ariel_os_network_config() -> embassy_net::Config;
             }
             unsafe { __ariel_os_network_config() }
-        } else if #[cfg(feature = "dhcpv4")] {
-            embassy_net::Config::dhcpv4(embassy_net::DhcpConfig::default())
-        } else if #[cfg(not(context = "ariel-os"))] {
-            // For platform-independent tooling.
-            embassy_net::Config::default()
+        } else {
+            ariel_os_network_config()
         }
     }
 }
@@ -171,15 +168,9 @@ impl embassy_net::driver::RxToken for DummyDriver {
     }
 }
 
-#[cfg(any(
-    feature = "network-config-ipv4-static",
-    feature = "network-config-ipv6-static"
-))]
-// SAFETY: the compiler prevents from defining multiple functions with the same name in the
-// same crate; the function signature is checked by the compiler as it is in the same crate as the
-// FFI declaration.
-#[unsafe(no_mangle)]
-fn __ariel_os_network_config() -> embassy_net::Config {
+#[cfg(not(feature = "network-config-override"))]
+fn ariel_os_network_config() -> embassy_net::Config {
+    #[allow(unused_mut, reason = "conditional compilation")]
     let mut config = embassy_net::Config::default();
 
     cfg_if::cfg_if! {
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -162,8 +162,8 @@ timer-generic-queue-64 = ["embassy-time?/generic-queue-64"]
 ## Generic Queue with 128 timers
 timer-generic-queue-128 = ["embassy-time?/generic-queue-128"]
 
-network-config-ipv4-static = ["network-config-override"]
-network-config-ipv6-static = ["network-config-override"]
+network-config-ipv4-static = []
+network-config-ipv6-static = []
 network-config-override = []
 override-usb-config = []
 rcc-config-override = ["ariel-os-hal/rcc-config-override"]


src/ariel-os-embassy/src/net.rs
anlavandier: This only compiles the function is either `network-config-ipv4-static` or `network-config-ipv6-static` are enabled

src/ariel-os-embassy/src/net.rs
```diff
@@ -182,42 +182,38 @@ impl embassy_net::driver::RxToken for DummyDriver {
...
-                );
-            }
-
-            config.ipv4 = embassy_net::ConfigV4::Static(embassy_net::StaticConfigV4 {
-                address: embassy_net::Ipv4Cidr::new(ipaddr, PREFIX_LEN),
-                dns_servers: heapless::Vec::new(),
-                gateway: Some(gw_addr),
-            });
-        } else if #[cfg(feature = "dhcpv4")] {
-            config.ipv4 = embassy_net::ConfigV4::Dhcp(embassy_net::DhcpConfig::default());
```
anlavandier, line 223: This was never compiled because `dhcpv4` does not go through this function

src/ariel-os-embassy/src/net.rs
anlavandier: I guess that technically someone could have had the features: `network-config-override`, `dhcpv4`, `network-config-ipv6-static` enabled and this PR changes the resulting configuration from an IPV4 DHCP to a IPV6 static configuration. This is however not allowed within laze build and clearly not something that users should do or rely on.

src/ariel-os-embassy/src/net.rs
nponsard: This can actually happen, if run a dual ipv4-ipv6 stack you can have the ipv4 address resolved through dhcpv4. When it comes to ipv6, a static IP is the only option available currently.

src/ariel-os-embassy/src/net.rs
nponsard: > This is however not allowed within laze build and clearly not something that users should do or rely on.

This actually happens when you select `ipv4` and `ipv6` on networking examples: 

```
laze -C examples/tcp-client/ build -b nrf52840dk -s ipv4 -s ipv6 
```

src/ariel-os-embassy/src/net.rs
```diff
@@ -87,8 +87,8 @@ pub(crate) fn config() -> embassy_net::Config {
                 fn __ariel_os_network_config() -> embassy_net::Config;
             }
             unsafe { __ariel_os_network_config() }
-        } else if #[cfg(feature = "dhcpv4")] {
-            embassy_net::Config::dhcpv4(embassy_net::DhcpConfig::default())
+        } else if #[cfg(context = "ariel-os")] {
+            ariel_os_network_config()
         } else if #[cfg(not(context = "ariel-os"))] {
```
nponsard:
```suggestion
        } else {
```

src/ariel-os-embassy/src/net.rs
```diff
@@ -171,15 +171,8 @@ impl embassy_net::driver::RxToken for DummyDriver {
...
-#[cfg(any(
-    feature = "network-config-ipv4-static",
-    feature = "network-config-ipv6-static"
-))]
-// SAFETY: the compiler prevents from defining multiple functions with the same name in the
-// same crate; the function signature is checked by the compiler as it is in the same crate as the
-// FFI declaration.
-#[unsafe(no_mangle)]
-fn __ariel_os_network_config() -> embassy_net::Config {
+#[cfg(all(not(feature = "network-config-override"), context = "ariel-os"))]
```
nponsard: I don't think feature-gating is needed now for this function.

It was there to avoid having two `__ariel_os_network_config` symbols for the final binary.

I think it's ok to always have this function compiled.

anlavandier: Another reason for feature gating this function was to make `clippy` happy. This would otherwise require an `#[allow(dead_code)]` for the function and most likely also an `#[allow(unused_mut)]` for the `config`.

src/ariel-os-embassy/src/net.rs
```diff
@@ -87,9 +87,9 @@ pub(crate) fn config() -> embassy_net::Config {
...
             }
             unsafe { __ariel_os_network_config() }
-        } else if #[cfg(feature = "dhcpv4")] {
-            embassy_net::Config::dhcpv4(embassy_net::DhcpConfig::default())
-        } else if #[cfg(not(context = "ariel-os"))] {
+        } else if #[cfg(context = "ariel-os")] {
+            ariel_os_network_config()
+        } else {
             // For platform-independent tooling.
             embassy_net::Config::default()
```
ROMemories: I haven't really tried this, but can we simplify this even more as follows:
```suggestion
        } else {
            ariel_os_network_config()
```
(the `#[cfg]` on `ariel_os_network_config()` also needs to be removed).

That would also have the benefit of not relying on the laze context for normal operation.

ROMemories (commented):
I wasn't too sure at first of what this was trying to solve (was maybe a bit confused with the other somewhat related in-flight network PRs), but I like the simplification a lot if we can do that. Haven't tested it yet, will do in the next round if we can indeed simplify this even further. The testing table from #915 can be used as reference.

src/ariel-os-embassy/src/net.rs
anlavandier: I have tested this (bc10cf3) on the `udp-echo` with all 5 configurations on an RPI PICO2W with wifi and the 3 statis-only configurations on an nrf52840dk. Also did some quick testing on the `dfrobot-firebeetle2-esp32-c6`

PR #1818, opened 2026-02-24
docs(gen-book): add ability to insert prose snippet in book pages

jkr-ariel:
# Description

This PR adds the ability to insert prose snippets in book pages. This will be useful when we want to document board (or chip) specific information such as the flashing process or other caveats.

## Issues/PRs References

Linked to #1815.

## Changelog Entry

It is now possible to add arbitrary prose snippets in book or chip pages by adding a `prose` field to the relevant object in `doc/support_matrix.yml`.

## Change Checklist

- [ ] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [ ] I have followed the [Coding Conventions][coding-conventions].
- [ ] I have tested and performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

Issue #1815, opened 2026-02-19
Ability to insert prose snippets in support documentation


ROMemories:
It would be useful to be able to insert "prose snippets" in the chip and board documentation we have. This would be particularly useful to provide guidance on how to flash boards, as it is not always that simple (see #876).

In particular, it should be possible to:

- [x] Insert one-off prose in some predefined locations in board pages. #1818
- [x] Define re-usable prose snippets once in the `support_matrix.yml`, and be able to use them for multiple boards. #1833

PR #1833, opened 2026-02-27
docs(support-matrix): add reusable prose snippets

jkr-ariel:
# Description

This adds introduces reusable prose snippets in the support matrix.

## Issues/PRs References

Linked to #1815.
Follow up to #1818.

## Changelog Entry

Contributors can now add prose snippets in `doc/support_matrix.yml` that can be reused within the support matrix.

## Change Checklist

### DIFF
--- a/doc/gen_book.rs
+++ b/doc/gen_book.rs
@@ -681,115 +681,146 @@ fn validate_input(matrix: &schema::Matrix) -> Result<(), Error> {
 }
 
 fn gen_board_functionalities(matrix: &schema::Matrix) -> Result<Vec<BoardInfo>, Error> {
-    let boards =
-        matrix.boards.iter().map(|board_info| {
-            let builders =
-                board_info.builders.iter().map(|builder_technical_name| {
-                    let builder_info = matrix.builders.get(builder_technical_name).ok_or(vec![
-                        Error::InvalidBuilderName {
-                            found: builder_technical_name.to_owned(),
-                            board: board_info.name.to_owned(),
-                        },
-                    ])?;
+    let boards = matrix.boards.iter().map(|board_info| {
+        let builders = board_info.builders.iter().map(|builder_technical_name| {
+            let builder_info = matrix.builders.get(builder_technical_name).ok_or(vec![
+                Error::InvalidBuilderName {
+                    found: builder_technical_name.to_owned(),
+                    board: board_info.name.to_owned(),
+                },
+            ])?;
 
-                    // Implement chip info inheritance
-                    let chip_info = matrix.chips.get(&builder_info.chip).ok_or(vec![
-                        Error::InvalidChipName {
-                            found: builder_info.chip.to_owned(),
-                            board: board_info.name.to_owned(),
-                        },
-                    ])?;
+            // Implement chip info inheritance
+            let chip_info =
+                matrix
+                    .chips
+                    .get(&builder_info.chip)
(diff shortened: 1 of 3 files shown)


jkr-ariel:
Added a test commit to see how the note snippets are rendered.

jkr-ariel:
Currently the content of the note snippet is concatenated with the existing note (if any). We might want to have a clearer separation between the note and the different note snippets.

doc/support_matrix.yml
```diff
@@ -166,6 +170,8 @@ chips:
       wifi: not_available
       ble: supported
       ethernet: not_available
+    note_snippets:
+      - test
```
ROMemories: It seems to me that this note, being added on a chip, isn't rendered, is it? We do need to be able to add these notes on chips to replace the current notes about the MAC address generation for Ethernet on STM32 chips.

doc/gen_book.rs
```diff
@@ -681,116 +681,163 @@ fn validate_input(matrix: &schema::Matrix) -> Result<(), Error> {
...
+                let errors = snippets
+                    .clone()
+                    .into_iter()
+                    .filter_map(Result::err)
+                    .collect::<Vec<_>>();
+                if !errors.is_empty() {
+                    return Err(errors);
+                }
+            }
+            // Safe to unwrap here since we check of errors above
```
ROMemories:
```suggestion
            // Safe to unwrap here since we checked for errors above
```

ROMemories (commented):
This seems very useful on the builders, and whitespace seems to be appropriately handled, but I can't get it to work for the chips.

doc/support_matrix.yml
jkr-ariel: Yes, I realise that I did not implement it for chips...

jkr-ariel: Fixed

doc/gen_book.rs
```diff
@@ -851,12 +927,29 @@ fn gen_chip_functionalities(matrix: &schema::Matrix) -> Result<Vec<ChipInfo>, Er
...
+                tmp.push_str(&notes);
+                tmp.push_str("\n\n");
+                tmp.push_str(&snippets.as_slice().join("\n\n"));
+                merged_notes = Some(tmp)
+            } else {
+                merged_notes = Some(snippets.as_slice().join("\n\n"));
+            }
+        } else {
+            merged_notes = notes;
+        }
```
ROMemories: I think this is exactly the same thing as above; can you factor this out?

ROMemories (commented):
Tested added note snippets on chips, that does work now (and builders still work as expected). Also tested adding both free-form notes and snippets on the same item. Should be good to go after this small refactor.

doc/gen_book.rs
```diff
@@ -909,6 +969,23 @@ fn add_custom_filters(env: &mut Environment) {
     env.add_filter("capitalize_first", capitalize_first);
 }
 
+fn merge_notes(notes: Option<String>, note_snippet_content: Option<Vec<String>>) -> Option<String> {
```
ROMemories, line 972: For reference, it would be better to take an `Option<&str>` and use `.deref()` at the call sites; but that doesn't really matter in this script, no need to change it.

ROMemories (approved):
LGTM, please remove the test commit and we'll merge this. I'll then do a follow-up PR to leverage this for the duplicate note that we have.

PR #1687, opened 2026-01-14
feat: driver for the SHT3x series of sensors

anlavandier:
# Description
 A driver for the [SHT3X](https://sensirion.com/media/documents/213E6A3B/63A5A569/Datasheet_SHT3x_DIS.pdf) series of humidity and temperature sensors from Sensiron.

## Testing
Testing done on an Ulanzi TC001 smart clock (ESP32-WROOM-32D)
Now can be done easily with `laze build -b ulanzi-tc001 -C examples/sensors-debug run`

## Changelog Entry

A sensor driver for the SHT3X series of sensors, compatible with the sensor API, is now available.

## Change Checklist

- [ ] I have cleaned up my [commit history][conventional-commits] and squashed fixup commits.
- [ ] I have followed the [Coding Conventions][coding-conventions].
- [ ] I have tested and performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html

PR #1841, opened 2026-03-02
fix(sensors): add built-in sensors to the workspace explicitly

ROMemories:
# Description

As pointed out in https://github.com/ariel-os/ariel-os/pull/1687#issuecomment-3982986508, sensor driver crates were not properly added as workspace members. This explicitly adds the crates to the workspace.

## Testing

Using `ci-build:stm32`, but really this doesn't change anything compilation-wise, and the lockfile is not modified.

## Change Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,6 +146,11 @@ ariel-os-storage = { path = "src/ariel-os-storage" }
 ariel-os-threads = { path = "src/ariel-os-threads" }
 ariel-os-utils = { path = "src/ariel-os-utils", default-features = false }
 
+# Built-in sensor drivers.
+ariel-os-sensor-lis2du12 = { path = "src/sensors/ariel-os-sensor-lis2du12" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -27,6 +27,9 @@ members = [
   "src/lib/coapcore",
   "src/lib/rbi",
   "src/lib/ringbuffer",
+  "src/sensors/ariel-os-sensor-lis2du12",
+  "src/sensors/ariel-os-sensor-lps22df",
+  "src/sensors/ariel-os-sensor-stts22h",
```
ROMemories, line 32: I would rather use `src/sensors/*` (which would in particular avoid potential conflicts here), but @kaspar030 didn't [seem too big on the idea](https://github.com/ariel-os/ariel-os/pull/1361#discussion_r2358880642).

ROMemories: I guess having to list them as dependencies makes this moot anyway.

kaspar030: (I still wouldn't want globs there.)

PR #1851, opened 2026-03-03
refactor: replace `cfg_if!` with the now built-in `cfg_select!`

ROMemories:
# Description

`cfg_select!` has just been stabilized, and [is part of Rust 1.95](https://doc.rust-lang.org/stable/core/macro.cfg_select.html). This PR replaces almost all occurrences of the `cfg_if!` macro (from the `cfg-if` crate) and removes the dependency on `cfg-if` everywhere, except in `ariel-os-boards`, which is generated by `sbd-gen` and is therefore out of scope of this PR. This PR also only takes care of replacing usage of `cfg_if!`, other refactors stemming from the new availability of `cfg_select!`, if any, are also out of scope.

**Note that, unlike `cfg_if!`, `cfg_select!` compile-time fail if no branch matches.** In a few places it was therefore necessary to add a default branch to preserve the behavior.

## How to review this PR

This PR is better reviewed commit by commit. I can easily split the PR if needed.

## Testing

```sh
laze -C tests/i2c-controller/ build -b stm32u083c-dk run
```

## Issues/PRs References

- Follows #1875

## Change Checklist


src/ariel-os-embassy/src/net.rs
```diff
@@ -210,9 +210,11 @@ fn ariel_os_network_config() -> embassy_net::Config {
                 dns_servers: Default::default(),
                 gateway: Some(gw_addr),
             });
-        } else if #[cfg(feature = "dhcpv4")] {
+        }
+        feature = "dhcpv4" => {
             config.ipv4 = embassy_net::ConfigV4::Dhcp(embassy_net::DhcpConfig::default());
         }
+        _ => {}
```
ROMemories, line 218: This default branch is now necessary, otherwise this compile-time fails for IPv6 for instance.

src/ariel-os-embassy/src/net.rs
```diff
@@ -68,6 +69,7 @@ pub(crate) fn unique_seed() -> u64 {
                 return u64::from_le_bytes(seed);
             }
         }
```
nponsard, line 71: Should the 1234 constant go in the default variant ? Can be done in another refactor.

nponsard (approved):
Looks good to me, the small refactor can be done in another PR.

src/ariel-os-embassy/src/net.rs
ROMemories: I've considered it, not that's not possible because of the `if` in the second branch.

nponsard: We may need to throw an error here, a const seed is not unique.

PR #1864, opened 2026-03-05
feat: add support for st-nucleo-f072rb

kaspar030:
# Description

This adds support for the Nucleo stm32f072rb.

This PR was mostly (first two commits) generated using Opus 4.6.

## Testing

Did successfully test:

- examples/blinky
- examples/gpio
- examples/i2c-scanner
- examples/storage
- tests/uart-loopback
- tests/spi-loopback

(with some i2c sensor on d14/d15, and connected d8/d9 and d11/d12)

## Changelog Entry

The STM32F072RB MCU and the ST NUCLEO-STM32F072RB board are now supported.

## Change Checklist


chrysn:
Please separate tool-generated changes (by updating the build matrix) from the changes in the edited source; this makes it easier to decide what is review-worthy changes and what is generated text (that is only in our files because of tooling difficulties).

kaspar030:
> Please separate tool-generated changes (by updating the build matrix) from the changes in the edited source; this makes it easier to decide what is review-worthy changes and what is generated text (that is only in our files because of tooling difficulties).

I'm not sure I know what exactly you want.
The first commit is tool generated, the other two are human made. Within the first commit, github hides the non-llm generated changes by default. Do you want that first commit split between LLM generated and other tool generated?

chrysn:
GitHub is somewhat arbitrary in what it renders by default: I've seen it not load files just because it thinks that the change is too big to show initially, and when I first looked at this, I did not get the "Some generated files are not rendered by default." message that I see now. Checking the behavior you mention, those are marked as "linguist-generated", a GitHub specific flag. (I've previously seen the `-diff` attribute in use for the same purpose, which works also when reviewing locally, which gives more flexibility w/rt --color-words / --ignore-all-space).

So yes, the request was to split that commit into one that contains source changes and one that contains only the output of the Ariel-built-in tools. If the gitattributes are consistent, I can also review this one based on those attributes (especially currently where GitHub is presenting a consistent view), but I think it's generally preferable to have that kind of checked-in changes in separate commits (or, ideally, just not have checked-in files unless they are indeed containing more information than a plain regenerate would give, like our Cargo.lock is).

kaspar030:
> So yes, the request was to split that commit into one that contains source changes and one that contains only the output of the Ariel-built-in tools.

Understood & done. I actually re-created those files, just to be sure I got it right. The force-push shows no changes.

doc/support_matrix.yml
```diff
@@ -531,6 +534,26 @@ chips:
...
+      hwrng: not_available
+      i2c_controller: supported
+      spi_main: supported
+      uart: supported
+      logging: supported
+      storage:
+        status: supported
+        comments:
+          - removing items not supported
+      user_usb: supported
```
kaspar030, line 552: this actually `needs_testing`

doc/support_matrix.yml
```diff
@@ -1278,6 +1308,11 @@ boards:
     builders:
       - st-nucleo-f042k6
 
+  - name: ST NUCLEO-F072RB
+    url: https://www.st.com/en/evaluation-tools/nucleo-f072rb.html
```
kaspar030, line 1312: needs archive.org url

PR #1860, opened 2026-03-05
docs(boards): provide dfu instructions for the `st-steval-mkboxpro`

bergzand:
# Description

This extends the initial instructions for the `st-steval-mkboxpro` with instructions how to use dfu-flash to make the flashing instructions complete.

## Testing

Follow the documented steps from this PR.

## Issues/PRs References

Follow-up to #1853 

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #1866, opened 2026-03-06
docs(book): clarify that dfu-util is required for flashing through DFU

ROMemories:
# Description

This clarifies that dfu-util needs to be installed on the host to use the `flash-dfu` laze task, and that this task is only available for STM32 (the task itself literally isn't defined otherwise).

## Testing

- I have now tested the instructions introduced by #1860.

## Change Checklist

### DIFF
--- a/book/src/build-system.md
+++ b/book/src/build-system.md
@@ -24,7 +24,7 @@ Tasks available in Ariel OS include:
 
 - `run`: Compiles, flashes, and runs an application. The [debug output](./debug-console.md) is printed in the terminal.
 - `flash`: Compiles and flashes an application.
-- `flash-dfu`: Compiles and flashes an application via USB DFU.
+- `flash-dfu`: (Currently only available for STM32.) Compiles and flashes an application via USB DFU.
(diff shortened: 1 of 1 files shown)


bergzand:
Thank you for adding this.

Do we support other stm32 boards with DFU in the bootloader, or can we restrict this to `st-steval-mkboxpro`-only instead of stm32-only?

ROMemories:
> Do we support other stm32 boards with DFU in the bootloader, or can we restrict this to `st-steval-mkboxpro`-only instead of stm32-only?

I haven't tried any other trough this task (even though I have used DFU/the system bootloader on other STM32 MCUs before), but I don't think it has to be restricted right now. I think the only thing that could go wrong would be if the firmware had to go at the [different address than `0x0800_0000`](https://github.com/ariel-os/ariel-os/blob/e32ee0eedfce806a442e422b8f15b0c3430a9aa1/laze-project.yml#L536) (I'm not aware of such STM32 MCU).

ROMemories:
For reference, the relevant STM32 application note is [AN2606](https://www.st.com/resource/en/application_note/cd00167594-stm32-microcontroller-system-memory-boot-mode-stmicroelectronics.pdf).

PR #1868, opened 2026-03-06
ci: check for "ci-host:github" label

nponsard:
# Description

Allows to override where the Build CI is run for a PR.
The default runner label selected is "self-hosted-ubuntu", it can be overriden by the repository level variable BUILD_RUNNER, that can be overridden by the PR label "ci-host:github" (sets the runner label to "ubuntu-24.04").

This would allow us to test the GitHub-hosted build ci and test toolchain upgrades without having an effect on the rest of the Build CI runs (in self-hosted, we can only have one esp toolchain preinstalled).

## Testing

Toggle on and off the label, see if it correctly changes the runners used.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,11 +121,6 @@ jobs:
       # newest standard library. It may also need to be aligned with the clang
       # version which Xtensa provides.
       CLANG_PATH: /usr/bin/clang-18
-      CACHE_THING_LOCATION: /cache/cache-thing
-      RUSTC_WRAPPER: sccache
-      CARGO_NET_GIT_FETCH_WITH_CLI: true
-      SCCACHE_DIR: /cache/sccache
-      SCCACHE_CACHE_SIZE: 128G
 
     strategy:
       fail-fast: true
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -43,12 +43,6 @@ jobs:
       # newest standard library. It may also need to be aligned with the clang
       # version which Xtensa provides.
       CLANG_PATH: /usr/bin/clang-18
-      CACHE_THING_LOCATION: /cache/cache-thing
-      RUSTC_WRAPPER: sccache
-      SCCACHE_DIR: /cache/sccache
-      SCCACHE_CACHE_SIZE: 128G
-
-      CARGO_NET_GIT_FETCH_WITH_CLI: true
 
     steps:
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -106,33 +121,40 @@ jobs:
...
+            "11/12",
+            "12/12",
           ]
 
     steps:
       - name: reset git-cache update barrier
+        if: runner.environment == 'self-hosted'
         run: pkill -f -SIGUSR1 /usr/local/bin/git-cache-daemon | true
 
+      - name: Set up self-hosted runner environment
```
kaspar030: shouldn't these be set on the runner?

.github/workflows/build.yml
```diff
@@ -106,33 +121,40 @@ jobs:
...
+            "3/12",
+            "4/12",
+            "5/12",
+            "6/12",
+            "7/12",
+            "8/12",
+            "9/12",
+            "10/12",
+            "11/12",
+            "12/12",
```
ROMemories, line 141: Can you split this out? I think this is somewhat unrelated and I can approve it easily separately.

kaspar030:
(this ran on github runners, I've removed the label to see if next push builds on the self-hosted ones)

.github/workflows/build.yml
nponsard: Moved to #1880

ROMemories:
> I've removed the label to see if next push builds on the self-hosted ones

I think this ran on our self-hosted runners as expected.

nponsard:
You can check that the workflow is running on self-hosted either in "setting up job" where the selected runner name would start by "garm-" or that the "Free up space" step is skipped.

.github/workflows/build.yml
kaspar030: I think this has been merged & deployed in the worker images. @nponsard please confirm an drop here!

kaspar030: (outdated, was replying on a stale window)

nponsard: The last CI run wasn't with the updated runners (you can see the misconfigured cache-thing). I'm re-running the CI now.

nponsard:
CI is being a bit slow as the cache is from the wrong rust toolchain

.github/workflows/build.yml
nponsard: CI completed successfully with the correct configuration.

PR #1879, opened 2026-03-09
ci(cache): don't skip the build when clearing the cache

nponsard:
# Description

There is a possibility where the `build-cache` job can be skipped if there was no file change since the last job run. This would also skip the cleaning of the cache, something we want to happen regularly.

This disables the skipping mechanism when the option to clean the cache is selected.  

## Change Checklist

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -67,16 +67,24 @@ jobs:
           path: .tree-hash
           key: build-cache-success-${{ steps.get-tree-hash.outputs.hash }}-${{ matrix.node }}
 
+      - name: Whether we should clear the cache
+        id: clear-cache
+        run: |
+          day=$(date +'%u')
+          # Clear the cache on Sundays and when manually selected.
+          if [ "$day" = "7" ] || [ "${{ inputs.clear_cache }}" = "true" ];
+          then
+            echo "result=true" >> $GITHUB_OUTPUT
+          else
+            echo "result=false" >> $GITHUB_OUTPUT
+          fi
+
       - name: Whether to skip next steps
         id: should-skip
         run: |
-          result="result=${{ steps.result-cache.outputs.cache-hit == 'true' }}"
+          result="result=${{ steps.result-cache.outputs.cache-hit == 'true' && steps.clear-cache.outputs.result != 'true' }}"
           echo "${result}"
           echo "${result}" >> $GITHUB_OUTPUT
-      - name: Get day of week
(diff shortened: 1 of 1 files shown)


.github/workflows/build-cache.yml
```diff
@@ -67,30 +67,38 @@ jobs:
...
 
-      - uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
         id: result-cache
         with:
           path: .tree-hash
           key: build-cache-success-${{ steps.get-tree-hash.outputs.hash }}-${{ matrix.node }}
 
+      - name: If we should clear the cache
+        id: clear-cahce
```
ROMemories:
```suggestion
        id: clear-cache
```

.github/workflows/build-cache.yml
```diff
@@ -67,30 +67,38 @@ jobs:
...
           echo "hash=$(cat .tree-hash)" >> $GITHUB_OUTPUT
 
-      - uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
         id: result-cache
         with:
           path: .tree-hash
           key: build-cache-success-${{ steps.get-tree-hash.outputs.hash }}-${{ matrix.node }}
 
+      - name: If we should clear the cache
```
ROMemories:
```suggestion
      - name: Whether we should clear the cache
```

nponsard:
> How can we test this? Alternatively, when will we know whether this works as expected? Do we have to wait for next Monday?

We can try this by manually triggering the workflow https://github.com/ariel-os/ariel-os/actions/workflows/build-cache.yml with the option to clean the cache enabled. To test that the skip condition is not triggered there should have no code change on main between the two runs. 

Alternatively the CI workflow in this PR can be modified to trigger a "clean" cache build. But the requirement for no files being changed in main would remain.

nponsard:
Last Sunday the cache build was skipped because there was no changes, this means it didn't get cleaned up: https://github.com/ariel-os/ariel-os/actions/runs/23102867072/job/67106640189

The cache is now taking 266.7GB (should be ~150GB after the clean up)

nponsard:
Seems to have worked as expected: https://github.com/ariel-os/ariel-os/actions/runs/23395014930/job/68056365964

Cache usage today is 139GB

Issue #1855, opened 2026-03-05
Obtaining the reset reason


ROMemories:
### Description

MCUs usually expose the reason why the MCU has reset (power-on, brown-out reset, wake-up from standby low-power mode, etc.). There should be an API to obtain that reason.

### Motivation

This would be particularly useful in conjunction with #510, to differentiate power-on resets from wake-ups from standby mode.

### Requirements

On some MCUs (e.g., STM32s), the bits representing that reason are only cleared by hardware on a *power-on* reset and must be manually cleared after boot to be able to differentiate between reasons (otherwise they "accumulate" across non-power-on resets). This requires some initialization logic, called as early as possible in the boot sequence, to obtain the reset reason and clear these bits.

### Are you planning to do it yourself in a pull request?

Yes

PR #1883, opened 2026-03-09
feat(power): expose the reset reason on nRF

ROMemories:
# Description

This exposes the reason why the microcontroller has reset inside `ariel_os::power`, see https://github.com/ariel-os/ariel-os/issues/1855 for details. This is limited to nRF MCUs only for now, but I'll expand support to other families in follow-ups.

## Testing

- The `SoftwareReset` and `ResetPin` reasons can be tested using the following snippet in the `blinky` example (replacing `ResetPin` with `SoftwareReset`):

  ```rust
  let mut led0 = if ariel_os::power::reset_reason() == ariel_os::power::ResetReason::ResetPin {
      Output::new(peripherals.led0, Level::Low)
  } else {
      Output::new(peripherals.led1, Level::Low)
  };
  ```

  Running the example using probe-rs (which seemingly resets the MCU through software) will make one of the LED blink, while pressing the RESET button on the DK will make the other LED blink instead.
- The combination with `ariel_os::power::reboot();` can also be tested.

## Issues/PRs References

- Split out from #1863

## Change Checklist

### DIFF
--- a/src/ariel-os-power/src/reset.rs
+++ b/src/ariel-os-power/src/reset.rs
@@ -1,3 +1,9 @@
+#[cfg(context = "nrf")]
+use portable_atomic::{AtomicU8, Ordering};
+
+#[cfg(context = "nrf")]
+static RESET_REASON: AtomicU8 = AtomicU8::new(ResetReason::PowerOnReset as u8);
+
 /// Indicates why the microcontroller has reset.
 ///
 /// # Note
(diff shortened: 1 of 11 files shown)


src/ariel-os-power/src/reset.rs
```diff
@@ -1,3 +1,150 @@
...
+                    embassy_nrf::pac::POWER.resetreas().write_value(clear_value);
+                }
+            }
+
+        }
+    }
+}
+
+/// Returns the reason why the microcontroller has reset.
+#[cfg(context = "nrf")]
```
ROMemories, line 23: This is only feature-gated to allow adding support one HAL at a time. Eventually it will not be feature-gated at all.

src/ariel-os-power/src/reset.rs
```diff
@@ -1,3 +1,150 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum ResetReason {
+    #[default]
+    PowerOnReset,
+    ResetPin,
+    SoftwareReset,
+    StandbyWakeup,
+    PowerFailure,
```
ROMemories: It seems that [ESPs are able to distinguish brownout resets](https://docs.espressif.com/projects/rust/esp-hal/1.0.0/esp32c6/esp_hal/rtc_cntl/enum.SocResetReason.html#variant.SysBrownOut).

PR #1885, opened 2026-03-10
docs(contributing): require attribution of AI tools

chrysn:
# Description

Recent events such as the Git Workflow incident ([article with images](https://www.pcgamer.com/software/ai/microsoft-uses-plagiarized-ai-slop-flowchart-to-explain-how-github-works-removes-it-after-original-creator-calls-it-out-careless-blatantly-amateuristic-and-lacking-any-ambition-to-put-it-gently/), [original author's blog post](https://nvie.com/posts/15-years-later/)) indicate that contributions involving large AI models might easily violate copyright in ways that are not trivial for the operator to recognize.

Note that this is not ruling out contributions with AI assistance (or, really, completely AI-automated PRs for that matter), it merely requires transparency about them.

So far, we have a good grasp of our code's provenance. Let's keep it that way.

## Issues / References

We discussed some of this in today's weekly meeting; the text in the PR is adapted from the text there.

## Open questions

* (How) should we maintain a list of tools/uses that do not need per-item declaration?

## Changelog Entry

Contributing guidelines now require that AI tools used in contributions are declared.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- n/a I have tested and performed a self-review of my own code.
- n/a I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #1887, opened 2026-03-10
docs(contributing): undo require attribution of AI tools

kaspar030:
Reverts ariel-os/ariel-os#1885

needs more discussion


kaspar030:
marking "changelog:skip", other PR should as well

chrysn:
Seeing PRs https://github.com/ariel-os/ariel-os/pull/1786 and https://github.com/ariel-os/ariel-os/pull/1870 closed immediately after the original PR, yes. I think we should stick with the direction, but reverting might help fend off immediate damage, and encourage wider participation in a re-opened #1885.

PR #1890, opened 2026-03-10
docs(networking): clarify the role of USB CDC-NCM

ROMemories:
# Description

This introduces a glossary entry about USB CDC-NCM, and goes back over the Networking page to correct and clarify the relationship between Ethernet over USB and USB CDC-NCM.

## Change Checklist

### DIFF
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -8,13 +8,13 @@ Overriding this default selection is possible by explicitly selecting the desire
 
 ## Network Link Selection
 
-Ariel OS currently supports two different networking links: Ethernet-over-USB (aka CDC-NCM) and Wi-Fi.
+Ariel OS currently supports two different networking links: USB CDC-NCM (i.e., Ethernet over USB) and Wi-Fi.
 Boards may support both of them, only one of them, or none of them. However, currently the network stack supports at most one interface.
 
 Which link layer is used for networking is selected at compile time,
 through [laze modules][laze-modules-book].
 
-- `usb-ethernet`: Selects Ethernet-over-USB.
+- `usb-ethernet`: Selects Ethernet over USB (currently using USB CDC-NCM).
 - `wifi-cyw43`: Selects Wi-Fi using the CYW43 chip along an RP2040 or RP235x MCU (e.g., on the Raspberry Pi Pico W or Pico 2 W).
 - `wifi-esp`: Selects Wi-Fi on an ESP32 MCU.
 
@@ -105,9 +105,9 @@ ip address show dev <interface>
 Replace `<interface>` with the name of the used network interface.
 To find out the name of your interface you can use a command such as `ip address`.
 
-### Ethernet-over-USB
+### Ethernet over USB
 
-For Ethernet-over-USB, ensure that, in addition to the USB cable used for flashing
+For Ethernet over USB, ensure that, in addition to the USB cable used for flashing
 and debugging, the *user* USB port is also connected to the host computer with
 a second cable.
 


book/src/networking.md
```diff
@@ -8,13 +8,13 @@ Overriding this default selection is possible by explicitly selecting the desire
 
 ## Network Link Selection
 
-Ariel OS currently supports two different networking links: Ethernet-over-USB (aka CDC-NCM) and Wi-Fi.
+Ariel OS currently supports two different networking links: USB CDC-NCM (i.e., Ethernet over USB) and Wi-Fi.
```
ROMemories, line 11: I think I was initially confused when I wrote that: USB CDC-NCM is *one* protocol allowing to do Ethernet over USB (as explained in the linked Wikipedia page), not the other way around.

book/src/networking.md
```diff
@@ -8,13 +8,13 @@ Overriding this default selection is possible by explicitly selecting the desire
...
 Boards may support both of them, only one of them, or none of them. However, currently the network stack supports at most one interface.
 
 Which link layer is used for networking is selected at compile time,
 through [laze modules][laze-modules-book].
 
-- `usb-ethernet`: Selects Ethernet-over-USB.
+- `usb-ethernet`: Selects Ethernet over USB (currently using USB CDC-NCM).
```
ROMemories, line 17: The "currently" is there so that this doesn't preclude from alternatively supporting Ethernet over USB using other protocols.

book/src/networking.md
```diff
@@ -105,9 +105,9 @@ ip address show dev <interface>
 Replace `<interface>` with the name of the used network interface.
 To find out the name of your interface you can use a command such as `ip address`.
 
-### Ethernet-over-USB
+### Ethernet over USB
 
-For Ethernet-over-USB, ensure that, in addition to the USB cable used for flashing
+For Ethernet over USB, ensure that, in addition to the USB cable used for flashing
```
bugadani, line 110: I think it would be more useful to call out that the device either needs a DHCP server, or the PC must be configured with a static IP in order for this to work. It wasn't obvious to me at all, and these things aren't as trivial as "use the right USB port".

book/src/networking.md
```diff
@@ -105,9 +105,9 @@ ip address show dev <interface>
 Replace `<interface>` with the name of the used network interface.
 To find out the name of your interface you can use a command such as `ip address`.
 
-### Ethernet-over-USB
+### Ethernet over USB
```
ROMemories, line 108: There is no reason to put hyphens there, unless it's used as an adjective.

book/src/networking.md
ROMemories: I believe this is addressed by the section above this one (at least for the static-IP case)? This is not related to the network link itself I think.

bugadani: That section says that the host CAN be configured with a static IP. That is irrelevant for Wi-Fi, the router will assign the IP usually. But with USB, the host will expect the MCU to provide the IP address. As far as I remember from a few weeks ago, Windows will flat out refuse to use the USB interface unless hand-configured, or the MCU runs a DHCP server.

This whole chapter implies that these interfaces are largely equivalent, while that's not the case and it takes considerable experience to understand that.

Also, it's not clear why the MCU using a static IP implies anything about the host. If the MCU connects to a Wi-Fi network, and has a static IP, "the host computer can be configured", but the chapter doesn't explain why it should be configured with a static IP.

Some of these are out of scope to be sure, but perhaps not every one of these are.

PR #1895, opened 2026-03-11
 fix: resolve smoltcp build failure when specifying only a link layer 

bergzand:
# Description

When building an application with a specific network link in the laze.yml, but without specifying the `network` module, the build fails on smoltcp:

```
error: You must enable at least one of the following features: proto-ipv4, proto-ipv6, proto-sixlowpan
  --> /home/koen/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/lib.rs:99:1
   |
99 | compile_error!("You must enable at least one of the following features: proto-ipv4, proto-ipv6, proto-sixlowpan");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: If you enable the socket feature, you must enable at least one of the following features: socket-raw, socket-udp, socket-tcp, socket-icmp, socket-dhcpv4, socket-dns
   --> /home/koen/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.12.0/src/lib.rs:112:1
    |
112 | compile_error!("If you enable the socket feature, you must enable at least one of the following features: socket-raw, socket-udp, socket-tcp, socket-icmp, socket-dhcpv4, socket-dns");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This can of course be resolved by also adding the `network` module, but this PR adds the network module as `selects` to the specific link layers. 

## Testing

Issue can be reproduced by adding `wifi-esp` to the hello-world example:

```diff
diff --git i/examples/hello-world/laze.yml w/examples/hello-world/laze.yml
index 8643593a..63826e1e 100644
--- i/examples/hello-world/laze.yml
+++ w/examples/hello-world/laze.yml
@@ -1,2 +1,4 @@
 apps:
   - name: hello-world
+    selects:
+      - wifi-esp
```

and building for any espressif board.

This PR resolves the build failure.

## Issues/PRs References

None

## Open Questions

None

## Changelog Entry

The `network` laze module is now automatically enabled by network-link laze modules (e.g., `usb-ethernet`, `wifi-esp`). 

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1353,6 +1353,7 @@ modules:
   - name: eth-stm32
     selects:
       - has_eth_stm32
+      - network
     provides_unique:
       - network_device
     env:


laze-project.yml
```diff
@@ -1370,6 +1371,7 @@ modules:
       - network_device
     selects:
       - usb
+      - network
```
ROMemories: Is `usb` required to come before `network` here? If so there should be a comment about it, otherwise these should be in lexicographical order (because the order *is* important to laze):
```suggestion
      - network
      - usb
```

bergzand: Agreed, fixed with the latest push

laze-project.yml
ROMemories, line 1355: Actually we should do this one too (and we should now work towards documenting support for it, as we can separate board support from MCU support now):
```suggestion
      - has_eth_stm32
      - network
```
(Should have seen it before.)

bergzand: I only checked the list currently mentioned on the book page. Added this one to the commits

PR #1896, opened 2026-03-11
build: make use of laze ${outfile} for Cargo output

kaspar030:
# Description

laze 0.1.39 introduced variable interpolation within the `$outfile` variable.
This allows directly using Cargo's output from the board specific target folder, no need for the extra copy to .elf anymore.

## Testing

I did a quick test with the current ariel-os-hello, this seems to still work.

## Changelog Entry

The minimum required laze version is now 0.1.39. This internally allows to use Cargo's output directly without copying to a file with the `.elf` extension.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1,4 +1,4 @@
-laze_required_version: 0.1.37
+laze_required_version: 0.1.39
 
 contexts:
   # base context that all other contexts inherit from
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -31,6 +31,7 @@ contexts:
       - ?stable
       - ?nightly
     env:
+      outfile: ${build-dir}/bin/${builder}/cargo/${RUSTC_TARGET}/${PROFILE_DIR}/${app}
```
ROMemories, line 36: Because I can't find the documentation of `outfile`, I have no idea what behavior to expect from this.

kaspar030: I've added a comment.

kaspar030:
> Should the changelog entry be saying "copying the .elf" instead of "to .elf"?

Technically no. Cargo outputs elf files but without .elf suffix. Laze was previously expecting a .elf suffix, so we manually copied the non-.elf to .elf. This is not necessary anymore with this PR.

laze-project.yml
ROMemories: Actual documentation in laze would be better, but I'll take the comment for now.

PR #1897, opened 2026-03-11
refactor(ariel-os-embassy)!: make `net` only enable the time driver

ROMemories:
# Description

[The docs of `embassy-net`](https://docs.embassy.dev/embassy-net/git/default/index.html#interoperability) only mention that it needs a time driver, not the entirety of `embassy-time` enabled in the build. This PR makes the `net` Cargo feature of `ariel-os-embassy` only enable `ariel-os-hal/time` which, as now documented in #1808, enables the HAL-specific time driver.

## Testing

Tested the `http-client` example on espressif-esp32-c6-devkitc-1 and rpi-pico-w, and the `udp-echo` example (with DHCP disabled) on nrf52840dk and stm32u083c-dk.

## Issues/PRs References

- Depends on #1899

## Changelog Entry

Timing functionality is now not indirectly enabled by enabling networking anymore. Applications that were relying on this simply need to enable the `time` Cargo feature.

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -89,8 +89,8 @@ uart = ["ariel-os-embassy-common/uart", "ariel-os-hal/uart"]
 usb = ["dep:embassy-usb", "ariel-os-hal/usb"]
 usb-hid = ["dep:usbd-hid", "embassy-usb?/usbd-hid", "usb"]
 
-# embassy-net requires embassy-time and support for timeouts in the executor
-net = ["dep:embassy-net", "time"]
+# `embassy-net` requires a time driver, which HALs provide.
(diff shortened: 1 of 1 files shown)


src/ariel-os-embassy/Cargo.toml
```diff
@@ -89,8 +89,8 @@ uart = ["ariel-os-embassy-common/uart", "ariel-os-hal/uart"]
 usb = ["dep:embassy-usb", "ariel-os-hal/usb"]
 usb-hid = ["dep:usbd-hid", "embassy-usb?/usbd-hid", "usb"]
 
-# embassy-net requires embassy-time and support for timeouts in the executor
```
ROMemories, line 92: When this was added in #191, I think we didn't have the granularity to do this.

PR #1865, opened 2026-03-06
build: bump MSRV to 1.94 and update CI toolchain

kaspar030:
# Description

This bumps Ariel's toolchain to 1.94. This stabilized the config include, so has the potential to close a couple of issues.

## Issues/PRs References

Corresponding worker image PR: https://github.com/ariel-os/garm-worker-image/pull/14

Last bump to 1.93: #1830.

- Closes #1811

## Open Questions

- [x] do we also need to bump the nightly? -> not now

## Changelog Entry

Ariel OS's MSRV is now 1.94. This version stabilizes Cargo's config `include` key, allowing Ariel OS's configuration to be discovered automatically by Cargo, without the need for either providing `--config` or using the nightly version of Cargo, including in out-of-tree applications.

## Change Checklist

- [ ] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [ ] I have followed the [Coding Conventions][coding-conventions].
- [ ] I have tested and performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #1898, opened 2026-03-11
build: remove the use of `--config` in laze

anlavandier:
# Description

This PR removes the use of `--config` now that config includes have been stabilized in 1.94

## Testing

In tree: `laze build ... cargo tree` fails. Now it doesn't
 O-O-T: the same 

## Issues/PRs References
Enabled by #1865 

## Open Questions
Should the laze `cargo` task exist ? Seems like it is doomed to always be fairly limited because not everything that follows will expect its arguments in the same places. 

Also OOT doing `laze build -b ... run/clippy/tree` will work on the OOT application by default (as expected) whereas `laze build -b ... cargo tree` will work on the default member of Ariel OS's workspace *i.e.* the `hello-world` example. 

I'm open to my second commit being too opinionated and thus needing to be removed. I will say though that without it, you can't `laze build cargo run` or `laze build cargo clippy` (but then again OOT if you don't remember to do `--apps my-app` you'll actually run those against the `hello-world` example and not your application plus there are already laze tasks for that).

Could be a release blocker I guess.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -189,7 +189,7 @@ contexts:
       vscode-config:
         help: Generate a VSCode configuration for selected board (select a board with -b)
         build: false
-        # The --config arg in CARGO_ARGS is pointing to a file relative to the appdir, it makes more sense to generate the config in the appdir
+        # Place the config in the application directory as it's the configuration for this application.
         workdir: ${appdir}
         export:
           - FEATURES


laze-project.yml
```diff
@@ -129,7 +127,7 @@ contexts:
       cargo:
         workdir: ${relpath}
         cmd:
-          - ${CARGO_PRESHELL} ${CARGO_ENV} ${CARGO} ${CARGO_TOOLCHAIN} ${CARGO_ARGS} $@
+          - ${CARGO_PRESHELL} ${CARGO_ENV} ${CARGO} ${CARGO_TOOLCHAIN} ${CARGO_ARGS} $@ ${FEATURES}
```
ROMemories: This needs testing, but I think `$@` should go at the end:

```suggestion
          - ${CARGO_PRESHELL} ${CARGO_ENV} ${CARGO} ${CARGO_TOOLCHAIN} ${CARGO_ARGS} ${FEATURES} $@
```

laze-project.yml
ROMemories: This comment needs to be updated, but I'm not sure how.

cc @nponsard

laze-project.yml
ROMemories: I also still need to test the effect of this.

anlavandier: I think that this is not true in most cases. `clippy` expects to be called with `cargo clippy --features` so does `run` so does `doc` so does `tree`. You can see this from the other tasks below.

chrysn: Hard part is when it would need to go in the middle; best example that comes to mind (although it's moot due to the `run` target) is `cargo run ${FEATURES} -- --probe usb:serial:number`.

anlavandier: Yes that's why I believe that this task is essentially doomed. If you don't include the features it's basically useless, if you do, then you have to decide where you put them and then some things will not work as a result

chrysn: It'd work easily if https://github.com/rust-lang/cargo/issues/4829 were implemented, but AIU that is "just" stuck on nobody prioritizing it high enough to make it happen. (Maybe if we team up with bacon?)

But indeed, for now our users might be better served with no `laze build cargo` and that rather erring out (maybe giving helpful pointers).

kaspar030:
> I'm open to my second commit being too opinionated and thus needing to be removed.

Yes, better drop that for now.

laze-project.yml
kaspar030: This task can be re-used. Do it without features, then for the next, use:

```
task:
 cargo-tree:
   cmd:
    - :cargo --features ${FEATURES}
```

Note the `:` that can be used to call a task from another.

But, I think this should be done in another PR. I think I have a branch actually somewhere...

anlavandier:
Done

laze-project.yml
nponsard: I'm looking into it.

nponsard:
```suggestion
        # Place the config in the application directory as it's the configuration for this application.
```

nponsard: I think something like that could work @ROMemories. The workspace `.cargo/config.toml` file is correctly found when testing inside `examples/hello-world`.

nponsard: Also tested with an application generated from the template. Works as expected.

PR #1899, opened 2026-03-11
docs(examples): use proper `Duration` types

ROMemories:
# Description

In #1321 we documented that types from `ariel_os::time` (especially `Duration`) must only be used "standalone", i.e., not passed to functions from crates that expect a specific version of `embassy_time` types, as these are *currently* re-exports from that crate. This was done to make the fact that these types were re-exports an implementation detail.

This PR updates the examples (and tests, but none require changes) to meet that policy: `Duration` that are passed to functions from crates that we also re-export (`embassy_net`) now come from `ariel_os::reexports::embassy_time` (because *we* make sure these crates work together), and the example using `picoserve` now adds an explicit dependency on `embassy-time` as this is actually decoupled form our re-export.

## How to review this PR

## Change Checklist

### DIFF
--- a/examples/ble-advertiser/src/main.rs
+++ b/examples/ble-advertiser/src/main.rs
@@ -2,15 +2,12 @@
 #![no_main]
 #![no_std]
 
+use ariel_os::{debug::log::info, reexports::embassy_time, time::Timer};
 use embassy_futures::join::join;
+use embassy_time::Duration;
 use trouble_host::advertise::{
     AdStructure, Advertisement, AdvertisementParameters, BR_EDR_NOT_SUPPORTED,
     LE_GENERAL_DISCOVERABLE,
-};
-
-use ariel_os::{
-    debug::log::info,
-    time::{Duration, Timer},
 };
 
 #[ariel_os::task(autostart)]
--- a/examples/tcp-echo/src/main.rs
+++ b/examples/tcp-echo/src/main.rs
@@ -1,8 +1,9 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::{debug::log::*, net, reexports::embassy_net, time::Duration};
+use ariel_os::{debug::log::*, net, reexports::embassy_net, reexports::embassy_time};
 use embassy_net::tcp::TcpSocket;
+use embassy_time::Duration;
 use embedded_io_async::Write;
 
 // Setting this to a small value would make packet handling slow and choppy, but would not cause
--- a/examples/ble-scanner/src/main.rs
+++ b/examples/ble-scanner/src/main.rs
@@ -4,18 +4,15 @@
 
 use core::cell::RefCell;
 
+use ariel_os::{debug::log::info, reexports::embassy_time, time::Timer};
 use embassy_futures::join::join;
+use embassy_time::Duration;
 use heapless::Deque;
 use trouble_host::{
     Host,
     connection::{PhySet, ScanConfig},
     prelude::{BdAddr, EventHandler, LeAdvReportsIter},
     scan::Scanner,
-};
-
-use ariel_os::{
-    debug::log::info,
-    time::{Duration, Timer},
 };
 
 #[ariel_os::task(autostart)]


examples/ble-advertiser/src/main.rs
```diff
@@ -3,15 +3,13 @@
...
 use trouble_host::advertise::{
     AdStructure, Advertisement, AdvertisementParameters, BR_EDR_NOT_SUPPORTED,
     LE_GENERAL_DISCOVERABLE,
 };
 
-use ariel_os::{
-    debug::log::info,
-    time::{Duration, Timer},
-};
+use ariel_os::{debug::log::info, reexports::embassy_time, time::Timer};
```
anlavandier: I feel like this should be above `use embassy_time::duration` otherwise when reading from top to bottom it's not clear that this isn't a direct dependency.

ROMemories: Agreed, I moved the `ariel_os` `use` statement with the existing import block, which places it first due to sorting. I don't see a reason why it should be a separate block anyway.

examples/ble-advertiser/src/main.rs
```diff
@@ -2,17 +2,14 @@
 #![no_main]
 #![no_std]
 
+use ariel_os::{debug::log::info, reexports::embassy_time, time::Timer};
 use embassy_futures::join::join;
+use embassy_time::Duration;
```
kaspar030, line 7: I want to think about this, I don't think I agree here.

ROMemories: The types from `ariel_os::time` cannot be used for this, as the fact that they currently come from `embassy_time` is an implementation detail. However, as long as we have this re-export of `embassy_time`, we pretty much guarantee that it will work well with types of other crates that we expose, `trouble_host` in this case, so I think that's what we use here. Unless your point is that we should instead use a direct dependency on `embassy-time` (that is compatible with `trouble-host`) in this example?

kaspar030: > The types from `ariel_os::time` cannot be used for this, as the fact that they currently come from `embassy_time` is an implementation detail.

This is the wrong approach. We solved the documentation issue in #1229 by lowering the level of integration we want to achieve, by limiting what the reexported types can be used for, vie #1321.

This will lead to cluttered examples and user code, and requiring *users* to match versions, taking the opportunity of Ariel OS to curate them.

ROMemories: I'm not exactly sure what you're saying: do you think we should reconsider #1321? #1229 was actually mostly a symptom of the issue fixed by #1321, not the other way around.

> requiring _users_ to match versions

Using the crates from the `reexports` module does *not* require users to match versions (that's the point).

Users only have to match the versions of dependencies *they* bring in (e.g., `picoserve` with `embassy-time`), which is a totally normal thing to do. You can't try to control the entire ecosystem.

examples/http-server/Cargo.toml
```diff
@@ -8,6 +8,8 @@ publish = false
 ariel-os = { path = "../../src/ariel-os" }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 embassy-sync = { workspace = true }
+# Version compatible with `picoserve`.
+embassy-time = "0.5.0"
```
kaspar030: this is what bugged me - we're on embassy-time 0.5.0, there's no need to import this here directly. Should use the reexport.

examples/ble-advertiser/src/main.rs
kaspar030: > Using the crates from the `reexports` module does _not_ require users to match versions (that's the point).

Agreed. Let's not open #1321 at this time, but at lease the reexports need to be used - they're not below.

examples/http-server/Cargo.toml
ROMemories: I'm having a hard time following your comments on this PR: are you saying that this is line is actually the one that you were concerned about, and *not* the one on which you opened the first comment thread?

---

Actually, I've just found out that picoserve v0.18.0 does now expose a `Duration` type (it didn't before), so this is what should be used here (otherwise having this import here *would* still be the proper way of doing that, unfortunately). I'll push a fixup.

examples/http-server/Cargo.toml
```diff
@@ -5,7 +5,7 @@ license.workspace = true
 publish = false
 
 [dependencies]
-ariel-os = { path = "../../src/ariel-os" }
+ariel-os = { path = "../../src/ariel-os", features = ["time"] }
```
ROMemories, line 8: This now enables the `time` Cargo feature so that a time driver is provided, as will be required by #1897.

examples/http-server/Cargo.toml
ROMemories: I've pushed the fixup (and will rename the commit when squashing).

examples/tcp-echo/src/main.rs
```diff
@@ -1,8 +1,9 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::{debug::log::*, net, reexports::embassy_net, time::Duration};
+use ariel_os::{debug::log::*, net, reexports::embassy_net, reexports::embassy_time};
 use embassy_net::tcp::TcpSocket;
+use embassy_time::Duration;
```
kaspar030, line 6: Here the reexport Duration is used, shouldn't rather tcp::TcpSocket be imported (and previously exported) in ariel_os::net?

kaspar030: (what I mean here, using the reexport is unfortunate IMO and should be at best temporary)

examples/http-server/Cargo.toml
anlavandier: I think that the idea is that in the example, `embassy_time` types are only used by `picoserve`. So in the future there could be a configuration where the version of `embassy_time` used in `picoserve` and the one from Ariel OS are not the same and thus the types not compatible. Importing `embassy_time` directly makes it clear that we make no attempts and deliver no guarantee of us following `picoserve`'s `embassy time`

examples/tcp-echo/src/main.rs
ROMemories: You mean providing `ariel_os::net::TcpSocket` and using the existing `ariel_os::time::Duration` with it?

Maybe, but I think what the provide in `ariel_os::net` is orthogonal to this PR, and we don't have a precedent of passing `ariel_os::time` types to other functions we also provide, so it's a separate decision. In the meantime, using the `reexports` for both is consistent and correct, as my expectations are that we guarantee that types from our `reexports` work well together.

kaspar030 (approved):

PR #1903, opened 2026-03-12
fix: warnings in dummy hal

luukoudshoorn:
# Description
These are the fixes I need to apply to have a build for native without warnings on a fairly basic out-of-tree application. 
- Removes 2 unused imports.
- Allows unused code to be unused. Most of these are explicitly marked as unimplemented.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/dummy/executor.rs
+++ b/src/ariel-os-hal/src/hal/dummy/executor.rs
@@ -1,3 +1,5 @@
+#![allow(unused, reason = "used by documentation only")]
+
 use embassy_executor::SpawnToken;
 
 #[doc(hidden)]
@@ -30,5 +32,7 @@ impl Spawner {
     pub fn spawn<S>(&self, _token: SpawnToken<S>) -> Result<(), ()> {
         unimplemented!();
     }
+
+    #[allow(clippy::unused_self)]
     pub fn must_spawn<S>(&self, _token: SpawnToken<S>) {}
 }
--- a/src/ariel-os-hal/src/hal/dummy/mod.rs
+++ b/src/ariel-os-hal/src/hal/dummy/mod.rs
@@ -12,6 +12,7 @@
     clippy::needless_pass_by_value,
     reason = "this dummy module mimics manufacturer-specific crates"
 )]
+#![allow(unused, reason = "used by native and documentation only")]
 
 mod executor;
 
--- a/src/ariel-os-hal/src/hal/dummy/identity.rs
+++ b/src/ariel-os-hal/src/hal/dummy/identity.rs
@@ -1,3 +1,4 @@
 use ariel_os_embassy_common::identity;
 
+#[allow(unused)]
 pub type DeviceId = identity::NoDeviceId<identity::NotImplemented>;


luukoudshoorn:
Don't know why my username didn't get linked to those commits, I'll look into it

luukoudshoorn:
It's a bit unclear to me how to reproduce the CI behavior locally. 
@bergzand suggested laze -C examples/alloc build or laze -C examples/gpio/ build to me, but that's just what happens to trigger the troubles for this PR. I'd probably be guessing again the next time.
It would be nice to have a bit more guidance (preferably in the manual) about what to run locally to get the CI pass more smoothly. It's feels a bit wasteful to use the CI in the trial-and-error way this PR has been going so far :).

kaspar030:
> It would be nice to have a bit more guidance (preferably in the manual) about what to run locally to get the CI pass more smoothly. It's feels a bit wasteful to use the CI in the trial-and-error way this PR has been going so far :).

Agreed that would be useful, noted down. I guess it is not helping that b/c you didn't contribute to the repo yet, every build must be manually allowed, for security reasons...

luukoudshoorn:
Ah that makes sense and will definitely reduce the amount of babysitting in the future

src/ariel-os-hal/src/hal/dummy/mod.rs
```diff
@@ -52,14 +52,15 @@ pub mod uart;
 #[cfg(feature = "usb")]
 pub mod usb;
 
-pub use executor::{Executor, Spawner};
-pub use peripheral::{IntoPeripheral, OptionalPeripherals};
```
bergzand, line 56: The removal of `IntoPeripheral` here is what causes the build failure for rustdoc as far as I can tell.

luukoudshoorn: Interesting, I removed that because it triggered a warning that it wasn't used (and I agreed when shortly looking at the file). But I didn't consider the docs

src/ariel-os-hal/src/hal/dummy/executor.rs
```diff
@@ -1,34 +1,40 @@
 use embassy_executor::SpawnToken;
```
ROMemories, line 3: I think in this particular module, it would be fine to allow it at module-level, while providing a `reason`:
```suggestion
#![allow(unused, reason = "used by documentation only")]

use embassy_executor::SpawnToken;
```
(And then individual `allow(unused)` attributes can be removed.)

src/ariel-os-hal/src/hal/dummy/executor.rs
```diff
@@ -1,34 +1,40 @@
...
 
+#[allow(unused)]
 #[doc(hidden)]
 pub struct Executor;
 
 impl Executor {
-    #[allow(clippy::new_without_default)]
+    #[allow(clippy::new_without_default, unused)]
```
ROMemories: It would be better to use two different `allow` attributes here, as they are motivated by different `reason`s, but that shouldn't be necessary anymore when `allow`ing `unused` at module-level.

src/ariel-os-hal/src/hal/dummy/executor.rs
```diff
@@ -1,34 +1,40 @@
...
 pub struct Spawner;
 
 impl Spawner {
-    #[allow(clippy::result_unit_err)]
+    #[allow(clippy::result_unit_err, unused)]
     pub fn spawn<S>(&self, _token: SpawnToken<S>) -> Result<(), ()> {
         unimplemented!();
     }
+
+    #[allow(unused)]
```
ROMemories:
```suggestion
    #[allow(clippy::unused_self)]
```

kaspar030 (approved):
Did a full `laze build --global -b native` run, didn't spot any warnings.

src/ariel-os-hal/src/hal/dummy/mod.rs
```diff
@@ -52,7 +53,6 @@ pub mod uart;
 #[cfg(feature = "usb")]
 pub mod usb;
 
-pub use executor::{Executor, Spawner};
```
ROMemories, line 55: I don't think these should be removed in this PR (especially not with a blanket `allow(unused)` in this module).

src/ariel-os-hal/src/hal/dummy/mod.rs
```diff
@@ -12,6 +12,7 @@
     clippy::needless_pass_by_value,
     reason = "this dummy module mimics manufacturer-specific crates"
 )]
+#![allow(unused, reason = "used by documentation only")]
```
ROMemories: This `dummy` module is only used by native (the `reason` for the `executor` module was true enough because it would never get used for native):
```suggestion
#![allow(unused, reason = "used by native and documentation only")]
```

ROMemories (commented):

src/ariel-os-hal/src/hal/dummy/mod.rs
luukoudshoorn: I don't think they're strictly _used_ by native, right? Most implementations panic, so if they were ever actually used, things would break

src/ariel-os-hal/src/hal/dummy/mod.rs
luukoudshoorn: Fine by me, the `allow(unused)` solves my main problem (a host build that gives warnings). But out of curiosity: What's their purpose?

src/ariel-os-hal/src/hal/dummy/mod.rs
ROMemories: Sure, but this being an inner attribute, the `reason` applies to the *module* itself, which does get compiled for native.

src/ariel-os-hal/src/hal/dummy/mod.rs
ROMemories: The goal here was initially to maintain some consistency in the public items between the HAL crates (`ariel-os-esp`, `ariel-os-nrf`, `ariel-os-rp`, and `ariel-os-stm32`), and this `dummy` module. I can see this has partially drifted, but possibly removing these items for consistency is out of scope (if this is even needed) of this PR (whose primary goal is to suppress the warnings).

luukoudshoorn: Makes sense, I'll do that

ROMemories (approved):
Thanks!

PR #1910, opened 2026-03-16
fix(hal)!: seal the doc-only `IntoPeripheral` trait

ROMemories:
# Description

This seals the dummy/doc-only variant of `IntoPeripheral` to make sure users never implement it.

BREAKING CHANGE: invalidates inappropriate user implementations of the trait.

## Changelog Entry

The variant of the `IntoPeripheral` trait that is used only for documentation is now sealed. It should never be implemented on anything outside of Ariel OS.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/dummy/peripheral.rs
+++ b/src/ariel-os-hal/src/hal/dummy/peripheral.rs
@@ -5,13 +5,17 @@
 pub struct OptionalPeripherals;
 
 /// Helper trait to support both `Peri` style and singleton style peripherals.
-pub trait IntoPeripheral<'a, P> {
+// NOTE: the trait is sealed as it is only used for documentation, and should never be implemented
+// by users on anything.
+pub trait IntoPeripheral<'a, P>: private::Sealed {
     #[must_use]
     fn into_hal_peripheral(self) -> Self;
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


src/ariel-os-hal/src/hal/dummy/peripheral.rs
```diff
@@ -28,3 +32,7 @@ impl From<Peripherals> for OptionalPeripherals {
         Self {}
     }
 }
+
+mod private {
+    pub trait Sealed {}
```
ROMemories, line 37: The sealing implementation is inspired by [the C-SEALED's example](https://rust-lang.github.io/api-guidelines/future-proofing.html#c-sealed).

ROMemories:
We've considered out-of-band whether it would make sense to also seal the HAL-specific definitions of `IntoPeripheral` but, on ESP, it seems to me that would require adding a blanket implementation of `Sealed for T`, which may not be a good idea:

```rust
pub trait IntoPeripheral<'a, T>: private::Sealed {
    fn into_hal_peripheral(self) -> T;
}

impl<'a, T> private::Sealed for T {}
```

kaspar030:
> on ESP, it seems to me that would require adding a blanket implementation

We can still seal it for the other platforms. #1918

PR #1913, opened 2026-03-17
fix(stm32): enable `time` when needed for the USB CDC-ACM workaround

ROMemories:
# Description

This fixes an issue introduced by #1897: `embassy-time` is internally required in `ariel-os-embassy` when using USB CDC-ACM on STM32, for the workaround.

## Testing

`ci-build:full` should make sure this now builds successfully (fixing the nightly build failure).

## Issues/PRs References

- Fixes an issue introduced by #1897

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -91,7 +91,8 @@ usb-hid = ["dep:usbd-hid", "embassy-usb?/usbd-hid", "usb"]
 
 # `embassy-net` requires a time driver, which HALs provide.
 net = ["dep:embassy-net", "ariel-os-hal/time"]
-usb-ethernet = ["net", "usb"]
+# NOTE: `time` is only needed on STM32 for the workaround.
+usb-ethernet = ["net", "time", "usb"]
 ## Selects the network backend that goes through tun/tap (native only)
 tuntap = ["net"]
 


src/ariel-os-embassy/Cargo.toml
```diff
@@ -91,7 +91,8 @@ usb-hid = ["dep:usbd-hid", "embassy-usb?/usbd-hid", "usb"]
 
 # `embassy-net` requires a time driver, which HALs provide.
 net = ["dep:embassy-net", "ariel-os-hal/time"]
-usb-ethernet = ["net", "usb"]
+# NOTE: `time` is only needed on STM32 for the workaround.
+usb-ethernet = ["net", "time", "usb"]
```
ROMemories, line 95: Ideally `time` would only be enabled on the `stm32` context or, even better, only for `capability = "hw/stm32-usb-synopsys"`, but I'm not sure we can do that currently.

kaspar030: Could do it with laze. Make the `usb-ethernet` module depend on time, only when `context::stm32` is also part of the build, or a marker module for just the stm32-synopsys.
We'd need a time module then. :)

ROMemories: Do you want to block the PR on this or shall we merge it to unblock the nightly build?

kaspar030: unblock

PR #1916, opened 2026-03-17
docs(debug-console): clarify that `defmt` is the default

ROMemories:
# Description

This clarifies in the book that `defmt` is the default, but that it is still a good idea for applications to explicitly select it for explicitness and robustness.

## Changelog Entry

The book now states that `defmt` is the default logger when available for the target.

## Change Checklist

### DIFF
--- a/book/src/debug-console.md
+++ b/book/src/debug-console.md
@@ -46,6 +46,13 @@ if none of them are enabled, logging statements become no-operations.
 Enabling either the `defmt` or `log` [laze modules][laze-modules-book] allows selecting which logging facade and logger is used.
 defmt should be preferred when possible as it results in smaller binaries.
 
+> [!TIP]
+> The `defmt` laze module is favored and enabled by default when possible for
+> the target.
+> Applications that specifically depend on it still need to explicitly
+> [select][laze-modules-book] it to make the dependency explicit and increase
+> robustness to potential future changes.
+
 The precise set of formatting operations and traits required on formatted data
 depends on the selected backend.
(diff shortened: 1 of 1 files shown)


book/src/debug-console.md
```diff
@@ -46,6 +46,11 @@ if none of them are enabled, logging statements become no-operations.
 Enabling either the `defmt` or `log` [laze modules][laze-modules-book] allows selecting which logging facade and logger is used.
 defmt should be preferred when possible as it results in smaller binaries.
 
+> [!TIP]
+> The `defmt` laze module is enabled by default, but applications that
+> specifically depend on it should still [select][laze-modules-book] it to make
+> the dependency explicit and increase robustness to potential future changes.
```
chrysn: default: Is that so and we commit to it? Right now it's the case for most boards, but not for native. So unless an application is tied closely to the hardware, it should rely on the laze modules to select the right thing.

Maybe:

```suggestion
> The `defmt` laze module is enabled by default in most situations. Applications that
> specifically depend on it still need to [select][laze-modules-book] it to make
> the dependency explicit and increase robustness to potential future changes.
```

ROMemories: That's true, that was too broad a statement. I've rephrased slightly differently, trying to convey that there is a "bias" towards defmt, but that it still depends on the target. Tell me whether that fully addresses you concerns.

chrysn: Thanks, works for me. (My emphasis was on "needs to").

Issue #1811, opened 2026-02-18
Bumping MSRV to 1.94


ROMemories:
Rust 1.94 is expected to be released on March 5 and Cargo should see the stabilization of `-Zconfig-include` which we have been relying on so that our patches get discovered by Cargo, including in out-of-tree applications.

When this version is released, we should bump our MSRV to it. This should make our build configuration easier to use, as all Cargo tooling (including `cargo clippy` and `cargo doc`) would now discover the patches automatically, without the need to rely on nightly's *Cargo*, either through `cargo +<nightly-version>` or by pinning the nightly version in the directory through rustup.

I'm marking this as a release blocker for now, but if this doesn't have the effect I expect we could consider it non-blocking.

chrysn:
IIRC, we can likely not do that on March 5th yet, because the ESP toolchains do have some lag.

chrysn:
When or after that is done, we can finally make sure that `laze build -b myboard cargo doc` works. (Currently that's a mess due to `--config` being understood differently than in early nightlies).

Issue #1875, opened 2026-03-09
Bumping MSRV to 1.95


ROMemories:
Rust 1.95 is expected to be released on April 16.

It should bring [`cfg_select!`](https://doc.rust-lang.org/nightly/core/macro.cfg_select.html), which we'll be able to use to replace the `cfg-if` dependency with, as shown in #1851. This will allow us to remove a dependency and should make it easier to use that construct in places where we wouldn't want to bring that extra dependency in.

PR #1917, opened 2026-03-17
ci: use a stable toolchain for vetting

anlavandier:
# Description

Now that our MSRV is 1.94 and `config-include` has been stabilized we don't need the vetting to run on nightly anymore.

## Testing
The CI for Dependency Vetting should not fail.

## Issues/PRs References
#1811 

## Change Checklist

### DIFF
--- a/.github/workflows/vetting.yml
+++ b/.github/workflows/vetting.yml
@@ -16,9 +16,6 @@ jobs:
         with:
           persist-credentials: false
 
-      - id: get_toolchain
-        run: echo "toolchain=$(scripts/rust-toolchain.sh)" >> $GITHUB_OUTPUT
-
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master
         with:


.github/workflows/vetting.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master
         with:
-          toolchain: ${{ steps.get_toolchain.outputs.toolchain }}
+          toolchain: 1.94
```
ROMemories, line 22: Can we even make it just `stable`? That's another version number that we won't have to bump, and I don't *think* it has to match our MSRV, does it (I don't think `cargo vet` does dependency resolution)?
```suggestion
          toolchain: stable
```

.github/workflows/vetting.yml
ROMemories: And I suppose we can remove this as well?
```suggestion
```

.github/workflows/vetting.yml
anlavandier: I think that it has to do some kind of dependency resolution, otherwise why did it previously fail on `X requires feature Y but doesn't contain that feature` when using previous stable versions ?

ROMemories: I was under the impression that Cargo would almost always do that, but you're right, it does have to do dependency resolution to obtain the list of enabled packages. So nothing to do here, but we should remember to bump this next time (I'll add it to #1875).

PR #1918, opened 2026-03-17
docs: add `IntoPeripheral` trait documentation

### LINKED ISSUE #1734: Revisiting `IntoPeripheral`'s docs
The `IntoPeripheral` trait's dummy implementation, which is used for documentation currently looks like this:

```rust
pub trait IntoPeripheral<'a, P> {
    fn into_hal_peripheral(self) -> Self;
}
```

- [x] ~~This doesn't match the actual `IntoPeripheral`'s definitions provided by each individual HAL (at least regarding the `Self` return type), which may make it *look* unsound, because the trait isn't sealed (and therefore the trait could be implemented on *any* type and passed to drivers).~~ It's complicated: `ariel-os-esp` implements `IntoPeripheral for T` and therefore `IntoPeripheral::into_hal_peripheral()` returns `T`, while Ariel OS HALs that rely on Embassy HALs implement `IntoPeripheral` for `Peri` and `into_hal_peripheral()` returns `Peri` (aka `Self`).

- [x] The doc-only definition of `IntoPeripheral` should be sealed. #1910

- [x] The documentation of the trait should be written to be useful for users and, in particular, it should explain the reasons for its existence, and the fact should users never have it to implement it. #1918
- [x] The trait's method should have documentation, and `#[deny(missing_docs)]` added to the dummy module. #1918

kaspar030:
# Description

Seals `IntoPeripheral` traits where it make sense (not on ESP as there it is implemented on just `T`).
Adds docs about it.

## Testing

This is trait gymnastics, CI should catch if it goes wrong.

## Issues/PRs References

- Depends on #1910.
- Fixes #1734

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/dummy/peripheral.rs
+++ b/src/ariel-os-hal/src/hal/dummy/peripheral.rs
@@ -1,13 +1,28 @@
+#![deny(missing_docs)]
+
 /// Dummy type.
 ///
 /// See the `OptionalPeripherals` type of your Embassy HAL crate instead.
 #[doc(hidden)]
 pub struct OptionalPeripherals;
 
-/// Helper trait to support both `Peri` style and singleton style peripherals.
-// NOTE: the trait is sealed as it is only used for documentation, and should never be implemented
-// by users on anything.
+/// This is an adapter trait necessary for compatibility with the underlying HALs.
+///
+/// The underlying HALs manage peripherals differently: for instance, the drivers from
+/// Embassy HALs require a `Peri` instance constrained on the required peripheral ZST, while
+/// drivers from `esp-hal` require the peripheral ZST directly.
+///
+/// In practice, this trait allows passing the peripheral ZSTs [obtained through Ariel OS
+/// facilities][obtaining-peripheral-access-book] directly to Ariel OS drivers that have a trait
+/// bound on [`IntoPeripheral`].
+///
+/// Note: It is never necessary to implement this trait outside of Ariel OS, and the trait is
+/// therefore sealed.
+///
+/// [obtaining-peripheral-access-book]: https://ariel-os.github.io/ariel-os/dev/docs/book/application.html#obtaining-peripheral-access
+// NOTE: Each Ariel OS HAL needs to provide its own definition of this trait, which should be sealed when possible.
 pub trait IntoPeripheral<'a, P>: private::Sealed {
+    /// Converts this peripheral instance into the type required by the HAL.
     #[must_use]
     fn into_hal_peripheral(self) -> Self;
 }
--- a/src/ariel-os-nrf/src/lib.rs
+++ b/src/ariel-os-nrf/src/lib.rs
@@ -81,14 +81,14 @@ pub trait IntoPeripheral<'a, T: PeripheralType>: private::Sealed {
     fn into_hal_peripheral(self) -> Peri<'a, T>;
 }
 
+impl<T: PeripheralType> private::Sealed for Peri<'_, T> {}
+
 #[doc(hidden)]
 impl<'a, T: PeripheralType> IntoPeripheral<'a, T> for Peri<'a, T> {
     fn into_hal_peripheral(self) -> Peri<'a, T> {
         self
     }
 }
-
-impl<'a, T: PeripheralType> private::Sealed for Peri<'a, T> {}
 
 mod private {
     pub trait Sealed {}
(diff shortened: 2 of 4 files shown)


src/ariel-os-nrf/src/lib.rs
```diff
@@ -88,6 +88,12 @@ impl<'a, T: PeripheralType> IntoPeripheral<'a, T> for Peri<'a, T> {
     }
 }
 
+impl<T: PeripheralType> private::Sealed for Peri<'_, T> {}
```
ROMemories: optional: I would prefer that that implementation comes before the previous one: as it's required by the `IntoPeripheral` impl it makes sense for it to come first, and that makes it consistent with the `dummy` module.

src/ariel-os-hal/src/hal/dummy/peripheral.rs
```diff
@@ -4,15 +4,30 @@
...
+///
+/// - `ariel_os_esp::IntoPeripheral`
+/// - `ariel_os_nrf::IntoPeripheral`
+/// - `ariel_os_rp::IntoPeripheral`
+/// - `ariel_os_stm32::IntoPeripheral`
+///
+/// NOTE: Where possible, the trait is sealed as it should only be implemented inside those hal crates.
+pub trait IntoPeripheral<'a, P>: private::Sealed {
     #[must_use]
     fn into_hal_peripheral(self) -> Self;
```
ROMemories, line 27: This method still needs documentation, as pointed out in #1734. And then `deny(missing_docs)` needs to be added.

src/ariel-os-hal/src/hal/dummy/peripheral.rs
```diff
@@ -4,15 +4,30 @@
...
+/// This is actually a documentation-only placeholder implementation. On actual hardware platforms, one of
+/// these implementations is used:
+///
+/// - `ariel_os_esp::IntoPeripheral`
+/// - `ariel_os_nrf::IntoPeripheral`
+/// - `ariel_os_rp::IntoPeripheral`
+/// - `ariel_os_stm32::IntoPeripheral`
+///
+/// NOTE: Where possible, the trait is sealed as it should only be implemented inside those hal crates.
+pub trait IntoPeripheral<'a, P>: private::Sealed {
```
ROMemories: I've rewritten this in a way that I hope makes it more useful to users. In particular:

- The trait itself isn't internal, only the conversion it allows is.
- `OptionalPeripherals` is not part of the docs, on purpose, as it is an implementation detail.
- Listing the definitions from each Ariel OS HAL doesn't seem useful, will go stale, and is an implementation detail.
- Everything else about individual Ariel OS HALs is an implementation detail.

```suggestion
/// This is an adapter trait necessary for compatibility with the underlying HALs.
///
/// Indeed, they manage peripherals differently: for instance, the drivers from Embassy HALs require a `Peri` instance constrained on the required peripheral ZST, while drivers from `esp-hal` require the peripheral ZST directly.
/// In practice, this trait allows passing the peripheral ZSTs [obtained through Ariel OS facilities][obtaining-peripheral-access-book] directly to Ariel OS drivers that have a trait bound on [`IntoPeripheral`].
///
/// Note: It is never necessary to implement this trait outside of Ariel OS, and the trait is therefore sealed.
///
/// [obtaining-peripheral-access-book]: https://ariel-os.github.io/ariel-os/dev/docs/book/application.html#obtaining-peripheral-access
// NOTE: Each Ariel OS HAL needs to provide its own definition of this trait, which should be sealed when possible.
```
(This likely needs reflowing.)

ROMemories: That may have eaten the trait declaration :/

src/ariel-os-hal/src/hal/dummy/peripheral.rs
```diff
@@ -1,13 +1,27 @@
+#[deny(missing_docs)]
```
ROMemories:
```suggestion
#![deny(missing_docs)]

```

src/ariel-os-hal/src/hal/dummy/peripheral.rs
kaspar030: yup but that was easily spotted. I did reword line 26 slightly, maybe double check if you're happy with that.

kaspar030:
- marked `changelog:skip`, the mentionable change was #1920.
- technically, for `ariel-os-nrf/stm32/rp`, this is a breaking change. for `ariel-os` it is not, so not marking as such.

src/ariel-os-hal/src/hal/dummy/peripheral.rs
ROMemories: I think you've just rephrased the clause before the colon? Fine by me.

PR #1923, opened 2026-03-18
docs(release-checklist): bump version numbers after the changelog update

ROMemories:
# Description

In sbd-gen (https://github.com/ariel-os/sbd/pull/126#issuecomment-4037586524), we've realized that merely bumping the version numbers triggered [release-plz](https://github.com/release-plz/release-plz) to publish the packages. Even tough we don't yet use release-plz for `ariel-os*` crates, it seems safer to move bumping the version numbers *after* the changelog update in the release checklist.

## Change Checklist

### DIFF
--- a/book/src/development-processes.md
+++ b/book/src/development-processes.md
@@ -79,16 +79,6 @@ The following steps must be followed when preparing a new release of `ariel-os`:
 1. Verify there are no [release blocking bugs](https://github.com/ariel-os/ariel-os/issues?q=state%3Aopen%20label%3Arelease-blocker).
 1. Update the roadmap if necessary.
 1. Check whether deprecated items should be removed, if any.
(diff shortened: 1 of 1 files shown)


bergzand:
Do you think it makes sense to already add a reminder here that *when release-plz is used*, this step triggers the publish?

kaspar030:
I think changelog and version bumps should actually be *done by release-plz*. So I'd assume that checklist will change anyways, when we start using release-plz.

ROMemories:
> Do you think it makes sense to already add a reminder here that _when release-plz is used_, this step triggers the publish?

Not necessarily, because the release-plz bot should usually be the one bumping the versions; reordering the steps makes sure that everything is ready where it would be time to bump them.

Issue #1925, opened 2026-03-18
Documenting support for undocumented, board-specific functionality


ROMemories:
There is a couple of pieces of functionality that are currently undocumented because they are currently implemented in a way that's tied to specific boards, and we weren't able to document this before. Thanks to #1428 this is now possible.

The following could be added to the documentation and the support info:

- [x] Support for Ethernet on STM32 chips/boards #1926
- [ ] Support for logging-over-UART (previously incorrectly called debug-over-UART) on specific boards #1008 #2100

PR #993, opened 2025-04-12
feat(network): Add ethernet from nucleo-144 board family

Naviss:
# Description

This PR add the ability of providing network from the embedded PHY of the Nucleo-144 board family

## Open Questions

Each nucleo board have a different way to connect the PHY depending on the STM32 family used.
How do we properly abstract the pin definition of the ethernet PHY ?

## Change checklist
- [x] I have cleaned up my commit history and squashed fixup commits.
- [ ] I have followed the [Coding Conventions](https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html).
- [ ] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

PR #1927, opened 2026-03-20
fix(eth-stm32)!: randomize the Ethernet MAC address on every boot

ROMemories:
# Description

Support for Ethernet on STM32 boards was introduced in #993 (but is currently undocumented, see #1925), with a hard-coded MAC address. This replaces this fixed address, with a randomized address in the AAI format.

## Testing

Successfully tested the `udp-echo` example using:

```sh
laze -C examples/udp-echo/ build -s eth-stm32 -b st-nucleo-h755zi-q
```

A log statement can manually be added to print the generated MAC address and check it complies to the expected format.

## Issues/PRs References

- Needed before #1926

## Open Questions

- Should we rely on the device identity instead? It's unclear to me how to make sure that only one interface gets each MAC address from [`interface_eui48()`](https://ariel-os.github.io/ariel-os/0.4.0/docs/api/ariel_os/identity/fn.interface_eui48.html).

## Changelog Entry

(STM32) Ethernet support on STM32 MCUs is now using a MAC address randomized on every boot instead of a fixed one.

## Change Checklist

### DIFF
--- a/src/ariel-os-stm32/src/eth.rs
+++ b/src/ariel-os-stm32/src/eth.rs
@@ -7,12 +7,14 @@ bind_interrupts!(struct Irqs {
     ETH => eth::InterruptHandler;
 });
 
+type MacAddress = [u8; 6];
+
 pub type NetworkDevice = Ethernet<'static, ETH, GenericPhy>;
 
 pub fn device(peripherals: &mut crate::OptionalPeripherals) -> NetworkDevice {
     static PKTS: StaticCell<eth::PacketQueue<4, 4>> = StaticCell::new();
 
-    let mac_addr = [0xCA, 0xCC, 0xCC, 0xCC, 0xCC, 0xCC];
+    let mac_addr = generate_random_aai_mac_addr();
 
     Ethernet::new(
         PKTS.init(eth::PacketQueue::<4, 4>::new()),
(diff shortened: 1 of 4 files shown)


src/ariel-os-stm32/src/eth.rs
```diff
@@ -31,3 +33,16 @@ pub fn device(peripherals: &mut crate::OptionalPeripherals) -> NetworkDevice {
...
+
+fn generate_random_aai_mac_addr() -> MacAddress {
+    use rand_core::RngCore as _;
+
+    let mut eui48 = [0u8; _];
+    ariel_os_random::crypto_rng().fill_bytes(&mut eui48);
+
+    // Enforce the `?2-??-??-??-??-??` pattern of an AAI (Administratively Assigned Identifier).
+    eui48[0] &= 0xf0;
+    eui48[0] |= 0x02;
```
ROMemories, line 45: For now copied from:

https://github.com/ariel-os/ariel-os/blob/ddace5226cabc2417cdc819861c4b415d7275718/src/ariel-os-embassy-common/src/identity.rs#L162-L164

src/ariel-os-stm32/src/eth.rs
```diff
@@ -31,3 +33,16 @@ pub fn device(peripherals: &mut crate::OptionalPeripherals) -> NetworkDevice {
         mac_addr,
     )
 }
+
+fn generate_random_aai_mac_addr() -> MacAddress {
```
ROMemories, line 37: With some refactor, this could likely find its place in our public API, possibly in the `ariel_os::random` module.

ROMemories:
Superseded by #1942.

Issue #845, opened 2025-02-20
Testing for source availability


chrysn:
### Description

With the tooling we currently have, we provide no means for users (or ourselves) to assess whether we build our tool from available sources or not.

`cargo deny` does *not* provide this: It just checks for the license advertised in Cargo, and those by the letter just describe terms by which that software is distributed, not whether those terms are actually useful.

### Motivation

Case in point: On any ESP running with WiFi, we have a dependency on the blobs published by Espressif. Those are published under some permissive open source license (which among other things enables reverse engineering without copyright trouble), but nonetheless they do not come with actual source, making them by all definitions of the term not open source software (or FLOSS). Yet, that dependency never showed up in our previous scans.
 
* We may need this as part of our SBOM (but that currently has super vague requirements).
  * In particular, any blobs shipped under whichever license hide any components that would otherwise need to be listed in an SBOM (like, vulnerable versions of C libraries)
* Having nonfree sources (mandatory or optional, with different consequences) affects many areas of the project:
  * We present ourselves as a free software project; we could do misleading advertising here and not even notice it.
  * This can be a factor in whether or not contributors participate in the project.
  * This is relevant for downstream users: If a firmware is built that is not free software, it can not be shipped with Debian as easily.

### Requirements

1. There should be an option in per-build checks to verify whether or not the output contains non-free dependencies.
2. There should be an allow-list for which non-free components are accepted (possibly per-board), similar to how we explicitly allow-list licenses.

### Open questions

* Is cargo deny willing to check additional sources?
* Where would esp32-wifi-sys even declare this? (Technically the `license=` field is about copyright, and that is about a copyright indication; but OSI's and FSF's recognition of software being free software does not exclusively depend on using an approved, although in the normal case that is sufficient).

Initially assigning to @emmanuelsearch for triage as to whether this warrants any priority with regard to the SBOM aspect.

### Are you planning to do it yourself in a pull request?

Yes

PR #1933, opened 2026-03-20
refactor(cyw43): use the now published `cyw43-firmware` crate

ROMemories:
# Description

This leverages the `cyw43-firmware` crate that was published last year to replace the blobs we shipped directly. This makes it easier for us to keep them updated in the future, and makes license tracking more straightforward, as the blobs are now in a dedicated package along with the license files and the appropriate `license-file` manifest key.

## Testing

Tested the `http-client` (over Wi-Fi) and the `ble-scanner` examples on a rpi-pico2-w. Couldn't spot a difference in behavior. The `size` laze task says the `.text` section is now 4 bytes *smaller*.

Additionally, I have downloaded the [three blobs from the package](https://docs.rs/crate/cyw43-firmware/0.1.0/source/dist/cyw43-firmware/), and compared their SHA-256 hashes to that of the ones we were shipping: the three hashes match.

## Issues/PRs References

- Related to #845

## Open Questions

- ~~Should this go into the changelog? This is technically a completely internal refactor, but users might still be interested to know about it wrt license tracking.~~

## Changelog Entry

(CYW43) Firmware blobs required for the CYW43 chip are now provided through the `cyw43-firmware` crate, making license tracking easier.

## Change Checklist


src/ariel-os-rp/Cargo.toml
```diff
@@ -86,6 +87,8 @@ ble-cyw43 = [
   "_cyw43",
   "ariel-os-random/csprng",
   "ble",
+  "cyw43-firmware?/bluetooth",
+  "cyw43-firmware?/wifi",
```
ROMemories, line 91: It *might* be possible to not enable this, but this would require a small refactoring of how these blobs are used, and this is out of scope of this PR.

src/ariel-os-rp/src/cyw43.rs
bergzand, line 80: This comment needs a small update, at least the `include_bytes!` doesn't match the code anymore.

bergzand (approved):

src/ariel-os-rp/src/cyw43.rs
ROMemories: I've considered it, but I don't really see how to update it: `include_bytes!` *is* used by `cyw43-firmware` internally, and that's the only way to do this in Rust, so I think this is still understandable enough.

PR #1937, opened 2026-03-23
refactor!: replace `eth` with `ethernet`

ROMemories:
# Description

This renames the Cargo features and Rust modules named `eth` and `eth-stm32` to `ethernet` and `ethernet-stm32` respectively. These names were originally introduced in #993, but the abbreviation wasn't really motivated (and likely biased by STM32 peripheral naming), and this makes it more consistent with the `usb-ethernet` laze modules. The motivation for this rename has also been discussed with @kaspar030.

There is no renaming occurrences of `eth_` in the codebase, and the only remaining occurrences of `eth\b` are in the STM32-specific module, because the STM32 *periperhal* is named ETH.

## How to review this PR

This PR is better reviewed commit by commit (I can also easily split it if needed).

## Testing

```sh
laze -C examples/udp-echo/ build -s ethernet-stm32 -b st-nucleo-h755zi-q run
```

## Issues/PRs References

- Needed for #1926

## Changelog Entry

Even though the renames are technically breaking changes, it might not make sense for this to go in the changelog, as it was completely undocumented until now (and this will be documented in #1926). Existing applications that would have been using this undocumented functionality would simply fail to compile, and this wouldn't result in surprising/unwanted behavior.

## Change Checklist


src/ariel-os-embassy/src/eth.rs
ROMemories, line 1: This file has just been renamed to `ethernet.rs` and the feature-gate renamed, so Git doesn't detect it as a file rename.

ROMemories:
CI has built successfully for `ci-build:stm32` and `ci-build:small`.

PR #1941, opened 2026-03-24
feat: driver for the ASAIR AHT20 relative humidity/temperature sensor

anlavandier:
# Description

This PR adds a new built-in sensor for the [ASAIR AHT20](https://files.seeedstudio.com/wiki/Grove-AHT20_I2C_Industrial_Grade_Temperature_and_Humidity_Sensor/AHT20-datasheet-2020-4-16.pdf).

## Testing

Running the `sensors-debug` example on the Unihiker K10 using the ~~`espressif-esp32-s3-devkitc-1`~~ `unihiker-k10` builder. 

## Open Questions

Currently the testing "hijacks" a builder. It might be better to have a custom board description for the Unihiker. 
## Changelog Entry

A sensor driver for the AHT20, compatible with Ariel OS's sensor API, is now available.

## Change Checklist


examples/sensors-debug/src/pins.rs
```diff
@@ -1,12 +1,12 @@
...
-    i2c_sda: GPIO2,
-    i2c_scl: GPIO0,
-});
+// #[cfg(context = "esp")]
+// pub type SensorI2c = i2c::controller::I2C0;
+// #[cfg(context = "esp")]
+// ariel_os::hal::define_peripherals!(Peripherals {
+//     i2c_sda: GPIO2,
+//     i2c_scl: GPIO0,
+// });
```
anlavandier: I'm not sure that I understand why this is here in the first place. In any case changes to this example might be reverted before merging

examples/sensors-debug/Cargo.toml
```diff
@@ -11,6 +11,7 @@ ariel-os = { path = "../../src/ariel-os", features = [
   "time",
 ] }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os-sensor-aht20 = { workspace = true, features = ["debug"]}
```
anlavandier: Need to remove this debug feature altogether now that it's not needed

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+    },
+    signal::Signal as ReadingSignal,
+};
+use ariel_os_sensors_utils::AtomicState;
+use embassy_sync::{
+    blocking_mutex::raw::CriticalSectionRawMutex, mutex::Mutex, once_lock::OnceLock, signal::Signal,
+};
+#[cfg(not(test))]
+use embassy_time::Timer;
+use embedded_hal_async::i2c::{
```
anlavandier, line 20: Using Embassy Time's timer in the testing configuration leads to poison errors.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+
+            self.reading.signal(self.measure().await);
+        }
+    }
+
+    /// Trigger a measurement and process the results.
+    ///
+    /// # Errors
+    /// - When the I2C connection fails.
+    /// - When the CRC of either the temperature or the relative humidity are incorrect.
```
anlavandier: Need to change this to highlight that it's a single CRC for the whole package of data

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+            ],
+        )
+        .await
+        .map_err(|_| ReadingError::SensorAccess)?;
+        // Wait for at least 80 ms, see Section 5.4.3 of the Datasheet
+        #[cfg(not(test))]
+        Timer::after_millis(80).await;
+        let mut buf = [0u8];
+        loop {
+            match i2c.read(I2C_ADDRESS, &mut buf).await {
```
anlavandier, line 180: This uses a generic I2C read instead of a dedicated command because the `ReadStatusReg` command has unexpected side effects on the CRC

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+        }
+        // Now read |status|humi|humi|humi-temp|temp|temp|crc|
+        let mut buf = [0u8; 7];
+        i2c.read(I2C_ADDRESS, &mut buf)
+            .await
+            .map_err(|_| ReadingError::SensorAccess)?;
+
+        if !crate::check_crc(&buf[0..6], buf[6]) {
+            // FIXME: Use a better error type
+            return Err(ReadingError::SensorAccess);
```
anlavandier: I expect this will happen fairly often since CRC seems quite common in sensors. However IIUC, this would require updating all of the sensor ecosystem to use this new version of the trait. We might want to do this sooner rather than later though

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+        }
+
+        let humidity =
+            (u32::from(buf[1]) << 12) + (u32::from(buf[2]) << 4) + u32::from(buf[3]) / 16;
+        let temperature =
+            ((u32::from(buf[3]) % 16) << 16) + (u32::from(buf[4]) << 8) + u32::from(buf[5]);
```
anlavandier: The datasheet specifies that two 20 bit values are sent as 5 bytes hence the bit shifts used to recover the numerical values

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+
+        // H = (Sh / 2^20) * 100 % see Section 6.1 of the Datasheet.
+        // This naturally brings the resolution down the %.
+        let humi = (humidity.cast_signed() * 100) >> 20;
+
+        // T = (St / 2^20) * 200 - 50 see Section 6.2 of the Datasheet.
+        // We pre-multiply by 10 to keep the decimal digit
+        let temp = ((temperature.cast_signed() * 200 * 10) >> 20) - 500;
```
anlavandier: I chose these resolution to stay consistent with the accuracy as described by the datasheet

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+    // Accuracy of 2 °C between 60 °C and 85 °C.
+    else if 200 < temp && temp < 850 {
+        SampleMetadata::SymmetricalError {
+            deviation: 20,
+            bias: 0,
+            scaling: -1,
+        }
+    }
```
anlavandier, line 86: Maximum error over the ranges

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    // Otherwise we are outstide of the scope of work of the sensor
+    else {
+        unreachable!("We are outisde of the working conditions of this sensor");
+    }
+}
+
+fn h_accuracy(humi: i32) -> SampleMetadata {
```
anlavandier, line 93: Same as for the temperature accuracy

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+/// Calculate the CRC of a 5 byte value.
+/// The algorithm is described in section 5.4.4
+/// The important informations are:
+/// - the polynomial: X^8 + X^5 + X^4 + 1,
+/// - and the initial value: 0xFF,
+///
+/// Usual CRC notation doesn't take into account the most significant
+/// coefficient and hence here it would be `0b0011_0001` = 0x31.
+fn calculate_crc(data: &[u8]) -> u8 {
+    let crc = Crc::<u8>::new(&CRC_8_NRSC_5);
```
anlavandier, line 122: It could be that the `crc_any` crate is better as it might not pull in the entire CRC catalogue of the `crc` crate.

Cargo.toml
```diff
@@ -154,6 +154,9 @@ ariel-os-utils = { path = "src/ariel-os-utils", default-features = false }
 ariel-os-sensor-lis2du12 = { path = "src/sensors/ariel-os-sensor-lis2du12" }
 ariel-os-sensor-lps22df = { path = "src/sensors/ariel-os-sensor-lps22df" }
 ariel-os-sensor-stts22h = { path = "src/sensors/ariel-os-sensor-stts22h" }
+ariel-os-sensor-aht20 = { path = "src/sensors/ariel-os-sensor-aht20" }
+
+
```
anlavandier:
```suggestion
```

Cargo.toml
ROMemories, line 32:
```suggestion
  "src/sensors/ariel-os-sensor-ath20",
  "src/sensors/ariel-os-sensor-lis2du12",
```

Cargo.lock
```diff
@@ -4227,7 +4244,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6a793df0d7afeac54f95b471d3af7f0d4fb975699f972341a4b76988d49cdf0c"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
```
ROMemories, line 4255: This (and the bump of `windows-sys`) are unrelated, even though Cargo really wants to bump them.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+
+    fn init_sensor(sht3x: &'static Aht20<I2cDeviceMock>) {
+        embassy_futures::block_on(async {
+            let peripherals = Peripherals {};
+            let i2c_device = I2cDeviceMock::default();
+
+            sht3x.init(peripherals, i2c_device).await;
+        });
+    }
+}
```
ROMemories, line 624: From what I can see, these are redundant with that of the STTS22H sensor: these tests were primarily meant to test the internal signaling logic, but if it's reused here it's not really necessary to duplicate those tests. The `fetch_reading` is still useful though I think, as it validates the internal conversion logic.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: I do think the `SensorAccess` variant is well suited for this: it's not possible to attribute a failing CRC check to anything in particular, and it could well be because of a bit flip on the bus. The documentation of that variant could be improved/expanded a bit to clarify that failing CRC checks *do* fall under that variant though.

src/sensors/ariel-os-sensor-aht20/src/lib.rs
ROMemories: I've used `crc-any` before, but using `crc` here seems fine (especially as it's already in our tree). Bringing `crc-catalog` doesn't seem like an issue, it's a very small crate.

src/sensors/ariel-os-sensor-aht20/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+] }
+embassy-futures = { workspace = true }
+embassy-time = { workspace = true, features = ["std"] }
+
+[features]
+_test = []
+
+# Feature that enables one-shot measurements without requiring
+# the sensor to run in a separate task
+no_runner = ["dep:embassy-futures"]
```
ROMemories: I would strongly prefer we remove this feature (and the associated logic) for now, and we see how we handle this in a follow-up.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+            .map_err(|_| ())?;
+        Ok(())
+    }
+
+    /// Reads the status register and return it as a u8.
+    ///
+    /// # Errors
+    /// - When the I2C connection fails for any reason.
+    ///
+    pub async fn read_status(&'static self) -> Result<u8, ()> {
```
ROMemories: I'm not sure what that public method is for: it seems like it's not called in the module, and it's not part of the usual de facto interface of sensor drivers. If it's needed for something else, I would prefer we do this separately.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+
+            self.state.set(State::Enabled);
+        }
+    }
+
+    /// Soft resets the device.
+    ///
+    /// # Errors
+    /// - When the I2C connection fails.
+    pub async fn reset(&self) -> Result<(), ()> {
```
ROMemories:
```suggestion
    async fn reset(&self) -> Result<(), ()> {
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,640 @@
...
+            label,
+            i2c: OnceLock::new(),
+            #[cfg(not(feature = "no_runner"))]
+            signaling: Signal::new(),
+            reading: ReadingSignal::new(),
+        }
+    }
+
+    /// Initializes the driver.
+    pub async fn init(&'static self, _peripherals: Peripherals, mut i2c_device: I2C) {
```
ROMemories: This should likely call `reset()` somewhere.

ROMemories (commented):
A very incomplete review for now, I'll do another pass later.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Yes I copied them from the STTS22H and adapted them. I think that it's still fine to keep the signaling related tests as it's an implementation detail of the driver.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: I don't understand why I wouldn't want to expose a `reset` function to users of my driver

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: It implements a command from the datasheet so IMO it has its place in a sensor driver

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: I disagree on that basic that `ReadingError::SensorAccess` is not suited for cases where the sensor replies with garbage data. It's quite different from the i2c bus failing or the sensor not being powered on. CRC errors might also be bit flips on the bus as you said which are simple please retry errors

Cargo.toml
anlavandier: Done in [0b32325](https://github.com/ariel-os/ariel-os/pull/1941/commits/0b323252a239b3b52d5bd8a015d70d2e6c513027)

src/sensors/ariel-os-sensor-aht20/Cargo.toml
anlavandier: Done in [0b32325](https://github.com/ariel-os/ariel-os/pull/1941/commits/0b323252a239b3b52d5bd8a015d70d2e6c513027)

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: I don't think I agree necessarily. Also, I don't like that this would mean that `reset` has to take an I2C device as an argument

Cargo.lock
anlavandier: Everytime I run any code this get bumped so I don't know what the best remedy to this right now. Probably I'll do a final rebase at the end right before merging and manually remove these bumps if they are still there

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    fn set_mode(&self, mode: SensorMode) -> Result<State, SetModeError> {
+        let new_state = self.state.set_mode(mode);
+
+        if new_state == State::Uninitialized {
+            Err(SetModeError::Uninitialized)
+        } else {
+            Ok(new_state)
+        }
```
ROMemories: This is now simpler following #1960:
```suggestion
        self.state.set_mode(mode)
```

anlavandier: Done in 2320ae8

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// This command has unclear side effects on the CRC
+    /// when used after a measurement has been triggered.
+    /// Issuing a general Read command and reading a single byte
+    /// has the same effect and thus should be preferred.
+    ReadStatusReg = 0x71,
+}
+
+// STATUS register bits, see Table 10 of the
+/// Set to 1 if the sensor is calibrated.
+pub const CALIBRATION_STATUS: u8 = 1 << 3;
```
ROMemories: All these consts should be private.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: These drivers are not primarily intended as generic sensor drivers: they are meant to be used by/through the `ariel_os_sensors`/`ariel_os::sensors`'s interface/registry, which means that adding methods on the inherent `impl` block is not useful as they won't be usable (at least not through the registry). This only works for `new()` and `init()`, if they follow the implicit interface, because calls to these are intended to be codegened.

In this particular case, the *device* reset should definitely not be exposed as is, as that would allow corrupting the device+driver state. There is also no reason to reset the device after bootup, and all lifecycle management should be done through `Sensor::set_mode()`. However, the device should also definitely be reset when the driver is initialized, as resetting the MCU will be not automatically reset the sensor device (this only happens when power cycling). Therefore, there should be a `reset()` method, called from `init()`, but not exposed.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: See https://github.com/ariel-os/ariel-os/pull/1941/changes#r3051809114.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,614 @@
...
+    blocking_mutex::raw::CriticalSectionRawMutex, mutex::Mutex, once_lock::OnceLock, signal::Signal,
+};
+#[cfg(not(test))]
+use embassy_time::Timer;
+use embedded_hal_async::i2c::{
+    Error as _, ErrorKind as I2CErrorKind, I2c, NoAcknowledgeSource::Data,
+};
+
+#[cfg(feature = "no_runner")]
+use embassy_futures::block_on;
```
ROMemories: I think that's a leftover:

```suggestion
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: See https://github.com/ariel-os/ariel-os/pull/1941/changes#r3051809114: the goal is not to expose the sensor device functionality as is, but through the `Sensor` API.

ROMemories (commented):
I'll have a look at the rest of the implementation later.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: I understand your last point about cycle management. I will make the method in private and change the arguments to fit the expected calling context.

Regarding your second paragraph I agree that the drivers are primarily meant to be used through the registry, I however don't think that we should assume that this is the only context in which they will be used. The main limitation I see with the registry approach is that it relies on the `Sensor` trait so it can't account for relevant sensor details like configuration options, 

This is relevant at the very least in debugging contexts which is I believe why the "codegened" `sensors.rs` in `example/sensors-debug` reexport the actual sensor refs. 

On this particular sensor there isn't much to configure so I'm going to comply with what you suggest but on other case where there is a wide variety of configurable usage conditions that impact accuracy and/or power consumption or some sanity checking features (e.g. the onboard heater of the STH3x temperature sensor)I think there is value in having a more general interface living alongside the `Sensor` trait stuff.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: After thinking a bit more about this, I think that this still makes the cut because it is useful for debugging. I can feature-gate if you think that this shouldn't be here by default.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Made `reset` private and part of `init` in df5d234a

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Correct, thanks for spotting this. Fixed in e939de44

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in 256a6c5c

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in df5d234

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Done in b4669a8c

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+};
+#[cfg(not(test))]
+use embassy_time::Timer;
+use embedded_hal_async::i2c::{
+    Error as _, ErrorKind as I2CErrorKind, I2c, NoAcknowledgeSource::Data,
+};
+
+// The I2C address is fixed
+const I2C_ADDRESS: u8 = 0x38;
+
```
ROMemories, line 26: There needs to be a `Config` struct (modeled after that of STTS22H's driver), even if it's empty, for consistency with the other sensor drivers and for codegen.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+            ReadingChannel::new(Label::Temperature, -1, MeasurementUnit::Celsius),
+        ])
+    }
+
+    fn label(&self) -> Option<&'static str> {
+        self.label
+    }
+
+    fn display_name(&self) -> Option<&'static str> {
+        Some("humidity/temperature sensor")
```
ROMemories: I would prefer we keep slashes for alternatives; relative humidity sensors do both at the same time by construction:
```suggestion
        Some("humidity & temperature sensor")
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+        let t_accuracy = crate::t_accuracy(temp);
+        let sample_temp = Sample::new(temp, t_accuracy);
+
+        let samples = Samples::from_2(self, [sample_humi, sample_temp]);
+
+        Ok(samples)
+    }
+}
+
+impl<I2C: I2c + Send> Sensor for Aht20<I2C> {
```
ROMemories: If the `I2c` is not needed, it's more idiomatic not to have it:
```suggestion
impl<I2C: Send> Sensor for Aht20<I2C> {
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: As we've discussed, I really don't think this should be exposed, even more so as the return type is so weekly typed and returns a raw register value.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+            Timer::after_millis(20).await;
+
+            // soft reset the sensor
+            if Self::reset(self, &mut i2c_device).await.is_err() {
+                return;
+            }
+            // The Datasheet says that the sensor takes at most 20ms to enter idle state
+            // after initial power-on or software reset
+            #[cfg(not(test))]
+            Timer::after_millis(20).await;
```
ROMemories: This is only needed because of the reset, right? If so that should be moved inside the `reset()` method.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+            state: AtomicState::new(State::Uninitialized),
+            label,
+            i2c: OnceLock::new(),
+            signaling: Signal::new(),
+            reading: ReadingSignal::new(),
+        }
+    }
+
+    /// Initializes the driver.
+    pub async fn init(&'static self, _peripherals: Peripherals, mut i2c_device: I2C) {
```
ROMemories: This should take the `Config` struct, even if it's empty.

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use ariel_os_sensors::sensor::SampleMetadata;
+
+use crc::{CRC_8_NRSC_5, Crc};
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+/// Commands that can be sent to the sensor. Described in Table 9 and in section 5.4.
+enum Command {
+    /// Initializes the sensor
+    /// In particular this set the calibration bit.
```
ROMemories:
```suggestion
    /// Initializes the sensor.
    /// In particular this sets the calibration bit.
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+// STATUS register bits, see Table 10 of the datasheet
+/// Set to 1 if the sensor is calibrated.
+const CALIBRATION_STATUS: u8 = 1 << 3;
+
+/// Set to 1 if the sensor is busy.
+const BUSY_BIT: u8 = 1 << 7;
+
+/// Magic argument to be sent with the [`Command::TriggerMeasurement`].
+/// Described in Section 5.3.
```
ROMemories:
```suggestion
/// Described in Section 5.4.
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+const MEASUREMENT_ARG_0: u8 = 0x33;
+/// Magic argument to be sent with the [`Command::TriggerMeasurement`].
+/// Described in Section 5.3.
```
ROMemories:
```suggestion
/// Described in Section 5.4.
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Magic argument to be sent with the [`Command::Initialize`].
+/// Described in Section 5.4 of v1.0 of the datasheet.
+const INITIALIZE_ARG_1: u8 = 0x00;
+
+const PART_NUMBER: &str = "AHT20";
+
+fn t_accuracy(temp: i32) -> SampleMetadata {
+    // See Table 3 and Figure 3 of the datasheet.
+    // Accuracy of 0.3 °C between around 20 °C and 60 °C.
+    if 200 < temp && temp < 600 {
```
ROMemories: Would otherwise panic if `temp == 200`:
```suggestion
    if 200 <= temp && temp <= 600 {
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Accuracy of 1.5 °C between -40 °C and 20 °C.
+    else if -400 < temp && temp < 200 {
+        SampleMetadata::SymmetricalError {
+            deviation: 15,
+            bias: 0,
+            scaling: -1,
+        }
+    }
+    // Accuracy of 2 °C between 60 °C and 85 °C.
+    else if 200 < temp && temp < 850 {
```
ROMemories: Given the `if` cascade this should be equivalent but this is less surprising wrt the comment:
```suggestion
    else if 600 < temp && temp <= 850 {
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Accuracy of 0.3 °C between around 20 °C and 60 °C.
+    if 200 < temp && temp < 600 {
+        SampleMetadata::SymmetricalError {
+            deviation: 3,
+            bias: 0,
+            scaling: -1,
+        }
+    }
+    // Accuracy of 1.5 °C between -40 °C and 20 °C.
+    else if -400 < temp && temp < 200 {
```
ROMemories: *Technically* the maximum values are still part of the range:

```suggestion
    else if -400 <= temp && temp < 200 {
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+    // Accuracy of 5 %RH otherwise.
+    else {
+        SampleMetadata::SymmetricalError {
+            deviation: 5,
+            bias: 0,
+            scaling: 0,
+        }
+    }
+}
```
ROMemories, line 111: The above accuracy calculations mix the typical and max values depending on the range, but I suppose this is fine enough.

ROMemories (commented):
Still a partial review, now moving to the core of the implementation.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+                return;
+            }
+            // The Datasheet says that the sensor takes at most 20ms to enter idle state
+            // after initial power-on or software reset
+            #[cfg(not(test))]
+            Timer::after_millis(20).await;
+
+            // Check for calibration status
+            loop {
+                // The read the status byte to check
```
ROMemories:
```suggestion
                // Read the status byte to check
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+        let mut buf = [0u8];
+        loop {
+            match i2c.read(I2C_ADDRESS, &mut buf).await {
+                // Busy
+                Err(e) if e.kind() == I2CErrorKind::NoAcknowledge(Data) => {}
+                Err(_) => {
+                    return Err(ReadingError::SensorAccess);
+                }
+                Ok(()) => {
+                    // Check if the busy bit of the status register is set to 1
```
ROMemories: "Set" already implies "to 1" (otherwise the bit is "cleared"):
```suggestion
                    // Check if the busy bit of the status register is still set.
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+        let mut buf = [0u8; 7];
+        i2c.read(I2C_ADDRESS, &mut buf)
+            .await
+            .map_err(|_| ReadingError::SensorAccess)?;
+
+        if !crate::check_crc(&buf[0..6], buf[6]) {
+            // FIXME: Use a better error type
+            return Err(ReadingError::SensorAccess);
+        }
+
```
ROMemories, line 212: Needs to be properly indented; this is essential to mention as it's not clear from the datasheet:
```suggestion

// The relative humidity and temperature are each encoded on 20 bits, over 5 bytes.
```

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+        let humidity =
+            (u32::from(buf[1]) << 12) + (u32::from(buf[2]) << 4) + u32::from(buf[3]) / 16;
```
ROMemories: This should be rewritten using `i32::from_be_bytes()` and a bit mask+a bit shift for the third byte: the intent would be clearer, avoid the cast to `i32` later, and bitshifts are cheaper than divisions.

Edit: `i32::from_be_bytes()`, not `i32::from_bytes()`

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: Yes, that should be a source code comment (added a suggestion above).

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+        let temperature =
```
ROMemories: Same, as for `humidity`.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
```diff
@@ -0,0 +1,617 @@
...
+                            // Status byte
+                            rbuf[0] = 0b0001_1000;
+                            // Humidity bits 20..12
+                            rbuf[1] = ((samples.0 % (1 << 20)) >> 12) as u8;
+                            // Humidity bits 12..4
+                            rbuf[2] = ((samples.0 % (1 << 12)) >> 4) as u8;
+                            // Humidity bits 4.. in the upper crumb of the byte
+                            rbuf[3] = ((samples.0 % (1 << 4)) as u8) << 4;
+
+                            // Temperature bits 20..16 in the lower crumb of the byte
```
ROMemories: Sounds like a nibble:
```suggestion
                            // Temperature bits 20..16 in the lower nibble
```

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Changed and made clearer in 899456ead4d6

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Same as above 899456ead4d6

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Done in 0f93eb622b2

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Done in 0f93eb622b2

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Done in 0f93eb622b2

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in 1ca348f6507cb4d45

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in 1ca348f6507cb4d45

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in 2235e14541c263fe3d

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: I agree on your point about the weak typing of the result, I still disagree about whether this should be exposed. I think that we should bring it up in the next weekly that we both participate in (so May 12th) to see what the other maintainers think

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in 77cc3bbec9345152

ROMemories (commented):
I've only *very* quickly skimmed over the tests. This should be good to go when the opened comment threads are addressed (and I'm still to try it in hardware).

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in 244a17e5

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: Done in ef76f3df

examples/sensors-debug/src/pins.rs
anlavandier: Resolved during the rebasing process thanks to #2024

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Fixed in d1c9d08f

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Fixed in d1c9d08f

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in d1c9d08f

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Done in d1c9d08f

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Fixed in dbd0b0f7

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Fixed in https://github.com/ariel-os/ariel-os/commit/d1c9d08fb5446680dab54003c983973d67a83ff2

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: I meant `i32::from_be_bytes()`, it should be much simpler with it as fewer bitshifts would be needed.

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: Again I *really* don't think this should be exposed: this status is a low-level property of the sensor *device*, not of the sensor driver, and making the driver expose it would make the abstraction very leaky. Additionally, being low-level, this concept is not defined as part of the abstraction, and users of the sensor driver have no way of acting on the result (nor they should). This could also lead users to think they should normally be accessing driver instances directly (as that's the only way of reaching that method), while really going through the registry should be the normal way.

If this is for debugging the sensor driver itself, the status should instead be printed with `debug!()` or `trace!()` log statements, but helpers for debugging the driver itself should not be part of the API.

examples/sensors-debug/src/pins.rs
```diff
@@ -111,3 +111,11 @@ ariel_os::hal::define_peripherals!(Peripherals {
     i2c_sda: PB9,
     i2c_scl: PB8,
 });
+
+#[cfg(any(context = "unihiker-k10"))]
```
ROMemories, line 115: Following #2024, this board needs to be added to the list of contexts allowed by the example.

anlavandier: Done in [4a797a0](https://github.com/ariel-os/ariel-os/pull/1941/commits/4a797a00775094f105b4c3e15dfa969935af24ee)

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Is something like what I did in [a8bfab7](https://github.com/ariel-os/ariel-os/pull/1941/commits/a8bfab772404fa207c7c27e2143431be18c4726d) what you were thinking about ?

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: See [a8bfab7](https://github.com/ariel-os/ariel-os/pull/1941/commits/a8bfab772404fa207c7c27e2143431be18c4726d)

src/sensors/ariel-os-sensor-aht20/src/lib.rs
anlavandier: I'm happy to change in either direction if you think that it's better. For the record my rationale was that in "normal" operating ranges users are probably care more about the typical performances while in already "degraded" operating conditions its best to be humble and assume that worst case scenario

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: It's not meant only  to debug the driver but also the sensor itself. In any case I agree that this can be achieved with `debug!` log statements. 

> This could also lead users to think they should normally be accessing driver instances directly (as that's the only way of reaching that method), while really going through the registry should be the normal way.

I understand your point, however I think that this will need to happen and maybe sooner rather than later. For example, I believe that it is currently impossible to change the configuration through the registry which could come up soon-ish if only for power management reasons (to be clear, I myself don't have a concrete use case for this)

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
ROMemories: Exactly, that is *so much* cleaner (and likely quite reduces the instruction count)!

src/sensors/ariel-os-sensor-aht20/src/i2c.rs
anlavandier: Added log statements in 94a5c92a

Cargo.lock
anlavandier: Seems like this disappeared during one of the rebases

anlavandier:
Not sure that I understand what the issue is with the build CI

ROMemories:
> Not sure that I understand what the issue is with the build CI

The lockfile is not up to date.

ROMemories (approved):
LGTM, tested on a Unihiker K10: values seem reasonable and vary in the expected direction when the immediate physical environment is modified.

Please vet and squash!

ROMemories (approved):
I've switched the build label to `ci-build:esp` because the unhiker-k10 isn't in `ci-build:small`.

PR #1950, opened 2026-03-25
docs(http-client): make the app not enable `time` as that is unnecessary

ROMemories:
# Description

This makes the `http-client` example not enable the `time` Cargo feature as it is simply not needed: neither `ariel_os::time` `ariel_os::reexports::embassy_time` are used, and `embassy-time` is not separately imported by the example either, so it doesn't directly need an `embassy-time` time driver. This would have made it easier to find #1949.

## Testing

Currently this makes the `http-client` panic at startup on ESP due to #1949.

## Issues/PRs References

- Depends on #1949

## Changelog Entry

The `http-client` example no longer needlessly enables the `time` Cargo feature.

## Change Checklist

### DIFF
--- a/examples/http-client/Cargo.toml
+++ b/examples/http-client/Cargo.toml
@@ -10,7 +10,6 @@ ariel-os = { path = "../../src/ariel-os", features = [
   "dns",
   "mdns",
   "tcp",
-  "time",
 ] }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 


ROMemories:
@kaspar030 You can observe that the example currently panics at boot on ESP. It should work after rebasing this after #1968 lands.

ROMemories:
Adding this to the changelog in case it got needlessly cargo culted in users' applications.

kaspar030 (approved):

Issue #1806, opened 2026-02-13
Documenting support for already-supported boards


ROMemories:
There are a few boards that we do support but for which we have no support documentation currently. We should either add such documentation for each of them, or drop the board entirely.

These boards currently are:

- [ ] `ai-c3` (it's unclear which exact board this is)
- [ ] `dwm1001`
- [ ] `espressif-esp32-devkitc`
- [ ] `nrf52840-mdk`
- [ ] `particle-xenon`
- [ ] `st-nucleo-f767zi`
- [ ] `st-nucleo-h753zi`

PR #1959, opened 2026-03-26
feat(boards): add Waveshare ESP32 S3 board

chrysn:
# Description

This adds a rather trivial board described [at the manufacturer's page](https://www.waveshare.com/wiki/ESP32-S3-Matrix). (Gist: 25x25mm board, USB-C connector, 8x8 WS281x LEDs at the front, accelerometer at the back, several solderable GPIOs).

## Testing

I've run hello-world on it, the http-client, and a more elaborate task driving the LEDs in a to-be-pushed branch.

## Open questions

It's really [an FH4R2](https://documentation.espressif.com/esp32-s3_datasheet_en.pdf#page=13) and not a wroom. Where does this need to be changed? Just the laze and sbd files? Or is this ongoing anyway and I get away with lying here?

## Future work

I hope that at some point soon, we'll start advertising in SBDs that there is a smart-pixel string at GPIO14, and so that #1947 can pick it up. But for the time being, that's left to applications / examples.

## Changelog Entry

The Waveshare ESP32-S3-Matrix board is now supported.

## Change Checklist

  <del>Initially, this should go into https://github.com/ariel-os/ariel-os/issues/1806 to spool there.</del>


boards/waveshare-esp32-s3-matrix.yaml
```diff
@@ -0,0 +1,8 @@
+version: 0.3.0
+targets:
+  waveshare-esp32-s3-matrix:
+    # website: https://www.waveshare.com/wiki/ESP32-S3-Matrix
+    chip: esp32-s3-wroom-1 # FIXME: Actually ESP32-S3FH4R2
```
ROMemories: Similarly to the `esp32s3fx8` laze context we have, a `esp32s3fx4r2` context is needed for this:
```suggestion
    chip: esp32s3fx4r2
```

chrysn: Thanks, added now, and threaded through ariel-chips.yaml and laze-project.yml.

(Almost straight fixups: I'll just need to shuffle them into the logical order when squashing.)

ROMemories (commented):
Support info for the chip and the board/builder should be added in `doc/support_matrix.yml`.

chrysn:
Why do other boards get to use #1806 for spooling docs (ideally until after maintaining the support matrix has become less tedious) and this one not?

Now added.

I was a bit unsure whether we need to document any workarounds for flashing, but my personal trouble mainly stem from having an Ulanzi TC-001 connected at the same time, which is kind of a niche situation, and now tracked upstream as https://github.com/esp-rs/espflash/issues/1022, with a path towards doing better at https://github.com/esp-rs/espflash/issues/1023.

Can I squash? (Because that's a nontrivial squash in this case).

doc/support_matrix.yml
```diff
@@ -1169,6 +1190,11 @@ builders:
     tier: "2"
     support:
 
+  waveshare-esp32-s3-matrix:
```
ROMemories, line 1283: This needs a `boards` entry as well, you can see the board not being rendered at all in the docs.

doc/support_matrix.yml
```diff
@@ -491,6 +491,27 @@ chips:
...
+      uart: needs_testing
+      logging: supported
+      storage:
+        status: not_currently_supported
+        comments:
+          - requires partitioning support
+      user_usb: supported
+      ethernet_over_usb: supported
+      wifi: supported
+      ble: needs_testing
```
ROMemories: Could you test this by running the `ble-scanner` example?

boards/waveshare-esp32-s3-matrix.yaml
```diff
@@ -0,0 +1,16 @@
...
+    chip: esp32s3fx4r2
+    flags:
+      - has_usb_device_port
+    # To be added, eg. through ariel.global_env.ESPFLASH_ARGS and once
+    # <https://github.com/esp-rs/espflash/issues/1023> is resolved:
+    #
+    # The device presents as VID/PID 303a:1001, and it'd be great to pass that
+    # information on to espflash so it won't give us boards where we can know
+    # they are *not* this device. (Might be even general enough to put in the
+    # chip's config, not sure for lackof other ESP32S3 devices).
```
ROMemories:
```suggestion
    # chip's config, not sure for lack of other ESP32-S3 devices).
```

doc/support_matrix.yml
chrysn: Tested, set to supported.

doc/support_matrix.yml
chrysn: Thanks, now the diffs are much bigger.

ROMemories (commented):
Please rebase on top of main to take into account the newly introduced Ethernet functionality (as CI is saying).
Edit: that might not even be needed, but the `ethernet` functionality needs to be specified in the chip info.

chrysn:
I did have to add it in the chip, it is inferred from there into the board.

All squashed and CI passing now.

ROMemories (approved):
LGTM after resolving the conflicts with the now landed #1954.

chrysn:
Thanks. Rebased and resolved by rebuilding the book. (Fortunately, the conflict was just in an automatically generated file anyway).

PR #1968, opened 2026-03-30
fix(esp): make timers compatible with internally-enabled threading

ROMemories:
# Description

This fixes #1949: `embassy-time-queue-utils?/_generic-queue` was not enabled (in `ariel-os-esp`) by enabling `threading` because `embassy-time-queue-utils` was only enabled *inside*  `ariel-os-esp` and not from inside `ariel-os-embassy` where the `embassy-time-queue-utils` + `embassy-time-queue-utils?/_generic-queue` logic is.

## Testing

The `udp-echo` example (which does not enable `time` at application-level) now works on ESPs.

## Issues/PRs References

- Depends on #1964
- Fixes #1949

## Changelog Entry

(ESP32) It is now possible to use the built-in Wi-Fi and BLE radio without enabling the `time` Cargo feature at the same time. 

## Change Checklist

### DIFF
--- a/src/ariel-os-esp/Cargo.toml
+++ b/src/ariel-os-esp/Cargo.toml
@@ -120,11 +120,11 @@ threading = [
   "embassy-time-queue-utils?/_generic-queue",
 ]
 
-## Enables timers
+## Enables a time driver.
 time = [
-  "dep:embassy-time",
   "dep:embassy-time-driver",
   "dep:embassy-time-queue-utils",
+  # `esp-sync` is used for the time driver.
   "dep:esp-sync",
 ]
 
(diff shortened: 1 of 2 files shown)


src/ariel-os-esp/Cargo.toml
```diff
@@ -120,9 +120,8 @@ threading = [
   "embassy-time-queue-utils?/_generic-queue",
 ]
 
-## Enables timers
+## Enables a time driver.
 time = [
-  "dep:embassy-time",
```
ROMemories, line 118: The `time` Cargo feature should not enable the `embassy-time` crate used only internally (i.e., not re-exported), as is done in other HALs. Instead, individual features that need it should enable it, allowing for better tracking of the dependencies.

kaspar030 (approved):
trusting your testing.

PR #1969, opened 2026-03-30
refactor(embassy-common): enable `embassy-time` only when needed

ROMemories:
# Description

`embassy-time` currently is a non-optional dependency of `ariel-os-embassy-common`, which makes it always part of the build, even though it is only necessary (in that crate) for I2C. In applications that do not enable the `time` Cargo feature, and that do not otherwise transitively rely on `embassy-time`, `embassy-time` was only part of the build because of `ariel-os-embassy-common`, as can be seen with the following:

```sh
laze -C examples/log build -b stm32u083c-dk tree --no-dedupe --invert embassy-time
```

This PR enables `embassy-time` in `ariel-os-embassy-common` only when `i2c` is enabled, as that is the only thing that requires it in that crate.  

## Testing

Grepping in `ariel-os-embassy-common` can be useful to check that `i2c` is only the only thing that needs `embassy-time`.

Successfully tested the `log` and `sensors-debug` (which uses I2C) on the stm32u083c-dk.

## Changelog Entry

`embassy-time` is no longer *always* part of the build of Ariel OS. It was previously compiled even when it was not actually necessary.

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy-common/src/lib.rs
+++ b/src/ariel-os-embassy-common/src/lib.rs
@@ -31,6 +31,7 @@ pub mod reexports {
     //! Crate re-exports.
 
     // Used by macros provided by this crate.
+    #[cfg(feature = "i2c")]
     pub use embassy_time;
     pub use embedded_hal_async;
 }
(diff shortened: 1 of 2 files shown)


src/ariel-os-embassy-common/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub mod reexports {
     //! Crate re-exports.
 
     // Used by macros provided by this crate.
+    #[cfg(feature = "i2c")]
     pub use embassy_time;
     pub use embedded_hal_async;
```
ROMemories, line 36: The same can be done for `embedded-hal-async` (which I plan to do in a follow-up), but is not *as* useful (because it only provides traits, and because many other crates still depend on it anyway).

kaspar030 (approved):

Issue #1805, opened 2026-02-13
Supporting debug logging over USB CDC-ACM


ROMemories:
Currently debug logging is only supported over SWD, JTAG, and UART. This works very well when these debug ports can be accessed, but this is often not the case on "gadgets", i.e., off-the-shelf products. Some of these however provide a USB port which may be used for flashing (which we don't currently always support). We should add the ability to also use the USB port for debug logging.

- [ ] [`embassy-usb-logger`](https://docs.rs/embassy-usb-logger/latest/embassy_usb_logger/) seems to provide such functionally; we should integrate it as an alternative *logger* for the `log` logging facade.

PR #1983, opened 2026-04-01
feat(debug-log): Add custom log handler

luukoudshoorn:
# Description
Add a laze module that allows the user to provide their own backend for log output. 

## Testing
I created example applications for native and esp32 that simply output to std::print and esp_println::print. No actual unit or integration tests have been added. Guidance on a better way of testing this is appreciated.

## Open Questions
I weighed defining a trait and injecting a dyn reference instead of a function pointer but decided the function pointer would be better because it's less overhead and easy to implement. Feel free to challenge me on this.

The most difficult part of the implementation (taking up most of my time) was figuring out how to enable and disable the correct laze modules for the corresponding builds. I'm not actually sure I'm done here:
- For native builds the std output gets selected automatically, I've gated that in ariel-os-debug lib.rs. 
- For ESP32 builds esp-println got selected automatically for the target. I moved that to a separate module that gets selected by default and is disabled when custom-log-handler is selected
- For probe-rs builds I'm not actually sure what will happen. Guidance on how to test this is appreciated, big chance that breaks now when the custom-log-handler is selected. If I understand correctly you select either RTT or UART as a backend for that normally? This should be a third acceptable backend for such a build.

I've added documentation in the book about the different backends including an example for the custom-log-handler. Please check the documentation about the other backends as I'm not 100% sure about what I wrote there.

## Changelog Entry

A custom log backend can now be selected

## Change Checklist

### DIFF
--- a/book/src/debug-console.md
+++ b/book/src/debug-console.md
@@ -8,6 +8,65 @@ The [`ariel_os::debug::println!()`][println-macro-rustdoc] macro is used to prin
 When the debug console is enabled, panic messages are automatically printed to it.
 If this is unwanted, the `panic-printing` [laze module][laze-modules-book] can be disabled.
 
+## Debug Output Backends
+
+The debug console backend determines where `ariel_os::debug::println!()`
+output, panic messages, and `log` output are sent.
+This is a separate choice from selecting the logging facade (`defmt` or `log`).
+
+In typical Ariel OS builds, the backend is usually selected by laze as part of
+the target and runner setup, and can be overridden through laze modules when
+needed.
+Only one backend can be selected at a time.
+
+| Backend                           | What it does                                                                                                                                     | How to select it                                                                                                                                                 |
+|-----------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| RTT (`defmt-rtt` / `rtt-target`)  | Sends output over RTT. Use `defmt-rtt` with the `defmt` facade, and `rtt-target` with the `log` facade for human-readable debug console output.  | `defmt-rtt` is selected automatically when using `defmt` with `probe-rs`. `rtt-target` can be selected explicitly when you want RTT output for the `log` facade. |
+| `debug-uart`                      | Sends human-readable output over the board's configured debug UART. This is a good choice when you want logs on a serial console.                | Select the `debug-uart` laze module (only for the `log` facade).                                                                                                 |
+| `esp-println`                     | Uses `esp-println` on ESP targets. This integrates well with `espflash monitor`; with `defmt`, it also supports `defmt-espflash`.                | Selected automatically for ESP targets.                                                                                                                          |
+| `custom-log-handler`              | Calls an application-provided function for every `println!()` call and every `log` record. Until a handler is installed, it is a no-op.          | Select the `custom-log-handler` laze module.                                                                                                                     |
+| `std`                             | Sends output to the host standard output stream using `std::println!()`. This is the backend used by native builds.                              | Selected automatically for native builds.                                                                                                                        |
+
+### Using `custom-log-handler`
+
+`custom-log-handler` is intended for `println!()` output and for the `log`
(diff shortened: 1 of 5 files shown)


ROMemories:
Thanks a lot for opening this! I'm also actively working on facilitating this kind of things (and adding documentation) as part of #1925, #1008, and #1805.

As I think you have found, there is currently a missing distinction between "debug output transports" (currently only RTT, but semihosting could be used too, even if it's much slower) and "debug logging transports" (currently only the debug output transport itself, but #1008 adds UART, and #1805 will do USB CDC-ACM). So what you are proposing to add is a "debug logging custom transport handler" ("debug logging" to distinguish from "user logging").

As part of the mentioned issues, I'm in the process of revisiting the set of laze modules/Cargo features that handle this, in the light of this new distinction. Ideally this would also support using multiple "debug logging transports" simultaneously, but I don't know if that will be part of the first iterator of that refactor.  

Regarding this PR, I'll think we'll be able to land the custom handler (possibly after renaming, so it fits into the above terminology), but please don't spend much time on the documentation, as I'll tackle it as part of the refactor mentioned above.

---

As an aside, having a single concept of "backend" (instead of distinguishing between "debug output transports" and "debug logging transports") currently makes the origin/definition of `println!()` unclear, as `println!()` can technically be implemented both as debug logging functionality *and/or* as debug output (directly over RTT or semihosting). I'm working on resolving this: clearly the `ariel_os::debug::log::println!()` should only use debug logging functionality, but it's not as clear for `ariel_os::debug::println!()`, especially depending on whether debug logging is enabled or not.

Edit: as an additional constraint, it should also be possible to use the debug logging transports when debugging functionality is disabled in hardware (e.g., in production).

luukoudshoorn:
Thanks for the explanation! 

To clarify: Is your intention that we add this first and reshape later, or reshape first and add this afterwards?

Either way works for me, I can use this functionality already by keeping it in the not-upstreamed part of our fork, so let me know what works for you.

ROMemories:
> reshape first and add this afterwards

That's the plan currently, #1987 acts as a (WIP) design documentation of what it should look like. When the design is approved, we'll first need to align the implementation with it, and that should make it more straightforward to add such a custom handler, including in a more generic manner (e.g., I don't think the custom handler has to be ESP-specific).

ROMemories:
The relevant parts of the refactor (#2007, #2008, #2013, #2026, #2030, #2036) are now done; could you rebase this PR and make the necessary adjustments stemming from the refactor (which will likely result in a slightly smaller PR)?

Here are the changes that should be relevant:

- What was called "debug logging" is now just called "logging," so what this PR is adding is a `custom-logging-handler`. I'd rather we use "logging" instead of "log" here as we are trying to keep `log` specifically referring to the `log` logging facade. Similarly, I think the `LOG_HANDLER`/`LogHandler`/`log_handler` should be called a "logging handler". (The `debug-uart` functionality that you may have taken inspiration from has not been renamed *yet*, but it is completely internal.)
- The `ariel-os-debug-log` crate is no more; these changes should be moved to the `ariel-os-log` crate. They shouldn't affect `ariel-os-debug` in any way as that crate is now only concerned with the debug channel and semihosting-like functionality (e.g., `exit()`), which this PR is not touching.
- There is no concept of "debug backend" anymore. What this PR adds is a new "logging transport". This should make it easier to not make the new logging transport ESP-only.
- You'll find more documentation about the logging transports in the new Logging book page, which is also the section that this PR should now update to document the handler.
- What has been transiently called the "debug output" is now called the "debug channel", documentation about it is part of #1987.

ROMemories:
Thanks to the work done by @nponsard, we now support setting up a custom logging transport as a first-class logging transport from applications. The documentation for this would be in the [Logging book page](https://ariel-os.github.io/ariel-os/dev/docs/book/logging.html#logging-transports), and we also have [an example](https://github.com/ariel-os/ariel-os/tree/dbfaebd5db4f617073180a65dfcbe686e3d6a645/examples/usb-serial-logger). If you end up using this, do tell us how that works for you!

PR #1985, opened 2026-04-02
fix: don't use semihosting with espflash

### LINKED ISSUE #1756: (ESP32-S3, ESP32) "Unhandled interrupt on ProCpu"
When testing out the [v0.3.0-rc.1](https://github.com/ariel-os/ariel-os/releases/tag/v0.3.0-rc.1), I've found the following panic again, when running the `http-example` on espressif-esp32-s3-devkitc-1, after the HTML document is successfully fetched:

```
</html> (http_client http-client/src/main.rs:88)
[ERROR] panicked at 'Unhandled interrupt on ProCpu' (esp_hal src/interrupt/mod.rs:90)
```

The panic does *not* happen when removing the `exit()` call at the end of the example, as suggested before by @anlavandier.

nponsard:
# Description

I searched the espflash codebase and found no mention of semihosting. I got confirmation in the esp-rs matrix chat espflash wasn't meant to be a debugger and handle semihosting calls.

## Testing

Shouldn't panic :

```
laze -C examples/hello-world build -b espressif-esp32-s3-devkitc-1 run 
```

Should still print "Firmware exited successfully": 

```
laze -C examples/hello-world build -b espressif-esp32-s3-devkitc-1 -s probe-rs run
```

## Issues/PRs References

- Depends on #1988
- Fixes #1756 

## Open Questions

There's no specific `espflash` laze module where `probe-rs` has one, should we make one ?

## Changelog Entry

Semihosting is now only enabled by default when probe-rs is selected as host tool during compilation. This fixes the "Unhandled interrupt" panic on Xtensa when calling `ariel_os::debug::exit()` when using `espflash`. Semihosting can be manually enabled using the `semihosting` laze module if needed.

## Change Checklist

### DIFF
--- a/book/src/debug-console.md
+++ b/book/src/debug-console.md
@@ -21,6 +21,14 @@ The laze configuration automatically enables semihosting on the target when the 
 When the flashing tool does not, support for semihosting can still be enabled in the firmware by selecting the `semihosting` [laze module][laze-modules-book].
 This is needed to later be able to attach a semihosting-enabled host tool to the target.
 
+> [!NOTE]
+> When semihosting is enabled on the target and no host tool supporting semihosting (or a debugger) is connected, calling [`ariel_os::debug::exit()`][debug-exit-fn-rustdoc] may result in a panic.
+> For example on ESP using `espflash` you would get:
+>
+> ```
+> [ERROR] panicked at 'Unhandled interrupt on ProCpu' (esp_hal src/interrupt/mod.rs:90)
+> ```
+
 ## Logging
 
 Ariel OS supports logging on all platforms and it is enabled by default with the `logging-facade` [laze module][laze-modules-book].
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -24,7 +24,6 @@ contexts:
       - ?critical-section
       - ?hw/device-identity
       - ?lto
-      - ?semihosting
 
       # Selecting default toolchain here. try xtensa, stable, nightly.
       - ?xtensa
@@ -1160,6 +1159,7 @@ modules:
     help: use probe-rs as runner
     selects:
       - ?debug-console
+      - ?semihosting
       - defmt:
           - defmt-rtt
       - log:
@@ -1202,6 +1202,8 @@ modules:
           - probe-rs reset --chip ${PROBE_RS_CHIP} $@
 
   - name: openocd
+    selects:
+      - ?semihosting
     tasks:
       flash:
         cmd:


ROMemories (changes requested):
I don't think we should do this now: even if `espflash` is used for flashing, this doesn't preclude later using probe-rs (or another semihosting-enabled tool) to use semihosting. Also semihosting may be required for running `embedded-tests` tests.

I am however actively working on documenting all this, at which point we may indeed want to do that, but we need to first expose and document how to enable semihosting for users that would need it (which I am working on).

nponsard:
I see, should I try instead to add an handler so it's not marked as unhandled when using `espflash` and doesn't look like an error ?

kaspar030:
On ARM there is a way to figure out if a debugger is connected. Maybe esp's can do the same, then we could make the semihosting calls depending on that.

nponsard:
I changed it to be disabled by default and enabled when a compatible debugger is used for flashing so it follows the book.

I also added a note in the book about the error when enabling `semihosting` with `espflash`, but this can be moved in another PR if you want.

ROMemories:
I've switched to `ci-build:small` as moving the `?semihosting` impacts everything.

book/src/debug-console.md
```diff
@@ -21,6 +21,14 @@ The laze configuration automatically enables semihosting on the target when the
 When the flashing tool does not, support for semihosting can still be enabled in the firmware by selecting the `semihosting` [laze module][laze-modules-book].
 This is needed to later be able to attach a semihosting-enabled host tool to the target.
 
+> [!NOTE]
+> When semihosting is enabled on the target and a monitoring tool that doesn't support semihosting is used, calling [`ariel_os::debug::exit()` function][debug-exit-fn-rustdoc] may trigger an exception.
```
ROMemories: #1987 uses the more generic "host tool" instead of "monitoring tool" for now. But is it actually caused by a host tool being used or not? Could it be that it always happens but we just can't see it?

Alternatively, instead of removing "function", "the" could be added before exit():


```suggestion
> When semihosting is enabled on the target and a host tool that doesn't support semihosting is used, calling [`ariel_os::debug::exit()`][debug-exit-fn-rustdoc] may trigger an exception.
```

ROMemories (commented):
Successfully tested the following to check that semihosting is still enabled otherwise:

```sh
laze -C examples/hello-world build -b stm32u083c-dk run
```
(The `hello-world` example `exit`s.)

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/http-client/ build -b espressif-esp32-s3-devkitc-1 run
```

which doesn't exit (as expected) and doesn't print the panic.

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/http-client/ build -s probe-rs -b espressif-esp32-s3-devkitc-1 run
```

which does exit as expected (documenting the `probe-rs` laze module is part of #1987).

I've also tested enabling `semihosting` when using `espflash`, which panics as before as expected.

book/src/debug-console.md
nponsard: The interrupt is probably fired anyway but we don't see it, this note is mainly to avoid getting bug reports about it.

ROMemories: > The interrupt is probably fired anyway but we don't see it

It's not just about the interrupt, it's about the subsequent panic stemming from it. I still have qualms about that phrasing because that would somewhat imply that it does *not* panic when no host tool is connected while we think it does, so that would be a bug.

nponsard: I changed the phrasing to be "when no compatible debugger is connected"

book/src/debug-console.md
```diff
@@ -21,6 +21,14 @@ The laze configuration automatically enables semihosting on the target when the
 When the flashing tool does not, support for semihosting can still be enabled in the firmware by selecting the `semihosting` [laze module][laze-modules-book].
 This is needed to later be able to attach a semihosting-enabled host tool to the target.
 
+> [!NOTE]
+> When semihosting is enabled on the target and no compatible debugger is connected, calling [`ariel_os::debug::exit()`][debug-exit-fn-rustdoc] may trigger an exception.
+> For example on ESP using espflash you would get:
```
ROMemories: I think I've found the issue I actually had: my understanding is that the exception is always generated, that's how semihosting works (from the [semihosting specification](https://developer.arm.com/documentation/dui0471/m/what-is-semihosting-/what-is-semihosting-)):

> Semihosting is implemented by a set of defined software instructions, for example SVCs, that generate exceptions from program control. The application invokes the appropriate semihosting call and the debug agent then handles the exception. The debug agent provides the required communication with the host.

What matters is whether the exception results in a panic or not, and that does depend on whether the exception is handled by a compatible host tool:
 
```suggestion
> When semihosting is enabled on the target and no host tool supporting semihosting (or a debugger) is connected, calling [`ariel_os::debug::exit()`][debug-exit-fn-rustdoc] may result in a panic.
> For example on ESP using `espflash` you would get:
```

book/src/debug-console.md
ROMemories: Continuing the discussion in another thread, because the original comment is now outdated and I can't make suggestions.

book/src/debug-console.md
nponsard: That makes sense, I was also tempted to replace the "may" into a "will".

ROMemories: > I was also tempted to replace the "may" into a "will".

I'm not sure at all, this doesn't seem to be the case on other MCU families (when using logging-over-UART), or even on ESP on some examples, for instance on the `hello-world` example (which I don't actually understand).

ROMemories (commented):
I should that should be good to go after rebasing now.

ROMemories (approved):

PR #1986, opened 2026-04-02
refactor(flash-dfu)!: rename the `flash-dfu` laze task to `flash-dfuse`

ROMemories:
# Description

This renames the `flash-dfu` laze task to `flash-dfuse` as it is only available on STM32 devices and *is* DfuSe-specific. [DfuSe is based on USB DFU, but is ST's non-standard protocol and is incompatible](https://dfu-util.sourceforge.net/dfuse.html) with [standard USB DFU](https://www.usb.org/sites/default/files/DFU_1.1.pdf). 

## Testing

Successfully tested with:

```sh
laze -C examples/blinky/ build -b st-steval-mkboxpro flash-dfuse
```

There is no remaining occurrences of `flash-dfu\b` in the codebase.

## Changelog Entry

The `flash-dfu` laze task has been renamed to `flash-dfuse` for clarity, as it is only available on DfuSe devices, i.e., STM32 devices.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #1987, opened 2026-04-02
docs(book): add a page about flashing and debugging

ROMemories:
# Description

This adds a page about flashing and the debug channel to the book (actual debugging can be tackled later).

## Future work

- Leveraging #1833, we could add a note to the relevant board pages to explain how to use the available USB port(s) (especially for boards with two of them, like the Espressif devkits).

## Issues/PRs References

- Depends on #1986
- Depends on #1988
- Depends on #2013
- Depends on #2034

## Changelog Entry

Documentation about how flashing works has been added to the book.

## Change Checklist


book/src/flashing-debugging.md
```diff
@@ -0,0 +1,210 @@
...
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print debug logs.
+Currently, the firmware automatically switches at runtime between using USB CDC-ACM or UART as debug logging transport.
+
+> [!WARNING]
+> This is likely to change in the future, and it may become necessary to select a specific laze module to choose which debug logging transport to enable and use.
+
+## Additional Host-Related Functionality
+
+On top of providing a debug output, [semihosting](#semihosting) also allows the implementation of other I/O and host-related functionality.
+In particular, [`ariel_os::debug::exit()`][debug-console-exit-book] is currently implemented through semihosting on embedded platforms.
```
ROMemories, line 124: Needs some back-linking after #1988 is in.

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,210 @@
...
+| ----------------- | ---------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `run`             | ESP32 devices                      | Compiles, flashes, and runs an application. [Debug logs](#debug-logging-transports) (not the debug output) are printed in the terminal.                                                                                                          |
+| `flash-dfuse`     | DfuSe devices, i.e., STM32 devices | Compiles and flashes an application via DfuSe, the non-standard ST protocol based on USB DFU, before rebooting the target. Requires bootloader support for DfuSe in the board or microcontroller, and [dfu-util][dfu-util-homepage] on the host. |
+
+<!-- TODO: consider introducing an `espflash` laze module -->
+
+## Debug Output Transports
+
+<!-- TODO: rename support functionality "Logging" to "Debug Logging" -->
+<!-- TODO: mark "Debug Output" as not available on native -->
```
ROMemories: Also on boards that don't expose a debug interface: Ulanzi TC001? Others?

book/src/build-system.md
```diff
@@ -22,15 +22,8 @@ For example, to run the hello-world example from the `ariel-os` directory, the c
 
 Tasks available in Ariel OS include:
 
-- `run`: Compiles, flashes, and runs an application. The [debug output](./debug-console.md) is printed in the terminal.
```
ROMemories, line 25: The documentation for these tasks is moved at the relevant places in the Flashing & Debugging page.

ROMemories:
@kaspar030 This could now use a first review as a design document: not all of this is currently implement/aligned with this documentation, but as soon as we agree that is what we want, I can start opening the gates on the implementation. The phrasing of the docs can be reviewed later.

Edit: I'm especially interested in the "debug output" vs "debug console" vs "debug logging" distinction.

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+SWD is a variant of JTAG with a reduced pin count but, being an Arm technology, it is only found on Arm-based microcontrollers, while JTAG is vendor-agnostic.
+Besides being a debug protocol, JTAG is actually more generic and, in particular, also enables boundary scans (automatically checking the traces of a PCB by taking direct control of the chip pins present on the board), which were originally its primary purpose.
+
+Debug protocols interact with the microcontroller through a *debug interface*, which may be a physical interface or an interface internal to the microcontroller, connected to the processor and/or the microcontroller buses.
+
+As host computers do not have support for these debug protocols, a debug probe is necessary.
+*Debug probes* are USB devices that allow using these debug protocols, either with standard (e.g., [CMSIS-DAP][cmsis-dap]) or vendor-specific USB classes.
+In some cases, debug probes can also be built into the microcontrollers themselves, behind a USB interface.
+
+Ariel OS currently uses [probe-rs][probe-rs-tool-probe-rs-docs] to interact with debug probes.
```
kaspar030, line 22: mention espflash?

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+### Real Time Transfer (RTT)
+
+[RTT][segger-rtt-wiki] output relies on in-memory buffers which are written to by the firmware on the target and read, in the background (when the microcontroller supports it), by the debug probe.
+RTT supports having multiple such buffers, allowing to implement multiple channels.
+In addition, RTT supports channels in both directions: from the target to the host ("up channels"), and from the host to the target ("down channels"), but the latter are not used for the debug output.
+RTT also requires an in-memory RTT Control Block, which stores the locations of the in-memory channel buffers.
+The RTT-enabled host tool either knows the location of the control block in memory, or scans the memory to find the magic bytes ("ID") the control block starts with.
+
+> [!TIP]
+> probe-rs automatically prints the RTT output when used in the firmware.
```
kaspar030, line 119:
```suggestion
> probe-rs automatically prints using the RTT output when used in the firmware.
```

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+- `debug-logging-over-uart`
+-->
+
+<!-- TODO: comment about the (non?) mutual exclusiveness of the debug logging transports: likely mutually exclusive at first, then simultaneous transports supported -->
+
+<!-- TODO: document where panics get printed [#1061](https://github.com/ariel-os/ariel-os/issues/1061) -->
+
+> [!IMPORTANT]
+> When using [`defmt` as debug logging facade][defmt-facade-book], a `defmt`-enabled host tool must be used so that logs are rendered correctly, as `defmt` uses its own encoding on the wire.
+> probe-rs and `espflash` both support `defmt`'s encoding transparently.
```
kaspar030:
```suggestion
> `probe-rs` and `espflash` both support `defmt`'s encoding transparently.
```

book/src/flashing-debugging.md
ROMemories: espflash doesn't do debug (edit: interface) protocols, that's the thing; however it could be mentioned explicitly in the documentation of the `run` laze task in the Flashing Through Bootloaders section.

book/src/flashing-debugging.md
ROMemories: I'm primarily interested in the review of the design for now; also the [probe-rs website](https://probe.rs/) does *not* use a monospace font for its name, while [`espflash`'s docs do (mostly)](https://github.com/esp-rs/espflash/tree/main/espflash).

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+
+Ariel OS makes it easy to flash and debug applications, by unifying the different existing flashing mechanisms and debug interface protocols and selecting the most suitable one, based on the target board and development and production requirements.
+
+> [!NOTE]
+> The terms "debug", "debugging", and "debugger" tend to refer to a wide array of overlapping concepts.
+> In the following, we define and use phrases that use narrower meanings, with the hope to make these clearer.
+
+## Debug Interface Protocols, Debug Interfaces, and Probes
+
+*Debug interface protocols* allow reading and writing from and to system memory and processor registers, setting breakpoints, and stepping through the program execution, among other things.
```
ROMemories, line 11: Following some out-of-band discussion, I've replaced "debug protocols" with "debug interface protocols", as "debug protocols" may possibly be misunderstood with other types of debug protocols.

book/src/build-system.md
bergzand: Should we keep something in this section to keep the section complete and just refer to the new section here?

bergzand: e.g. 
> - `run`, `flash`, `flash-dfuse`: See [Flashing & Debugging](./flashing-debugging) for flash- and debug-related tasks.

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+The two debug interface protocols supported by Ariel OS tooling are [JTAG][jtag-wikipedia] and [Serial Wire Debug (SWD)][swd-arm-spec]:
+SWD is a variant of JTAG with a reduced pin count but, being an Arm technology, it is only found on Arm-based microcontrollers, while JTAG is vendor-agnostic.
+Besides being a debug interface protocol, JTAG is actually more generic and, in particular, also enables boundary scans (automatically checking the traces of a PCB by taking direct control of the chip pins present on the board), which were originally its primary purpose.
```
bergzand, line 14: If I understand the terminology correctly, there's some overlap between the debug interface protocol and the debug interface itself. The debug interface itself is the physical part and the protocol is the exchange over the wires. e.g.: SWD talks SWD protocol over the wires.

If that's correct, I would remove the mention of pin count and PCB traces here to prevent confusion. Or alternatively, swap this paragraph with the paragraph below on Debug Interfaces and add the explanation on the differences there.

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+The RTT-enabled host tool either knows the location of the control block in memory, or scans the memory to find the magic bytes ("ID") the control block starts with.
+
+> [!TIP]
+> probe-rs automatically prints the RTT output when used in the firmware.
+
+<!-- TODO: document the to-be-introduced laze module that enables RTT: `debug-output-rtt` -->
+
+## Debug Logging Transports
+
+Orthogonally to the debug output, debug logging allows to output debug logs.
```
bergzand: My suggestion here would be to not name this something with debug in the name. This to remove confusion between the logging and the debug-output. Given how this book chapter is shaped, it is easy to connect debug probe to debug output (correct) and to debug logging (not correct). Personally I would stick to plain 'logging' and 'logging transports' (or 'logger' and 'logger transports'). 

Logging is not just about debugging, but also about catching errors logged via `error!()`, and those could be transported to some aggregator via $protocol. Calling it *debug* logging omits that part for me

bergzand (commented):
Here's my initial review on the terminology

book/src/flashing-debugging.md
ROMemories: > This to remove confusion between the logging and the debug-output.

I agree with the potential source of confusion, but it was my impression that *these* logs were specifically debug logs, and that we'd have a separate system/mechanism/API for user logs later (that's why the module path is `ariel_os::debug::log`). If this is the case and if we renamed this to just "logging", we'd later have "logging" and "user logging", and I fear that that would be confusing too (but maybe I misunderstood the distinction between debug logs and user logs).

bergzand: I think what I'm missing in our documentation or in this discussion is the purpose of the debug logging, and how it relates to user logging.

book/src/flashing-debugging.md
ROMemories: I'm mostly attempting to paraphrase Arm's specification here; for instance the pin count is part of the definition of the SWD protocol (see section 4.3 of [the specification](https://developer.arm.com/documentation/ihi0031/latest/)):

> The SWD protocol uses a synchronous serial interface, which comprises a single bidirectional data signal, and a clock signal.

In other words, SWD defines at least the data link layer and the network layer (if we want to map that to the OSI layer); but I'm not sure whether that addresses your concern.

bergzand: Mapping to OSI layer, I get that SWD defines both data link and network layer. My concern is that this subsection is about the network layer, but emphasizes differences between the two options on the data link layer. I think that's going to be confusing to the reader on the terminology you're defining here.

book/src/flashing-debugging.md
ROMemories: Debug logging is documented in the Debug Console book page but I agree there isn't really a definition of its scope or purpose.

To me user logs are aimed at users of the products that are built with Ariel OS: they may be shown on-device (on a display) or accessed through the network for instance, and may be stored on persistent storage, while debug logs are intended for developers of the products, or for maintenance once deployed in production.

bergzand: > Debug logging is documented in the Debug Console book page but I agree there isn't really a definition of its scope of purpose.

I checked there first :grin: 

> To me user logs are aimed at users of the products that are built with Ariel OS: they may be shown on-device (on a display) or accessed through the network for instance, and may be stored on persistent storage, while debug logs are intended for developers of the products, or for maintenance once deployed in production.

Thanks for the clarification, that makes sense. Let me think a bit on this.

book/src/flashing-debugging.md
ROMemories: > My concern is that this subsection is about the network layer

I don't think this subsection (or paragraph) is about the network layer specifically, the first paragraph is about everything related to debug interface protocols, and the second one is about debug interfaces, but they're a logical component of MCUs, not a different protocol layer. The first two paragraphs are about the MCU point of view, then the third about debug probes (which bridge between MCUs and hosts), and the fourth about host-side tools.

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,216 @@
...
+Debug interface protocols as introduced above also allow providing an additional piece of functionality: a debug output.
+Two main techniques exist to implement such debug output over debug interface protocols: [semihosting][arm-semihosting-docs], and [Real Time Transfer (RTT)][segger-rtt-wiki].
+Even though originally vendor-specific technologies, they have been extended to other architectures and vendors (e.g., [semihosting on RISC-V][riscv-semihosting-spec]), and can be used on every microcontroller currently supported by Ariel OS.
+
+<!-- TODO
+It is currently unclear whether we want to allow have two potentially different
+`println!()` macros: one for the debug output, and one the debug logging.
+This may also depend on whether debug logging is enabled.
+-->
+The *debug output* provides an implementation for a `println!()` macro.
```
ROMemories: We've agreed out-of-band to simply remove the `ariel_os::debug::println!()` macro, and only keep what currently is `ariel_os::debug::log::println!()` (but we also discussed renaming "debug logging" to "logging", so that will become `ariel_os::log::println!()`).

ROMemories:
Following the out-of-band discussion, I'll now start with renaming `ariel-os-debug-log` to `ariel-os-log`, and `ariel_os::debug::log` to just `ariel_os::log`.

ROMemories:
@kaspar030 @bergzand I've given the terminology some more thought: what do you think we call the "debug output" the "debug channel"/"debug channel output" (depending on whether we are referring to the channel itself or its result as printed by host tools) instead? That's what [RTT calls it](https://www.segger.com/products/debug-probes/j-link/technology/about-real-time-transfer/#:~:text=Uses%20debug%20channel%20for%20communication).

Edit: I've updated this PR to have a preview of what that looks like.

ROMemories:
@kaspar030 agreed in a out-of-band discussion that "debug channel" was better than "debug output", or at least not as potentially confusing. I'll rename the few occurrences of "debug output" on main and start using that term henceforth.

book/src/flashing-debugging.md
ROMemories: Done

book/src/flashing-debugging.md
ROMemories: Done on native and the Ulanzi TC001.

book/src/flashing-debugging.md
ROMemories: #2039 mentions `espflash` in the Logging page.

book/src/build-system.md
ROMemories: > Should we keep something in this section to keep the section complete and just refer to the new section here?

I don't think so: I think these are better grouped semantically where they belong as we do for laze modules, and different laze tasks can have the same name in different contexts (as for the `run` task, which is different for `espflash`).

book/src/flashing-debugging.md
ROMemories: Why? *probe-rs* prints the output fetched over RTT.

book/src/flashing-debugging.md
ROMemories: I think this was addressed by #2008.

book/templates/SUMMARY.md.tmpl
```diff
@@ -19,6 +19,7 @@
 - [Getting Started](./getting-started.md)
 - [Build System](./build-system.md)
 - [Building an Application](./application.md)
+- [Flashing & Debugging](./flashing-debugging.md)
 - [Debug Console](./debug-console.md)
 - [Logging](./logging.md)
```
ROMemories, line 23: We might actually prefer having the Logging page right before Debug Console now.

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,148 @@
+# Flashing & Debugging
+
+Ariel OS makes it easy to flash and debug applications, by unifying the different existing flashing mechanisms and debug interface protocols and selecting the most suitable one, based on the target board and development and production requirements.
+
+> [!NOTE]
+> The terms "debug", "debugging", and "debugger" tend to refer to a wide array of overlapping concepts.
```
anlavandier, line 6: This page is unfinished and does not deal with debugging or debuggers yet

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,148 @@
...
+
+| laze tasks        | Description                                                                                                                   |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| `run`             | Compiles, flashes, and runs an application. The [debug channel output](#debug-channel-transports) is printed in the terminal. |
+| `flash`           | Compiles and flashes an application, before rebooting the target.                                                             |
+| `flash-erase-all` | Erases the entire flash memory, including user data. Unlocks it if locked.                                                    |
+| `reset`           | Reboots the target.                                                                                                           |
+
+> [!TIP]
+> Debug interface protocols also allow writing the firmware to RAM (inside of flash) and rebooting from there, which could be useful during development as flash endurance is limited.
```
anlavandier:
```suggestion
> Debug interface protocols also allow writing the firmware to RAM (instead of flash) and rebooting from there, which could be useful during development as flash endurance is limited.
```

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,148 @@
...
+However, to specifically choose probe-rs as the host tool, the `probe-rs` [laze module][laze-modules-book] can be selected.
+
+> [!NOTE]
+> probe-rs is currently focused on *debug interface protocols* only.
+> It does not support other serial protocols used for [flashing through bootloaders](#flashing-through-bootloaders).
+
+<!-- NOTE: We refer to flashing a *board*, not just a microcontroller, as the flash memory may be outside the microcontroller. -->
+## Flashing a Board
+
+In general, there are two ways of writing the firmware to the flash memory (or memories) where the processor(s) execute(s) from---i.e., to flash the board: either by using a debug interface protocol, or by booting into a bootloader available on the board and then using one of its supported methods.
```
anlavandier, line 34:
```suggestion
In general, there are two ways of writing the firmware to the flash memory (or memories) where the processor(s) execute(s) from---_i.e._, to flash the board: either by using a debug interface protocol, or by booting into a bootloader available on the board and then using one of its supported methods.
```

book/src/flashing-debugging.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Flashing Trough Debug Interface Protocols
+
+As debug interface protocols allow arbitrarily writing to system memory, they allow downloading the firmware into the flash memory, at the necessary location.
+When the debug interface is available (e.g., during development), this is generally the preferred option when SWD is available.
```
anlavandier: Why only when SWD is available and not when JTAG is ?

anlavandier (approved):
LGTM after dealing with the suggested changes

book/src/flashing-debugging.md
anlavandier: I think that latin words should be emphasized but this might only be true when writing in french

ROMemories: "i.e." is common enough in English to not require italicization, see for instance The Chicago Manual of Style.

book/src/flashing-debugging.md
ROMemories: Done.

book/src/flashing-debugging.md
ROMemories: You're right, that wasn't actually the right criterion: I've replaced it with a condition on flashing speed instead. This is mostly to explain/allow why `espflash` (and therefore flashing through the bootloader) is preferred over probe-rs on ESP32.

book/src/build-system.md
anlavandier: I agree that it feels weird to remove tasks from this especially since the paragraphs above says:
> For tasks like flashing and debugging, Ariel OS uses laze tasks. laze tasks currently have the syntax laze build -b <builder> [other options] <task-name>. For example, to run the hello-world example from the ariel-os directory, the command would be:

Koen's suggestion might be the simplest way to have it make sense

book/src/flashing-debugging.md
anlavandier: Resolving to make the PR mergeable. I don't have a strong opinion either way here.

book/src/flashing-debugging.md
anlavandier: I agree with @ROMemories so I'm resolving this comment

anlavandier (changes requested):
See the first comment about `build-system.md`

book/src/build-system.md
ROMemories: I've integrated the suggestion (and completed the list of relevant tasks).

anlavandier (approved):
LGTM, please squash

book/src/flashing-debugging.md
anlavandier: Again, I agree with @ROMemories's rationale hence why I'm resolving this conversation.

PR #1989, opened 2026-04-03
docs(log): use `debug::log::println!()` in applications when relevant

ROMemories:
# Description

This uses `ariel_os::debug::log::println!()` instead of `ariel_os::debug::println!()` for the two relevant occurrences in the codebase. `ariel_os::debug::log::println!()` is not *always* the right choice (especially as the future of `ariel_os::debug::println!()` is simply unclear), but it is in these cases.

On top of being semantically the right choice, this helps reduce the size of the binary in some cases, as it enables using `defmt` (which may or may not be used with `ariel_os::debug::println!()`), and especially on ESP32.

## Testing

Comparing the binary size of the `log` example on ESP32-C6, this saves 116 bytes of `.text`. Additionally, it can be seen that the `println!()` message is now `defmt`-encoded, for instance by using a regular serial monitor on the UART output of an ESP32-C6 running the `log` example.

## Changelog Entry

I don't think this has to go into the changelog, this is just alignment with the upcoming documentation of debug logging transports. 

## Change Checklist

### DIFF
--- a/examples/log/src/main.rs
+++ b/examples/log/src/main.rs
@@ -1,7 +1,7 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::debug::{log::*, println};
+use ariel_os::debug::log::*;
 
 #[ariel_os::task(autostart)]
 async fn main() {
--- a/tests/benchmarks/bench_sched_yield/src/main.rs
+++ b/tests/benchmarks/bench_sched_yield/src/main.rs
@@ -1,7 +1,7 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::{debug::println, thread};
+use ariel_os::{debug::log::println, thread};
 
 #[ariel_os::thread(autostart)]
 fn thread0() {


examples/log/src/main.rs
```diff
@@ -1,7 +1,7 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::debug::{log::*, println};
+use ariel_os::debug::log::*;
```
ROMemories, line 4: This one just makes sense: this is the `log` example, so the `debug::log` version should be used.

tests/benchmarks/bench_sched_yield/src/main.rs
```diff
@@ -1,7 +1,7 @@
 #![no_main]
 #![no_std]
 
-use ariel_os::{debug::println, thread};
+use ariel_os::{debug::log::println, thread};
```
ROMemories, line 4: Even if the future of `ariel_os::debug::println!()` is unclear, it seems just better to use the debug logging version in this test: it may makes the binary a bit smaller, marginally more efficient at runtime and, as this is just a test, there is no reason for debug logging to be disabled.

ROMemories:
An implementation of `println!()` is missing from `ariel_os::debug::log` *when using the `log` logging facade*: we'll need to add one first.

ROMemories:
Should be superseded by #2013, will close when it (or another PR addressing this) lands.

PR #1991, opened 2026-04-07
refactor: use `ariel-os-debug-log` directly when possible

ROMemories:
# Description

This replaces every internal uses of `ariel_os_debug::log` with a direct dependency on `ariel_os_debug_log`. This makes the exact relationships between crates clearer, often allows to remove the dependency on `ariel-os-debug`, and should make it easier to potentially invert the dependency relationship between `ariel-os-debug` and `ariel-os-debug-log`.

There are no remaining occurrences of `ariel_os_debug::log` in the codebase.

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -32,6 +32,7 @@ embedded-hal-async = { workspace = true }
 ariel-os-boards = { workspace = true }
 ariel-os-buildinfo = { workspace = true }
 ariel-os-debug = { workspace = true }
+ariel-os-debug-log = { workspace = true }
 ariel-os-embassy-common = { workspace = true }
 ariel-os-hal = { path = "../ariel-os-hal" }
 ariel-os-identity = { path = "../ariel-os-identity" }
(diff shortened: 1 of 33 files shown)


src/ariel-os-embassy/Cargo.toml
```diff
@@ -32,6 +32,7 @@ embedded-hal-async = { workspace = true }
 ariel-os-boards = { workspace = true }
 ariel-os-buildinfo = { workspace = true }
 ariel-os-debug = { workspace = true }
```
ROMemories, line 34: In this crate the dependency on `ariel-os-debug` cannot be removed at the moment, as it is used to set up the UART debug logging handler.

src/ariel-os-rt/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license.workspace = true
 [dependencies]
 ariel-os-alloc = { workspace = true, optional = true }
 ariel-os-debug = { workspace = true }
```
ROMemories, line 10: The dependency cannot be removed as it is needs to be initialized there (among other things).

anlavandier (approved):
LGTM,

Confirmed that there are no occurences of `ariel_os_debug::log` left in the codebase using `grep -rI -e "ariel_os_debug::log" .`

Issue #1061, opened 2025-05-27
Dispatch panic! and assert*! to defmt/log


chrysn:
### Description

defmt provides its own set of [panic!](https://docs.rs/defmt/latest/defmt/macro.panic.html), unreachable/unimplemented/todo, the various asserts, and [unwrap!](https://docs.rs/defmt/latest/defmt/macro.unwrap.html)/expect.

When we are on a defmt backend, we should use them rather than the core versions. (Ideally, we'd even lint against using core's versions).

### Motivation

String formatting should never wind up in our binaries -- but using the core panic/assert easily pulls that in.

### Current Implementation

We string-format panics and assetions.

### Are you planning to do it yourself in a pull request?

Yes

chrysn:
Before going head-first in there, we should check whether it makes a difference on panics with actual argument formatting -- and where is the threshold on asserts?

PR #1994, opened 2026-04-08
refactor(debug-logging)!: remove the `defmt` re-export module

ROMemories:
# Description

This proposes to remove the `ariel_os::debug::log::defmt` module: it was added at a time where the items from `ariel_os::debug::log` were less portable (especially before #972 and #1485). In particular this makes the `ariel_os::debug::log` module clearer, and removes some items that were lacking documentation.

## Testing

CI does it.

```sh
laze -C examples/storage/ build -b stm32u083c-dk run
```

## Issues/PRs References

- Related to #1061
- Related to #1987
- Depends on #1999

## Changelog Entry

The `ariel_os::debug::log::defmt` module has been removed. The portable items from `ariel_os::log` should be used instead, or the `defmt` crate should be used directly instead if actually needed.

## Change Checklist

### DIFF
--- a/src/ariel-os-debug-log/src/lib.rs
+++ b/src/ariel-os-debug-log/src/lib.rs
@@ -17,23 +17,13 @@
 #[featurecomb::comb]
 mod _featurecomb {}
 
-#[cfg(feature = "defmt")]
-pub mod defmt {
-    //! Selected [`defmt`] items.
-
-    // This module is hidden in the docs, but would still be imported by a wildcard import of this
-    // crate's items.
-    #[doc(hidden)]
-    pub mod hidden {
-        // Required so the macros can access it.
-        #[doc(hidden)]
-        pub use defmt;
-    }
-
-    pub use defmt::{Debug2Format, Display2Format, Format};
-
-    // These are required "internally" by `defmt`.
-    pub use defmt::{Formatter, Str, export, unreachable};
+// This module is hidden in the docs, but would still be imported by a wildcard import of this
+// crate's items.
+#[doc(hidden)]
+pub mod hidden {
+    // Required so the macros can access it.
+    #[cfg(feature = "defmt")]
+    pub use defmt;
 }
 
 #[cfg(feature = "defmt")]
@@ -85,7 +75,7 @@ mod log_macros {
     #[macro_export]
     macro_rules! trace {
         ($($arg:tt)*) => {{
-            use $crate::defmt::hidden::defmt;
+            use $crate::hidden::defmt;
             if true {
                 defmt::trace!($($arg)*);
             } else {
@@ -98,7 +88,7 @@ mod log_macros {
     #[macro_export]
     macro_rules! debug {
         ($($arg:tt)*) => {{
-            use $crate::defmt::hidden::defmt;
+            use $crate::hidden::defmt;
             if true {
                 defmt::debug!($($arg)*);
             } else {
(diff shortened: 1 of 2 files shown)


src/ariel-os-debug-log/src/lib.rs
```diff
@@ -256,7 +246,7 @@ impl<T: AsRef<[u8]>> core::fmt::Display for Hex<T> {
 #[cfg(feature = "defmt")]
 impl<T: AsRef<[u8]>> defmt::Format for Hex<T> {
     fn format(&self, f: defmt::Formatter<'_>) {
-        ::defmt::write!(f, "{=[u8]:02x}", self.0.as_ref());
```
ROMemories, line 259: The `::` prefix was needed to disambiguate between the `defmt` crate and the `defmt` module; as this removes the module, this is not needed anymore.

src/ariel-os-debug-log/src/lib.rs
```diff
@@ -17,23 +17,13 @@
...
-        #[doc(hidden)]
-        pub use defmt;
-    }
-
-    pub use defmt::{Debug2Format, Display2Format, Format};
-
-    // These are required "internally" by `defmt`.
-    pub use defmt::{Formatter, Str, export, unreachable};
+// This module is hidden in the docs, but would still be imported by a wildcard import of this
+// crate's items.
```
ROMemories, line 21: This means that users using a wildcard would now get a module named `hidden`; not sure whether keeping the `defmt` module is better though.

examples/storage/src/main.rs
```diff
@@ -3,7 +3,7 @@
 
 use ariel_os::debug::{
     ExitCode, exit,
-    log::{Hex, defmt, info},
+    log::{Hex, info},
```
ROMemories, line 6: Depending on the crate directly actually fixes this example as the following wouldn't compile:

```sh
laze -C examples/storage/ build -s log -b stm32u083c-dk 
```

ROMemories: This is better addressed in #1999.

src/ariel-os-debug-log/src/lib.rs
chrysn: Maybe give it a very-unlikely-to-collide name just for cases of multiple wildcard imports? Like `_hidden_log_defmt_for_macros`?

(No strong preference though, especially not about the leading underscore, which I've sometimes seen in this use, but maybe it's a Pythonism)

chrysn (commented):
Makes sense, not only because compatibility has improved, but also because defmt is now a stable public dependency. (Back then you could have wanted to re-export the-very-version-of-defmt-that-Ariel-uses, now that is `^1`').

src/ariel-os-debug-log/src/lib.rs
chrysn: For clarification, this concern is purely to keep users from getting warnings that "both those things you wildcard-import have something named `hidden`, we're using this over that; it won't affect functionality anyway.

Might also consider having an `ariel_os::prelude` that only exports the relevant items from here and other places.

ROMemories: > Maybe give it a very-unlikely-to-collide name just for cases of multiple wildcard imports? Like `_hidden_log_defmt_for_macros`?

That clearly seems useful, I've renamed it to `hidden_for_logging_macros`: that should avoid the need for renaming it again in the future if its scope changed a bit, and I do think the underscore is a Pythonism, the leading underscore only means "unused" I think (the only counterexample I know is `_print`, internal to the std, because it dates back from before the 1.0.

chrysn (approved):

PR #1997, opened 2026-04-08
fix(debug-logging): add dummy `x2Format` struct

anlavandier:
# Description

This PR adds dummy `Debug/Display2Format` structs when no logging facades are used. 

This is done to:
- stay consistent with other items like `Hex`, `Cbor` or the log macros that also have dummy versions 
- allow dependents on `ariel-debug-log` that don't want to opt-in to a facade and use `Debug2Format` to run `clippy` on their crates which would otherwise fail on failed imports. 

## Testing

Import `ariel_os_debug_log::Debug2Format` in a crate that only has `ariel-os-debug-log` as a dependency without any feature enabled. Running `clippy`  should succeed with this PR when it woud previously fail. 

## Issues/PRs References

- Depends on #2013

## Change Checklist

- [n/a] I have made corresponding changes to the documentation.

### DIFF
--- a/src/ariel-os-log/src/lib.rs
+++ b/src/ariel-os-log/src/lib.rs
@@ -67,31 +67,10 @@ pub fn print_panic(info: &core::panic::PanicInfo<'_>) {
 #[cfg(feature = "log")]
 #[doc(hidden)]
 pub mod log {
-    use core::fmt::{Debug, Display, Formatter};
 
     // Re-export only the minimum set of items to minimize breaking changes in case `log`
     // adds/removes any items.
     pub use log::{debug, error, info, trace, warn};
-
-    /// No-op wrapper that formats the Debug trait (drop-in replacement for the equivalent `defmt`
-    /// type).
-    pub struct Debug2Format<T: Debug>(pub T);
-
-    impl<T: Debug> Debug for Debug2Format<T> {
-        fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), core::fmt::Error> {
-            self.0.fmt(f)
-        }
-    }
-
-    /// No-op wrapper that formats the Display trait (drop-in replacement for the equivalent
-    /// `defmt` type).
-    pub struct Display2Format<T: Display>(pub T);
-
-    impl<T: Display> Display for Display2Format<T> {
-        fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), core::fmt::Error> {
-            self.0.fmt(f)
-        }
-    }
 
     #[cfg(all(
         context = "ariel-os",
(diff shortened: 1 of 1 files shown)


src/ariel-os-log/src/lib.rs
```diff
@@ -72,6 +72,19 @@ pub mod log {
 #[cfg(feature = "log")]
 pub use log::{Debug2Format, Display2Format};
 
+#[cfg(not(any(feature = "log", feature = "defmt")))]
+#[doc(hidden)]
+mod dummy {
+    use core::fmt::{Debug, Display};
+    /// Dummy struct needed to run clippy on crates that don't select a logging facade.
+    pub struct Debug2Format<T: Debug>(pub T);
```
ROMemories, line 112: We might want to *consider* reusing the type definitions from the `log` implementation, as they *are* semantically the same.

ROMemories (commented):
This module is really in flux right now (through #1987 and the changes stemming from it), so it's a bit difficult for me to reason about it this at the moment, but I do think we want that (but can't be sure yet).

src/ariel-os-log/src/lib.rs
anlavandier: In terms of features and modules it seems complicated to make this not ugly. 
It could be like this I guess 
```rust

/// Module containing the structs needed only for defmt 
#[cfg(not(feature = "defmt"))]
mod dummy {
  use core::fmt::{Debug, Display};
  pub struct  Debug2Format<T: Debug>(pub T);
  pub struct  Display2Format<T: Display>(pub T);
}

#[cfg(not(feature = "defmt"))]
pub use dummy::{Debug2Format, Display2Format};

#[cfg(feature = "log")]
mod log {
  ... 
  impl<T: Debug> for super::Debug2Format<T> {
     ...
  }
}

Comment (ROMemories) [reply src/ariel-os-log/src/lib.rs:112]: I've given this some more thought, I think we can actually simplify things a lot: I think the existing `log::{Debug2Format, Display2Format}` types should simply be available when `defmt` is not enabled. The other benefit is that these types would always have an implementation of `Debug` and `Display` (respectively) because I've since realized that it would be odd that these dummy types didn't.

**Review by ROMemories — COMMENTED** [review]
I think this should be done, but in a slightly different (and simpler) way as explained in the inline comment.

The following can be used for testing:

```sh
cargo clippy -p ariel-os-sensor-nrf91-gnss --features 'nrf-modem/nrf9160' 
```

(This errors out on main, and only prints warnings with this PR.)

src/ariel-os-log/src/lib.rs
```diff
@@ -122,8 +101,33 @@ pub mod log {
     pub use crate::noop_println as println;
 }
 
-#[cfg(feature = "log")]
-pub use log::{Debug2Format, Display2Format};
+#[cfg(not(feature = "defmt"))]
+#[doc(hidden)]
+mod dummy {
```
ROMemories: I think this should be renamed, as these are not only dummy types anymore (even though the module is private, this could be a bit confusing):
```suggestion
mod format_wrappers {
```

src/ariel-os-log/src/lib.rs
```diff
@@ -122,8 +101,33 @@ pub mod log {
     pub use crate::noop_println as println;
 }
 
-#[cfg(feature = "log")]
-pub use log::{Debug2Format, Display2Format};
+#[cfg(not(feature = "defmt"))]
```
ROMemories, line 105: This could likely use a comment clarifying the intent of this feature-gating:
```suggestion
// NOTE: this module is used both for `log` and when no logging facades are enabled.
#[cfg(not(feature = "defmt"))]
```

ROMemories (commented):

src/ariel-os-log/src/lib.rs
anlavandier: Done in [4a6872b](https://github.com/ariel-os/ariel-os/pull/1997/commits/4a6872bfde27f25940f4b5096ca510b9fbf41e48)

src/ariel-os-log/src/lib.rs
anlavandier: Done in [4a6872b](https://github.com/ariel-os/ariel-os/pull/1997/commits/4a6872bfde27f25940f4b5096ca510b9fbf41e48)

ROMemories (approved):

```sh
laze -C examples/log build -b stm32u083c-dk run 
```

```sh
laze -C examples/log build -s log -b stm32u083c-dk run 
```

The following now only generates warnings:

```sh
cargo clippy -p ariel-os-sensor-nrf91-gnss --features 'nrf-modem/nrf9160'
```

---

Please squash!

PR #1999, opened 2026-04-09
docs(storage): impl `defmt::Format` if and only if needed

ROMemories:
# Description

This introduces a laze module in the `storage` example to enable `defmt` *in the application* if and only if needed. This makes this example now properly usable with both `defmt` and `log`.

## Testing

Successfully tested in hardware with:

```sh
laze -C examples/storage/ build -b stm32u083c-dk -v run
```

and

```sh
laze -C examples/storage/ build -s log -b stm32u083c-dk -v run
```

Printing the `size` allows to check that `defmt` is appropriately enabled only when needed:

```sh
$ laze -C examples/storage/ build -b stm32u083c-dk -v size
  53244      88   13544   66876   1053c build/bin/stm32u083c-dk/cargo/thumbv6m-none-eabi/release/example-storage
```

and

```sh
$ laze -C examples/storage/ build -s log -b stm32u083c-dk -v size
  64748      40   13612   78400   13240 build/bin/stm32u083c-dk/cargo/thumbv6m-none-eabi/release/example-storage
```

## Issues/PRs References

- Related to #1022 (and maybe completely addresses it)

## Changelog Entry

The `storage` example can now be used with both `defmt` or `log`.

## Change Checklist

### DIFF
--- a/examples/storage/laze.yml
+++ b/examples/storage/laze.yml
@@ -1,4 +1,14 @@
 apps:
   - name: example-storage
     selects:
+      - ?example-storage-defmt
       - sw/storage
+
+modules:
+  - name: example-storage-defmt
+    selects:
+      - defmt
+    env:
+      global:
+        FEATURES:
+          - defmt
--- a/examples/storage/src/main.rs
+++ b/examples/storage/src/main.rs
@@ -3,7 +3,7 @@
 
 use ariel_os::debug::{
     ExitCode, exit,
-    log::{Hex, defmt, info},
+    log::{Hex, info},
 };
 
 // Imports for using [`ariel_os::storage`]
@@ -13,7 +13,8 @@ use serde::{Deserialize, Serialize};
 /// Example object.
 ///
 /// The serde Serialize / Deserialize traits are required for storage.
-#[derive(Serialize, Deserialize, Debug, defmt::Format)]
+#[derive(Serialize, Deserialize, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 struct MyConfig {
     val_one: heapless::String<64>,
     val_two: u64,
@@ -69,7 +70,6 @@ async fn main() {
         .await
         .unwrap()
     {
-        // no `defmt::Format` for arrayvec, so just print length
         info!(
             "Attempting to retrieve string value as ArrayString: {}",
             Hex(string.as_bytes())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3225,6 +3225,7 @@ dependencies = [
  "ariel-os",
  "ariel-os-boards",
  "arrayvec",
+ "defmt 1.0.1",
  "heapless 0.9.1",
  "serde",
 ]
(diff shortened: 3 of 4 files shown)


examples/storage/src/main.rs
```diff
@@ -69,7 +70,6 @@ async fn main() {
         .await
         .unwrap()
     {
-        // no `defmt::Format` for arrayvec, so just print length
```
ROMemories, line 72: This comment is a leftover (but on-topic enough to be removed in this PR).

examples/storage/src/main.rs
```diff
@@ -3,7 +3,7 @@
 
 use ariel_os::debug::{
     ExitCode, exit,
-    log::{Hex, defmt, info},
```
ROMemories, line 6: This prepares for the removal of the module in #1994.

examples/storage/laze.yml
```diff
@@ -1,4 +1,14 @@
 apps:
   - name: example-storage
     selects:
+      - ?example-storage-defmt
       - sw/storage
+
+modules:
+  - name: example-storage-defmt
```
ROMemories, line 8: This is similar to:

https://github.com/ariel-os/ariel-os/blob/7d67839feb426cb44bbf8cc7394facb5fe251ed1/examples/sensors-debug/laze.yml#L26-L32

Unfortunately it seems that these modules have to be unique in the whole project, so I picked a different name (and the one from `sensors-debug` should likely be renamed later).

ROMemories:
The CI failure seems unrelated (it seems that ~~rustup~~Cargo is panicking?).
Edit: it seems to have been fixed upstream and deployed. CI now runs successfully it seems.

examples/storage/laze.yml
```diff
@@ -1,4 +1,14 @@
 apps:
   - name: example-storage
     selects:
+      - ?example-storage-defmt
```
chrysn, line 4: What this does might be a bit surprising for users: It doesn't mean "if we're running on defmt, enable this" (and then defmt gets derived and used), but "unless there's any good reason not to, use it" (and the only good reason is that native `disables: defmt`, which native does).

Do you have any idea how to make this understood more easily?

chrysn (approved):
Generally LGTM; didn't test but for this kind, CI covers that well.

See comment inline.

(Also, I still have hopes that we might eventually do this all better / differently, with really additive features and no spurious trait requirements that might come from innocent derives on generics, but that doesn't need to be resolved for this enhancement to pass).

examples/storage/laze.yml
ROMemories: > if we're running on defmt, enable this

I think that this what is doing actually, but I agree this could be clearer.

> the only good reason is that native `disables: defmt`, which native does

This is (or soon will) also be the case for any debug logging transport (see #1987 for definitions) with which we can't or don't want to do `defmt`-encoding, e.g., UART with a serial monitor we don't have control over.

ROMemories:
I'm merging this as this needed to unblock #1994 (working toward #1987), but I agree with the point you've raised. As the same thing would need to be done for the `sensors-debug` example, this can be done in a cross-example PR.

PR #2000, opened 2026-04-11
RFC: GPIO emulation in native

dcz-self:
# Description

This is a prototype version of GPIO on the native target.

Does everyone think that this is the right direction to take, at least before we get out-of-tree board definitions?

It's the simplest version I could think of, and it's not very invasive, but also not extremely ergonomic. It has one important thing, though: emulation is done using regular Rust channels, meaning everything stays in one binary/process, so building and running is really straightforward.

Background: my project will use a bunch of GPIO. I'd like to be able to test GPIO using the native board while I develop.

## Testing

This example shows how the native build is done. The changes versus an actual board should be self-explanatory
https://codeberg.org/dcz/ariel-os-native

## Issues/PRs References

I've chatted about it on Matrix

## Open Questions

- Should the channel be sync or async? 

Shortcomings:
- you can't use the exact same code for the native targets and the other targets because the native target needs an explicit handler. Perhaps this could be handled by some clever separation of modules, where only native board code ever sees the handler.
- only pin level supported
- if the emulation doesn't handle all the peripherals that actually get used, messages will queue up and presumably some application threads would stall

Technical stuff:
- I split out dummy peripherals into their own module, so that the native board can import them
- that was maybe unnecessary
- I'll clean up commit messages on request
- some things are not perfect (trait not sealed, leftover comments, dummy module with unneeded changes) and I'll clean it up on request

## Change Checklist


kaspar030:
happy 2k btw! :birthday:

kaspar030:
@dcz-self Thanks for working on this! I'm quite full but will take a closer look soon.

kaspar030:
I split out dummy peripherals into their own module, so that the native board can import them

May I suggest splitting that refactor out of this PR? Should be quick to review, and once you have *any* PR in, we don't have to approve your workflow manually anymore. :wink:

kaspar030:
Should the channel be sync or async?

Let's see how we'll use this - being on native means we're not resource starved, and there's plenty of channel options that support both async and sync.

dcz-self:
So I split out the spliting out of dummy peripherals: https://github.com/ariel-os/ariel-os/pull/2068

PR #2001, opened 2026-04-12
docs(support-matrix): a re-indent

stappersg:
book/templates/support-matrix.html.tmpl is re-indented. A "white space" change only, to make upcoming a change on same file cleaner.

Signed-off-by: Geert Stappers <stappers@stappers.it>


stappersg:
image

`Signed-off-by:` is with double _f_

stappersg:
There will be another attempt.

PR #2002, opened 2026-04-12
Draft Scroll table

stappersg:
# Description

Two commits that I don't to _squash_ into one commit.
Because the first commit, #2001,  is a white space change
that will pollute the actual change.

## Issues/PRs References

#1967


stappersg:
There will be another attempt.

PR #2005, opened 2026-04-13
issue template: feature versus freebee

stappersg:
For creating awareness that features don't happen magically.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -31,7 +31,10 @@ body:
       label: Are you planning to do it yourself in a pull request?
       options:
         - "Yes"
-        - "No"
         - Maybe
+        - "No, but I'm about to hire someone to do it"
+        - "No. Please contact me for funding this feature"
+        - "No. I'm only expressing a wish"
+        - "Yes, in very far future. Don't wait for me."
     validations:
       required: true


.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -31,7 +31,10 @@ body:
       label: Are you planning to do it yourself in a pull request?
       options:
         - "Yes"
-        - "No"
         - Maybe
+        - "No, but I'm about to hire someone to do it"
```
chrysn, line 35: I'd suggest rolling this into "yes" – it's just a wider sense of "you".

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -31,7 +31,10 @@ body:
...
+        - "No. Please contact me for funding this feature"
```
bergzand, line 36: This option is ambiguous, which way does the funding flow here? Should Ariel OS contact the contributor for funding or should Ariel OS contact the contributor when funding is available for an implementation

bergzand (changes requested):
> Thank you for the contribution.

One additional option that came to mind that is also missing here is a case where the author would want to implement the feature, but requires mentoring. Something along the lines of:

> - "Yes, but I would need mentoring to implement it"

Would you be okay with also adding that option?

PR #2007, opened 2026-04-14
refactor(ariel-os-debug-log): rename the crate to `ariel-os-log`

ROMemories:
# Description

As mentioned in https://github.com/ariel-os/ariel-os/pull/1987#issuecomment-4242681143, this renames the `ariel-os-debug-log` to `ariel-os-log`. I think we can consider this an internal change (and therefore not a breaking change), as we were not generating docs for `ariel-os-debug-log` and never advertised its standalone usage.

This *only* renames the crate and does not affect the API of the `ariel_os` crate.

## Testing

CI does most of it; I've also tested the `log` example in hardware.

## Issues/PRs References

- Changes to the `ariel_os` API will be done in another PR (#2008), that *will* be a breaking change.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2008, opened 2026-04-14
refactor(logging)!: rename "debug logging" to "logging"

ROMemories:
# Description

As mentioned in https://github.com/ariel-os/ariel-os/pull/1987#issuecomment-4242681143, this renames "debug logging" to "logging", and replaces `ariel_os::debug::log` with the top-level `ariel_os::log`.

## Testing

I've tested the `log` example in hardware, and searched for leftovers with the following case-insensitive regexes: `debug.log`, `debug$`.

Also tested `laze -C examples/log build -DLOG=warn -s log -b stm32u083c-dk -v run` to make sure log level filtering still works when use the `log` logger.

This PR only takes care of this, and in particular does not address `ariel_os::debug::println!()`, or the distinction between "debug output" vs "logging" introduced by #1987 (including moving items from `ariel-os-debug` that actually belong in `ariel-os-log`).

## Issues/PRs References

- Depends on #2007

## Changelog Entry

"Debug logging" has been renamed to "logging": `ariel_os::log` should be used instead of `ariel_os::debug::log`. In addition, the `ariel_os::debug::log::defmt` module has been removed entirely. The portable items from `ariel_os::log` should be used instead, or the `defmt` crate should be used directly instead if actually needed.

## Change Checklist

### DIFF
--- a/src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
+++ b/src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
@@ -18,7 +18,7 @@ pub enum GnssOperationMode {
 pub struct Config {
     /// The GNSS operating mode to use.
     pub operation_mode: GnssOperationMode,
-    /// Whether NMEA messages should be logged as debug logs (adds extra processing).
+    /// Whether NMEA messages should be logged as logs (adds extra processing).
     pub log_nmea: bool,
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,6 +149,7 @@ dependencies = [
  "ariel-os-embassy",
  "ariel-os-hal",
  "ariel-os-identity",
+ "ariel-os-log",
  "ariel-os-macros",
  "ariel-os-nrf",
  "ariel-os-power",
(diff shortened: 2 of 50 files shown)


ROMemories:
I wanted to make sure that this would not introduce/require cyclic dependencies for doing #2013 before marking this for review, but #2013 is now in a state that shows that it is not the case, so marking this for review.

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
```diff
@@ -18,7 +18,7 @@ pub enum GnssOperationMode {
 pub struct Config {
     /// The GNSS operating mode to use.
     pub operation_mode: GnssOperationMode,
-    /// Whether NMEA messages should be logged as debug logs (adds extra processing).
+    /// Whether NMEA messages should be logged as logs (adds extra processing).
```
bergzand, line 21: This looks duplicate:
```suggestion
    /// Whether NMEA messages should be logged (adds extra processing).
```

bergzand (approved):
One nitpick. Feel free to squash in or ignore if you disagree

src/sensors/ariel-os-sensor-nrf91-gnss/src/config.rs
ROMemories: I can see it being a bit redundant, but "should be logged" would still make me wonder where it would be (especially as these are separate drivers). Keeping as-is, it can always be changed later if we really want to revisit this.

PR #2013, opened 2026-04-15
fix(logging)!: provide `println!()` when using the `log` facade

ROMemories:
# Description

This heavily revisits the relationship between the `ariel-os-log` and `ariel-os-debug` crates. The motivation is mostly twofold: aligning the contents of the crates with their responsibility as mentioned in https://github.com/ariel-os/ariel-os/pull/1987#issuecomment-4242681143 and defined in the documentation of #1987, and providing the missing `println!()` macro from `ariel_os::log` when using the `log` facade (see https://github.com/ariel-os/ariel-os/pull/1989#issuecomment-4183480958).

Unfortunately I haven't found a way to split this PR in smaller chunks, as I couldn't find intermediate states where the various `println!()` macros were not more broken than on main. Instead, here is a list of semantic changes this PR does:

- Move the `log` logger from `ariel-os-debug` to `ariel-os-log`
- **Add a `println!()` macro to `ariel-os-log` when the `log` facade is used**
- **Remove `ariel-os-debug`'s `println!()` macro from the API** (as discussed out of band)
- Expose the debug output from `ariel-os-debug` to `ariel-os-log` when using `rtt-target`, by exposing a `println!()` macro to it (hidden from the public API)
- Make `ariel-os-log` depend on `ariel-os-debug` instead of the opposite
- Move the `panic-printing` logic into `ariel-os-log`, and **make it use the logging output** instead of the debug output directly (which means for instance panics could get printed on UART)
- Move the `debug-uart` logic into `ariel-os-log`
- Move the `esp-println` logic into `ariel-os-log`

(Changes that are visible in the API are in bold.)

Renaming `debug-uart` and revisiting the laze configuration to allow for more combinations (including an explicit `logging-over-debug-output`) is out of scope of this PR and will be done in another one.

## Future work

Introducing the following laze modules:

- `logging-over-debug-output`
- `logging-over-uart`
- `logging-over-usb`

and likely slightly realigning the set of Cargo features on `ariel-os`/`ariel-os-log`/`ariel-os-debug` when doing so.

### Testing the debug output and logging

The testing matrix of this is quite large, as we need `defmt-rtt`, `rtt-target` with `log` (and maybe `defmt` as well?), `esp-println` with both `defmt` and `log`, and `debug-uart` (soon to be renamed; always uses `log`) to still be working.

Here are the cases I've checked (they should have have one `println!()` and three log lines):

- `laze -C examples/log build -b stm32u083c-dk -v run`
- `laze -C examples/log build -s log -b stm32u083c-dk -v run`
- `laze -C examples/log build -s rtt-target -s log -b stm32u083c-dk -v run`
- `laze -C examples/log build -b espressif-esp32-c6-devkitc-1 -v run`
- `laze -C examples/log build -s log -b espressif-esp32-c6-devkitc-1 -v run`
- `laze -C examples/log build -s debug-uart -b stm32u083c-dk -v run`
- `laze -C examples/log build -s probe-rs -b espressif-esp32-c6-devkitc-1 -v run`
- `laze -C examples/log build -s rtt-target -s log -s probe-rs -b espressif-esp32-c6-devkitc-1 -v run`
- `laze -C examples/log build -b native run`

The nrf52840dk can be used instead of the stm32u083c-dk.
Observing command parameters printed by `-v`, the style of the output, and the binary size (using the `size` laze task) can be useful to make sure whether `defmt` or `log` is effectively used in each case.

When using `debug-uart`, the following can be used as serial monitor (the serial device may need to be adjusted):

```sh
picocom /dev/ttyACM0 --baud 115200
```

### Testing panics

Additionally, panics also need to be tested, especially as they are now printed on the logging output instead of only the debug output. To test this, a `panic!("test panic");` can be added at the end of the `log` example (for instance) before going the above list of commands again.

## Issues/PRs References

- Depends on #2008
- Closes #1989

## Open Questions

- Should `defmt` be usable with `rtt-target`? I think the laze config doesn't currently allow for that, but there are remnants of implementation from before `defmt-rtt` was added in #1603.

## Changelog Entry

The logging facilities have been revisited: `ariel_os::debug::println!()` has been removed, `ariel_os::log::println!()` must be used instead; `ariel_os::log::println!()` now also exists when using the `log` logging facade. Panics are always printed to the logging output. 

## Change Checklist

### DIFF
--- a/book/src/debug-console.md
+++ b/book/src/debug-console.md
@@ -6,7 +6,7 @@ The debug console is enabled by default and the corresponding [laze module][laze
 
 ## Printing on the Debug Console
 
-The [`ariel_os::debug::println!()`][println-macro-rustdoc] macro is used to print on the debug console.
+The [`ariel_os::log::println!()`][println-macro-rustdoc] macro is used to print on the debug console.
 
 When the debug console is enabled, panic messages are automatically printed to it.
 If this is unwanted, the `panic-printing` [laze module][laze-modules-book] can be disabled.
@@ -95,7 +95,7 @@ The defmt logger supports configuring the log level per crate and per module, as
 $ laze build -C examples/log --builders nrf52840dk -DLOG=info,ariel_os_rt=trace run
 ```
 
-Note: On Cortex-M devices, the order of `ariel_os::debug::println!()` output and
+Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
       `defmt` log output is not deterministic.
 
 #### [log]
@@ -106,5 +106,5 @@ Ariel OS's logger for `log` supports configuring the log level globally, but do
 [defmt documentation]: https://defmt.ferrous-systems.com/
 [log]: https://github.com/rust-lang/log
 [laze-modules-book]: ./build-system.md#laze-modules
-[println-macro-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/debug/macro.println.html
+[println-macro-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/log/macro.println.html
 [debug-exit-fn-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/debug/fn.exit.html
--- a/src/ariel-os-debug/Cargo.toml
+++ b/src/ariel-os-debug/Cargo.toml
@@ -6,71 +6,32 @@ rust-version.workspace = true
 repository.workspace = true
 license.workspace = true
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[dependencies]
-ariel-os-log = { workspace = true }
-ariel-os-utils = { workspace = true }
-const-str = { workspace = true }
-critical-section = { workspace = true, optional = true }
-defmt-rtt = { workspace = true, optional = true }
-embassy-sync = { workspace = true }
-featurecomb = { workspace = true }
-log = { workspace = true, optional = true }
-rtt-target = { workspace = true, optional = true }
-semihosting = { workspace = true, optional = true }
-
 [package.metadata.feature-groups]
-debug-backend.xor = { features = ["esp-println", "rtt-target", "uart"] }
+debug-output.xor = { features = ["defmt-rtt", "rtt-target", "std"] }
 
 [package.metadata.features]
-debug-console = { groups = ["debug-backend"] }
+debug-console = { groups = ["debug-output"] }
+
+[dependencies]
+defmt-rtt = { workspace = true, optional = true }
+featurecomb = { workspace = true }
+rtt-target = { workspace = true, optional = true }
+semihosting = { workspace = true, optional = true }
 
 [target.'cfg(context = "xtensa")'.dependencies]
 semihosting = { workspace = true, optional = true, features = [
   "openocd-semihosting",
 ] }
 
-[target.'cfg(context = "esp")'.dependencies]
-esp-println = { workspace = true, optional = true, features = [
-  "auto",
-  "colors",
-  "critical-section",
-] }
-
-[target.'cfg(context = "esp32")'.dependencies]
-esp-println = { workspace = true, optional = true, features = ["esp32"] }
-
-[target.'cfg(context = "esp32c3")'.dependencies]
-esp-println = { workspace = true, optional = true, features = ["esp32c3"] }
-
-[target.'cfg(context = "esp32c6")'.dependencies]
-esp-println = { workspace = true, optional = true, features = ["esp32c6"] }
-
-[target.'cfg(context = "esp32s2")'.dependencies]
-esp-println = { workspace = true, optional = true, features = ["esp32s2"] }
-
-[target.'cfg(context = "esp32s3")'.dependencies]
-esp-println = { workspace = true, optional = true, features = ["esp32s3"] }
-
 [features]
 debug-console = []
 
-defmt = [
-  "ariel-os-log/defmt",
-  "esp-println?/defmt-espflash",
-  "rtt-target?/defmt",
-]
-log = ["dep:critical-section", "dep:log", "ariel-os-log/log"]
-
 semihosting = ["dep:semihosting"]
-
-# Debug output backends
-defmt-rtt = ["dep:defmt-rtt"]
-esp-println = ["dep:esp-println"]
-rtt-target = ["dep:rtt-target"]
 std = []
-uart = []
+
+# Debug output transports.
+defmt-rtt = ["dep:defmt-rtt"]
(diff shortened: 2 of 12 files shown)


src/ariel-os-debug/Cargo.toml
```diff
@@ -6,71 +6,32 @@ rust-version.workspace = true
...
-[target.'cfg(context = "esp32s3")'.dependencies]
-esp-println = { workspace = true, optional = true, features = ["esp32s3"] }
-
 [features]
 debug-console = []
 
-defmt = [
-  "ariel-os-debug-log/defmt",
-  "esp-println?/defmt-espflash",
-  "rtt-target?/defmt",
```
ROMemories, line 62: Currently this is lost, it's unclear whether it is needed to bring it back (see the open question).

src/ariel-os-log/src/lib.rs
```diff
@@ -26,9 +35,31 @@ pub mod hidden_for_logging_macros {
...
+    let message = info.message();
+
+    // `PanicMessage` does not currently implement `defmt::Format`.
+    // We *need* to use the `Display` implementation and cannot use `PanicMessage::as_str()` as
+    // that would not work for dynamically formatted messages.
+    #[cfg(feature = "defmt")]
+    let message = Display2Format(&message);
+
+    // Mimics the `Display` implementation of `core::panic::PanicInfo`.
+    println!("panicked at {}:\n{}", location, message);
```
ROMemories, line 64: Warnings about will need to be addressed.

book/src/debug-console.md
```diff
@@ -6,7 +6,7 @@ The debug console is enabled by default and the corresponding [laze module][laze
 
 ## Printing on the Debug Console
```
ROMemories, line 7: We'll likely want to revisit this section (to distinguish between the debug output and logging), but the way this is phrased still seems correct enough, so it can be done in a follow-up.

src/ariel-os-log/src/log_logger.rs
ROMemories, line 1: This is really just moved from `ariel-os-debug`.

src/ariel-os-log/src/lib.rs
```diff
@@ -57,6 +92,24 @@ pub mod log {
...
+    pub use crate::uart_println as println;
+
+    /// Prints to the logging output, with a newline.
+    #[cfg(not(feature = "ariel-os"))]
+    #[macro_export]
+    macro_rules! noop_println {
+        ($($arg:tt)*) => {};
+    }
+    #[cfg(not(feature = "ariel-os"))]
+    pub use crate::noop_println as println;
```
ROMemories: This dispatching logic will be revisited in a follow-up that introduces the `logging-over-*` laze modules.

src/ariel-os-debug/src/lib.rs
```diff
@@ -45,11 +21,11 @@ mod backend {
...
 mod backend {
-    #[cfg(not(feature = "defmt"))]
-    pub use rtt_target::rprintln as println;
-
     #[cfg(feature = "defmt")]
-    pub use ariel_os_log::println;
+    pub use defmt::println as debug_output_println;
+
+    #[cfg(not(feature = "defmt"))]
+    pub use rtt_target::rprintln as debug_output_println;
```
ROMemories, line 28: I've added a commit that renames these to prevent users from mistakenly continuing to use `ariel_os::debug::println!()`, as that would otherwise still compile when using the `log` logging facade or native.

src/ariel-os-log/Cargo.toml
```diff
@@ -8,17 +8,50 @@ license.workspace = true
...
 [features]
 ## Enables `defmt` as logging facade.
-defmt = ["dep:defmt"]
+defmt = ["dep:defmt", "esp-println?/defmt-espflash"]
 ## Enables `log` as logging facade.
-log = ["dep:log"]
+log = ["dep:ariel-os-utils", "dep:const-str", "dep:critical-section", "dep:log"]
+
+# Logging transports.
+defmt-rtt = ["ariel-os-debug/defmt-rtt"]
```
bergzand, line 54: I'm missing something here. Why is there a feature for a specific debug output transport for the logging transports? Shouldn't this be a case where we want to use any (enabled) debug transport for the logging transport?

ROMemories: Good question: ~~if I remember correctly, that's just a temporary thing until #2030 introduces the `logging-over-debug-output` laze module and the Cargo feature of the same name. It is actually only needed to make sure this Cargo feature is not enabled at the same time as `esp-println` or `uart`. It's one of those things that can only be half-addressed in this PR because we're lacking the higher-level `logging-over-*` laze modules and Cargo features that #2030 introduces.~~

Edit: Scratch that: this is really only because `defmt-rtt` selects both a debug output transport *and a logging transport*, so it has to go though `ariel-os-log`.

ROMemories: Also `defmt-rtt` selects both a debug output transport and a logging transport, that's why it has to go into `ariel-os-log` as well (unlike `rtt-target`).

bergzand: > Good question: if I remember correctly, that's just a temporary thing until #2030 introduces the `logging-over-debug-output` laze module and the Cargo feature of the same name. It is actually only needed to make sure this Cargo feature is not enabled at the same time as `esp-println` or `uart`. It's one of those things that can only be half-addressed in this PR because we're lacking the higher-level `logging-over-*` laze modules and Cargo features that #2030 introduces.

Makes sense, thanks for the clarification

> Also `defmt-rtt` selects both a debug output transport and a logging transport, that's why it has to go into `ariel-os-log` as well (unlike `rtt-target`).

I didn't spot any conditional compilation (*in this PR*) for `defmt-rtt` in `ariel-os-log`. Am I missing something here?

ROMemories: > I didn't spot any conditional compilation (_in this PR_) for `defmt-rtt` in `ariel-os-log`.

Indeed, because there is nothing to do about it in `ariel-os-log`: the `package.metadata.feature-groups` will enforce that only one of `"defmt-rtt", "esp-println", "uart"` is enabled at a time, and the dispatching will be done by `defmt` itself.

ROMemories: I've added a quick comment about `defmt-rtt` in the manifest to maybe clarify this a bit.

bergzand: Thanks for clarifying!

src/ariel-os-debug/Cargo.toml
bergzand: No strong opinion from my side. We can always bring it back later on.

bergzand (approved):
Tested on the nrf52840dk and on an esp32c6 variant. Please squash and LGTM

PR #2016, opened 2026-04-15
refactor(logging): drop values using the underscore when syntax-checking

ROMemories:
# Description

This replaces the usage of `core::mem::drop()` with the `let _ =` construct in our `defmt` logging macros.

This fixes the following warning that appears in #2013 for some configurations:

```
   --> src/ariel-os-log/src/lib.rs:187:17
    |
 60 |     println!("panicked at {}:\n{}", location, message);
    |     -------------------------------------------------- in this macro invocation
...
187 |                 drop(format_args!($($arg)*));
    |                 ^^^^^----------------------^
    |                      |
    |                      argument has type `core::fmt::Arguments<'_>`
    |
    = note: `-D dropping-copy-types` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dropping_copy_types)]`
    = note: this error originates in the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)
help: use `let _ = ...` to ignore the expression or result
    |
187 -                 drop(format_args!($($arg)*));
187 +                 let _ = format_args!($($arg)*);
```

(This is because [`Location`](https://doc.rust-lang.org/stable/core/panic/struct.Location.html) is `Copy`.)

## Testing

The `if true { /* … */ } else { drop(/* … */) }` construct was used to enforce at compile time that only the common denominator of display hints of [`defmt` macros](https://defmt.ferrous-systems.com/hints) and of [`log`/`std::fmt`](https://doc.rust-lang.org/stable/std/fmt/index.html#formatting-parameters) is used in our logging macros. The following can be done to make sure this still works as expected:

- In the `log` example, add a `{=u8}` display hint and a `42` parameter to:
    - The `println!()` statement
    - Any of the actual logging macros
- This should *still* be failing at compile time with the following error:

    ```sh
    error: invalid format string: expected `}`, found `=`
    ```

## Changelog Entry

I don't think this was observable in any way, so not adding to the changelog.

## Change Checklist

### DIFF
--- a/src/ariel-os-log/src/lib.rs
+++ b/src/ariel-os-log/src/lib.rs
@@ -79,7 +79,7 @@ mod log_macros {
             if true {
                 defmt::trace!($($arg)*);
             } else {
-                drop(format_args!($($arg)*));
+                let _ = format_args!($($arg)*);
             }
         }};
     }
@@ -92,7 +92,7 @@ mod log_macros {
             if true {
                 defmt::debug!($($arg)*);
             } else {
-                drop(format_args!($($arg)*));
+                let _ = format_args!($($arg)*);
             }
         }};
     }
@@ -105,7 +105,7 @@ mod log_macros {
             if true {
                 defmt::info!($($arg)*);
             } else {
-                drop(format_args!($($arg)*));
+                let _ = format_args!($($arg)*);
             }
         }};
     }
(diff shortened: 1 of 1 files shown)


anlavandier (approved):
LGTM, I was able to reproduce the testing

PR #2025, opened 2026-04-17
docs(support): clarify that the plain ESP32 has no USB peripheral at all

ROMemories:
# Description

This clarifies in the support info that the plain ESP32 doesn't have any kind of USB peripheral (unlike the other ESP32s we support, which at least have a non-generic, JTAG+USB CDC-ACM USB peripheral), by removing the comment we use on ESP32s that have that non-generic USB peripheral.

## Issues/PRs References

- Follow-up to #1948

## Changelog Entry

No need for a changelog as the support info hasn't been released yet.

## Change Checklist

### DIFF
--- a/book/src/boards/ulanzi-tc001.md
+++ b/book/src/boards/ulanzi-tc001.md
@@ -31,7 +31,7 @@ laze build -b ulanzi-tc001
 |SPI Main Mode|<span title="needs testing">🚦</span>|
 |UART|<span title="needs testing">🚦</span>|
 |Ethernet|<span title="not available on this piece of hardware">–</span>|
-|User USB|<span title="not available on this piece of hardware">–</span>[^no-generic-usb-peripheral]|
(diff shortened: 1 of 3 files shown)


book/src/boards/ulanzi-tc001.md
```diff
@@ -31,7 +31,7 @@ laze build -b ulanzi-tc001
 |SPI Main Mode|<span title="needs testing">🚦</span>|
 |UART|<span title="needs testing">🚦</span>|
 |Ethernet|<span title="not available on this piece of hardware">–</span>|
-|User USB|<span title="not available on this piece of hardware">–</span>[^no-generic-usb-peripheral]|
+|User USB|<span title="not available on this piece of hardware">–</span>|
```
ROMemories, line 34: I don't even think this needs a comment: it clearly says *User* USB, and this is no different than development kits having a USB port only for the on-board debug probe.

anlavandier (approved):
LGTM

PR #2026, opened 2026-04-17
fix(espflash): allow using `espflash` when `esp-println` is not used

ROMemories:
# Description

This introduces an `espflash` laze module, similar to the `probe-rs` laze module, to decouple the usage of the `espflash` flashing and log printing tool from the `esp-println` laze module, as they are not actually related (at least regarding flashing). In particular this allows to still use `espflash` for flashing (both through USB and over USB CDC-ACM), even when `esp-println` is manually disabled.

As background, `esp-println` is only responsible for *logging*, either over UART or USB CDC-ACM (currently *both* of them are enabled, and whichever is used is selected automatically at runtime, but I hope we revisit that soon).

## Future work

The interaction between the flashing/log printing tools, the debug output and the logging output will be revisited very soon, after #2013.

## Testing

The following can be used to highlight the issue (on main, before this PR), and observe its resolution:

```sh
laze -C examples/log build -d esp-println -d debug-console -b espressif-esp32-c6-devkitc-1 run
```

when using the "UART"-labeled USB port, or, equivalently:

```sh
laze -C examples/log build -d esp-println -d debug-console -b ulanzi-tc001 run
```

This won't print anything as this disables the debug console, using another example with an effect observable without logging/the debug console may be a better idea depending on the board.

(Disabling `debug-console` is currently needed because of the internal, *current* interaction between logging and the debug console, but this will be addressed by #2013.)

## Changelog Entry

I think we can skip the changelog on this one: we didn't document the `esp-println` laze module (and don't plan to, I hope we eventually remove it), and no other module would disable it internally.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -358,6 +358,7 @@ contexts:
     parent: ariel-os
     selects:
       - ?debug-console
+      - ?espflash
       - ?esp-println
     provides:
       - has_hwrng
@@ -1138,6 +1139,14 @@ modules:
           - "-Clink-arg=-v"
         LINK_ARG_PREFIX: -Wl,
 
+  - name: espflash
+    help: use `espflash` as runner
+    provides_unique:
+      - flashing-tool
+    env:
+      global:
+        CARGO_RUNNER: '"espflash flash --monitor ${ESPFLASH_LOG_FORMAT}"'
+
   - name: probe-rs
     help: use probe-rs as runner
     selects:
@@ -1150,6 +1159,7 @@ modules:
           - ?debug-uart
     provides_unique:
       - embedded-test-runner
+      - flashing-tool
 
     env:
       global:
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -1138,6 +1139,14 @@ modules:
           - "-Clink-arg=-v"
         LINK_ARG_PREFIX: -Wl,
 
+  - name: espflash
+    help: use `espflash` as runner
+    provides_unique:
+      - flashing-tool
```
ROMemories, line 1145: Both probe-rs and `espflash` are more than just flashing tools (they are at least also log printers), but having them conflict on this seems sufficient for now.

nponsard (approved):
Looks good to me, was able to flash using `espflash` and then attach with `probe-rs` to read the logs: 

```
laze -C examples/log/ build -b espressif-esp32-s3-devkitc-1 -d esp-println  -s defmt-rtt run

laze -C examples/log/ build -b espressif-esp32-s3-devkitc-1 -d esp-println  -s defmt-rtt -s probe-rs attach
```

anlavandier (approved):
LGTM,

Confirmed that this PR fixes: `laze -C examples/log build -d esp-println -d debug-console -b ulanzi-tc001 run`.

PR #2027, opened 2026-04-17
feat(gpio)!: introduce a lifetime generic on GPIO types

### LINKED ISSUE #2017: Adding a lifetime generic on GPIO types
Our current GPIO types (e.g., `Input` and `Output`) have no generics at all, including no lifetime generics. That was a conscious decision at the time, with the goal of making these types as simple as possible, as they would likely be one of the first beginners would interact with.

However, we've since realized that being able to release the underlying peripheral mid-execution would very likely be needed for power management (see #510) and have also encountered other use cases where this would be useful (and different from bit-banging), for instance for an on-device self-test.

It would therefore be needed to add such lifetime generic on our GPIO types (i.e., `Input`, `IntEnabledInput`, and `Output`) and propagate the lifetime to the underlying Embassy-style HAL, so that it is possible to switch at runtime between using the GPIO as an input or an output.

This will of course unfortunately be a breaking change, since existing user types that would contain such GPIO types would need to get the extra lifetime generic (but such user types should be relatively rare enough).

nponsard:
# Description

This adds lifetimes to GPIOS

## Testing

Tested with:

```
laze -C examples/gpio-reuse/ build -b rpi-pico run
laze -C examples/gpio-reuse/ build -b nrf9160dk-nrf9160 run 
laze -C examples/gpio-reuse/ build -b stm32u083c-dk run
laze -C examples/gpio-reuse/ build -b seeedstudio-xiao-esp32c6 run
```

## Issues/PRs References

- Fixes #2017 

## Changelog Entry

`ariel_os::gpio::Input`, `ariel_os::gpio::IntEnabledInput` and `ariel_os::gpio::Output` now each have a generic lifetime parameter, allowing to re-use GPIOs. 

## Change Checklist

### DIFF
--- a/examples/gpio-reuse/README.md
+++ b/examples/gpio-reuse/README.md
@@ -1,0 +1,11 @@
+# gpio-reuse
+
+## About
+
+This application is demonstrating changing the role of a GPIO at runtime.
+
+## How to run
+
+In this directory, run
+
+    laze build -b nrf52840dk run
--- a/examples/gpio-reuse/src/main.rs
+++ b/examples/gpio-reuse/src/main.rs
@@ -1,0 +1,33 @@
+#![no_main]
+#![no_std]
+
+use ariel_os_boards::pins;
+
+use ariel_os::{
+    gpio::{Input, Level, Output, Pull},
+    log::info,
+    time::Timer,
+};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn blinky(mut peripherals: pins::LedPeripherals) {
+    loop {
+        // First, set the GPIO as input and read its level.
+        // Some HALs require dropping both the driver itself and the peripheral ZST before using it
+        // again for another driver.
+        {
+            let input = Input::new(peripherals.led0.reborrow(), Pull::Down);
+
+            let level = input.get_level();
+            info!("GPIO level: {:?}", level);
+        }
+
+        // Then, make the LED blink.
+        let mut led = Output::new(peripherals.led0.reborrow(), Level::Low);
+
+        for _ in 0..(2 * 3) {
+            led.toggle();
+            Timer::after_millis(500).await;
+        }
+    }
+}
--- a/src/ariel-os-hal/src/gpio.rs
+++ b/src/ariel-os-hal/src/gpio.rs
@@ -60,21 +60,21 @@ macro_rules! inner_impl_input_methods {
 ///
 /// If support for external interrupts is needed, use [`InputBuilder::build_with_interrupt()`] to
 /// obtain an [`IntEnabledInput`].
-pub struct Input {
-    input: HalInput<'static>,
+pub struct Input<'a> {
+    input: HalInput<'a>,
 }
 
-impl Input {
+impl<'a> Input<'a> {
     /// Returns a configured [`Input`].
-    pub fn new<P: HalInputPin + 'static>(pin: impl IntoPeripheral<'static, P>, pull: Pull) -> Self {
+    pub fn new<P: HalInputPin + 'a>(pin: impl IntoPeripheral<'a, P>, pull: Pull) -> Self {
         Self::builder(pin, pull).build()
     }
 
     /// Returns an [`InputBuilder`], allowing to configure the GPIO input further.
-    pub fn builder<T: IntoPeripheral<'static, P>, P: HalInputPin + 'static>(
+    pub fn builder<T: IntoPeripheral<'a, P>, P: HalInputPin + 'a>(
         pin: T,
         pull: Pull,
-    ) -> InputBuilder<T, P> {
+    ) -> InputBuilder<'a, T, P> {
         InputBuilder {
             pin,
             pull,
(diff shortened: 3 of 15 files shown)


ROMemories:
I've changed the changelog entry: these GPIO types are from the `ariel_os::gpio` module.

examples/gpio-lifetime/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+    log::info,
+    time::Timer,
+};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn blinky(mut peripherals: pins::LedPeripherals) {
+    // First use as input
+    // esp-hal lets us have both `input` and `led` in the same scope but embassy wants us to drop `btn` before using `peripherals.led0` again.
+    {
+        let input = Input::builder(peripherals.led0.reborrow(), Pull::Down).build();
```
ROMemories:
```suggestion
        let input = Input::new(peripherals.led0.reborrow(), Pull::Down);
```

examples/gpio-reuse/src/main.rs
ROMemories, line 1: I think the example name should focus more on the high-level functionality than the implementation mechanism: maybe `gpio-reuse` (example names should be nouns, and keeping `gpio` as a prefix should make the example easier to discover)?

examples/gpio-lifetime/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+
+        let level = input.get_level();
+        info!("level: {:?}", level);
+    }
+
+    // Then blink
```
ROMemories:
```suggestion
    // Then, make the LED blink.
```

examples/gpio-lifetime/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+use ariel_os::{
+    gpio::{Input, Level, Output, Pull},
+    log::info,
+    time::Timer,
+};
+
+#[ariel_os::task(autostart, peripherals)]
+async fn blinky(mut peripherals: pins::LedPeripherals) {
+    // First use as input
+    // esp-hal lets us have both `input` and `led` in the same scope but embassy wants us to drop `btn` before using `peripherals.led0` again.
```
ROMemories: I think we should avoid referring to underlying HALs if possible:
```suggestion
    // First, set the GPIO as input and read its level.
    // Some HALs require dropping both the driver itself and the peripheral ZST before using it again for another driver.
```

examples/gpio-lifetime/src/main.rs
ROMemories: For reference (because I initially struggled to find where it was coming from): `reborrow()` comes [from `Peri` on Embassy HALs](https://docs.embassy.dev/embassy-nrf/git/nrf52805/struct.Peri.html#method.reborrow) while it comes from [individual GPIO ZSTs on esp-hal](https://docs.espressif.com/projects/rust/esp-hal/1.0.0/esp32c6/esp_hal/peripherals/struct.GPIO0.html#method.reborrow). It would be better if this were more exposed in our documentation as it seems very necessary for reusing GPIOs (and maybe other peripheral ZSTs) and if we had an abstraction that made sure that the same application can be compiled for other HALs even if initially written for ESP, but I suppose this can be done later/in a follow-up.

examples/gpio-lifetime/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        let input = Input::builder(peripherals.led0.reborrow(), Pull::Down).build();
+
+        let level = input.get_level();
+        info!("level: {:?}", level);
```
ROMemories:
```suggestion
        info!("GPIO level: {:?}", level);
```

examples/gpio-lifetime/README.md
```diff
@@ -0,0 +1,11 @@
+# gpio-lifetime
+
+## About
+
+This application is demonstrating changing the role of a pin at runtime.
```
ROMemories: Reusing the pin for another type of peripheral is not in-scope and could be done in another example (if needed, but not now):
```suggestion
This application is demonstrating changing the role of a GPIO at runtime.
```

src/ariel-os-hal/src/gpio.rs
```diff
@@ -87,20 +87,20 @@ impl Input {
 }
 
 #[doc(hidden)]
-impl embedded_hal::digital::ErrorType for Input {
+impl embedded_hal::digital::ErrorType for Input<'_> {
     type Error = <HalInput<'static> as embedded_hal::digital::ErrorType>::Error;
```
ROMemories: Maybe this `'static` also needs to be replaced? I need to think about the exact semantics of this lifetime here.

ROMemories (commented):
I need to think for a bit about the behavior on ESP (both regarding the behavior of the peripheral ZST and of the driver), but if I don't find any major issue with it this should be good to go after this one.

examples/gpio-reuse/README.md
ROMemories, line 1: The new example also needs to be added to the examples' README.

examples/gpio-reuse/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+        info!("GPIO level: {:?}", level);
+    }
+
+    // Then, make the LED blink.
+    let mut led = Output::new(peripherals.led0, Level::Low);
+
+    loop {
+        led.toggle();
+        Timer::after_millis(500).await;
+    }
```
ROMemories: While exploring a bit more, I've realized that this example wasn't obviously testing that the GPIO could still be reused after being used as an output, and makes the input and output roles look asymmetrical as only one of them uses a `reborrow()`. I suppose we could use the following, which preserves the blinking functionality while measuring the input level every few blinking patterns:

```suggestion
    loop {
        // First, set the GPIO as input and read its level.
        // Some HALs require dropping both the driver itself and the peripheral ZST before using it
        // again for another driver.
        {
            let input = Input::new(peripherals.led0.reborrow(), Pull::Down);

            let level = input.get_level();
            info!("GPIO level: {:?}", level);
        }

        // Then, make the LED blink.
        let mut led = Output::new(peripherals.led0.reborrow(), Level::Low);

        for _ in 0..(2 * 3) {
            led.toggle();
            Timer::after_millis(500).await;
        }
    }
```

ROMemories: This also demonstrates reusing GPIOs depending on "execution order", not just "program order."

nponsard: Thanks for the suggestion, it shows how flexible it is to change the role of the GPIO

ROMemories (approved):
Checked that the `gpio` example was still working on stm32u083c-dk.

Thanks for adding an example, that made reviewing much easier! Please squash!

PR #2028, opened 2026-04-20
ci: measure and report link status in the API docs

anlavandier:
# Description

This PR checks the links in the API docs and reports it as a sticky comment in the PR. 

This PR has been opened from within the Ariel OS repository to allow the new pieces of CI to work directly in this PR. 

This PR has been changed to be based on top of ~~#2031~~ #2047  to use it measure the changes in terms of broken links without requiring the contents specific to this PR to be merged into main

## Issues/PRs References
~~Based on #2031~~
Based on  #2047 

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -68,6 +68,17 @@ jobs:
               -p ariel-os \
               -p ariel-os-sensors \
               -p coapcore \
+              -p ariel-os-bench \
+              -p ariel-os-buildinfo \
+              -p ariel-os-coap \
+              -p ariel-os-debug \
+              -p ariel-os-identity \
+              -p ariel-os-log \
+              -p ariel-os-power \
+              -p ariel-os-random \
+              -p ariel-os-rt \
+              -p ariel-os-storage \
+              -p ariel-os-thread \
               --features "
                   ariel-os-sensors/max-sample-min-count-12,
                   bench,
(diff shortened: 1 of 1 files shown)


anlavandier:
The last commit was made to try to make the new CI work for #2031. This did not work and this commit is meant to be removed

ROMemories:
For reference, this is moved/copied from https://github.com/ariel-os/docs/pull/2.

In a way this seems useful, and I like that kind of monitor as we go, but also I'm a bit wary of having a comment that we know will always report errors, and that requires us to manually compare it to a baseline count of errors. Not sure what we should do it about it.

anlavandier:
That's why I added a `<details>` html element so that actual error can be reported in a less intrusive way. I still am not super impressed why how this turns out but to me this metric seems useful in the context of improving the docs

anlavandier:
This PR probably won't be merged so I'm changing it to a draft PR

anlavandier:
@ROMemories, the edits show that the number of errors went down from 108 down to 43.

ROMemories:
So the reason why lychee can't find the fragment identifiers of source line ranges (e.g., `#1-42` in rustdoc docs) is because this is done dynamically in JavaScript: each line has an `N` `id` (where `N` is the line number), and client-side JavaScript is used to read the fragment identifier and highlight the range of lines dynamically. lychee will never be able to check these, but *maybe* there is a way to ignore some fragment identifiers based on a regex (`\d+-\d+`)?

PR #2030, opened 2026-04-20
feat(logging)!: revisit how logging and the debug channel interact

ROMemories:
# Description

This revisits how logging and the debug channel interact, by both introducing new laze modules and changing the definitions of existing ones, to align with the definitions of #1987. In particular this decouples the debug console from the logging output.

For now, it's not expected that `logging-over-*` laze modules other than `logging-over-debug-output` be used *directly* by users (even though `logging-over-usb` is selected internally on ESP)~~, so making them only selectable when allowed by the hardware is out of scope of this PR, and is deferred to a follow-up~~.

## How to review this PR

This PR is better reviewed commit by commit. I can easily split it over existing commits if needed.

Most of the new laze configuration is likely better reviewed by both reading it and checking that it expresses what we want at a high-level, checking that the properties documented as comments are indeed implemented correctly (e.g., mutual exclusion of modules), and by playing with various combinations of the modules, including some that are expected to be rejected.

## Future work

Documenting the *new* laze modules will be done in #2036 and #1987, only the already-documented laze modules that are renamed in this PR should be documented in this PR.
The Debug Console and Logging documentation has been restructured in #2034.

## Testing

The following commands can be used for testing, the `info-modules` and `size` laze tasks and running in hardware are interesting to check (`hello-world` is used as an example as it uses logging and also `exit()`s):

```sh
laze -C examples/hello-world build -b stm32u083c-dk -v run
```

```sh
laze -C examples/hello-world build -d panic-printing -b stm32u083c-dk -v run
```

```sh
laze -C examples/hello-world build -d debug-console -b stm32u083c-dk -v run
```

(This will only make the firmware not `exit()`, logging will still be enabled.)

```sh
laze -C examples/hello-world build -d logging -b stm32u083c-dk -v run
```

(This will only make the firmware `exit()`, logging will not be enabled.)

```sh
laze -C examples/hello-world build -d debug-console -d logging -b stm32u083c-dk -v run
```

(This will make the firmware not `exit()` and logging will also not be enabled.)

```sh
laze -C examples/hello-world build -s logging-over-uart -b stm32u083c-dk run
```

The output can be obtained using `picocom /dev/ttyACM0 --baud 115200` for instance.

(The nrf52840dk can be used instead of the stm32u083c-dk.)

```sh
laze -C examples/hello-world build -b espressif-esp32-c6-devkitc-1 -v run
```

```sh
laze -C examples/hello-world build -d panic-printing -b espressif-esp32-c6-devkitc-1 -v run
```

```sh
laze -C examples/hello-world build -s probe-rs -d debug-console -b espressif-esp32-c6-devkitc-1 -v run
```

(Enabling `probe-rs` is necessary as `espflash`, which is used by default on `esp`, cannot offer a debug console as it doesn't support debug interface protocols.)

```sh
laze -C examples/hello-world build -b native -v run
```

```sh
laze -C examples/hello-world build -d logging -b native -v run
```

## Issues/PRs References

- Depends on #2013
- Depends on #2052
- Closes #657

## Changelog Entry

"Debug logging" has been renamed to "logging" and the definitions of logging and of the debug console have been revisited: `ariel_os::log` should be used instead of `ariel_os::debug::log`, and `ariel_os::log::println!()` instead of `ariel_os::debug::println!()`, which has been removed. In addition, the `logging` laze module should now be used instead of `logging-facade` (formerly `debug-logging-facade`) when enabling or disabling logging as a whole. Panics are now always printed to the logging output. ([#2008](https://github.com/ariel-os/ariel-os/pull/2008)) ([#2013](https://github.com/ariel-os/ariel-os/pull/2013))

The changelog entry of #2013 should likely be merged into this one when both have landed.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -22,6 +22,8 @@ contexts:
     selects:
       - executor-default
       - ?critical-section
+      # Optional as we want to be able to disable it in applications.
+      - ?logging
       - ?hw/device-identity
       - ?lto
 
@@ -361,9 +363,7 @@ contexts:
   - name: esp
     parent: ariel-os
     selects:
-      - ?debug-console
       - ?espflash
-      - ?esp-println
     provides:
       - has_hwrng
     env:
@@ -502,6 +502,7 @@ contexts:
     parent: ariel-os
     selects:
       - infini-core
+      - logging-over-stdout
       - ?debug-console
     provides:
       - has_device_identity
@@ -959,16 +960,6 @@ modules:
           - CC="${CC}"
           - CFLAGS="${CFLAGS}"
 
-  - name: debug-console
-    context: ariel-os
-    selects:
-      - ?logging-facade-default
-      - ?panic-printing
-    env:
-      global:
-        FEATURES:
-          - ariel-os/debug-console
-
   - name: emit-stack-sizes
     conflicts:
       - stable
@@ -1021,45 +1012,6 @@ modules:
         CARGO_TOOLCHAIN: +nightly-2026-04-28
         RUSTFLAGS:
           - --cfg nightly
-
-  # This module should be hard-selected when an application *requires* logging.
-  - name: logging-facade-default
-    selects:
-      - ?defmt
-      - ?log
-      - logging-facade
-
-  - name: defmt
-    help: Use the `defmt` crate as the logging facade
-    provides_unique:
-      - logging-facade
-    env:
-      global:
-        FEATURES:
-          - ariel-os/defmt
-        RUSTFLAGS:
-          - -Clink-arg=-Tdefmt.x
-        ESPFLASH_LOG_FORMAT: "--log-format defmt"
-        CARGO_ENV:
-          - DEFMT_LOG=info,${LOG}
-
-  - name: log
-    help: Use the `log` crate as the logging facade
-    provides_unique:
-      - logging-facade
-    env:
-      global:
-        FEATURES:
-          - ariel-os/log
-        CARGO_ENV:
-          - LOG_LEVEL=${LOG}
-
-  - name: panic-printing
-    context: ariel-os
-    env:
-      global:
-        FEATURES:
-          - ariel-os/panic-printing
 
   - name: lto
     context: ariel-os
@@ -1155,12 +1107,6 @@ modules:
     help: use probe-rs as runner
     selects:
       - ?debug-console
-      - ?semihosting
-      - defmt:
-          - defmt-rtt
-      - log:
-          - ?rtt-target
-          - ?debug-uart
     provides_unique:
       - embedded-test-runner
       - flashing-tool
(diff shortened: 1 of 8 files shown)


laze-project.yml
```diff
@@ -1802,6 +1748,126 @@ modules:
...
+  - name: debug-output-stdout
+    help: use native's stdout as debug output transport
+    selects:
+      - context::native-chip
+    provides_unique:
+      - debug-output-transport
+
+  - name: logging-over-usb
+    help: use USB CDC-ACM as logging transport
+    selects:
```
ROMemories, line 1875: This needs to be selectable only on boards that have the appropriate hardware (i.e., USB peripheral or USB CDC-ACM-enabled USB peripheral, physical USB port connected to it).

laze-project.yml
```diff
@@ -1820,12 +1886,13 @@ modules:
         FEATURES:
           - ariel-os/rtt-target
 
-  - name: debug-uart
-    help: use UART in ariel-os-debug
+  - name: logging-over-uart
+    help: use UART as logging transport
     selects:
```
ROMemories: This needs to be selectable only on boards that have the appropriate hardware (a UART peripheral and a way to access it the UART pins, either directly or through a USB⟷UART adapter).

laze-project.yml
ROMemories: Actually I think this is better done in a follow-up, as doing this properly requires introducing a few extra laze modules.

laze-project.yml
ROMemories: Actually I think this is better done in a follow-up, as doing this properly requires introducing a few extra laze modules, and `debug-uart` was already selectable on hardware it wasn't available on anyway.

laze-project.yml
```diff
@@ -1018,45 +1010,6 @@ modules:
         RUSTFLAGS:
           - --cfg nightly
 
-  # This module should be hard-selected when an application *requires* logging.
-  - name: logging-facade-default
-    selects:
```
ROMemories, line 1027: The rest of this chunk is just moved around to keep it close to the related modules.

laze-project.yml
```diff
@@ -1802,48 +1749,164 @@ modules:
...
+    context:
+      - native-chip
+    provides_unique:
+      - debug-output-transport
+
+  - name: logging-over-usb
+    help: use USB CDC-ACM as logging transport
+    selects:
+      - esp:
+          - esp-println
```
ROMemories, line 1877: This will be revisited in a follow-up (#2039), by turning off the `auto` Cargo feature of `esp-println` and selecting `jtag-serial` or `uart` depending on wheter `logging-over-usb` or `logging-over-uart` is selected.

laze-project.yml
```diff
@@ -1802,48 +1750,176 @@ modules:
...
+
+  - name: logging-over-debug-output
+    help: use the debug output as logging transport
+    selects:
+      - ?debug-output-stdout
+      - rtt-enabled-host-tool:
+          - ?debug-output-rtt
+      # It would technically be possible to also support semihosting as debug
+      # output transport, even though it is extremely slow.
+      - debug-output-transport
```
ROMemories: Currently this doesn't work: what we want is to enable `debug-output-rtt` when using an RTT-enabled host tool at the time of build (and if the hardware allows) but that, if neither `debug-output-stdout` or `debug-output-rtt` got selected, then neither does `logging-over-debug-output`. Currently what happens is that the `selects: [debug-output-transport]` construct still selects `debug-output-rtt` anyway.

ROMemories: I ended up gating `?debug-output-rtt` `probe-rs` directly instead of `rtt-enabled-host-tool`; it seems that `rtt-enabled-host-tool` wasn't provided by `probe-rs` early enough in the resolution process.

laze-project.yml
```diff
@@ -309,6 +311,8 @@ contexts:
   - name: rp
     help: Raspberry Pi Pico (2) MCU support (based on embassy-rp)
     parent: ariel-os
+    selects:
+      - ?probe-rs
```
ROMemories, line 319: Factoring out this here is required because modules that select `has_*` modules, which are required (very) transitively by applications (e.g., `has_hwrng` through the `random` laze module), otherwise go through `rp` without enabling `probe-rs`, which makes the selection of `logging-over-debug-channel` fail.

laze-project.yml
ROMemories: After all I've opted for restricting the set of contexts this is available on, ad hoc. This should be revisited soonish.

laze-project.yml
ROMemories: After all I've opted for restricting the set of contexts this is available on, ad hoc. This will be revisited soonish.

laze-project.yml
```diff
@@ -1802,48 +1748,198 @@ modules:
...
+      - stm32u083c-dk
+    selects:
+      - context::esp:
+          - esp-println
+      - ?debug-uart
+    provides_unique:
+      - logging-transport
+
+  # Temporary module for boards that have ad hoc support.
   - name: debug-uart
```
ROMemories, line 1897: The initial was to rename this module to `logging-over-uart`, but I realized that we needed a module dedicated to our ad hoc implementation of logging-over-UART, at least so that `defmt` is not disabled on ESP and that the `debug-uart` Cargo feature is not passed to `ariel-os`. The plan is still eventually to rename "debug UART" and ideally to remove this dedicated module later (maybe by replacing `esp-println` for UART with a custom implementation actually).

kaspar030:
All tests working fine here.

Just two observations:

- `laze build -s panic-printing -d logging` does not do any panic printing. Can we make the former depend on the other, or `require` it?

- I think this is intended, but previously, `-d debug-console` disabled all logging, panic printing and rtt. I think the equivalent now is `-d logging -d debug-console`. -> fine, should get a release note.

kaspar030:
> * Can we make the former depend on the other, or `require` it?

I'll merge as-is, we can tackle that in a follow-up.

kaspar030 (approved):

PR #1894, opened 2026-03-11
fix: put the esp sources and the static-files in the docs + link check

anlavandier:
# Description

## Change Checklist

- [ ] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [ ] I have followed the [Coding Conventions][coding-conventions].
- [ ] I have tested and performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2031, opened 2026-04-20
fix(docs): move ariel-os-esp source files in the rustdoc docs 

anlavandier:
# Description

This PR fixes some broken in our API docs caused by our dual target rustdoc workflow (`riscv32imac` for `ariel-os-esp`, `host` for the rest). This is done primarily by including the `ariel-os-esp` sources in the docs.

~~This PR is against #2028 to use the PR'ed CI changes that would otherwise not have the correct permissions.~~
This PR was changed to be against main and so that #2028 depends on this instead of the other way around.

## Issues/PRs References
This is a follow-up/refactor on #1894 

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -162,6 +162,15 @@ jobs:
           mkdir -p ./_site/dev/docs/api
           mv target/riscv32imac-unknown-none-elf/doc/ariel_os_esp/ ./_site/dev/docs/api
           mv target/doc/* ./_site/dev/docs/api
+          mv target/riscv32imac-unknown-none-elf/doc/src/ariel_os_esp ./_site/dev/docs/api/src
+          mv target/riscv32imac-unknown-none-elf/doc/static.files/* ./_site/dev/docs/api/static.files
+          # get the ariel-os-esp sidebar contents
+          # save the first line of the src-files.js in a variable
+          TEXT=`head -n 1 target/riscv32imac-unknown-none-elf/doc/src-files.js`
+          # use sed keep only the first element of the JSON string (i.e. "ariel_os_esp")
+          TEXT=`echo $TEXT | sed "s/createSrcSidebar('//" - | sed "s/');//" - | jq '.[0]' -cr`
+          # hack ariel-os-esp source files in the source files sidebar
+          sed -i "s/\[\"ariel_os\",\[\"\",\[\],\[\"lib.rs\",\"sensors.rs\"\]\]\]/[\"ariel_os\",[\"\",[],[\"lib.rs\",\"sensors.rs\"]]],$TEXT/g" ./_site/dev/docs/api/src-files.js
           # hack ariel-os-esp in the sidebar
           sed -i 's/"ariel_os","ariel_os_nrf"/"ariel_os","ariel_os_esp","ariel_os_nrf"/g' ./_site/dev/docs/api/crates.js
 


anlavandier:
Sadly this did not permit me to have the link checker comment from #2028 in a PR originating from my fork...

.github/workflows/build-deploy-docs.yml
```diff
@@ -28,7 +28,6 @@ on:
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
   contents: read
-
```
ROMemories, line 31: I think we should keep this blank line.

anlavandier:
See #2028 for the improvements that this PR brings in terms of link-checking. 
Also this PR adds the ability to browse the source files of `ariel-os-esp`

.github/workflows/build-deploy-docs.yml
```diff
@@ -162,6 +161,10 @@ jobs:
           mkdir -p ./_site/dev/docs/api
           mv target/riscv32imac-unknown-none-elf/doc/ariel_os_esp/ ./_site/dev/docs/api
           mv target/doc/* ./_site/dev/docs/api
+          mv target/riscv32imac-unknown-none-elf/doc/src/ariel_os_esp ./_site/dev/docs/api/src
+          mv target/riscv32imac-unknown-none-elf/doc/static.files/* ./_site/dev/docs/api/static.files
+          # hack ariel-os-esp source files in the source files sidebar
+          sed -i 's/\["ariel_os",\["",\[\],\["lib.rs","sensors.rs"\]\]\]/["ariel_os",["",[],["lib.rs","sensors.rs"]]],["ariel_os_esp",["",[["i2c",[["controller",[],["mod.rs"]]],["mod.rs"]],["spi",[["main",[],["mod.rs"]]],["mod.rs"]]],["gpio.rs","lib.rs","time_driver.rs","uart.rs"]]]/g' ./_site/dev/docs/api/src-files.js
```
ROMemories: This definitely seems to be working for now, but it seems like it would be brittle and could break in a way that would be non-obvious to see if/when we expose more modules from Ariel OS HALs. Would there be a way to make this more robust?

ROMemories (commented):
This was definitely broken before, and this does fix every "Source" link I've check in `ariel_os_esp`. I've also quickly checked this didn't break those for other HALs.

.github/workflows/build-deploy-docs.yml
anlavandier: After many dumb mistakes, I finally have something that's a bit more robust. Here is relies on hardcoding the name of the end of the source files related to ariel-os-esp. In practice currently this would only need to change if either a top level-module/file of `ariel-os-esp` is lexicographically after `uart.rs`. Wdyt ? (please don't mind the extraneous comments and debug prints)

.github/workflows/build-deploy-docs.yml
anlavandier: Will fix when squashing

.github/workflows/build-deploy-docs.yml
```diff
@@ -162,6 +162,22 @@ jobs:
...
+          TEXT=`head -n 1 target/riscv32imac-unknown-none-elf/doc/src-files.js`
+          # use sed keep only the fille/module hierarchy
+          TEXT=`echo $TEXT | sed "s/createSrcSidebar('\[//" - | sed "s/\]');//" -`
+          # Cut off after "uart.rs"
+          TEXT=`echo $TEXT | sed 's/"uart.rs"]]]/"uart.rs"]]]\n/g' - | head -n 1 -`
+          # Debug
+          echo $TEXT
+          # DEBUG 2
+          # hack ariel-os-esp source files in the source files sidebar
+          sed -i "s/\[\"ariel_os\",\[\"\",\[\],\[\"lib.rs\",\"sensors.rs\"\]\]\]/[\"ariel_os\",[\"\",[],[\"lib.rs\",\"sensors.rs\"]]],$TEXT/g" ./_site/dev/docs/api/src-files.js
```
ROMemories: This seems to be mostly JSON processing; could you use jq instead? It seems to [already be present in the image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools) (so no need to install it), and that would make this clearer and more robust (e.g., it seems this would still break if we added another module after `uart`).

.github/workflows/build-deploy-docs.yml
```diff
@@ -162,6 +162,22 @@ jobs:
           mkdir -p ./_site/dev/docs/api
           mv target/riscv32imac-unknown-none-elf/doc/ariel_os_esp/ ./_site/dev/docs/api
           mv target/doc/* ./_site/dev/docs/api
+          mv target/riscv32imac-unknown-none-elf/doc/src/ariel_os_esp ./_site/dev/docs/api/src
+          mv target/riscv32imac-unknown-none-elf/doc/static.files/* ./_site/dev/docs/api/static.files
+          # get the ariel-os-esp sidebar contents
+          # save the first line of the src-files.js in a variable
+          TEXT=`head -n 1 target/riscv32imac-unknown-none-elf/doc/src-files.js`
+          # use sed keep only the fille/module hierarchy
```
ROMemories:
```suggestion
          # use sed to keep only the file/module hierarchy
```

anlavandier: Closed as irrelevant since I changed to wording in the 2c72e44

.github/workflows/build-deploy-docs.yml
anlavandier: Done in [2c72e44](https://github.com/ariel-os/ariel-os/pull/2031/commits/2c72e4478adcac9762af7d50008dbf2e3b01af68)

ROMemories (approved):
Seems robust enough now, I've checked the output again, thanks for tackling this! Please squash!

PR #2034, opened 2026-04-21
docs(book): dedicate a page to logging

ROMemories:
# Description

Following #2008, this splits the logging section from the Debug Console page into a new Logging page. The main goal is to avoid conflating the two as #1987 clearly defines them as orthogonal. The Logging page will also be expanded, possibly by moving the discussion about logging transports from #1987 into it.

## How to review this PR

This PR is better reviewed commit by commit: as this mostly moves an entire existing section, the first commit only does that, and the following fixup commits then take care of the necessary adjustments. I'll squash the fixup commits after review.

## Future work

The Logging page will then be updated following #2030.

## Changelog Entry

"Debug logging" has been renamed to "logging" and the definitions of logging and of the debug console have been revisited: `ariel_os::log` should be used instead of `ariel_os::debug::log`, and `ariel_os::log::println!()` instead of `ariel_os::debug::println!()`, which has been removed. In addition, the `logging` laze module should now be used instead of `logging-facade` (formerly `debug-logging-facade`) when enabling or disabling logging as a whole. Panics are now always printed to the logging output. A new page about logging has been added to the book. ([#2008](https://github.com/ariel-os/ariel-os/pull/2008)) ([#2013](https://github.com/ariel-os/ariel-os/pull/2013)) ([#2030](https://github.com/ariel-os/ariel-os/pull/2030))

## Change Checklist


bergzand (approved):
LGTM. Would be perfect if you could give the move of the Async Executors chapter it's own commit.

PR #2035, opened 2026-04-22
fix(ci): update time of files changed since last cache build

nponsard:
# Description

Some of the files going through the CI may have a commit date earlier than the date the cache was built, this made cargo skip building more files than was intended, keeping outdated information.

This PR changes the file time attribution to force all files that are different from the cached commit to be considered as new (using file modification time).

## Testing

This will be a bit tricky to test. The cache needs to be built with the new `/home/runner/metadata` directory (containing an easily accessible git ref to compare to). 

Then create commits that are older than the cache build time modifying some core libraries (e.g. `ariel-os-embassy`) and a commit newer than the cache build time that modifies an example to use the modified core library. 

No build error should occur.

Failing CI reproducing the bug: https://github.com/ariel-os/ariel-os/actions/runs/24767820546/job/72466372510?pr=2035
After manually adding the commit hash in `/home/runner/metadata/cache-ref` done in commit 9e6200dbf1052e206fadbfaf093e0f28ac520d77, the build succeeds: https://github.com/ariel-os/ariel-os/actions/runs/24768016545/job/72466943707?pr=2035

## Changelog Entry

Fixed an issue where the CI was using cached files instead of the files in the PR. 

## Change Checklist

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -95,27 +95,36 @@ jobs:
         if: steps.clear-cache.outputs.result == 'true'
         run: echo "FALLBACK_KEY=none" >> "$GITHUB_ENV"
 
-      # Allows to reuse more cache
-      - name: Touch file time to commits time
-        if: steps.should-skip.outputs.result != 'true'
-        # https://stackoverflow.com/questions/21735435/git-clone-changes-file-modification-time
-        run: |
-          git ls-tree -r --name-only HEAD | while read filename; do
-            unixtime=$(git log -1 --format="%at" -- "${filename}")
-            touchtime=$(date -d @$unixtime +'%Y%m%d%H%M.%S')
-            touch -t ${touchtime} "${filename}"
-          done
-
       - name: Cache with post-job cleanup
         if: steps.should-skip.outputs.result != 'true'
         uses: gacts/run-and-post-run@81b6ce503cde93862cec047c54652e45c5dca991 # v1
         with:
           run: |
-            cache-thing pull -f build -f /home/runner/.cargo/git -f /home/runner/.cargo/registry --prefix global --fallback-key ${FALLBACK_KEY} || true
+            cache-thing pull -f /home/runner/metadata -f build -f /home/runner/.cargo/git -f /home/runner/.cargo/registry --prefix global --fallback-key ${FALLBACK_KEY} || true
 
           post: |
             export  RUST_LOG=trace
             cache-thing clean --prefix global || true
+
+      # Allows to reuse more cache
+      - name: Touch file time to commits time
+        if: steps.should-skip.outputs.result != 'true'
+        # https://stackoverflow.com/questions/21735435/git-clone-changes-file-modification-time
+        run: |
+          if [ -f /home/runner/metadata/cache-ref ]; then
+            git ls-tree -r --name-only HEAD | while read filename; do
+              unixtime=$(git log -1 --format="%at" -- "${filename}")
+              touchtime=$(date -d @$unixtime +'%Y%m%d%H%M.%S')
+              touch -t ${touchtime} "${filename}"
+            done
+
+            # Treat all files modified after the cached commit as "changed", by touching them.
+            git diff --name-only $(cat /home/runner/metadata/cache-ref) | while read filename; do
+              touch "${filename}"
+            done
+          else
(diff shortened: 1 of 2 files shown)


.github/workflows/build-cache.yml
```diff
@@ -106,16 +117,10 @@ jobs:
...
-        if: steps.should-skip.outputs.result != 'true'
-        uses: gacts/run-and-post-run@81b6ce503cde93862cec047c54652e45c5dca991 # v1
-        with:
-          run: |
-            cache-thing pull -f build -f /home/runner/.cargo/git -f /home/runner/.cargo/registry --prefix global --fallback-key ${FALLBACK_KEY} || true
-
-          post: |
-            export  RUST_LOG=trace
-            cache-thing clean --prefix global || true
+          # make all files changed since the cached commit considered new
```
kaspar030:
```suggestion
          # Treat all files modified after the cached commit as "changed", by touching them.
```

.github/workflows/build-cache.yml
```diff
@@ -95,6 +95,17 @@ jobs:
         if: steps.clear-cache.outputs.result == 'true'
         run: echo "FALLBACK_KEY=none" >> "$GITHUB_ENV"
 
+      - name: Cache with post-job cleanup
+        if: steps.should-skip.outputs.result != 'true'
+        uses: gacts/run-and-post-run@81b6ce503cde93862cec047c54652e45c5dca991 # v1
+        with:
+          run: |
+            cache-thing pull -f /home/runner/metadata -f build -f /home/runner/.cargo/git -f /home/runner/.cargo/registry --prefix global --fallback-key ${FALLBACK_KEY} || true
```
kaspar030, line 103: Ah, so `-f /home/runner/metadata` makes that folder also cached. Nifty.

nponsard: Yeah, I had to change it's position so the script changing the file time would have access to it.

.github/workflows/build-cache.yml
```diff
@@ -106,16 +117,10 @@ jobs:
...
-        uses: gacts/run-and-post-run@81b6ce503cde93862cec047c54652e45c5dca991 # v1
-        with:
-          run: |
-            cache-thing pull -f build -f /home/runner/.cargo/git -f /home/runner/.cargo/registry --prefix global --fallback-key ${FALLBACK_KEY} || true
-
-          post: |
-            export  RUST_LOG=trace
-            cache-thing clean --prefix global || true
+          # Treat all files modified after the cached commit as "changed", by touching them.
+          git diff --name-only $(cat /home/runner/metadata/cache-ref) | while read filename; do
```
kaspar030: Should this handle the case that file (`cache-ref`) does not exist?

nponsard: right now if the file doesn't exist the error is silently ignored (see https://github.com/ariel-os/ariel-os/actions/runs/24767546841/job/72465402535), I don't know if a fallback is better.

kaspar030: hmok, so if there is no `cache-ref`, it errors silently and shows the broken behavior as before (which took weeks to notice, so not that bad).
What's the behavior when there is no cache? It starts from scratch, right? That would be fine.
Otherwise, it might be better to just error out if that file does not exist?

kaspar030 (approved):
LGTM!

Conceptually fine. Merge if you are happy with testing.

.github/workflows/build-cache.yml
nponsard: Yes with no cache at all it starts from scratch, this would be quite slow.

I don't like the idea of putting a hard requirement on cache, but it could also work, we would have to merge this PR with `ci-build:skip` then force the generation of a new cache.

kaspar030: I agree slow is better than not working. Let's go with it as is.

nponsard: Oh I may have an alternative that doesn't need to do a clean build if cache-ref is not present.

I could skip the whole step, that would make all git-tracked files be more recent than the cache. This would rebuild more than needed but never less than what really changed.

kaspar030: Lets go with something simple!

nponsard: added the check, should be good now.

kaspar030 (approved):

PR #2036, opened 2026-04-23
docs(book): document logging transports

ROMemories:
# Description

Following #2030, this documents the logging transports in the Logging page introduced by #2034. This new section is extracted from #1987. I think moving this section to the Logging page simply makes more sense, and helps emphasize the orthogonality to the debug console.

## Issues/PRs References

- Depends on #2030

## Changelog Entry

New laze modules have been introduced to select the transport used by logging: `logging-over-debug-channel`, `logging-over-usb`, and `logging-over-uart`. Not every transport is available on every MCU and board. Documentation about logging transports has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -76,7 +76,50 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
 
 Ariel OS's logger for `log` supports configuring the log level globally, but does not currently support per-crate filtering.
 
+## Logging Transports
+
+Logging can use various transports, but currently only one can be used at a time.
+The table below presents those supported in Ariel OS and which hardware and host tool are required:
+
+| Logging transport                        | Supported               | laze module                  | Required hardware                                                                         | Required host tool              |
+| ---------------------------------------- | :---------------------: | ---------------------------- | ----------------------------------------------------------------------------------------- | ------------------------------- |
+| Debug channel                            | Available on all chips  | `logging-over-debug-channel` | Debug probe attached to the debug interface                                               | Debug channel-enabled host tool |
+| [USB CDC-ACM][usb-cdc-acm-glossary-book] | On ESP32 MCUs only      | `logging-over-usb`           | USB cable attached to the user USB port                                                   | Serial monitor                  |
+| [UART][uart-glossary-book]               | On ESP32 MCUs only      | `logging-over-uart`          | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
+
+> [!IMPORTANT]
+> When using [`defmt` as logging facade](#defmt), a `defmt`-enabled host tool must be used so that logs are rendered correctly, as `defmt` uses its own encoding on the wire.
+> probe-rs and `espflash` both support `defmt`'s encoding transparently.
+>
+> When a separate serial monitor is needed, [`defmt-print`][defmt-print-cratesio] can be used as `defmt`-enabled serial monitor.
+> If this is not possible, `defmt` should be disabled and [`log`](#log) used instead as logging facade.
+
+<!-- TODO: link to "debug interface" when the page exists -->
+> [!TIP]
+> When a logging transport other than the debug channel is enabled, logging can still be used when the debug channel is disabled either in software (by disabling the `logging-over-debug-channel` laze module) or in hardware when the debug interface itself is disabled.
+> This means that logging can still be used in production, even if the debug interface has been disabled.
+>
+> If this is unwanted, logging can be disabled altogether by disabling the [`logging`](#logging) laze module.
+
+> [!NOTE]
+> Future plans for logging facilities:
+>
+> - Other logging transports will later be supported, including UART and USB CDC-ACM on non-ESP32 devices.
+> - Using multiple transports at the same time may be supported in the future.
+
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs.
+Currently, the firmware automatically switches at runtime between using USB CDC-ACM or UART as logging transport.
+
+> [!WARNING]
+> This is likely to change in the future, and it may become necessary to select a specific laze module to choose which logging transport to enable and use.
+
 [defmt]: https://github.com/knurling-rs/defmt
 [defmt documentation]: https://defmt.ferrous-systems.com/
 [log]: https://github.com/rust-lang/log
 [laze-modules-book]: ./build-system.md#laze-modules
+[usb-cdc-acm-glossary-book]: ./glossary.md#usb-cdc-acm
+[uart-glossary-book]: ./glossary.md#uart
+[log-mod-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/log/index.html
+[defmt-print-cratesio]: https://crates.io/crates/defmt-print
+[debug-console-debug-console-book]: ./debug-console.md#debug-console
+[espflah-cratesio]: https://crates.io/crates/espflash


book/src/logging.md
```diff
@@ -71,7 +71,50 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
...
+| ---------------------------------------- | ------------------------------------------------------------- | :--------------------------: | --------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- | ------------------------------ |
+| Debug output                             | Prints logs on the [debug output](#debug-output-transports).  | ✅                           | Enabled with the `logging-over-debug-output` laze module                    | [Debug probe](#debug-interfaces-protocols-and-probes) attached to the [debug interface](#debug-interfaces-protocols-and-probes) | Debug output-enabled host tool |
+| [USB CDC-ACM][usb-cdc-acm-glossary-book] | Prints logs through USB CDC-ACM.                              | Currently on ESP32 MCUs only | Enabled by the `logging-over-usb` laze module, enabled by default on ESP32  | USB cable attached to the user USB port                                                                                         | Serial monitor                 |
```
ROMemories: We may not want to document the `logging-over-usb` and `logging-over-uart` laze modules just yet, as #2030 does not currently prevent them from being selected on hardware where they are not actually available.

book/src/logging.md
```diff
@@ -71,7 +71,52 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
...
+> This means that logging can still be used in production, even if the debug interface has been disabled.
+>
+> If this is unwanted, logging can be disabled altogether by disabling the [`logging`](#logging) laze module.
+
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs.
+Currently, the firmware automatically switches at runtime between using USB CDC-ACM or UART as logging transport.
+
+> [!WARNING]
+> This is likely to change in the future, and it may become necessary to select a specific laze module to choose which logging transport to enable and use.
+
```
ROMemories, line 115: That's quite a few admonitions back to back, but half of them should be relatively temporary, so I hope this is fine enough.

book/src/logging.md
ROMemories: As #2030 now actually limits their availability, in a ad-hoc manner for now (#2061 should partially address that), these modules are now documented.

book/src/logging.md
```diff
@@ -76,7 +71,50 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
...
+<!-- TODO: link to "debug interface" when the page exists -->
+> [!TIP]
+> When a logging transport other than the debug channel is enabled, logging can still be used when the debug channel is disabled either in software (by disabling the `logging-over-debug-channel` laze module) or in hardware when the debug interface itself is disabled.
+> This means that logging can still be used in production, even if the debug interface has been disabled.
+>
+> If this is unwanted, logging can be disabled altogether by disabling the [`logging`](#logging) laze module.
+
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs.
+Currently, the firmware automatically switches at runtime between using USB CDC-ACM or UART as logging transport.
```
ROMemories, line 111: This documents the current behavior; hopefully #2039 should fix that.

book/src/logging.md
```diff
@@ -76,7 +71,50 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
 
 Ariel OS's logger for `log` supports configuring the log level globally, but does not currently support per-crate filtering.
 
+## Logging Transports
+
+Logging can use multiple transports; the table below presents those supported in Ariel OS and which hardware and host tool are required:
```
bergzand: This needs clarification here whether multiple at the same time or one of out multiple is currently supported.

book/src/logging.md
```diff
@@ -76,7 +71,50 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
...
+| Debug channel                            | Prints logs on the debug channel output.                      | ✅                           | Enabled with the `logging-over-debug-channel` laze module                              | Debug probe attached to the debug interface                                               | Debug channel-enabled host tool |
+| [USB CDC-ACM][usb-cdc-acm-glossary-book] | Prints logs through USB CDC-ACM.                              | Currently on ESP32 MCUs only | Enabled with the `logging-over-usb` laze module; currently enabled by default on ESP32 | USB cable attached to the user USB port                                                   | Serial monitor                  |
+| [UART][uart-glossary-book]               | Prints logs over UART.                                        | Currently on ESP32 MCUs only | Enabled with the `logging-over-uart`; currently enabled by default on ESP32            | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
```
bergzand: These descriptions do not add much as they all are along the lines of 'print over x' where x is the transport name.

My suggestion here would be to either remove the column or use it to clarify the module:

```suggestion
| Debug channel                            | Prints logs on the debug channel output.                      | ✅                           | Enabled with the `logging-over-debug-channel` laze module                              | Debug probe attached to the debug interface                                               | Debug channel-enabled host tool |
| [USB CDC-ACM][usb-cdc-acm-glossary-book] | prints logs over the user USB port.                              | Currently on ESP32 MCUs only | Enabled with the `logging-over-usb` laze module; currently enabled by default on ESP32 | USB cable attached to the user USB port                                                   | Serial monitor                  |
| [UART][uart-glossary-book]               | Prints logs over the UART peripheral.                                        | Currently on ESP32 MCUs only | Enabled with the `logging-over-uart`; currently enabled by default on ESP32            | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
```

bergzand: The check mark in the 'Supported' column is not immediately clear. I assume it means that this is supported on all boards / MCUs, but this is not stated anywhere. At this point I would just write out 'all platforms' here.

```suggestion
| Debug channel                            | Prints logs on the debug channel output.                      | All boards                   | Enabled with the `logging-over-debug-channel` laze module                              | Debug probe attached to the debug interface                                               | Debug channel-enabled host tool |
| [USB CDC-ACM][usb-cdc-acm-glossary-book] | Prints logs through USB CDC-ACM.                              | ESP32 MCUs only | Enabled with the `logging-over-usb` laze module; currently enabled by default on ESP32 | USB cable attached to the user USB port                                                   | Serial monitor                  |
| [UART][uart-glossary-book]               | Prints logs over UART.                                        | ESP32 MCUs only | Enabled with the `logging-over-uart`; currently enabled by default on ESP32            | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
```

bergzand: Does it make sense to rename the 'How to enable' column to 'Laze module' and only mention the module name there? The default for esp32 could be moved outside the table.

bergzand (changes requested):
The current text looks to me like it suffers from the expected future changes. This makes it harder than necessary for readers to grasp what they can expect.

My suggestion is: 
1. Document the current supported state, and remove most of the 'currently' and 'in the future' remarks.
2. Add a subsection to document the future plans. Alternatively this could also be in the roadmap or even a separate issue linked from this document.

This also solves the issue of the many back-to-back admonitions as a number of them describe future behavior.

book/src/logging.md
ROMemories: I've changed "multiple" to "various" to avoid confusion, and expanded the sentence a bit.

book/src/logging.md
ROMemories: > These descriptions do not add much

Agreed, they used to be less clear; I've removed the column entirely, this only makes the table easier to read.

ROMemories: I've removed the checkmark and added "Available on all chips" instead. This is a bit complicated because the debug interface might not be accessible on all *boards* (e.g., the Ulanzi TC001), but we don't actually restrict it on these, so this reflect the current behavior.

ROMemories: I renamed and simplified the column. I've also completely removed the mention of how defaults work on ESP32 because this will be better in #2039 (which is almost ready) anyway.

ROMemories:
> The current text looks to me like it suffers from the expected future changes. This makes it harder than necessary for readers to grasp what they can expect.

I've reworked this a bit, I think having the future plans mentioned on this page is pretty important, as this helps not only clarify what should be supported in the future but also make it clear that the current state is not intrinsic to the laze modules' definitions (especially for `logging-over-uart`, `logging-over-usb`).

This section is also updated in #2039.

book/src/logging.md
bergzand: > Available on all chips except where it is not supported

:grin:

bergzand:
> I've reworked this a bit, I think having the future plans mentioned on this page is pretty important, as this helps not only clarify what should be supported in the future but also make it clear that the current state is not intrinsic to the laze modules' definitions (especially for `logging-over-uart`, `logging-over-usb`).

Agreed. Your change of grouping future plans in an admonition is fine imho.

> This section is also updated in #2039.

:+1:

bergzand (approved):
Ack, please squash and lgtm!

PR #2038, opened 2026-04-23
docs(gpio): explain how to re-use a pin

nponsard:
# Description

This edits the documentation of the GPIO module to update the note about re-using a GPIO pin and explaining how to do it.

## Testing

Documentation builds

```
laze -C examples/gpio-reuse/ build -b nrf52840dk       
```

## Issues/PRs References

- Follow-up of #2027 

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/gpio.rs
+++ b/src/ariel-os-hal/src/gpio.rs
@@ -2,10 +2,9 @@
 //!
 //! # Note
 //!
-//! This API does not currently provide a way of using the same GPIO pin as an input and an output
-//! alternatively.
-//! If you have a use case for this, especially if this is not regarding bit-banging, please open
-//! an issue on our repository.
+//! To use the same GPIO pin as an input and an output alternatively, you can use `.reborrow()` on the HAL pin type to give an instance of the pin peripheral with a shorter lifetime (a "child" pin instance) to [`Output::new()`] and [`Input::new()`].
+//!
+//! Make sure to drop both the driver and the "child" pin instance before using the pin in another driver as some HALs require it while others don't enforce it.
 #![allow(missing_docs)]
 
 use core::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


src/ariel-os-hal/src/gpio.rs
ROMemories, line 79: I think we need the same thing here.

src/ariel-os-hal/src/gpio.rs
```diff
@@ -2,10 +2,9 @@
 //!
 //! # Note
 //!
-//! This API does not currently provide a way of using the same GPIO pin as an input and an output
-//! alternatively.
-//! If you have a use case for this, especially if this is not regarding bit-banging, please open
-//! an issue on our repository.
+//! To use the same GPIO pin as an input and an output alternatively, you can use `.reborrow()` on the HAL pin type to give an instance of the pin peripheral with a shorter lifetime to [`Output::new()`] and [`Input::new()`].
+//!
+//! Make sure to drop both the driver and the "child" pin instance before using the pin in another driver as som HALs require it and others don't enforce it.
```
ROMemories:
```suggestion
//! Make sure to drop both the driver and the "child" pin instance before using the pin in another driver as some HALs require it while others don't enforce it.
```

ROMemories: Giving this some more thought I think "underlying" would be better than "child" (here and in the other doc comments). "Child" would refer to some inheritance relationship to me.

nponsard: For me "underlying" doesn't work, I used the "child" term as it's [used in embassy](https://docs.embassy.dev/embassy-nrf/0.9.0/nrf9160-s/struct.Peri.html#method.reborrow).

src/ariel-os-hal/src/gpio.rs
```diff
@@ -2,10 +2,9 @@
 //!
 //! # Note
 //!
-//! This API does not currently provide a way of using the same GPIO pin as an input and an output
-//! alternatively.
-//! If you have a use case for this, especially if this is not regarding bit-banging, please open
-//! an issue on our repository.
+//! To use the same GPIO pin as an input and an output alternatively, you can use `.reborrow()` on the HAL pin type to give an instance of the pin peripheral with a shorter lifetime to [`Output::new()`] and [`Input::new()`].
```
ROMemories: Ah I misread, the "child" pin instance is the instance with a shorter lifetime. This could be introduced this way I think:
```suggestion
//! To use the same GPIO pin as an input and an output alternatively, you can use `.reborrow()` on the HAL pin type to give an instance of the pin peripheral with a shorter lifetime (a "child" pin instance) to [`Output::new()`] and [`Input::new()`].
```

src/ariel-os-hal/src/gpio.rs
ROMemories: Indeed, see [my new suggestion](https://github.com/ariel-os/ariel-os/pull/2038#discussion_r3131174391).

ROMemories (approved):
Thanks for updating this!

PR #2039, opened 2026-04-23
refactor(logging-esp)!: enable each `esp-println` output individually

ROMemories:
# Description

This leverages the `logging-over-usb` and `logging-over-uart` laze modules introduced by #2030 to only enable the required output of `esp-println` to avoid monopolizing both the JTAG+USB CDC-ACM USB peripheral and UART peripheral all the time and (slightly) reduce the size of the binary. This also arguably makes it less surprising both during development and production, as at least one of the outputs could unknowingly stay enabled.

## Future work

- `esp-println` should later be replaced with custom implementations using the drivers available in `esp-hal`. This will (a) fix potential issues as `esp-println` just steals peripherals and (b) remove the dependency on `esp-println` which [is expected to be eventually archived](https://redirect.github.com/esp-rs/esp-hal/issues/4666#issuecomment-4353684721).
- Leveraging #1833, we could add a note to the relevant board pages to explain how to use the available USB port(s) (especially for boards with two of them, like the Espressif devkits).

## Testing

The `log` example can be tested on ESP32 devkits with two USB ports: one labeled "USB" and the other labeled "UART". Using the "USB"-labeled port doesn't require any change, as `logging-over-usb` is selected by default when `espflash` is used (which is the default), but using the "UART"-labeled port now requires selecting `logging-over-uart` explicitly.

Undocumented support for logging-over-UART still works on other boards (e.g., nRF52840-DK, STM32U083C-DK)

## Issues/PRs References

- Depends on #2030
- Depends on #2036

## Changelog Entry

(ESP32) As the runtime auto-selection of the logging transport has been disabled on these MCUs, it must now be selected at build time using either `logging-over-usb` or `logging-over-uart` (`logging-over-usb` is the default when available).

## Change Checklist

### DIFF
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -87,7 +87,8 @@ The table below presents those supported in Ariel OS and which hardware and hos
 | [USB CDC-ACM][usb-cdc-acm-glossary-book] | On ESP32 MCUs only      | `logging-over-usb`           | USB cable attached to the user USB port                                                   | Serial monitor                  |
 | [UART][uart-glossary-book]               | On ESP32 MCUs only      | `logging-over-uart`          | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
 
-On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs.
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs, whose usage is determined by the `espflash` [laze module][laze-modules-book].
+When `espflash` is selected at the time of compilation, `logging-over-debug-channel` is not enabled and one of the other available logging transports is used instead.
 
 > [!IMPORTANT]
 > When using [`defmt` as logging facade](#defmt), a `defmt`-enabled host tool must be used so that logs are rendered correctly, as `defmt` uses its own encoding on the wire.
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -87,6 +87,8 @@ The table below presents those supported in Ariel OS and which hardware and hos
 | [USB CDC-ACM][usb-cdc-acm-glossary-book] | On ESP32 MCUs only      | `logging-over-usb`           | USB cable attached to the user USB port                                                   | Serial monitor                  |
 | [UART][uart-glossary-book]               | On ESP32 MCUs only      | `logging-over-uart`          | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
 
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs.
+
 > [!IMPORTANT]
 > When using [`defmt` as logging facade](#defmt), a `defmt`-enabled host tool must be used so that logs are rendered correctly, as `defmt` uses its own encoding on the wire.
 > probe-rs and `espflash` both support `defmt`'s encoding transparently.
@@ -106,12 +108,6 @@ The table below presents those supported in Ariel OS and which hardware and hos
 >
 > - Other logging transports will later be supported, including UART and USB CDC-ACM on non-ESP32 devices.
 > - Using multiple transports at the same time may be supported in the future.
-
-On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs.
-Currently, the firmware automatically switches at runtime between using USB CDC-ACM or UART as logging transport.
-
-> [!WARNING]
-> This is likely to change in the future, and it may become necessary to select a specific laze module to choose which logging transport to enable and use.
 
 [defmt]: https://github.com/knurling-rs/defmt
 [defmt documentation]: https://defmt.ferrous-systems.com/
--- a/src/ariel-os-log/Cargo.toml
+++ b/src/ariel-os-log/Cargo.toml
@@ -8,7 +8,11 @@ license.workspace = true
 
 [package.metadata.feature-groups]
 logging-facade.xor = { features = ["defmt", "log"] }
-logging-transport.xor = { features = ["defmt-rtt", "esp-println", "uart"] }
+logging-transport.xor = { features = [
+  "defmt-rtt",
+  "logging-over-uart",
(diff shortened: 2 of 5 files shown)


src/ariel-os-log/Cargo.toml
```diff
@@ -48,12 +47,15 @@ defmt = ["dep:defmt", "esp-println?/defmt-espflash"]
...
+esp-println = ["dep:esp-println"]
+
 # Logging transports.
 # `defmt-rtt` forces the logging transport to be the debug output, so it has to
 # go through this crate.
 defmt-rtt = ["ariel-os-debug/defmt-rtt"]
-esp-println = ["dep:esp-println"]
-uart = ["dep:embassy-sync"]
+logging-over-usb = ["esp-println?/jtag-serial"]
+uart = ["dep:embassy-sync", "esp-println?/uart"]
```
ROMemories: I hope we can later use our regular logging-over-UART mechanism (from `ariel-os-embassy`) instead of relying on `esp-println` for it.

book/src/logging.md
```diff
@@ -76,7 +76,47 @@ Note: On Cortex-M devices, the order of `ariel_os::log::println!()` output and
...
+
+| Logging transport                        | Supported               | laze module                  | Required hardware                                                                         | Required host tool              |
+| ---------------------------------------- | :---------------------: | ---------------------------- | ----------------------------------------------------------------------------------------- | ------------------------------- |
+| Debug channel                            | Available on all chips  | `logging-over-debug-channel` | Debug probe attached to the debug interface                                               | Debug channel-enabled host tool |
+| [USB CDC-ACM][usb-cdc-acm-glossary-book] | On ESP32 MCUs only      | `logging-over-usb`           | USB cable attached to the user USB port                                                   | Serial monitor                  |
+| [UART][uart-glossary-book]               | On ESP32 MCUs only      | `logging-over-uart`          | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
```
ROMemories, line 88: On ESP32-C3 and ESP32-S3 (but not ESP32-C6), `esp-println` *also* prints to USB CDC-ACM when UART is enabled. This is a [known bug that seems to stem from ROM functions](https://redirect.github.com/esp-rs/esp-hal/issues/4510) (which I can reproduce on both MCUs), that we may need to document, but this may be better in a follow-up on the chip and board pages.

src/ariel-os/Cargo.toml
```diff
@@ -221,7 +221,9 @@ no-boards = [
 ]
 
 # These are selected by laze
-debug-uart = ["ariel-os-embassy/debug-uart", "ariel-os-log/uart"]
+debug-uart = ["ariel-os-embassy/debug-uart", "ariel-os-log/debug-uart"]
```
ROMemories, line 224: The ill-named `debug-uart` is really a temporary thing (as much as the laze module of the same name): the hope is to later replace `esp-println` with a custom implementation using a regular UART driver from `esp-hal` on ESP, and collapse `debug-uart` into `logging-into-uart` (fortunately `debug-uart` is completely internal).

src/ariel-os-log/Cargo.toml
```diff
@@ -22,7 +26,6 @@ log = { workspace = true, optional = true }
 
 [target.'cfg(context = "esp")'.dependencies]
 esp-println = { workspace = true, optional = true, features = [
-  "auto",
```
ROMemories, line 25: This is what enabled auto-selection of the transport at runtime.

book/src/logging.md
```diff
@@ -87,6 +87,9 @@ The table below presents those supported in Ariel OS and which hardware and hos
 | [USB CDC-ACM][usb-cdc-acm-glossary-book] | On ESP32 MCUs only      | `logging-over-usb`           | USB cable attached to the user USB port                                                   | Serial monitor                  |
 | [UART][uart-glossary-book]               | On ESP32 MCUs only      | `logging-over-uart`          | USB ⟷ UART adapter attached to the supported UART pins (may already be part of the board) | Serial monitor                  |
 
+On ESP32 devices, Ariel OS uses [`espflash`][espflah-cratesio] by default to obtain and print logs, whose usage is determined by the `espflash` [laze module][laze-modules-book].
+When `espflash` is selected at the time of compilation, `logging-over-debug-channel` is not enabled and one of the other available logging transports is used instead.
```
bergzand, line 91: Does the selection of the logging transport (`logging-over-uart` vs `logging-over-usb`) depend on what is available on the exact esp32 board? Does that need clarification here or somewhere else?

bergzand (approved):

bergzand:
Tested on my `espressif-esp32-c6-devkitc-1`, works!

book/src/logging.md
ROMemories: > Does the selection of the logging transport (`logging-over-uart` vs `logging-over-usb`) depend on what is available on the exact esp32 board?

Doesn't depend on the board currently (but I don't think there is a single board that we support that has a USB-enabled ESP32 but instead *only* exposes the UART, that would be a bit odd), but you won't be able to select `logging-over-usb` if *the MCU* doesn't have a USB CDC-ACM/JTAG USB peripheral. This is refined in #2061.

ROMemories: > I don't think there is a single board that we support that has a USB-enabled ESP32 but instead _only_ exposes the UART

Actually that's not true, the [Heltec WiFi LoRa 32 V3](https://web.archive.org/web/20250807184214/https://heltec.org/project/wifi-lora-32-v3/), which we do support, has only one USB port, which is connected to a USB ⟷ UART adapter and the USB peripheral is *not* exposed, despite using an ESP32-S3. This should be addressed by #2061 anyway.

bergzand: > Actually that's not true, the [Heltec WiFi LoRa 32 V3](https://web.archive.org/web/20250807184214/https://heltec.org/project/wifi-lora-32-v3/), which we do support, has only one USB port, which is connected to a USB ⟷ UART adapter and the USB peripheral is _not_ exposed, despite using an ESP32-S3. This should be addressed by #2061 anyway.

This sounds cursed and I hope this stays as an exception.

PR #2047, opened 2026-04-24
ci: explictly document crates that are otherwise documented as 'modules'

anlavandier:
# Description

This PR explictly documents the various Ariel crates that are direct dependencies of Ariel OS (at least in the version compiled for documentation). Doing this correctly generates an `all.html` file for each crates listing of the items. Otherwise a broken link is generated. 

## Testing
See the impact on link status on #2028 

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -68,6 +68,17 @@ jobs:
               -p ariel-os \
               -p ariel-os-sensors \
               -p coapcore \
+              -p ariel-os-bench \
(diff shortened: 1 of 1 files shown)


anlavandier:
This did not do what I expected.. I need to investigate

ROMemories (changes requested):
We'll want to do this, but not before we add documentation to clarify which are considered internal.

anlavandier:
I was more trying to fix some links but it turns out that it didn't. I get that the problem is the `#[doc(inline)]` on the `use ariel_os_x` items

PR #2051, opened 2026-04-27
fix(net): require that IPv4 or IPv6 be enabled for networking

ROMemories:
# Description

The `ipv6` laze modules was introduced by #915, but it wasn't known at the at time how to express an "at least one" relation between the `ipv4` and `ipv6` laze modules. The build would fail if networking was enabled but `ipv4` (otherwise enabled by default) was manually disabled, as `embassy-net` requires at least one of them. This PR encodes that constraint in the laze configuration so the build can fail earlier.

## Testing

The `info-modules` laze task can be used to print the enabled laze modules:

```sh
laze -C examples/udp-echo/ build -b rpi-pico info-modules
```

```sh
laze -C examples/udp-echo/ build -d ipv4 -b rpi-pico info-modules
```

(This build should now fail.)

```sh
laze -C examples/udp-echo/ build -s ipv6 -b rpi-pico info-modules
```

(Enables both IPv4 and IPv6.)

```sh
laze -C examples/udp-echo/ build -d ipv4 -s ipv6 -b rpi-pico info-modules
```

(Enables only IPv6.)

## Changelog Entry

I don't think this deserves an entry as the build would have failed anyway.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1227,7 +1227,14 @@ modules:
   - name: network
     selects:
       - network_device
+      # Enables IPv4 by default.
       - ?ipv4
+      # Requires at least one of {`ipv4`, `ipv6`}.
+      # This works only if `ipv4` or `ipv6` has been selected before this
+      # `network` module.
+      - ?no-ipv4-ipv6
+      - no-ipv4-ipv6:
+          - ipv4-or-ipv6-required
 
   - name: network-config-ipv4-default
     help: use default network configuration method
@@ -1272,10 +1279,16 @@ modules:
         FEATURES:
           - ariel-os/network-config-ipv4-static
 
+  # Disabled by modules that provide one of these.
+  - name: no-ipv4-ipv6
+    help: private
+
   - name: ipv4
     selects:
       - network
       - network-config-ipv4-default
+    disables:
+      - no-ipv4-ipv6
     env:
       global:
         FEATURES:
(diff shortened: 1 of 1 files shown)


laze-project.yml
```diff
@@ -1227,7 +1227,14 @@ modules:
   - name: network
     selects:
       - network_device
+      # Enables IPv4 by default.
       - ?ipv4
+      # Requires at least one of {`ipv4`, `ipv6`}.
+      # This works only if `ipv4` or `ipv6` has been selected before this
+      # `network` module.
```
ROMemories, line 1234: Some out-of-band discussion with @kaspar030 showed that this technique only works if `ipv4` or `ipv6` are selected before `network` (as it happens when selecting these modules on command line, unless purposefully selecting `network` *before* them). I think this is still useful, so I've added a comment about this.

bergzand (approved):
Tested, LGTM!

@kaspar030 we can add this to the list of obscure Laze error messages:

```console
$ laze -C examples/udp-echo/ build -d ipv4 -b rpi-pico info-modules
laze: error: no matching target for task "info-modules" found.
```

kaspar030:
> ```
> laze -C examples/udp-echo/ build -d ipv4 -b rpi-pico info-modules
> ```

works for me on main? *edit* ah, if it doesn't build with ipv4 enabled, that target is unavailable, I see.

PR #2053, opened 2026-04-27
fix(ci): touch only files that are present

nponsard:
# Description

The step "Touch file time to commits time" tries to touch all modified files between the cached commit and the commit in the PR, if a folder is not preset in the checked out tree but files inside it have been modified between the two commits, touch will throw an error like in this run: https://github.com/ariel-os/ariel-os/actions/runs/24995583111/job/73191643712 . 

This PR fixes that by first checking if the file exists before touching it.

## Change Checklist


ROMemories (approved):
Hard to actually test but that should definitely help, thanks!

PR #2054, opened 2026-04-28
build: bump MSRV to 1.95

### LINKED ISSUE #1875: Bumping MSRV to 1.95

anlavandier:
# Description

This bumps Ariel's toolchain to 1.95. This introduced [`cfg_select!`](https://doc.rust-lang.org/stable/core/macro.cfg_select.html) which should allow us to get rid of our dependency on `cfg-if`. See #1851 

## Issues/PRs References
- Waiting on an image update PR https://github.com/ariel-os/garm-worker-image to test the build CI on our machines https://github.com/ariel-os/garm-worker-image/pull/20
- Depends on #2057
- Closes #1875
- Depends on #2059 

## Open Questions
- [ ] Do we also want to bump the nightly ? 

## Changelog Entry

Ariel OS's MSRV is now 1.95.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1991,7 +1991,7 @@ builders:
           - rustup ${CARGO_TOOLCHAIN} target add riscv32imc-unknown-none-elf
           - rustup ${CARGO_TOOLCHAIN} target add riscv32imac-unknown-none-elf
           - rustup ${CARGO_TOOLCHAIN} component add rust-src llvm-tools
-          - if command -v espup >/dev/null; then espup update --toolchain-version 1.94.0.2 --targets=esp32,esp32s2,esp32s3; fi
+          - if command -v espup >/dev/null; then espup update --targets=esp32,esp32s2,esp32s3; fi
 
       update-book:
         build: false
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1014,7 +1014,7 @@ modules:
     env:
       global:
         # this pins our nightly version
-        CARGO_TOOLCHAIN: +nightly-2026-01-20
+        CARGO_TOOLCHAIN: +nightly-2026-04-28
         RUSTFLAGS:
           - --cfg nightly
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -280,7 +280,7 @@ jobs:
         # configure MSRV here:
         uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master
         with:
-          toolchain: 1.94
+          toolchain: 1.95
           targets: riscv32imc-unknown-none-elf,riscv32imac-unknown-none-elf,thumbv6m-none-eabi,thumbv7m-none-eabi,thumbv7em-none-eabi,thumbv7em-none-eabihf,thumbv8m.main-none-eabi,thumbv8m.main-none-eabihf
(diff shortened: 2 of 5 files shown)


ROMemories:
> Do we also want to bump the nightly ?

I think we do; the current version is `nightly-2026-01-20`, but 1.95 was [branched from](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html) main [in february](https://releases.rs/docs/1.95.0/) (the unofficial releases.rs is useful to know that).

laze-project.yml
```diff
@@ -1991,7 +1991,7 @@ builders:
           - rustup ${CARGO_TOOLCHAIN} target add riscv32imc-unknown-none-elf
           - rustup ${CARGO_TOOLCHAIN} target add riscv32imac-unknown-none-elf
           - rustup ${CARGO_TOOLCHAIN} component add rust-src llvm-tools
-          - if command -v espup >/dev/null; then espup update --toolchain-version 1.94.0.2 --targets=esp32,esp32s2,esp32s3; fi
+          - if command -v espup >/dev/null; then espup update --toolchain-version 1.95.0.0 --targets=esp32,esp32s2,esp32s3; fi
```
ROMemories: It seems that the Xtensa toolchain has [had an actual release](https://github.com/esp-rs/rust-build/releases) (not just a pre-release), so we should not have to force the toolchain version anymore (so future versions can automatically be picked up, as happens on other targets); this needs testing though:
```suggestion
          - if command -v espup >/dev/null; then espup update --targets=esp32,esp32s2,esp32s3; fi
```

ROMemories:
I think the tiny refactor to make Clippy happy is better done in a separate PR, to keep this PR focused so it can be used as a reference for the next time we need to bump.

laze-project.yml
ROMemories: Checked locally that this command (without `toolchain-version`) indeed installs the 1.95 version of the Xtensa toolchain.

anlavandier: Done

ROMemories:
I'll take care of updating the `trybuild` tests.

src/ariel-os-macros/tests/ui/spawner/missing_peripherals_param.stderr
```diff
@@ -10,4 +10,4 @@ error: ... because this function has a second parameter
  --> tests/ui/spawner/missing_peripherals_param.rs:9:9
   |
 9 | fn main(_spawner: Spawner, _peripherals: Peripherals) {}
-  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ROMemories: Technically that was a better span, but I don't think that matters much.

ROMemories:
Temporarily removing the `ci-build:full` label until we get the non-Build jobs passing.

ROMemories (approved):
I'm approving the three commits that bump version numbers. This will need to be rebased before landing, but not before the current CI run terminates.

PR #2057, opened 2026-04-29
ci: use a stable toolchain where possible

ROMemories:
# Description

Since #1811 it should be possible for the `cargo-test` job to use a stable toolchain. This is useful for #2054 as the `trybuild` snapshots apparently flip-flop between Rust 1.95 and the current nightly (nightly-2026-04-29).

## Testing

CI does it.

## Change Checklist

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -126,10 +126,6 @@ jobs:
              echo "/home/runner/metadata/cache-ref not found, skipping step"
           fi
 
-      - id: get_toolchain
-        if: steps.should-skip.outputs.result != 'true'
-        run: echo "toolchain=$(scripts/rust-toolchain.sh)" >> $GITHUB_OUTPUT
-
       - name: "Show installed versions"
         run: |
           set -x
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,13 +25,10 @@ jobs:
         with:
           persist-credentials: false
 
-      - id: get_toolchain
-        run: echo "toolchain=$(scripts/rust-toolchain.sh)" >> $GITHUB_OUTPUT
-
       - name: Install toolchain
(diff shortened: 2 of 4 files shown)


.github/workflows/build-cache.yml
```diff
@@ -126,10 +126,6 @@ jobs:
              echo "/home/runner/metadata/cache-ref not found, skipping step"
           fi
 
-      - id: get_toolchain
-        if: steps.should-skip.outputs.result != 'true'
-        run: echo "toolchain=$(scripts/rust-toolchain.sh)" >> $GITHUB_OUTPUT
```
ROMemories, line 131: I think this job wasn't used; should we keep it anyway?

.github/workflows/main.yml
```diff
@@ -25,13 +25,10 @@ jobs:
...
           persist-credentials: false
 
-      - id: get_toolchain
-        run: echo "toolchain=$(scripts/rust-toolchain.sh)" >> $GITHUB_OUTPUT
-
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master
         with:
-          toolchain: ${{ steps.get_toolchain.outputs.toolchain }}
+          toolchain: stable
```
ROMemories, line 31: I think we can safely always use the latest stable here; vetting is still pinned following https://github.com/ariel-os/ariel-os/pull/1917#discussion_r2945497636.

.github/workflows/build-cache.yml
nponsard: Yeah this is not used, I'm fine with deleting it.

.github/workflows/main.yml
nponsard: This should be fine as long as building with MSRV is tested somewhere else.

ROMemories:
After rebasing #2054 on this, this does allow the `cargo-test` job to pass on #2054.

nponsard (approved):
Looks good to me !

PR #2061, opened 2026-04-29
refactor(logging): introduce a `usb-cdc-acm-esp` laze module

ROMemories:
# Description

This introduces the necessary laze module to model the non-generic USB CDC-ACM(/JTAG) USB peripheral found on some (but not all) ESP32s and limits the availability of the `logging-over-usb` module introduced by #2030 accordingly. It is the intent that `logging-over-usb` be eventually implemented using the generic USB peripheral when available (#1805), but this is out of scope.

This also addresses an issue introduced in #2039 (see https://github.com/ariel-os/ariel-os/pull/2039#discussion_r3193850801) on the [Heltec WiFi LoRa 32 V3](https://web.archive.org/web/20250807184214/https://heltec.org/project/wifi-lora-32-v3/), as that board does *not* expose the USB peripheral at all (despite using an ESP32-S3).

For reference:

- The "plain" ESP32 (and the ESP32-C2) have no USB peripheral at all.
- The ESP32-S2 has a generic USB peripheral but no USB CDC-ACM/JTAG USB peripheral.
- The ESP32-S3 has both a generic USB peripheral and a USB CDC-ACM/JTAG USB peripheral.
- The ESP32-C3, (ESP32-C5, ESP32-C61) and ESP32-C6 have a USB CDC-ACM/JTAG USB peripheral but no generic USB peripheral.

(Chips we don't currently support are mentioned in parentheses.)

The USB CDC-ACM/JTAG USB peripheral and the generic USB peripheral cannot be used at the same time on the same USB port (it is possible to add an external PHY and use both peripherals with separate USB ports, but supporting this is not a priority) so this PR disables `logging-over-usb` when USB is otherwise used by the application for now (I hope we can have a composite USB device later).

## How to review this PR

The changes to the SBD files could be moved to a dedicated PR.

## Testing

Should laze-time fail:

```sh
laze -C examples/log/ build -s logging-over-usb -b rpi-pico
laze -C examples/log/ build -s logging-over-usb -b espressif-esp32-devkitc
laze -C examples/log/ build -s logging-over-usb -b espressif-esp32-s2-devkitc-1
laze -C examples/udp-echo/ build -s logging-over-usb -s usb-ethernet -b espressif-esp32-s3-devkitc-1
```

(The last failure is because the USB CDC-ACM/JTAG USB peripheral cannot be used at the same time as the generic USB peripheral over the same USB port.)

Should compile:

```sh
laze -C examples/log/ build -s logging-over-usb -b espressif-esp32-c3-devkit-rust-1
laze -C examples/log/ build -s logging-over-usb -b espressif-esp32-c6-devkitc-1
laze -C examples/log/ build -s logging-over-usb -b espressif-esp32-s3-devkitc-1
```

The following can be used to check that the [Heltec WiFi LoRa 32 V3](https://web.archive.org/web/20250807184214/https://heltec.org/project/wifi-lora-32-v3/) *defaults* to using the UART transport:

```sh
laze -C examples/log/ build -b heltec-wifi-lora-32-v3 info-modules
```

## Issues/PRs References

- Depends on #2030

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -406,6 +406,7 @@ contexts:
       - riscv
     provides:
       - has_ble_esp
+      - has_usb_cdc_acm
       - sw/benchmark
     provides_unique: [c-function-abort]
     env:
@@ -423,6 +424,7 @@ contexts:
       - riscv
     provides:
       - has_ble_esp
+      - has_usb_cdc_acm
       - sw/benchmark
     provides_unique: [c-function-abort]
     env:
@@ -476,6 +478,7 @@ contexts:
       # TODO: re-enable as soon as upstream peripheral drivers become `Send`
       # in combination with multi-core.
       # - has_multi_core_support
+      - has_usb_cdc_acm
       - sw/benchmark
     env:
       RUSTC_TARGET: xtensa-esp32s3-none-elf
@@ -1440,6 +1443,12 @@ modules:
       global:
         FEATURES:
           - ariel-os/usb
+
+  # Provided by ESP32 MCUs with a non-generic USB CDC-ACM peripheral (they may
+  # *also* have a generic USB peripheral).
+  - name: has_usb_cdc_acm
+    selects:
+      - doc-only
 
   # Provided by boards having a USB port connected to a USB peripheral
   # supporting USB CDC-ACM while not being a generic USB peripheral.
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1437,12 +1437,25 @@ modules:
           - ariel-os/nrf91-modem
 
   - name: usb
+    disables:
+      # Currently the generic USB should take priority.
+      - usb-cdc-acm-esp
     selects:
       - hw/usb-device-port
     env:
       global:
         FEATURES:
           - ariel-os/usb
+
+  # Available on ESP32 boards with ESP32 having a non-generic USB CDC-ACM
+  # peripheral (they may *also* have a generic USB peripheral) and a USB port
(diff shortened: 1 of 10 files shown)


bugadani:
> The ESP32-C*s have a USB CDC-ACM/JTAG USB peripheral but no generic USB peripheral.

I'll have to note, that, while you don't support ESP32-C2 yet, it also does NOT have USB Serial/JTAG.

ROMemories:
> I'll have to note, that, while you don't support ESP32-C2 yet, it also does NOT have USB Serial/JTAG.

I had checked the ESP32-C5 and ESP32-C61, which we don't support yet either, but missed the ESP32-C2. The PR doesn't need any changes but I've updated the description, thanks a lot!

laze-project.yml
```diff
@@ -1434,13 +1437,39 @@ modules:
...
+  - name: usb-cdc-acm-esp
+    context:
+      - esp
+    selects:
+      - has_usb_cdc_acm
+      - has_usb_cdc_acm_port
+
+  # Provided by ESP32 MCUs with a non-generic USB CDC-ACM peripheral (they may
+  # *also* have a generic USB peripheral).
+  - name: has_usb_cdc_acm
```
ROMemories, line 1462: This PR introduces a dedicated laze module for the *MCU* peripheral itself, but we don't do the same for generic USB. Is this fine? Should we do the same for generic USB actually?

kaspar030 (approved):

PR #2062, opened 2026-04-30
fix(ci): don't try to write the git ref when we skipped the build

nponsard:
# Description

The build-cache task is failing when there are no changes since the last execution because the directory `/home/runner/metadata` isn't created in this case. 
The solution is to not execute this step when the build has been skipped.

## Change Checklist


ROMemories (approved):
LGTM, should we wait for the ongoing scheduled Build to terminate before landing this?

nponsard:
This change doesn't affect the `Build` CI.

PR #2063, opened 2026-04-30
fix(ci): ignore the PR label on scheduled build

nponsard:
# Description

Ignores the label when the job is triggered by a schedule.

## Change Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,8 +82,8 @@ jobs:
               // Read the environment variable, if none use the self-hosted default.
               let runnerLabel = process.env.BUILD_RUNNER || SELF_HOSTED_RUNNER_LABEL;
 
-              // Override the previous value if we have the corresponding PR label.
-              if (prLabels.includes("ci-host:github")) {
+              // Override the previous value if we have the corresponding PR label (doesn't apply when it's a scheduled job).
+              if (prLabels.includes("ci-host:github") && "${{ github.event_name }}" != "schedule") {
                 runnerLabel = GITHUB_HOSTED_RUNNER_LABEL;
               }
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,9 +50,10 @@ jobs:
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
         env:
           BUILD_RUNNER: ${{ vars.BUILD_RUNNER }}
+          COMMIT_ID: ${{ steps.get-commit-id.outputs.commit_id }}
         with:
-          script: | # zizmor: ignore[template-injection] input is always a commit SHA
-            const commitId = "${{ steps.get-commit-id.outputs.commit_id }}";
+          script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -82,8 +82,8 @@ jobs:
               // Read the environment variable, if none use the self-hosted default.
               let runnerLabel = process.env.BUILD_RUNNER || SELF_HOSTED_RUNNER_LABEL;
 
-              // Override the previous value if we have the corresponding PR label.
-              if (prLabels.includes("ci-host:github")) {
+              // Override the previous value if we have the corresponding PR label (doesn't apply when it's a scheduled job).
+              if (prLabels.includes("ci-host:github")) && "${{ github.event_name }}" != "schedule" {
```
```suggestion
              if (prLabels.includes("ci-host:github") && "${{ github.event_name }}" != "schedule") {
```

ROMemories: This is one of those cases where we should be careful about injections: this is only ok because `event_name` is completely defined by GitHub.

ROMemories (approved):
LGTM after moving the parenthesis.

.github/workflows/build.yml
nponsard: Yeah this isn't Rust :sweat_smile:

nponsard: Zizmor would catch that, right ?

ROMemories: I hope that [`template-injection`](https://docs.zizmor.sh/audits/#template-injection) would indeed trigger otherwise; just wasn't too sure because this is mixed with both YAML and JavaScript in this case.

nponsard: Is 
```
# zizmor: ignore[template-injection] input is always a commit SHA
```
applying to the whole script or only the next line ?

ROMemories: [To the whole string block it seems](https://docs.zizmor.sh/usage/#with-comments) you're right; can you update the zizmor comment?

ROMemories: Actually updating the comment only makes sense if it *would* trigger zizmor.

nponsard: What if we don't need to have a `# zizmor: ignore` ?

.github/workflows/build.yml
```diff
@@ -50,9 +50,10 @@ jobs:
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
         env:
           BUILD_RUNNER: ${{ vars.BUILD_RUNNER }}
+          COMMIT_ID: ${{ steps.get-commit-id.outputs.commit_id }}
         with:
-          script: | # zizmor: ignore[template-injection] input is always a commit SHA
-            const commitId = "${{ steps.get-commit-id.outputs.commit_id }}";
+          script: |
+            const commitId = process.env.COMMIT_ID;
```
ROMemories, line 56: This accesses the env var through Node.js's API (*not* by using template expansion), as recommend by [zizmor](https://docs.zizmor.sh/audits/#remediation_27) 👍

ROMemories (approved):
Even better!

PR #2068, opened 2026-05-01
feat: Move dummy HAL to its own crate

dcz-self:
Now it's possible to use selected dummy devices in a custom board. This is needed for native peripheral emulation.

# Description

Splitting out dummy from hal is necessary if a new board wants to use some. Otherwise it results in a circular import: hal->board->hal::dummy. I want to use dummy devices in the definition of the native board to emulate only some peripherals (GPIO).

## Testing

I ran `laze build -DCARGO_ARGS+='--locked' --global -b native` as far as I could on a stable toolchain.

The general concept (but not exactly the same code) was tested https://github.com/ariel-os/ariel-os/pull/2000

## Issues/PRs References
Will be needed for https://github.com/ariel-os/ariel-os/pull/2000

## Change Checklist


dcz-self:
This is as far as I can make the CI happy. The last 2 checks are about labels which I don't understand and I suspect don't have the permissions to fix.

src/ariel-os-dummy/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+name = "ariel-os-dummy"
+version = "0.4.0"
+edition.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+license.workspace = true
+
+[dependencies]
+ariel-os-embassy-common = { workspace = true }
+embassy-executor = { workspace = true }
```
dcz-self, line 12: I had to move it from optional (under `[target(not(any(...))]`) to obligatory because of this https://github.com/ariel-os/ariel-os/actions/runs/25227241164/job/73973717481?pr=2068

I don't know otherwise what is different. It's the same rule that worked in ariel-os-hal/Cargo.toml

src/ariel-os-hal/Cargo.toml
```diff
@@ -6,6 +6,7 @@ rust-version.workspace = true
 license.workspace = true
 
 [dependencies]
+ariel-os-dummy = { workspace = true }
```
dcz-self, line 9: Not optional. I could see that in the future, someone will want a dummy peripheral while working on some bare metal target, and it's easier if they don't have to modify hal dependencies for that.

kaspar030:
> This is as far as I can make the CI happy.

Oh no, I should have enabled one of the "ci-build::X" labels, otherwise CI is only doing linting. Sorry. Did that now.

dcz-self:
It looks like there's still something wrong with the labels.

dcz-self:
The builds unskipped for some reason.

kaspar030:
> The builds unskipped for some reason.

Sometimes the label check doesn't find the PR (Issue https://github.com/ariel-os/ariel-os/issues/1945). I'm re-starting that check manually.

src/ariel-os-dummy/Cargo.toml
dcz-self: This gave me more trouble, so I made more dependencies obligatory..
I don't really understand under what conditions the CI tries to build stuff, so I am unable to test it.
Meanwhile, having dependencies unconditional works whether they are actually used or not.

I think getting rid of conditional compilation is an additional win here: the test matrix can shrink.

PR #2078, opened 2026-05-06
refactor(rtt-target): remove unused remnants of `defmt` support

ROMemories:
# Description

In #2013, the ability to use `defmt` with `rtt-target` (not RTT in general, specifically the `rtt-target` implementation) was knowingly broken/removed (see https://github.com/ariel-os/ariel-os/pull/2013#discussion_r3084770748). In addition, #2030 set the `help` message of the `rtt-target` laze module to "private" and that laze module was never documented. It is my opinion that there is no reason to keep two implementations of `defmt` over RTT, as `defmt-rtt` is otherwise used (and has been the default since its introduction in #1603), and as using the `rtt-target` implementation increases the binary size by a couple of hundred bytes. This PR therefore proposes to resolve the inconsistency by removing the remnants of that implementation.

Alternatively, the implementation of `defmt`+`rtt-target` could be fixed, and the `rtt-target` laze module documented in the book as an alternative implementation.

## Testing

```sh
laze -C examples/log build -b stm32u083c-dk run
```

(Unchanged by this PR, uses `defmt-rtt`.)

```sh
laze -C examples/log build -s log -b stm32u083c-dk run
```

(Unchanged by this PR, uses `log` and `rtt-target`.)

```sh
laze -C examples/log build -s rtt-target -b stm32u083c-dk run
```

(Forces the *undocumented* `rtt-target` laze module, technically unchanged by this PR, but different from the behavior from 0.4.0: uses `log` and `rtt-target`.)

## Changelog Entry

To me this is not a breaking change (the alternative implementations were never given any visibility), so this doesn't need an entry.

## Change Checklist

### DIFF
--- a/src/ariel-os-debug/src/lib.rs
+++ b/src/ariel-os-debug/src/lib.rs
@@ -21,38 +21,13 @@ mod backend {
 
 #[cfg(feature = "rtt-target")]
 mod backend {
-    #[cfg(feature = "defmt")]
-    pub use defmt::println as debug_output_println;
-
-    #[cfg(not(feature = "defmt"))]
     pub use rtt_target::rprintln as debug_output_println;
 
     #[doc(hidden)]
     pub fn init() {
-        #[cfg(not(feature = "defmt"))]
-        {
-            use rtt_target::ChannelMode::NoBlockTrim;
+        use rtt_target::ChannelMode::NoBlockTrim;
 
-            rtt_target::rtt_init_print!(NoBlockTrim);
-        }
-
-        #[cfg(feature = "defmt")]
-        {
-            use rtt_target::ChannelMode::NoBlockSkip;
-            const DEFMT_BUFFER_SIZE: usize = 1024;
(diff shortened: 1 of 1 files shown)


kaspar030:
I'm not sure we should drop this - it does enable a second non-defmt channel next to regular defmt. When I added `defmt-rtt`, we still had those linker bugs, which somehow didn't affect `defmt-rtt` but did `rtt-target`'s defmt implementation. Apart from increasing the testing now that we're refactoring everything, is this hurting a lot?

ROMemories:
> it does enable a second non-defmt channel next to regular defmt

I'm not sure I'm following here: I think the [`rtt_init!`](https://docs.rs/rtt-target/0.6.1/rtt_target/macro.rtt_init.html) call only creates one channel (which is the `defmt` channel)? But in any case the `rtt-target` laze module is completely undocumented and it doesn't seem necessary when using `defmt`.

kaspar030 (approved):

PR #2079, opened 2026-05-08
feat(boards): Add nRF52840 MDK USB Dongle

SimonIT:
# Description

Added nRF52840 MDK USB Dongle as board. I also added the missing green and blue for nrf52840-mdk.

## Testing

https://wiki.makerdiary.com/nrf52840-mdk-usb-dongle/hardware/

https://wiki.makerdiary.com/nrf52840-mdk/

## Changelog Entry

The Makerdiary nRF52840 MDK USB Dongle board is now supported.

## Change Checklist

### DIFF
--- a/boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
+++ b/boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
@@ -1,0 +1,26 @@
+version: 0.3.0
+targets:
+  makerdiary-nrf52840-mdk-usb-dongle:
+    # website: https://wiki.makerdiary.com/nrf52840-mdk-usb-dongle/
+    chip: nrf52840
+    leds:
+      led0:
+        # note: this is the RGB LED red channel
+        color: red
+        pin: P0_23
+        active: low
+      led1:
+        # note: this is the RGB LED green channel
+        color: green
+        pin: P0_22
+        active: low
+      led2:
+        # note: this is the RGB LED blue channel
+        color: blue
+        pin: P0_24
+        active: low
+    buttons:
+      # button0:
+      #   # note: this is configurable via UICR
+      #   pin: P0_18
+      #   active: low
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -1022,6 +1022,13 @@ builders:
       user_usb: not_available
       ethernet_over_usb: not_available
 
+  makerdiary-nrf52840-mdk-usb-dongle:
+    chip: nrf52840
+    tier: "3"
+    support:
+      user_usb: supported
+      ethernet_over_usb: supported
+
   native:
     chip: native-chip
     tier: "1"
@@ -1361,6 +1368,11 @@ boards:
     url: https://web.archive.org/web/20250109121140/https://microbit.org/new-microbit/
     builders:
       - bbc-microbit-v2
+
+  - name: Makerdiary nRF52840 MDK USB Dongle
+    url: https://web.archive.org/web/20260508214910/https://wiki.makerdiary.com/nrf52840-mdk-usb-dongle/
+    builders:
+      - makerdiary-nrf52840-mdk-usb-dongle
 
   - name: native
     url: "#"
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -18,6 +18,7 @@
   - [Espressif ESP32-S2-DevKitC-1](./boards/espressif-esp32-s2-devkitc-1.md)
   - [Espressif ESP32-S3-DevKitC-1](./boards/espressif-esp32-s3-devkitc-1.md)
   - [Heltec WiFi LoRa 32 V3](./boards/heltec-wifi-lora-32-v3.md)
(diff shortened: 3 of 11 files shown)


boards/nrf52840-mdk.yaml
ROMemories, line 1: Could you move this to a separate PR? I should then be able to review this quickly.

boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
ROMemories, line 1: Could you add the necessary support info for this new board, as [explained in the porting guide](https://ariel-os.github.io/ariel-os/0.4.0/docs/book/adding-board-support.html#adding-support-for-a-board-1)? (We should do the same for the nRF52840-MDK, but that's tracked by #1806 and out of scope of this PR.)

ROMemories (commented):
Thanks for the PR! I haven't looked at the SBD info yet, I'll do that while looking at the support info at the same time.

boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
SimonIT: I think it's all there now

src/ariel-os-nrf/Cargo.toml
```diff
@@ -50,7 +50,7 @@ embassy-nrf = { workspace = true, features = ["nrf52833"] }
 nrf-sdc = { workspace = true, features = ["nrf52833"], optional = true }
 
 [target.'cfg(context = "nrf52840")'.dependencies]
-embassy-nrf = { workspace = true, features = ["nrf52840"] }
+embassy-nrf = { workspace = true, features = ["nrf52840", "reset-pin-as-gpio"] }
```
ROMemories: Unfortunately we're not going to be able to enable this here as that would set this for nRF52840s on every board. I do see that it's used for the user button, so it's fine to keep the pin information in the SBD file however.

```suggestion
embassy-nrf = { workspace = true, features = ["nrf52840"] }
```

For now if you need to use the user button in a application, the easier solution would be to depend on `embassy-nrf` and to enabled `reset-pin-as-gpio` there (which will enable it through feature unification). We should later have a mechanism to configure this through the SBD file.

doc/support_matrix.yml
```diff
@@ -1064,6 +1064,13 @@ builders:
     support:
       ble: supported
 
+  nrf52840-mdk-usb-dongle:
+    chip: nrf52840
+    tier: "3"
```
ROMemories: Indeed a Tier 3 board for now. 👍

boards/nrf52840-mdk-usb-dongle.yaml
```diff
@@ -0,0 +1,26 @@
...
+      led2:
+        # note: this is the RGB LED blue channel
+        color: blue
+        pin: P0_24
+        active: low
+    buttons:
+      button0:
+        # note: this is configurable via UICR
+        pin: P0_18
+        active: low
```
ROMemories: Checked the pins and active states [with the docs](https://wiki.makerdiary.com/nrf52840-mdk-usb-dongle/hardware/#button-and-leds). 👍

boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
ROMemories, line 1: Considering this is not an official board from Nordic, could you rename the board and builder to `makerdiary-nrf52840-mdk-usb-dongle`? (I know this is not consistent with `nrf52840-mdk`, but we'll rename that one as well.)

ROMemories (changes requested):
Should be good to go after renaming the board and removing the extra feature. The comments with a 👍 are just things I've checked, they don't required any action from your side.

src/ariel-os-nrf/Cargo.toml
SimonIT: I only did this because the command from the docs, `CONFIG_WIFI_NETWORK='test' CONFIG_WIFI_PASSWORD='password' laze build --global -b nrf52840-mdk-usb-dongle`, fails without it

ROMemories: Ok, that's because `P0_18` doesn't exist otherwise, and having it in the SBD file attempts to take that pin. You can comment out the definition in the SBD file then, and we'll try to find a general solution to that later.

boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
ROMemories: Thanks for even reordering the renamed builders!

ROMemories (approved):
Thanks a lot!

ROMemories:
GitHub Actions seem to be having a hard time at the moment: it might take a while for this to land, we'll retry later if necessary.

boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
SimonIT: It scratches my brain when everything except one thing is not in order 😄

PR #2091, opened 2026-05-13
build(espflash): introduce an `attach` laze task

ROMemories:
# Description

This introduces the missing `attach` laze task for `espflash` (there was already one for probe-rs), and documents it by introducing a section in the Logging page.

## Testing

Successfully tested the `attach` tasks by first modifying the `hello-world` example by adding a loop with a few iterations and delays (while keeping the `exit()` call after the bounded loop) and then using the following:

```sh
laze -C examples/hello-world build -b stm32u083c-dk flash
laze -C examples/hello-world build -b stm32u083c-dk attach
```

(The second command must be run while the aforementioned loop is still running, before `exit()` is called, otherwise the core is already locked up.)

```sh
laze -C examples/hello-world build -s logging-over-uart -b espressif-esp32-c6-devkitc-1 run
laze -C examples/hello-world build -s logging-over-uart -b espressif-esp32-c6-devkitc-1 attach
```

(The port allowing access to UART0 must be used. We have to use `run` for now as there is no `flash` task yet for these; it'll be added in a soon-to-be-opened PR. The `run` task can be interrupted after flashing is done, before using the `attach` task.)

```sh
laze -C examples/hello-world build -b espressif-esp32-c6-devkitc-1 run
laze -C examples/hello-world build -b espressif-esp32-c6-devkitc-1 attach
```

(The "USB"-labeled USB port must be used.)

```sh
laze -C examples/hello-world build -s probe-rs -b espressif-esp32-c6-devkitc-1 run
laze -C examples/hello-world build -s probe-rs -b espressif-esp32-c6-devkitc-1 attach
```

(The "USB"-labeled USB port must be used. `attach` should terminate on its own due to the call to `exit()`, however this doesn't work for me (neither does it on an ESP32-C3), while it does work with the `run` command and also when using `attach` on an ESP32-S3, so I believe this is a probe-rs bug. I'll file an issue upstream later but I don't consider this a blocker for this PR.)

```sh
laze -C examples/hello-world build -b unihiker-k10 run
laze -C examples/hello-world build -b unihiker-k10 attach
```

(Works as expected.)

```sh
laze -C examples/hello-world build -s probe-rs -b unihiker-k10 run
laze -C examples/hello-world build -s probe-rs -b unihiker-k10 attach
```

(`attach` does terminate on the ESP32-S3 due to the `exit()` call.)

## Issues/PRs References

- Depends on #1987

## Changelog Entry

An `attach` laze task is now available to fetch and display the logs from a running target.

## Change Checklist

### DIFF
--- a/book/src/build-system.md
+++ b/book/src/build-system.md
@@ -22,7 +22,7 @@ For example, to run the hello-world example from the `ariel-os` directory, the c
 
 Tasks available in Ariel OS include:
 
-- `run`, `flash`, `flash-dfuse`, `flash-erase-all`, `reset`: See [Flashing & Debugging][flashing-debugging-book] for flashing and debug-related tasks.
+- `run`, `flash`, `flash-dfuse`, `flash-erase-all`, `attach`, `reset`: See [Flashing & Debugging][flashing-debugging-book] and [Logging][logging-book] for flashing and logging-related tasks.
 - `debug`: Starts a GDB debug session for the selected application.
   The application needs to be flashed using the `flash` task beforehand.
 - `tree`: Prints the application's `cargo tree`.
@@ -140,3 +140,4 @@ apps:
 [laze-module-selects-book]: https://kaspar030.github.io/laze/dev/reference/module/selects.html
 [laze-module-conflicts-book]: https://kaspar030.github.io/laze/dev/reference/module/conflicts.html
 [flashing-debugging-book]: ./flashing-debugging.md
+[logging-book]: ./logging.md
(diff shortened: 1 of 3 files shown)


book/src/build-system.md
```diff
@@ -22,7 +22,7 @@ For example, to run the hello-world example from the `ariel-os` directory, the c
 
 Tasks available in Ariel OS include:
 
-- `run`, `flash`, `flash-dfuse`, `flash-erase-all`, `reset`: See [Flashing & Debugging][flashing-debugging-book] for flashing and debug-related tasks.
+- `run`, `flash`, `flash-dfuse`, `flash-erase-all`, `attach`, `reset`: See [Flashing & Debugging][flashing-debugging-book] for flashing and debug-related tasks.
```
ROMemories: Actually the `attach` task is from the Logging page, let me change that quickly.

ROMemories: Done.

anlavandier (approved):
LGTM,
I successfully tested this on the `espressif-esp32-devkitc` and the `unihiker-k10`.

Issue #2088, opened 2026-05-12
restrict creating objects when not supported.


stappersg:
Parking partial discussion from today's Matrix chat:

* assumed that `CoreAffinity` did anything, turns out that it only does if you use the `core-affinity` feature from ariel-os
* prevent importing/using the `CoreAffinity` type when the feature is not enabled

The idea of this issue is avoiding that an improvement proposal gets lost (in the good discussion)

anlavandier:
Thank you for opening the issue and archiving this conversation

jvarias02:
Hi, just wanted to ask about the status of this issue. 

I see #2092 is waiting on a decision whether to document the Cargo feature or the laze module. In any case, I believe the restriction of `CoreAffinity` type usage when the feature is not "on" can still be implemented, maybe left as a draft until documentation is resolved.

anlavandier:
I think that this issue can be considered closed by #2092.

PR #2092, opened 2026-05-13
fix(threading): Improve error messages in the thread proc-macro and feature gate `CoreAffinity`

anlavandier:
# Description

This PR expands the `core-affinity` section of the book to clarify the situation raised in #2088. it also makes code like 
```rust
#[ariel_os::thread(autostart, affinity = CoreAffinity::one(CoreId::new(1)), priority = 2)]
fn main() { ... }
```
A compile time error when the `core-affinity` is not enabled.

## Testing

Removing the `core-affinity` feature from the `threading-multicore` examples results in a compile time error with the following error message
```sh
error: Providing a core affinity does nothing unless the 'core-affinity' feature is enabled.
 --> examples/threading-multicore/src/main.rs:9:42
  |
9 | #[ariel_os::thread(autostart, affinity = CoreAffinity::one(CoreId::new(1)), priority = 2)]
  |                                          ^^^^^^^^^^^^
```

## Issues/PRs References
Advances #2088 

## Changelog Entry

The `#[ariel_os::thread]` attribute macro now makes sure that the `core-affinity` Cargo feature is enabled when attempting to pin a thread.

## Change Checklist

### DIFF
--- a/book/src/multithreading.md
+++ b/book/src/multithreading.md
@@ -58,7 +58,7 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 ### Core Affinity
 
 Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
-It allows to restrict the execution of a thread to a specific core and prevent it from being scheduled on another one.
+It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one. This is only enabled when the `core-affinity` [laze module][laze-modules-book] is selected. Otherwise, the `affinity` parameter of the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc] is ignored.
 See the [`threading-multicore` example][threading-multicore-example-repo] for a usage example.
 
 [Embassy]: https://embassy.dev/
--- a/book/src/multithreading.md
+++ b/book/src/multithreading.md
@@ -58,7 +58,8 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 ### Core Affinity
 
 Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
-It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one. This is only enabled when the `core-affinity` [laze module][laze-modules-book] is selected. Otherwise, the `affinity` parameter of the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc] is ignored.
+It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one. This is only enabled when the `core-affinity` [laze module][laze-modules-book] is selected.
+When disabled the `affinity` parameter of the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc] is ignored.
 See the [`threading-multicore` example][threading-multicore-example-repo] for a usage example.
 
 [Embassy]: https://embassy.dev/
--- a/book/src/multithreading.md
+++ b/book/src/multithreading.md
@@ -59,7 +59,7 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 
 Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
 It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one. This is only enabled when the `core-affinity` [laze module][laze-modules-book] is selected.
-When disabled the `affinity` parameter of the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc] is ignored.
+When disabled, the `affinity` parameter of the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc] is ignored.
 See the [`threading-multicore` example][threading-multicore-example-repo] for a usage example.
 
 [Embassy]: https://embassy.dev/
--- a/book/src/multithreading.md
+++ b/book/src/multithreading.md
@@ -58,8 +58,8 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 ### Core Affinity
 
 Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
-It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one. This is only enabled when the `core-affinity` [laze module][laze-modules-book] is selected.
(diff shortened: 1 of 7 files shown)


book/src/multithreading.md
```diff
@@ -58,7 +58,8 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
 ### Core Affinity
 
 Core affinity, also known as core pinning, is optionally configurable for each thread using the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc].
-It allows to restrict the execution of a thread to a specific core and prevent it from being scheduled on another one.
+It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one. This is only enabled when the `core-affinity` [laze module][laze-modules-book] is selected.
```
ROMemories, line 61: Currently there is a *documented* `core-affinity` Cargo feature and an undocumented `core-affinity` laze module, which is internally used for Wi-Fi on ESP, which is *only* available when the MCU is multicore. We need to decide whether we want to keep documenting only the Cargo feature (which should then be the one mentioned here), document only the laze module (which would be a breaking change, but that's fine), or somehow document both. I think this boils down to whether we think *the concept* of core affinity should only be available on multicore MCUs; I need to think about this for a bit.

ROMemories: So we could consider feature-gating `CoreAffinity` on `core-affinity`, but this isn't easy because `thread::create()` always takes a core affinity. Instead, could you look into making the `thread` macro fail at compile time if the `core-affinity` feature is *not* enabled while a `CoreAffinity` is given? Simply feature-gating a `compile_error!` might be enough.

anlavandier: This would only prevent using `CoreAffinity` in the `ariel_os::thread` macro and not manually spawned thread though

anlavandier:
Requesting the review for the compile time failure. I didn't change the documentation part yet

anlavandier:
See [here](https://github.com/ariel-os/ariel-os/commit/849336c230d72652da1568a57f0d5c3feada62ba#diff-717aeb9a3a3d8aae680ad215e2fe277851b045b2170479b5348d67d4e70ab4b8) v1 of the affinity error message and [here](72070523830e128e748fef35ee357bc0d632ff12) for v2 and  the changed missing `autostart` message

src/ariel-os-macros/tests/ui/thread/core-affinity-without-the-feature.rs
```diff
@@ -0,0 +1,8 @@
+// #![no_std]
+#![no_main]
+
+use ariel_os::thread::{CoreAffinity, CoreId};
+
+// FAIL: the `autostart` parameter is mandatory
```
ROMemories: This comment needs to be updated.

ROMemories: This comment needs to be updated.

src/ariel-os-macros/src/thread.rs
```diff
@@ -78,7 +74,11 @@ pub fn thread(args: TokenStream, item: TokenStream) -> TokenStream {
         stack_size,
         priority,
         affinity,
-    } = Parameters::from(attrs);
+    } = match Parameters::try_from(attrs) {
+        Ok(p) => p,
+        Err(ParametersError::CoreAffinityNotEnabled) => return syn::Error::new_spanned(<TokenStream as Into<proc_macro2::TokenStream>>::into(args.clone()), "Providing a core affinity does nothing unless the 'core-affinity' feature is enabled.").into_compile_error().into(),
```
ROMemories: These error messages should not be capitalized nor end with a period:
```suggestion
        Err(ParametersError::CoreAffinityNotEnabled) => return syn::Error::new_spanned(<TokenStream as Into<proc_macro2::TokenStream>>::into(args.clone()), "providing a core affinity does nothing unless the `core-affinity` feature is enabled").into_compile_error().into(),
```

src/ariel-os-macros/Cargo.toml
```diff
@@ -30,6 +30,7 @@ trybuild = "=1.0.116"
 
 [features]
 _test = []
+core-affinity = []
```
ROMemories, line 33: What I had in mind at first was that it would be enough have the macro generate a feature-gated `compile_error!`, but that actually doesn't make sense as the application (or wherever the macro would be called) wouldn't have the `core-affinity` Cargo feature itself.

book/src/multithreading.md
ROMemories: Yes, I think we can consider this separate.

book/src/multithreading.md
```diff
@@ -58,7 +58,8 @@ When an idle thread is scheduled, it prompts the current core to enter sleep mod
...
+When disabled, the `affinity` parameter of the [`#[ariel_os:thread]` attribute macro][thread-attr-macro-rustdoc] is ignored.
```
ROMemories:
```suggestion
It allows restricting the execution of a thread to a specific core and prevents it from being scheduled on another one.
This can be enabled using the `core-affinity` Cargo feature.
```

ROMemories (commented):
Should be good to go after this one

src/ariel-os-macros/Cargo.toml
anlavandier: Yes. The only way I can think of would be to inspect dependency features (but I don't think that it's possible) or to go through a laze module but that's clunky. It's also less clean IMO

book/src/multithreading.md
anlavandier: In the rest of the book pages, it's mostly the laze modules that are documented e.g. `multi-core` and `sw/threading`. Is there any reason why we want to highlight the cargo feature here instead ?

src/ariel-os-macros/src/thread.rs
anlavandier: Fixed in [4af840d](https://github.com/ariel-os/ariel-os/pull/2092/commits/4af840d0802bb0e67a23d2b4eb7fbce5c6a7dccd)

src/ariel-os-macros/tests/ui/thread/core-affinity-without-the-feature.rs
anlavandier: Fixed in [0e58610](https://github.com/ariel-os/ariel-os/pull/2092/commits/0e58610e1061f2de0a8a1daf7566141be15f3d53)

book/src/multithreading.md
ROMemories: Currently we do already document the `core-affinity` Cargo feature in the rustdoc docs, so switching to documenting the laze module *instead* would be a breaking change, which doesn't seem necessary. None of that is very obvious because the individual scopes and interaction of the Cargo feature and the laze module are ill-defined right now, but I think it's better not to try to address that technically orthogonal issue in this PR.

anlavandier: Fixed in 3633d7d

ROMemories (approved):
LGTM, please squash (the docs could likely go into their own commit)!

PR #2093, opened 2026-05-15
fix: let `run` task go over all applications

dcz-self:
When using -multiple-tasks, multiple tasks were generated, but each command would be identical and would end up running the default binary. After this change, each task runs a different binary, selected from the application name.

# Description

I was trying to run a separate binary under the native target as another application. `laze-project.yml`:

```
apps:
  - name: ariel-os-hello
  - name: interrupts
```

This still doesn't let me choose which binary I want to run. Or maybe I can't figure it out. Thankfully, my default binary is not an infinite loop.

## Testing

Download the hello world app. Modify `laze-project.yml` like above. Then:

```
mkdir src/bin
cp src/main.rs src/bin/interrupts.rs
laze build -m -b native run
```

This should run 2 separate binaries, not just one.

## Changelog Entry

`laze build run` now works properly when there are multiple binaries within a single application directory.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -130,7 +130,7 @@ contexts:
         build: false
         workdir: ${appdir}
         cmd:
-          - ${CARGO_PRESHELL} ${CARGO_ENV} ${CARGO} ${CARGO_TOOLCHAIN} ${CARGO_ARGS} run --profile=${PROFILE} ${FEATURES} $@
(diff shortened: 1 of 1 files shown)


kaspar030 (approved):
ACK.

I tested behavior on a "real" board - the flash and run tasks already do the right thing.

I pushed a commit that passes linting - mainly, wrapping its description lines.

Thanks!

Issue #284, opened 2024-05-02
Wall time and trusted time


chrysn:
When using ACE/OAuth for authentication, the tokens (CWTs) typically carry expiry information. Currently, we don't have wall time, let alone trusted wall time.

A document I'm working on in T2TRG ([raytime](https://www.ietf.org/archive/id/draft-amsuess-t2trg-raytime-02.html)) describes what is about a minimal thing a device can do, but it'd still be about keeping some time. (In essence, it's about just keeping a lower bound of time, bumping it guided by seen tokens, and pushing it along while the device has a timer running).

So, what do we need in general, what do we want, and what should be default?

* Do we expect that securely networked devices are used in a mode that can do with no time at all?
  * If yes, how should this be made opt-in so that we are still secure by default (with "secure" following general guidance on not granting infinitely long-lived access)?
* Do we want to offer wall time as an OS-wide service? How many of them / at which trust levels?
  * Some secure time source authorized by the owner? [The AS itself](https://datatracker.ietf.org/doc/html/draft-navas-ace-secure-time-synchronization-00)?
  * What precisions do we expect?
  * Which reliability v. security trade-offs are acceptable for which clock?
  * Can we even protect high precision time sources? (GNSS such as GPS can give us "perfect" time without authentication, AS provided time is ultimately trusted but goes through a full network round trip and might include extra times when the AS is heavily guarded)
  * How does an application select "the" time to use, eg. for storing timestamps of sensor values?
* Do we want to check token validity against a wall-time-ish service, or do we better split those, let the token security parts do their own weird raytime thing, and trust the time multiplexer to keep things efficient hardware-wise?

chrysn:
One approach I start favoring is that the OS manages something like cumulative experienced-time. This is a time counter on some constant scale, and it is monotonous across reboots. As this would be backed by flash writes that consume power and wear down flash, it may make sense to model this as an interval (guaranteed-monotonous, best-effort-latest), the monotony only applies to the lower bound, but the application querying time can request this interval to be arbitrarily small (but requesting a small interval may force the OS to do a flash write, so applications please use lose bounds), or may contribute to the system wide policy on how large that window may become. (Early guesses are that when the time is requested, an application might ask for a maximum 10-minute window, which would create a flash write every 10 minutes while under load, and a maximum of 1-day when idle).

[...]

PR #2094, opened 2026-05-15
feat: Current wall time synchronized with SNTP

SimonIT:
# Description

I created a module called ariel-os-sntp to retrieve the current wall time. You can either make a manual request with the `update_global_clock` function, or simply use the `start` function with an interval to update the clock automatically. The `now` function returns the current time. This is calculated using the latest SNTP timestamp, to which the duration since synchronisation with embassy-time is added. Therefore, you can obtain the current time at any time. I have included a check for updating SNTP requests to verify their plausibility.

## Testing

I tested it on an ESP32C6 using the included example. When syncing hourly, I got a diff of around 30 ms.

## Open Questions

Feel free to say what you think of this feature. I think it's useful. What do you think about the structure? I'm not sure about the values of the constants.

## Changelog Entry

Added a wall clock based on SNTP.

## Change Checklist


SimonIT:
Well... For defmt, the newest sntpc version is needed, but we use an old embassy-net version, so we need an older sntpc version... Unless #1939 is merged

src/ariel-os-sntp/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+pub async fn start(stack: Stack<'static>, addr: net::SocketAddr, interval: Duration) {
+    // SAFETY: same pattern used throughout ariel-os-embassy
+    #![expect(unsafe_code)]
+    let spawner = unsafe { embassy_executor::Spawner::for_current_executor().await };
+    spawner.must_spawn(sntp_task(stack, addr, interval));
+}
+
+/// Returns the current Unix timestamp from [`GLOBAL_CLOCK`], or `None` if not
+/// synchronized yet.
+pub fn now() -> Option<Instant> {
```
chrysn, line 194: With this as a pub item, do GLOBAL_CLOCK and GlobalClock need to be public?

examples/sntp-router/README.md
```diff
@@ -0,0 +1,17 @@
+# sntp-router
+
+## About
+
+This example synchronizes `ariel-os-sntp` with an NTP server on the local
+router (default: `192.168.188.1:123`) and updates the global SNTP clock.
```
chrysn: Whose default is that? Maybe:

```suggestion
router (e.g.: `10.42.0.1:123` in many NetworkManager setups) and updates the global SNTP clock.
```

chrysn (commented):
At a high level, this looks good to me. There are some themes I'd suggest for consideration, some of which have more background in https://github.com/ariel-os/ariel-os/issues/284:

* Statements around properties.

  User visible functions such as `now()` should have a statement, or refer to a module level statement, what can and can not be expected from such a timestamp, in particular.

  - security: Unencrypted network time is essentially hearsay. Users should not do things like time based access control with it. (I expect we'll have a `now_secure()` at some point).
  - monotony: This has no monotony guarantees.
  - The API describes reporting a UNIX timestamp, which means UTC with POSIX epoch. Does SNTP provide information for getting TAI too? (That'd be monotonic at least in intention, although sync artifacts and lying time sources can still cause decreases).

* Generalization: On the long run, I'd like Ariel to have a now() / now_secure() function that is not tied (by module name and functionality) to SNTP; in particular, there can be GNSS time sources and BLE providers.

  This doesn't have to happen in this PR, though, but can be generalized when the next one comes in -- and may be selected by the build system. (Like, if `net` is used, go SNTP; if a GNSS receiver is present, etc.).

* Ease of use: The current PR offers a one-shot mechanism, and a task. When a user wants hassle-free operation, I don't think they should set up the task like in `sntp-router`, but instead just enable the relevant Cargo/laze feature, and trust the system to auto-start the task.

  Selecting the address is then a system configuration topic -- but there can be selectable configurations, and if routers are generally expected to provide NTP, maybe that is a good default -- at least for now. (Our current default for IPv6 is hardcoded addresses, so don't be afraid to make an initially sensible mechanism the default and change it later when there are better options such as listening to DHCP fields).

SimonIT:
Thanks for the review! I will make the changes when #2125 is merged — sntpc needs the newer version of embassy-net

SimonIT:
I think I have made all the mentioned changes. What do you think?

Regarding the vetting: Should I do it? Or would you like to do it?

SimonIT:
Hey, is there anything I can do to help get it merged?

kaspar030:
Hey @SimonIT! This looks almost g2g. I'll deal with the vetting, and we need to change the used Mutex - the thread Mutex will silently do the wrong thing when used from an interrupt (-executor). I'll also deal with that. Stay tuned.

SimonIT:
A huge thank you! I just saw that an update to sntpc has been released, version 0.11.0. Perhaps we could sneak that in?

kaspar030:
> Perhaps we could sneak that in?

will do, thanks for the hint!

kaspar030:
@SimonIT I have bumped to the current sntp crates, fixed the mutex issue, cleaned up imports, updated vetting and renamed the example to sntp-client. Getting someone else to do a last review now.

src/ariel-os-sntp/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+/// A larger difference causes [`PlausibilityError::JumpTooLarge`].
+pub const MAX_PLAUSIBLE_DIFF: Duration = Duration::from_millis(100);
+
+/// The sleep interval for between synchronizations
+pub const SYNC_INTERVAL: Duration = Duration::from_secs(60 * 60);
+
+/// Errors that can occur while fetching SNTP time.
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Error {
```
bergzand, line 65: This should implement [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html).

src/ariel-os-sntp/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+            Self::Protocol => f.write_str("SNTP protocol error"),
+            Self::Plausibility(e) => write!(f, "plausibility check failed: {e}"),
+        }
+    }
+}
+
+/// Reason why a received SNTP timestamp was rejected.
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PlausibilityError {
```
bergzand, line 95: This should implement [core::error::Error](https://doc.rust-lang.org/core/error/trait.Error.html).

src/ariel-os-sntp/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+/// SNTP packets are 48 bytes, but the buffer is intentionally a bit larger.
+pub const DEFAULT_TX_BUFFER_SIZE: usize = 64;
+
+/// Maximum tolerated absolute difference between local estimate and SNTP.
+///
+/// A larger difference causes [`PlausibilityError::JumpTooLarge`].
+pub const MAX_PLAUSIBLE_DIFF: Duration = Duration::from_millis(100);
+
+/// The sleep interval for between synchronizations
+pub const SYNC_INTERVAL: Duration = Duration::from_secs(60 * 60);
```
bergzand: Should these be `pub`? They're not used outside this file right?

bergzand (changes requested):
Some last comments from my side. Looks good otherwise!

SimonIT:
Thanks for the review! I made the according changes

SimonIT:
So... any updates?🙈 @bergzand

bergzand:
Sorry for the delay, and thanks for the ping :see_no_evil: 

As far as I can see, the only thing remaining before this can be merged is to squash the commits into a reasonable set.

SimonIT:
Awesome! What do you think is a reasonable set? I'm not really sure

PR #2087, opened 2026-05-12
docs(debug-channel): document the RTT blocking behavior

ROMemories:
# Description

This documents and guarantees the default non-blocking behavior of RTT, and how it can be affected by RTT-enabled host tools.

## Issues/PRs References

- Depends on #1987
- Depends on #2078

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2056, opened 2026-04-29
docs(support): leverage the new `note_snippets` where possible

ROMemories:
# Description

This leverages the new `note_snippets` functionality introduced by #1833 to factor out the documentation of the identity-based generation of the MAC address for use with Ethernet.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventional-commits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2097, opened 2026-05-17
Logging docs improvements

dcz-self:
# Description

I wanted to see some logs and I noticed that it would have been much harder if I was a newbie:

- how to edit the laze file?
- what option to choose?
- how to connect to the board?
- how to diagnose unexpeced outcomes?

The actual command to connect to the logger is implied but I haven't figured it out yet. So I submit only what I could find out.

## Testing

I copied the relevant source code from my working config.

## Change Checklist

### DIFF
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -1,6 +1,6 @@
 # Logging
 
-Ariel OS supports logging on all platforms and it is enabled by default with the `logging` [laze module][laze-modules-book].
+Ariel OS supports logging on all platforms and it is enabled by default thanks to the `logging` [laze module][laze-modules-book].
 Logging offers a set of macros that print on the debug console with helpful logging formatting.
 
 ## Printing Panics
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -71,6 +71,12 @@ $ laze build -C examples/log --builders nrf52840dk -DLOG=info,ariel_os_rt=trace 
 
 ### [log]
 
+To build your project with plain logging support rather than defmt, use the `--select=log` option on the command line. For example:
+
+```
+$ laze build -C examples/log --builders nrf52840dk --select=log
+```
+
 Ariel OS's logger for `log` supports configuring the log level globally, but does not currently support per-crate filtering.
 
 ## Logging Transports
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -93,6 +93,11 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 >
 > When a separate serial monitor is needed, [`defmt-print`][defmt-print-cratesio] can be used as `defmt`-enabled serial monitor.
 > If this is not possible, `defmt` should be disabled and [`log`](#log) used instead as logging facade.
+>
+> This is what you may see instead of your messages if you use `defmt` logging but use a serial monitor that is not `defmt`-enabled:
+> ```
+> ~� z� j� z� j� z� j� z� j� z� j
+> ```
 
 <!-- TODO: link to "debug interface" when the page exists -->
 > [!TIP]
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -1,6 +1,6 @@
 # Logging
 
-Ariel OS supports logging on all platforms and it is enabled by default with the `logging` [laze module][laze-modules-book].
+Ariel OS supports logging on all platforms and it is enabled by default with the `logging` [laze module][laze-modules-book].
 Logging offers a set of macros that print on the debug console with helpful logging formatting.
 
 ## Printing Panics
@@ -23,7 +23,7 @@ async fn main() {
 
 ## Filtering Logs
 
-In Ariel OS, the log level defaults to `info`. It can be configured using the
+In Ariel OS, the log level defaults to `info`. It can be configured using the
 laze variable `LOG`.
 Depending on the logger, it may be possible to configure different levels per crate or per module.
 
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -124,6 +124,15 @@ note_snippets:
       It can also be used to obtain logs when the [`logging-over-uart`](../logging.md#logging-transports) laze module is enabled.
 
       Note: On the ESP32-C3, enabling `logging-over-uart` will [not prevent the logs from also being printed using the USB CDC-ACM USB peripheral](https://github.com/esp-rs/esp-hal/issues/4510).
+    - name: espressif-uart-logging-configuration
+    content: |-
+      #### UART logging
+
+      The UART used for logging is configured to use:
+      - 115200 baud,
+      - 8 data bits,
+      - 1 stop bit,
+      - no parity.
 
 # Encodes support status for each chip.
 chips:
@@ -429,6 +438,8 @@ chips:
           - no generic USB peripheral
       ethernet_over_usb: not_available
       ethernet: not_available
+    note_snippets:
+      - espressif-uart-logging-configuration
 
   # Currently identical to esp32c6.
   esp32c6fx4:
@@ -455,6 +466,8 @@ chips:
           - no generic USB peripheral
       ethernet_over_usb: not_available
       ethernet: not_available
+    note_snippets:
+      - espressif-uart-logging-configuration
 
   esp32s2:
     name: ESP32-S2


dcz-self:
I can see that there's something in laze about espflash: 

```
CARGO_RUNNER: '"espflash flash --monitor ${ESPFLASH_LOG_FORMAT}"'
```

which would mean that the base command is just `laze run`?

On the other hand, ESP32 with 2 ports adds complexity, where the user connects to JTAG and successfully flashes with probe-rs but then doesn't get any logs because it's UART.

Additionally, once UART gets used on other boards, there's no longer a guarantee that flashing and watching logs can be done in one command.

That means connecting to logs serial console should be documented separately. I know how to do it for `log`, but what's the command for `defmt`?

src/ariel-os-esp/src/uart.rs
```diff
@@ -28,6 +28,7 @@ pub struct Config {
 
 impl Default for Config {
     fn default() -> Self {
+        // Update the logging documentation when changing this.
```
ROMemories: *Currently* this UART driver isn't used for logging (this may well changed in the future however); [`esp-println`](https://docs.rs/esp-println/latest/esp_println/) is used instead.
```suggestion
```

book/src/logging.md
```diff
@@ -94,6 +105,12 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 > When a separate serial monitor is needed, [`defmt-print`][defmt-print-cratesio] can be used as `defmt`-enabled serial monitor.
 > If this is not possible, `defmt` should be disabled and [`log`](#log) used instead as logging facade.
 
+> [!NOTE]
+> This is what you may see if you use `defmt` logging but use a serial monitor that is not `defmt`-enabled:
+> ```
+> �       jespressif-esp32-c6-devkitc-1�!z!z!j!z!z!j!z!j!z!
```
ROMemories: I think that would be useful to have: could you merge this note with the above admonition and replace the output with one that is more board-agnostic? I think that of the `log` example will do.

book/src/logging.md
```diff
@@ -21,9 +21,20 @@ async fn main() {
 }
 ```
 
+In the `laze-project.yml` file, add the appropriate modules (see table below). For an ESP32-based board, the resulting app could look like:
```
ROMemories: Can you clarify what pitfall this is intended to avoid? We shouldn't recommend enabling the `log` facade by default (that is bound to be cargo-culted). Also sometimes it's better to select the logging transport on the command line as this depends on which USB port you feel like using on boards that have multiple. It should be selected in the laze application file only when it is intrinsic to the application (which it may be, but that depends).

book/src/logging.md
```diff
@@ -107,6 +124,14 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 > - Other logging transports will later be supported, including UART and USB CDC-ACM on non-ESP32 devices.
 > - Using multiple transports at the same time may be supported in the future.
 
+### UART
+
+The UART logger is configured to use:
```
ROMemories: I'm torn on this one: that would be useful to document, but currently the only *documented* support for `logging-over-uart` is on ESP32s, through `esp-println`, and [`esp-println` doesn't itself document these parameters](https://docs.rs/esp-println/latest/esp_println/) (even though I do agree they may be required), so I'm not sure what's guaranteed. This may additionally be MCU/board-dependent so it might be a better fit for the chip/board pages.  You can either remove this from this PR or add such documentation in the relevant MCU/board pages, leveraging the `note_snippets` as demonstrated in #2056.

ROMemories (changes requested):
Thank you for the PR! As I'm sure you've seen, this area is undergoing quite some changes, so none of this documentation is set in stone and potential remaining issues may well be addressed by later PRs (e.g., #1987 and #2087, and as part of #1008).

ROMemories:
Regarding https://github.com/ariel-os/ariel-os/pull/2097#issuecomment-4470759762, if I've understood your concerns correctly, I think that should be mostly addressed by the new documentation section that #2091 should introduce. We'll also introduce `flash` and `reset` laze tasks for `espflash` (I have a branch with that, just waiting for some of the in-flight PRs to land before opening a PR for it).

book/src/logging.md
dcz-self: Moved to board section. I didn't know which boards exactly are covered by esp-println and UART, so I only added the one I test.

book/src/logging.md
dcz-self: This was to make sure the user knows what to do when they got a defmt mismatch. I don't know how to use defmt, so this is the advice I can offer.
I changed things to show how to use this module on the command line.

book/src/logging.md
dcz-self: Replaced with slightly different garbage.

src/ariel-os-esp/src/uart.rs
dcz-self: Done.

book/src/logging.md
ROMemories: > I don't know how to use defmt

I'm not sure what you mean by this: do you mean using `defmt` in the firmware on the target, or on the host? The firmware part is documented on this page, and the host part is part of #1356.

book/src/logging.md
```diff
@@ -21,9 +21,19 @@ async fn main() {
...
+In the `laze-project.yml` file, add the appropriate modules (see table below). For an ESP32-based board, the resulting app could look like:
+
+```yaml
+apps:
+  - name: main
+    selects:
+      - logging-over-uart
+      - has_buttons
+```
+
```
ROMemories: Please remove this chunk, we shouldn't refer to the `logging-over-uart` module before the logging transports are explained below, nor recommend hard-coding the dependency on logging-over-UART specifically (without further assumptions).
```suggestion
```

book/src/logging.md
```diff
@@ -71,6 +81,12 @@ $ laze build -C examples/log --builders nrf52840dk -DLOG=info,ariel_os_rt=trace
 
 ### [log]
 
+To build your project with plain logging support rather than defmt, use the `--select=log` option on the command line. For example:
+
+```
+$ laze build -C examples/log --builders nrf52840dk --select=log
+```
+
```
ROMemories, line 79: The `log` laze module (and the `defmt` laze module) are already documented in the "Logging Facades and Loggers" section. We could add a sentence in both the the "defmt" and "log" sections to make these stand out a bit more, but no need for a complete command line.

book/src/logging.md
```diff
@@ -93,6 +109,11 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 >
 > When a separate serial monitor is needed, [`defmt-print`][defmt-print-cratesio] can be used as `defmt`-enabled serial monitor.
 > If this is not possible, `defmt` should be disabled and [`log`](#log) used instead as logging facade.
+>
+> This is what you may see instead of your messages if you use `defmt` logging but use a serial monitor that is not `defmt`-enabled:
+> ```
+> ~� z� j� z� j� z� j� z� j� z� j
+> ```
```
ROMemories, line 106: I think that works now 👍

doc/support_matrix.yml
```diff
@@ -124,6 +124,15 @@ note_snippets:
       It can also be used to obtain logs when the [`logging-over-uart`](../logging.md#logging-transports) laze module is enabled.
 
       Note: On the ESP32-C3, enabling `logging-over-uart` will [not prevent the logs from also being printed using the USB CDC-ACM USB peripheral](https://github.com/esp-rs/esp-hal/issues/4510).
+    - name: espressif-uart-log-configuration
```
ROMemories: We try to keep "log" to specifically refer to the `log` logging facade and use "logging" for the general concept of logging:
```suggestion
  - name: espressif-uart-logging-configuration
```

doc/support_matrix.yml
```diff
@@ -124,6 +124,15 @@ note_snippets:
...
+    content: |-
+      ##### UART logging
+
+      The UART logger is configured to use:
```
ROMemories: I would prefer to keep "logger" referring to the "backend" of `log` or `defmt`, independent from the logging transport, as defined in the Logging page:
```suggestion
      The UART used for logging is configured to use:
```

doc/support_matrix.yml
```diff
@@ -1158,6 +1167,7 @@ builders:
     support:
     note_snippets:
       - espressif-devkit-two-usb-ports-no-usb-otg
+      - espressif-uart-log-configuration
```
ROMemories: Can you add this note on the chips instead of on the boards? This would make this more generic. `esp-println` [supports using UART on every single ESP32 MCU](https://docs.rs/esp-println/0.17.0/esp_println/).

doc/support_matrix.yml
```diff
@@ -124,6 +124,15 @@ note_snippets:
       It can also be used to obtain logs when the [`logging-over-uart`](../logging.md#logging-transports) laze module is enabled.
 
       Note: On the ESP32-C3, enabling `logging-over-uart` will [not prevent the logs from also being printed using the USB CDC-ACM USB peripheral](https://github.com/esp-rs/esp-hal/issues/4510).
+    - name: espressif-uart-log-configuration
+    content: |-
+      ##### UART logging
```
ROMemories: The heading level likely needs to be adjusted when moving this to the chips (see my other comment).

doc/support_matrix.yml
ROMemories, line 1: If you prefer, you can split changes to the support info into a separate PR, but it's also totally fine to keep it in this PR.

doc/support_matrix.yml
```diff
@@ -124,6 +124,15 @@ note_snippets:
...
+
+      The UART logger is configured to use:
+      - 115200 baud,
```
ROMemories, line 132: Even though this doesn't seem documented by `esp-println`, this does seem correct. I've checked by using `picocom /dev/ttyUSB0 --baud 115200` manually (instead of using the integrated `espflash`).

book/src/logging.md
dcz-self: This is the main improvement of this changeset: when I wanted to start logging, instead of focusing to code I had to stop coding, jump into the laze documentation rabbit hole and spend the next 2 hours discovering what to do from first principles, without knowing where to look for examples.

The idea behind this is to give people an example, which, even if not entirely fitting their use case, lets them know what the rough shape of the solution is. Note that this doesn't call for copy-pasting, but only presents an example.

I think front-loading assumptions can also be fine if you want to teach the tools, but I tried to move this page in the direction of giving people an easier way to deal with their current problem.

book/src/logging.md
dcz-self: Same above.

book/src/logging.md
ROMemories: I understand what you're saying but I'm unfortunately still not understanding *what* this example is intended to clarify: logging is already enabled by default (this is mentioned at the very top of the page) and we don't want users to specifically enable logging over UART if that's not what they need.

book/src/logging.md
dcz-self: I don't know how to use defmt on host. Thanks for the link. I'll update this once I learn more.

book/src/logging.md
dcz-self: Okay, I think this is a bad wording elsewhere then. I parsed "it is enabled by default with the logging [laze module](https://ariel-os.github.io/ariel-os/dev/docs/book/build-system.html#laze-modules)." as "by default, the responsibility of the logging module is to enable logging". I guess I didn't stop to think about this harder because I discarded it as "it doesn't say anything about logging itself, only about what the module does".

Maybe I should remove the example and reword the above instead.

That being said, what transport is enabled by default?

doc/support_matrix.yml
dcz-self: Thanks for pointing this out. I moved it to the chip I tested it with.

book/src/logging.md
dcz-self: While logging enabled by default makes the laze.yml example unneeded, I still believe this one is still needed - either that or an example for how to use defmt.

PR #2100, opened 2026-05-18
docs(boards): document ad hoc support for logging over UART

### LINKED ISSUE #1925: Documenting support for undocumented, board-specific functionality

ROMemories:
# Description

This documents the ad-hoc support for logging over UART that was added by #1029, now that the `logging-over-uart` laze module exists and that we have the infrastructure to document this functionality on the small set of boards supported. Even if the implementation *will* evolve later (by being generalized and by leveraging SBD files), I believe we can now document it because these future evolutions should result in few breaking changes (if any) and because this reduces the surface of undocumented features.

## TODO

- [ ] Consider documenting the UART settings, see https://github.com/ariel-os/ariel-os/pull/2097#discussion_r3258163846.
- [ ] Consider how this interacts with the `run` and `attach` laze tasks from #2091.

## Testing

Successfully tested the `log` example on stm32u083c-dk and nrf52840dk.

## Issues/PRs References

- Closes #1925
- Works towards #1008 (which aims to implement this in a more generic manner)

## Changelog Entry

Logging over UART is now supported on a small set of boards (on top of the existing support for ESP32 MCUs).

## Change Checklist

### DIFF
--- a/doc/support_matrix.yml
+++ b/doc/support_matrix.yml
@@ -124,6 +124,18 @@ note_snippets:
       It can also be used to obtain logs when the [`logging-over-uart`](../logging.md#logging-transports) laze module is enabled.
 
       Note: On the ESP32-C3, enabling `logging-over-uart` will [not prevent the logs from also being printed using the USB CDC-ACM USB peripheral](https://github.com/esp-rs/esp-hal/issues/4510).
+  - name: logging-over-uart-board-debug-probe
+    content: |-
+      ##### Logging over UART
+
+      Using UART as the [logging transport](../logging.md#logging-transports) is supported on this board.
+      Currently only the UART peripheral and the UART pins connected to the onboard, built-in debug probe can be used.
+  - name: logging-over-uart-rpi-pico
+    content: |-
+      ##### Logging over UART
+
+      Using UART as the [logging transport](../logging.md#logging-transports) is supported on this board.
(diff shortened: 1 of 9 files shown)


doc/support_matrix.yml
```diff
@@ -124,6 +124,12 @@ note_snippets:
       It can also be used to obtain logs when the [`logging-over-uart`](../logging.md#logging-transports) laze module is enabled.
 
       Note: On the ESP32-C3, enabling `logging-over-uart` will [not prevent the logs from also being printed using the USB CDC-ACM USB peripheral](https://github.com/esp-rs/esp-hal/issues/4510).
+  - name: logging-over-uart-board-debug-probe
```
ROMemories, line 127: The current set of supported boards is the one from [that module](https://github.com/ariel-os/ariel-os/blob/b57531b660ee9bf8103b6eb025f2fb218fd97218/src/ariel-os-embassy/src/debug_uart.rs#L59).

PR #2101, opened 2026-05-18
feat(net): allow users to implement custom networking

chrysn:
# Description

Currently, applications have to use one of the provided network backends; an out-of-tree application can't just say "yeah I have an Ethernet interface, and I'll wire it up".

We can't fix that easily in general, but we can for implementations of embassy-net-driver that use embassy-net-driver-channel: For the general case we'd need to have Ariel call something from the user crate that is an impl trait, we currently even need a named type. But drivers that use the simpler/higher-level embassy-net-driver-channel just use that crate's types.

There are some limitations still -- we need to agree on the MTU (scalar features anyone?) and number of rx/tx slots, and probably also the medium. (But that we can probably get rid of, because the user can select the embassy-net/medium-\* feature, and could pass the 2nd argument to new in `user_net_device()`).

## Testing

TBD

## Issues/PRs References

This is currently how I implement networking for hophop in https://github.com/ariel-os/hophop/pull/35.

On the long run, I shouldn't need it there, for hophop can become an optional dependency of Ariel OS and then just offer a NetworkDevice type, but it's useful for experimentation, and I guess that over time, more people will want to add their network back-ends to Ariel in a low-threshold way.

## Open Questions

* Is this a good approach in the first place?
* The USER_NET_RUNNER (fortunately private) type is a mouthful.

## Changelog Entry

TBD

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1713,6 +1713,19 @@ modules:
         FEATURES:
           - ariel-os/tuntap
 
+  - name: user-net-driver-channel
+    help: Network backend is provided by the user.
+
+      While Ariel OS can not let user define an arbitrary embassy-net driver
+      (as that type needs to be named), an embassy_net_driver can be
+      implemented.
+    provides_unique:
+      - network_device
+    env:
+      global:
+        FEATURES:
+          - ariel-os/user-net-driver-channel
+
   - name: idle-threads
     help: create idle-threads to be taken when no other threads are ready
     env:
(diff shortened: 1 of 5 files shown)


chrysn:
An extra use case that came up during [chat](https://matrix.to/#/!cINbgbHLemsitmThSh:matrix.org/$5d3HjD56JqaJDbWDJn7mtyvZpEyt-7j61yP9Ipj0dmE?via=matrix.org&via=utwente.io&via=fehler-in-der-matrix.de) was that this would allow native users to tap into userspace network simulations.

PR #2106, opened 2026-05-21
build(espflash): introduce a `reset` laze task

ROMemories:
# Description

This introduces the missing `reset` laze task for `espflash` (there was already one for probe-rs).

## Future Work

A similar PR will follow to introduce a task for flashing only with `espflash`, but its exact name is not yet determined.

## Testing

Successfully tested on unihiker-k10 and espressif-esp32-c6-devkitc-1 (on both USB ports: `espflash` over UART and USB CDC-ACM).

## Changelog Entry

(ESP32) A `reset` laze task is now available to reset a connected target over UART or USB CDC-ACM.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2108, opened 2026-05-21
 build(espflash): introduce a `flash` laze task

ROMemories:
# Description

This introduces the missing `flash` laze task for `espflash` (there was already one for probe-rs).

I've pondered what this task should be called (especially w.r.t. `flash-dfuse`). I think it makes sense to call this one just `flash` for the following reasons:

- It relies on a tool (`espflash`) that we require as part of the Getting Started guide.
- That tool is also used for `run` tasks (unlike `flash-dfuse`).
- The host tool and MCU are able to enter the bootloader without any extra user interaction, independently from the firmware running on the target, unlike the STM32 bootloader which requires having the boot pattern (e.g., having the BOOT0 pin activated) activated during boot. (On USB CDC-ACM, this is done through special semantics associated with the RTS and DTR signals, see Tables 32.3-2 and 32.4-1 from the ESP32-C6 RM v1.2.)

## Testing

Successfully tested on unihiker-k10 and espressif-esp32-c6-devkitc-1 (on both USB ports: `espflash` over UART, by enabling `logging-over-uart`, and USB CDC-ACM).

## Issues/PRs References

- Depends on #2106

## Changelog Entry

(ESP32) A `flash` laze task is now available to flash a board over UART or USB CDC-ACM.

## Change Checklist


bergzand (approved):
Ack, tested flashing via both usb ports on my espressif-esp32-c6-devkitc-1

PR #2080, opened 2026-05-08
Add ESP32 storage support and refactor SPI example

jadsongmatos:
This pull request introduces support for storage functionality on ESP32 and other platforms, refactors the SPI minimal example into a multi-core SPI loopback demo, and adds a new example for testing minimal RAM/ROM usage with persistent storage. The changes also improve platform-specific linker script generation, update dependencies, and enhance feature gating and modularity for storage across the codebase.

**Storage support and platform integration:**

* Added and configured the `esp-storage` crate for ESP32, including feature flags and workspace integration in `Cargo.toml` files, and enabled storage support in `ariel-os-esp` and `ariel-os-hal` with new `storage.rs` modules and feature gating. [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R118-L119) [[2]](diffhunk://#diff-2f1dc69bf535fbdc0231dee19575e2feb8488ad474aeea92f1a414fced425a61R34) [[3]](diffhunk://#diff-2f1dc69bf535fbdc0231dee19575e2feb8488ad474aeea92f1a414fced425a61R60) [[4]](diffhunk://#diff-2f1dc69bf535fbdc0231dee19575e2feb8488ad474aeea92f1a414fced425a61R143-R145) [[5]](diffhunk://#diff-d1eed5d70d3c0eedc001ac1466ca991f4aecd656f7a1f5193d27b8641835df6dR53-R56) [[6]](diffhunk://#diff-2984a7a45aff6674f36adaf7c7f7ec8ec46c2ba93e0c87684af7ae215048320eR1-R10) [[7]](diffhunk://#diff-e61d3391a5736894e88cd95c094f9f2312ef2ed28f89616bef978d5860c66fdbL130-R130) [[8]](diffhunk://#diff-c2bee477c2dff2d1804044a8234583ff1770f8a22e714eb7e35b636bf28a32d7L45-R51) [[9]](diffhunk://#diff-11ea0949847f22f2612a99be029cad84eb74c3b43d1e80eaa89b3d81a1e346f4R1-R4)
[...]

PR #2110, opened 2026-05-22
Add ESP32 persistent storage support

jhirschler:
# Description

This draft adds Persistent Storage support for ESP32-based Ariel OS targets.

The implementation wires `ariel-os-storage` into the ESP HAL backend using `esp-storage`, and provides an ESP partition table with a dedicated `ariel_store` flash partition.

The storage area is represented as an explicit ESP flash partition instead of a linker-reserved section inside the app partition. This makes the flash layout visible to the ESP tooling and should cause flashing to fail if the application image no longer fits into its app partition.

For now, applications that enable storage on ESP need to provide their own ESP partition table. The configured table is passed both to `espflash` and to the `ariel-os-storage` build script. The build script parses the table, looks up the configured storage partition, validates it, and emits the `__storage_start`/`__storage_end` linker symbols from that partition.

The default storage partition name is `ariel_store` and can be overridden through `ARIEL_ESP_STORAGE_PARTITION`.

Since the partition table is now application-provided, this PR no longer relies on a single default flash layout. Storage support is therefore enabled for all ESP MCUs, with the concrete flash layout left to the application.

## Testing

I have successfully tested it using the storage example on:

- ESP32-C3 with 4 MB flash
- ESP32-C6 with 4 MB flash

## Open Questions

This PR now avoids selecting a default ESP flash layout by requiring applications to provide a partition table with a named storage partition.

The remaining interface question is how this should fit with the partitioning work tracked in #639.

Feedback would be very welcome.

## Change Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -684,6 +684,7 @@ dependencies = [
  "embassy-sync 0.7.2",
  "embassy-time",
  "embedded-storage-async",
+ "esp-idf-part",
  "postcard",
  "sequential-storage",
  "serde",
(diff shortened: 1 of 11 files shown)


kaspar030:
Hey, thanks for this contribution and sorry for the long silence!

There are now three PRs adding esp storage support (#1786, #2080 and #2110).
As they need some handling of the esp flash partitioning, we decided to wait until https://github.com/ariel-os/ariel-os/issues/639 has produced some results. Once we have more clarity on how that turns out, we'll sort through those esp storage PRs. Stay tuned!

jhirschler:
Thanks for the update! 

That makes sense to me. The flash partitioning question is also the main open point I ran into while working on this.

I pushed an update that removes the fixed 4 MB ESP partition table assumption from the PR. For ESP targets, the storage build script now expects the application to provide a partition table, so it can derive the storage linker symbols from the configured storage partition. 

That still leaves the final interface question open, of course. The changes should make the PR easier to adapt once #639 concludes.

PR #2074, opened 2026-05-05
fix(debug): fix usage of featurecomb's `features` table

ROMemories:
# Description

The `features` table of featurecomb of `ariel-os-log` was misused and wasn't enforcing that a debug channel was selected when the `debug-channel` Cargo feature of `ariel-os-log` was enabled. This fixes it.

## Testing

It can be tested that the table is now properly picked up by featurecomb by introducing a misspelling in the name of the `debug-channel` Cargo feature used as the key in the table and checking that featurecomb complains about the non-existing feature. Because manifest's values are aggressively cached by Cargo, the manifest can for instance be invalidated by commenting out the `lints` section. 

## Issues/PRs References

- That featurecomb's `features`'s table was added in #1029.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2114, opened 2026-05-26
fix(debug-console): only select a transport when logging is enabled

ROMemories:
# Description

This fixes a bad interaction between the `logging` and `debug-console` laze modules, which was already present in the refactor from #2030 (specifically https://github.com/ariel-os/ariel-os/pull/2030/changes/2b2d64c9817af38f8370a252d16cdc0f244b8747), but has only been surfaced by #2074: trying to compile with the following command on main:

```sh
laze -C examples/hello-world/ build -d logging -b stm32u083c-dk
```

would result in the following (legitimate) featurecomb error:

```sh
error: feature `debug-channel` requires that some features from feature group `debug-channel` be enabled.
       Candidate features: `defmt-rtt`, `rtt-target`
 --> src/ariel-os-debug/src/lib.rs:7:1
  |
7 | #[featurecomb::comb]
  | ^^^^^^^^^^^^^^^^^^^^
  |
```

This is because the `debug-console` laze module still selects the `logging-over-debug-channel` laze module even if `logging` is disabled (e.g., on the CLI). Because no logging *facade* gets selected when `logging` is disabled, no RTT implementation gets enabled by the `debug-channel-rtt` laze module:

https://github.com/ariel-os/ariel-os/blob/bddd93f81a61c000cdbfa71a6241fd97d8e7ee46/laze-project.yml#L1864-L1873

When that happens, featurecomb errors out (as expected), as `ariel-os-debug`'s `debug-channel` Cargo feature *requires* an RTT implementation be selected.

This PR fixes this by:

- Making the `debug-console` laze module only enable `logging-over-debug-channel` when `logging` is selected.
- Making the `logging-over-debug-channel` laze module enable the `ariel-os/debug-channel` Cargo feature, instead of having the `debug-console` laze module enabling that Cargo feature (which just made little sense).
- Making the `native-chip` context not select `debug-console` as that apparently prevents a logging facade from being selected, thus breaking the logging macros. Logging is still enabled by default as per the "global" optional selection of the `logging` module.

## Testing

See #2030 for the testing procedure: only the parts about `stm32u083c-dk` and `native` are relevant.

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -505,8 +505,9 @@ contexts:
     parent: ariel-os
     selects:
       - infini-core
+      # NOTE: semantically this should only be enabled when `logging` is
+      # enabled, but laze does not support conditions on contexts.
       - logging-over-stdout
-      - ?debug-console
     provides:
       - has_device_identity
       - has_hwrng
@@ -1812,16 +1813,13 @@ modules:
     help: enable support for the debug console
     context: ariel-os
     selects:
-      # Mandatory.
-      - logging-over-debug-channel
+      - logging:
+          # Mandatory when logging is otherwise enabled.
+          - logging-over-debug-channel
       # Optional for now, at least until multiple logging transports can be used
       # at the same time: allows using semihosting while a logging transport
       # other than `logging-over-debug-channel` is used.
       - ?semihosting
-    env:
-      global:
-        FEATURES:
-          - ariel-os/debug-channel
 
   - name: logging
     help: enables logging functionality
@@ -1856,6 +1854,10 @@ modules:
           - debug-channel-transport-required
     provides_unique:
       - logging-transport
+    env:
+      global:
+        FEATURES:
+          - ariel-os/debug-channel
 
   # Disabled by modules that provide a `debug-channel-transport`.
   - name: no-debug-channel-transport


nponsard (approved):
Looks good to me !

ROMemories:
I've rebased on top of main to trigger the CI, which was stuck due to yesterday's GH Actions incident.

ROMemories:
It seems that this results in no logging facade being selected on native, I'll look into this.

laze-project.yml
```diff
@@ -505,8 +505,9 @@ contexts:
     parent: ariel-os
     selects:
       - infini-core
+      # NOTE: semantically this should only be enabled when `logging` is
+      # enabled, but laze does not support conditions on contexts.
       - logging-over-stdout
-      - ?debug-console
```
nponsard, line 509: This seems to work, but I don't understand it, how is enabling `debug-console` (after having enabled `logging-over-stdout`) disabling the `log` feature on `ariel-os-log` ?

ROMemories: Good question. I used the following command to clarify this:

```sh
laze -C examples/hello-world build -b native -vv info-modules
```

What happens is that `debug-console` hard-selects `logging-over-debug-channel` (when `logging` is enabled), which makes it mutually exclusive with `logging` on `native-chip` as only `logging-over-stdout` can be enabled on `native-chip`.

nponsard: Thanks for the explanation, it seems that laze's module selection doesn't work as I expected.

nponsard (approved):
Can be merged after squashing

PR #2115, opened 2026-05-26
feat(coap): replace demo keys with runtime-only identity

chrysn:
# Description

The fact that the demokeys have a hardcoded policy has time and again confused experienced developers (and yours truly). Also, they are an eyesore under security anyway.

The reason they can't easily go is that storage is not universally available (eg. https://github.com/ariel-os/ariel-os/pull/2110, https://github.com/ariel-os/ariel-os/pull/2080, https://github.com/ariel-os/ariel-os/pull/1786).

The alternative added <del>(TBD: so far, this PR is just a first commit because I don't want to open an issue so I rather started an impl)</del> here is short-lived identity for servers. With current coapcore and aiocoap versions, that's well usable -- the user just has the choice of either updating the client's security expectations anew on every device startup, or to just treat the server as unauthenticated (which may be fine for some setups, and is still way better than -unprotected).

So the examples <del>would</del>[edit:]<ins>have been</ins> switch to using this as a fallback for lack of storage, and shipped aiocoap configs says "unauthenticated" with a note in the README.

## Testing

[edit, added:]

Run the examples in all the changed READMEs.

## Changelog Entry

* The `coap-server-config-demokeys` feature is removed; `coap-server-config-runtime-identity` is added.
* CoAP examples do not flash the shipped demo keys any more, but leave the devices to generate a key on every startup if no storage is available.

## Change Checklist

  - [x] all examples
  - [x] the book chapter on CoAP


chrysn:
The technical changes are in, and I've already overhauled the coap-server example, which should now be a better starting point for own applications.

There's a ton of stuff CI probably finds, but I'd appreciate a first high-level round of review, focusing on the high-level changes and the new coap-server example.

(Speaking of examples: With the level of configurability, I think we can then also promote coap-blinky from a test to being an example.)

bergzand:
Based on what I currently read in the readme some rambling from my side:

What is not clear to me is what the goal is of this example. What should a user have learned after using this example. Currently the example feels to me like a demonstrator and not something that can be easily used to extend.

- Is it in a shape where it can be used by users to extend with their own handlers.
- What is needed to extend this with other endpoints (adjust `./peers.yml` where required?)
- What is needed to deploy this on a device with storage, what are the instructions to make this secure.

bergzand:
I think the name `coap-server-config-runtime-identity` can be improved upon. There's ambiguity in that it might be *required* at runtime for an identity. Would `coap-server-config-ephemeral-identity` work for you or does that clash too much with the second E in ECDHE?

chrysn:
The "E" in "EDHOC" is why I didn't go with that; more generally, people have Requirements on ephemeral keys and this is not one in the sense that NIST and other folks who write Requirements with a captial R mean it.

Maybe "coap-server-config-shortlived" "-pseudonymous"? Argh, naming…

chrysn:
> Currently the example feels to me like a demonstrator and not something that can be easily used to extend.

I'll take that as permission to go beyond the scope of adjusting to the key changes. (Convenient: I already had some small changes in there in that direction because things have become obsolete, or wouldn't work that well any more as examples).

> Is it in a shape where it can be used by users to extend with their own handlers.

Yes; just added documentation on that.

> What is needed to extend this with other endpoints (adjust ./peers.yml where required?)

Covered in the latest addition.

> What is needed to deploy this on a device with storage, what are the instructions to make this secure.

Storage is used if available, so there's nothing to do.

Instructions to make it secure … it's tough because I'd rather not promise security: after all, none of the code has been audited, and keys are not stored in anything that has any memory protection, so it's not exactly what I'd call state of the art. But if at all, that'd probably better go into the book and not the example.

One thing I'd consider shifting around is the section in tests/coap/README on "using your own key". I'll move that over to the CoAP example; then, at the end of the example I can claim something like "if you've switched out the shipped key with one you generated on your own and configured server authentication (which is tedious unless your device has storage), you get the best that Ariel OS can deliver".

chrysn:
> What is needed to deploy this on a device with storage, what are the instructions to make this secure.

I've now moved the instructions for an own key into the main example (d5526ad606d423553d8faece6aed17d0b6d67668) and added a section to shape security expectations (1f5caebfb8ce5674beddb040f3729caf72615b4d), and did a bit more deduplication.

chrysn:
> Is it in a shape where it can be used by users to extend with their own handlers.

The updated README is rather verbose there. You mentioned in other places that the doc around coap-handler should be improved, and I fully agree. When that happens, maybe the example's README can shrink again and become a pointer there.

PR #2118, opened 2026-05-27
refactor(hal): unify define_peripherals!()

kaspar030:
# Description

`define_peripherals!()` had two copies only differing in one line since #1328.

I realized we'll need a fix also for the UART integration, so this is chopped out from that work.

## Testing

I sanity checked this with `examples/blinky` on `nrf52840dk`, and `examples/gpio` on the `espressif-esp32-c6-devkitc-1` (modified to not need leds), so both basic paths (esp and non-esp) are working.

Getting this to compile was fiddly macro language stuff. Now that it does I'm positive compile-testing is sufficient.

## Issues/PRs References

Fixing a TODO introduced in #1328.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -13,7 +13,6 @@
 ///
 // Inspired by https://github.com/adamgreig/assign-resources/tree/94ad10e2729afdf0fd5a77cd12e68409a982f58a
 // under MIT license
-#[cfg(not(context = "esp"))]
 #[macro_export]
 macro_rules! define_peripherals {
     (
@@ -31,8 +30,8 @@ macro_rules! define_peripherals {
         pub struct $peripherals {
             $(
                 $(#[$inner])*
-                pub $peripheral_name: $crate::hal::peripheral::Peri<'static, peripherals::$peripheral_field>
-            ),*
+                pub $peripheral_name: $crate::__peripheral_ty!($peripheral_field),
+            )*
         }
 
         $($(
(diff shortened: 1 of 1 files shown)


src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -53,46 +52,28 @@ macro_rules! define_peripherals {
     }
 }
 
-// This is almost identical to the version above, only the line that contains `Peri` is different.
-// TODO: unify
-#[cfg(context = "esp")]
+// This helper macro creates a peripheral type from it's name.
```
ROMemories:
```suggestion
// This helper macro creates a peripheral type from its name.
```

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -53,46 +52,28 @@ macro_rules! define_peripherals {
...
+// We need to variants: one for embassy hal `Peri`, one for the esp-hal style peripheral
```
ROMemories:
```suggestion
// We need two variants: one for embassy hal `Peri`, one for the esp-hal style peripheral
```

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -53,46 +52,28 @@ macro_rules! define_peripherals {
...
-                pub $peripheral_name: peripherals::$peripheral_field<'static>
-            ),*
-        }
-
-        $($(
-            #[allow(missing_docs, non_camel_case_types)]
-            pub type $peripheral_alias = peripherals::$peripheral_field;
-        )?)*
+#[doc(hidden)]
+macro_rules! __peripheral_ty {
```
ROMemories, line 66: The double-underscore prefix isn't really well-established (except for linker symbols), but I think it's fine in this case.

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -53,46 +52,28 @@ macro_rules! define_peripherals {
...
-// TODO: unify
-#[cfg(context = "esp")]
+// This helper macro creates a peripheral type from it's name.
+// We need to variants: one for embassy hal `Peri`, one for the esp-hal style peripheral
+// singletons.
+// This is split out of `define_peripherals` so the gating on `esp` is done at definition time in
+// this crate, and not at usage time, as that would make all crates using `define_peripherals` need
+// to add a check-cfg for `esp`.
+#[cfg(not(context = "esp"))]
 #[macro_export]
```
ROMemories, line 64: It seems these macros are not even importable from applications (unlike what I initially thought) as they are not part of the re-exported `ariel_os_hal::api` (which is good). It might be useful to mention that in a comment so this property doesn't get forgotten in future refactors.

ROMemories (approved):
LGTM after fixing the typos and ideally adding the extra comment.

src/ariel-os-hal/src/hal/define_peripherals.rs
kaspar030: I have added a comment.

kaspar030: ... and fixed it. 🙂

ROMemories (changes requested):
Some other changes seems to have ended up in this PR.

kaspar030:
> Some other changes seems to have ended up in this PR.

ups. (note to myself: use extra commits for ad-hoc testing)

ROMemories (approved):

PR #2119, opened 2026-05-27
feat(hal): Clarify crash from taken peripheral

dcz-self:
When the user claims the same peripheral in multiple tasks, the error message should help find the cause. Without this patch, it was a line number and:

```
called `Option::unwrap()` on a `None` value
```

# Description

".expect()" doesn't accept a format string, and I consider telling which exact resource caused this important. So I replaced it with an explicit "panic!()" which formats arguments.

## Testing

The blinky example, apart from the main task:

```
#[ariel_os::task(autostart, peripherals)]
pub async fn main(peripherals: Peripherals) {
```

have

```
#[ariel_os::task(autostart, peripherals)]
pub async fn crashy(peripherals: Peripherals) {}
```

and run it.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -45,7 +45,15 @@ macro_rules! define_peripherals {
                 $peripherals {
                     $(
                         $(#[$inner])*
-                        $peripheral_name: self.$peripheral_field.take().unwrap()
+                        $peripheral_name: match self.$peripheral_field.take() {
+                            Some(x) => x,
+                            None => panic!(
+                                "Peripheral {} ({}::{}) already taken",
+                                stringify!($peripheral_field),
+                                stringify!($peripherals),
+                                stringify!($peripheral_name),
+                            ),
+                        }
                     ),*
                 }
             }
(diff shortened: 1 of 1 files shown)


dcz-self:
> ac4f2ba: error Line is too long, 151 exceeds the max length of 80

I don't get which line this is referring to.

ROMemories:
> I don't get which line this is referring to.

This error is about the first line of the commit message, which exceeds the maximum allowed length.

ROMemories (commented):
Thank you for the PR! After discussing this with the team, we agree than the UX on this can be improved, but we are concerned by the impact on code size of the current proposition: for instance this results in 458 extra bytes on the `gpio` example on `nrf52840`, which is a bit too much for something that should never happen once the application is properly structured. Additionally, the impact on code size worsens with the number of peripherals.

To address the UX issue, we've come up with the following solution:

- The impact on code size mostly stems from the fact that every peripheral name needs to be stringified, so the solution would be to instead use an `.expect()` with a static string to reduce code size.
- To improve the debugging experience, a new Cargo feature (e.g., `debug-peripherals`, the exact name can be decided later) can be introduced, that would enable the more specific error message as currently proposed. That Cargo feature should be mentioned in the `expect()` message to be easily discoverable.
- Feature-gating *inside* the macro is not possible, as the feature-gate would be evaluated at the location where the macro is expanded (in the application), so a small dedicated macro should be introduced, similarly to what [#2118](https://github.com/ariel-os/ariel-os/pull/2118) did.

Could you evaluate that approach and measure the impact on code size on that example?

PR #2120, opened 2026-05-27
refactor(hal): make `define_peripherals!()`  use `$crate::hal::peripherals` internally

kaspar030:
# Description

All users of `define_peripherals!()` had to include `ariel_os_hal::hal::peripherals`, as `define_peripherals!()` internally uses `peripherals::` as prefix for peripherals.

This changes the macro to just use `$crate::hal::peripherals` directly.

WIP, depends on ~~#2118~~.
https://github.com/ariel-os/sbd/pull/153 is needed to remove "unused imports" warning.

Change looks scary, but it is essential two lines here, one line in sbd-gen, and the rest is fall-out.

The two lines here are in [`9d550c7` (this PR)](https://github.com/ariel-os/ariel-os/pull/2120/changes/9d550c79f6230286b2a19a16c537099e5195650f#r3313783233)

## Testing

Compilation should actually suffice here to check if anything broke.
Also, leftover includes might now warn. I've done some grep/cleanup on `tests/` and `examples/` to remove now obsolete imports, but may have missed some.

## Changelog Entry

When using `define_peripherals!()`, it is in most cases no longer necessary to import `ariel_os::hal::peripherals`.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -5,12 +5,6 @@
 /// the [`cfg`
 /// attribute](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute)),
 /// to define different setups for different boards.
-///
-/// # Note
-///
-/// The [`define_peripherals!`](crate::define_peripherals!) macro expects the
-/// `ariel_os::hal::peripherals` module to be in scope.
-///
 // Inspired by https://github.com/adamgreig/assign-resources/tree/94ad10e2729afdf0fd5a77cd12e68409a982f58a
 // under MIT license
 #[macro_export]
@@ -65,7 +59,7 @@ macro_rules! define_peripherals {
 #[doc(hidden)]
 macro_rules! __peripheral_ty {
     ($field:ident) => {
-        $crate::hal::peripheral::Peri<'static, peripherals::$field>
+        $crate::hal::peripheral::Peri<'static, $crate::hal::peripherals::$field>
     };
 }
 
@@ -74,7 +68,7 @@ macro_rules! __peripheral_ty {
 #[doc(hidden)]
 macro_rules! __peripheral_ty {
     ($field:ident) => {
-        peripherals::$field<'static>
+        $crate::hal::peripherals::$field<'static>
     };
 }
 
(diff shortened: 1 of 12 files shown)


src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -65,7 +65,7 @@ macro_rules! define_peripherals {
 #[doc(hidden)]
 macro_rules! __peripheral_ty {
     ($field:ident) => {
-        $crate::hal::peripheral::Peri<'static, peripherals::$field>
+        $crate::hal::peripheral::Peri<'static, $crate::hal::peripherals::$field>
```
kaspar030, line 62: This line and the corresponding one in line 77 are the actual change, the rest is fallout.

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -13,7 +13,6 @@
 ///
```
ROMemories: This should allow us to remove this note.
```suggestion
```

ROMemories (commented):
Haven't looked at the rest at all for now.

kaspar030:
TBC, as is this introduces build warnings from the generated board files:

```
warning: unused import: `ariel_os_hal::hal::peripherals`
 --> src/ariel-os-boards/src/nrf52dk.rs:4:9
  |
4 |     use ariel_os_hal::hal::peripherals;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default
```

These will need this sbd PR: with https://github.com/ariel-os/sbd/pull/153. As that is a breaking SBD change, I propose to live with the warnings here until we can bump sbd.

ROMemories (approved):
Successfully tested with:

```sh
laze -C examples/blinky/ build -b stm32u083c-dk run
```

This prints the "unused import" warning originating from the codegened `ariel-os-boards` as expected (which I think is fine for now).

This definitely makes the macro cleaner to use.

I did some quick grepping and couldn't find any other import to remove either.

---

I'm fine with landing this now, even with the warnings, as they are very innocuous and this makes the macro easier to use.

PR #2122, opened 2026-05-28
fix(deps): pin esp-hal minor version

### LINKED ISSUE #2117: esp-hal is not defined correctly as a dependency
### Summary

As per [our recommendation](https://github.com/esp-rs/esp-hal#support-policy), define esp-hal with the `~` operator to avoid auto bumping minor versions. Minor versions can and will contain breaking changes to unstable drivers - whilst unstable drivers don't follow semver, preventing breaking projects until the user is on the new HAL is preferable.

Experienced by a real ArielOS + esp-hal user here: https://github.com/esp-rs/esp-hal-community/issues/66.

### Relevant log output

```shell

```

### Hardware/ Setup

_No response_

### Possible Solution

_No response_

### Would you like to work on fixing this bug?

No

kaspar030:
# Description

As [reported](https://github.com/ariel-os/ariel-os/issues/2117) by upstream, it is recommended to pin the minor version of the `esp-hal` dependency, as we're using its `unstable` feature and that might otherwise lead to breaking changes even within minor version bumps.

Fixes #2117.

## Testing

Shouldn't change anything in-tree.

## Changelog Entry

The `esp-hal` dependency is now pinned to a specific minor version.

## Change Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,7 +106,7 @@ embedded-test = { version = "0.7.1", default-features = false, features = [
 
 esp-alloc = { version = "0.9.0", default-features = false }
 esp-bootloader-esp-idf = { version = "0.4.0", features = ["esp-rom-sys"] }
-esp-hal = { version = "1.0.0", default-features = false, features = [
+esp-hal = { version = "~1.0.0", default-features = false, features = [
   "unstable",
 ] }
 esp-metadata-generated = { version = "0.3.0", default-features = false }


ROMemories (approved):
I'm approving because this can't hurt but I'm not sure this is the right solution: shouldn't we be pinning (using `=`) every dependency we provide patches for anyway?

kaspar030:
> I'm approving because this can't hurt but I'm not sure this is the right solution: shouldn't we be pinning (using `=`) every dependency we provide patches for anyway?

Maybe. This is a slightly special case: The upstream `unstable` feature (which we opt-in) breaks the regular guarantees.

PR #2123, opened 2026-05-28
feat(boards): migrate to sbd schema with implicit led/button/uart names

kaspar030:
# Description

Updates the in-tree schema to sbd-gen 0.4.0.

- led/button defines
- version

Also, bumps the CI sbd-gen version to 0.4.0, and re-generates all files with that.

~~Depends on ariel-os/sbd#141.~~

~~Verified by running `sbd-gen generate-ariel --mode check` built from the PR-141 branch.~~

## Testing

~~Verified that the generated `src/ariel-os_boards` is byte-identical with this PR and ariel-os/sbd#141.~~ the new sbd-gen drops those obsolete imports now, otherwise identical.

## Changelog Entry

The in-tree board descriptions have been updated to sbd-gen 0.4.0.

## Change Checklist

### DIFF
--- a/boards/adafruit-feather-nrf52840-express.yaml
+++ b/boards/adafruit-feather-nrf52840-express.yaml
@@ -9,18 +9,16 @@ targets:
     flags:
       - has_usb_device_port
     leds:
-      # FIXME: led0/led1 names are purely synthetic to work with the current
-      # sbd::ariel convention of naming them sequentially.
-      led0:
-        color: red
+      # led0
+      - color: red
         pin: P1_15
         active: high
         aliases:
           - LED1 # bus name in schematic
           - D2 # LED part designator in schematic
           - D3 # silkscreen
-      led1:
-        color: blue
+      # led1
+      - color: blue
         pin: P1_10
         active: high
         aliases:
@@ -34,10 +32,8 @@ targets:
       # * orange D3 is hardwired to LiPo charge port STAT
 
     buttons:
-      # FIXME: "button0" name is purely synthetic to work with the current
-      # sbd::ariel convention of naming them sequentially.
-      button0:
-        pin: P1_02
+      # button0
+      - pin: P1_02
         # Note that there is no external pull-up.
         active: low
         aliases:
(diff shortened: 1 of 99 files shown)


ROMemories:
The porting guide from the book also needs to be updated.

boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -9,18 +9,16 @@ targets:
     flags:
       - has_usb_device_port
     leds:
-      # FIXME: led0/led1 names are purely synthetic to work with the current
-      # sbd::ariel convention of naming them sequentially.
```
chrysn, line 13: How will we carry that kind of information along?

AIU, we have the comment-on-generated-name there to stay greppable – and I am fine with doing this mechanical change now and only later going through the boards to set all the right aliases. When we do that, let's make sure to make "led0" an explicit alias for boards where that is an actual printed / documented name.

boards/makerdiary-nrf52840-mdk-usb-dongle.yaml
```diff
@@ -4,23 +4,22 @@ targets:
...
         pin: P0_24
         active: low
-    buttons:
-      # button0:
-      #   # note: this is configurable via UICR
-      #   pin: P0_18
-      #   active: low
+    # buttons:
+    #   - # note: this is configurable via UICR
+    #     pin: P0_18
```
chrysn, line 25:
```suggestion
    #   # button0
    #   # note: this is configurable via UICR
    #   - pin: P0_18
```

(for consistency)

chrysn (approved):
LGTM. Checked by scrolling through and looking in detail on everything highlighted by `git show kaspar/sbd-implicit-led-button-names --color-words='[^- #:]+'`.

AIU, this has no change w/rt which sbd-gen version we use, because we don't document or pin the currently used version; it is implicit that the latest is used.

Comments are non-blocking.

kaspar030:
> AIU, this has no change w/rt which sbd-gen version we use, because we don't document or pin the currently used version; it is implicit that the latest is used.

actually that needs to be part of this PR (update used SBD version, and update used schema version). Will add now.

ROMemories (approved):

PR #2124, opened 2026-05-28
docs(examples): point network instructions to the book

bergzand:
# Description

Save users a click by pointing them to the book from example readme 

## Testing

Click all the links!

## Open Questions

None

## Change Checklist

### DIFF
--- a/examples/coap-client/README.md
+++ b/examples/coap-client/README.md
@@ -10,7 +10,7 @@ On the CoAP client side, security is work in progress and not available for the 
 
 ## Running
 
-* [Set up networking](../README.md#networking).
+* [Set up networking](https://ariel-os.github.io/ariel-os/dev/docs/book/networking.html).
 * In one terminal, run `pipx run ./server.py`, which runs the server for the embedded device to interact with.
 * Run on any board with networking, eg. `laze build -b particle-xenon run`.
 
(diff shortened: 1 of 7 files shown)


nponsard:
I don't know if it's a good idea, a link relative to the current checkout will always be valid, where a link to the in-dev book may have differences between the checked out version and the latest version published.

One way to mitigate that would be to point to an historic version like https://ariel-os.github.io/ariel-os/0.4.0/docs/book/networking.html

bergzand:
> I don't know if it's a good idea, a link relative to the current checkout will always be valid, where a link to the in-dev book may have differences between the checked out version and the latest version published.

We have the same link already in the [readme in the examples dir](https://github.com/ariel-os/ariel-os/blob/main/examples/README.md#networking). 

> One way to mitigate that would be to point to an historic version like https://ariel-os.github.io/ariel-os/0.4.0/docs/book/networking.html

We don't do this currently in any other examples. I'm fine with changing that, but I would prefer to change that all together in a separate PR and add it to a post-release maintenance if we decide on this.

nponsard (approved):
> We don't do this currently in any other examples. I'm fine with changing that, but I would prefer to change that all together in a separate PR and add it to a post-release maintenance if we decide on this.

I feel like this kind of change needs to be discussed in the weekly meeting.

We can merge this today and change the links if needed after discussing it in the weekly.

PR #2125, opened 2026-05-28
chore(dep)!: update embassy-net

chrysn:
# Description

This drops [the one patch we had on increasing IP address logging verbosity to INFO](https://github.com/ariel-os/embassy/commit/f3ac56ee8fb8646d52520a1926c931944b0becb6): There has to be a better way of doing this than patching embassy-net.

(The other patch was bending paths, which has become moot when going with a released version).

## Testing

* Run your favorite network example.

## Issues/PRs References

This has SLAAAAAAAAAC!!!!!!!! See https://github.com/embassy-rs/embassy/pull/5623.

<small>(SLAAC for issue tracker internal SEO, but that doesn't convey my level of enthusiasm well enough).</small>

## Open Questions

<del>

* Do we use the opportunity to find that better way?
* Or do we just cherry-pick that branch into a new one and run the path-stripping script again?

</del>

[edit, added]: We went for the latter as the better way comes with its own caveats (details in https://github.com/ariel-os/ariel-os/pull/2125#issuecomment-4572456813 )

## Changelog Entry

`embassy-net` has been updated to v0.9.1: this is a breaking change as this crate is re-exported from `ariel_os::reexports`.

## Change Checklist


kaspar030:
> * Do we use the opportunity to find that better way?

Initially I had a small task for this (waiting on the link state channel provided by embassy-net), that works but uses a couple hundred bytes ram&flash. Better than outdated and patched dependency though.

chrysn:
I've added a small reporter here: on examples/coap-server for st-nucleo-wb55, it clocks in at 800 bytes text and 72 bytes BSS. Not great, not terrible -- options are:

- squeeze it
- try gating it behind LOG=info (which is not exactly well accessible)
- go back to patching

What should it be?

kaspar030:
> What should it be?

Run with it for now, try submitting the original patch upstream?

chrysn:
Works for me, I'll shape it up & adjust the comments & PR it (https://github.com/embassy-rs/embassy/pull/6195).

And then we may want to have a look at how to get that signal out, because the thing I'm hooking into here is *not* ideal: it's just level triggered "we have config" and "we have no config", but eg. in static-v6 + dhcp, it reports "on" immediately, and then gives no signal for "now v4 is here too". Not a big issue as long as our v6 is not really usable anyway, but as that's changing, worth doing something.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -409,6 +409,9 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
             unreachable!();
         }
 
+        #[cfg(any(feature = "defmt", feature = "log"))]
```
ROMemories: I think this can just be `logging` now, no need to check for individual logging facades (needs testing however):
```suggestion
        #[cfg(feature = "logging")]
```

src/ariel-os-embassy/src/net.rs
```diff
@@ -255,3 +255,58 @@ fn ariel_os_network_config() -> embassy_net::Config {
...
+///
+/// This is a workaround for <https://github.com/embassy-rs/embassy/pull/6195> that makes IP
+/// addresses visible at the "info" log level for the common case of users interacting with devices
+/// during development when no DNS integration or similar is set up. Once that issue is closed, it
+/// should be removed, also due to two shortcomings:
+///
+/// * It is relatively big (almost 1 KiB of code).
+/// * It does not report all configuration changes, but only the first change after the stack
+///   status goes from "no configuration" to "some configuration".
+#[cfg(any(feature = "defmt", feature = "log"))]
```
ROMemories: Same as above:
```suggestion
#[cfg(feature = "logging")]
```

ROMemories (commented):
Haven't looked at the rest.

src/ariel-os-embassy/src/lib.rs
chrysn: I took the condition from what triggered the compiler error, and that is this snippet of `src/ariel-os-log/src/lib.rs`:

```rust
// Define no-op macros in case no facade is enabled.
#[cfg(not(any(feature = "defmt", feature = "log")))]
mod log_macros {
    /// Logs a message at the trace level.
    #[macro_export]
    macro_rules! trace {
        ($($arg:tt)*) => {};
    }
```

And `logging` is not even a feature of ariel-os-embassy.

kaspar030:
(my brain parses "handlingo" as two words :wink:)

src/ariel-os-embassy/src/lib.rs
ROMemories: I thought you explicitly meant to feature-gate this on whether logging was enabled, but `logging` is indeed not even a Cargo feature of `ariel-os` :sweat_smile:

chrysn:
There's still a grave bug in this: the `wake_link_up()` / `wake_link_down()` function goes through an embassy-sync `WakerRegistration`, which has the unpleasant property of, while being formally correct in the sense that no wakeups are missed, causes a ping-pong wakeup between the tasks when different tasks keep waiting for it (effectively busy-looping albeit with yielding) -- and at least the flurry of trace level messages I got when using it indicates that this does indeed happen.

I'm looking into tools to avoid that, but it's the typical linked list territory :-/ – and maybe the practical resolution *is* to drop the task and cherry-pick a patch once more. (At least the thumbs-ups on https://github.com/embassy-rs/embassy/pull/6195 are not exclusively from here, but then again, I want a usable `wait_config_changed` even if we don't need it for debug output).

chrysn:
There are tools to do the wakeups better, but they're far from prime time (see https://github.com/embassy-rs/embassy/issues/3109 for the direction I think this should take). Thus reverting to the simpler path of using a patched-for-now-because-they-already-approved-the-change-for-main embassy-net.

ROMemories (approved):

```sh
laze -C examples/udp-echo/ build -d network-config-dhcp -b nrf52840dk run 
```

```sh
CONFIG_WIFI_NETWORK=<ssid> CONFIG_WIFI_PASSWORD=<pwd> laze -C examples/http-client/ build -b rpi-pico2-w run
```

---

This is a breaking change however, as `embassy-net` is re-exported as a whole from `ariel_os::reexports`: I've updated the changelog label, changelog entry, and the PR title.

Is there any feature/fix worth mentioning in the changelog entry brought by the upgrade itself? I suppose the SLAAC feature will be documented by #2126?

ROMemories:
It's unfortunate that the main change of that `embassy-net` version (the update of `smoltcp`) doesn't seem listed in the upstream changelog 😕

chrysn:
Indeed using SLAAC is for #2126 (with a lot of documentation updates). I am not aware of any relevant news item; the smoltcp update seems to be, as far as Ariel is concerned, "just" bug / compatibility fixes.

chrysn:
Had to rebase onto main to get https://github.com/ariel-os/ariel-os/pull/2133 in; other than that, the rebase was just autosquash.

PR #2126, opened 2026-05-28
feat(net): enable SLAAC

### LINKED ISSUE #501: Add SLAAC support to provide dynamic IPv6 addressing
### Description

Add SLAAC support to provide dynamic IPv6 addressing

### Motivation

SLAAC needed https://datatracker.ietf.org/doc/html/rfc4862

### Requirements

Dynamic IPv6 addressing support

### Open questions

_No response_

No

chrysn:
# Description

Our IPv6 story was previously static-only (that's not how it's supposed to be used), but @bergzand has driven updates all the way through smoltcp and embassy-net, and now it's there.

## Testing

[edit, added:]

* Pick any networking example.
* Be on a network where there are IPv6 prefixes being announced.
* Run with `-s network-config-ipv6-slaac`.
* Run the example's connection on the printed address, or a link-local address discovered by a ping to `ff02::1%yourinterface`.

## Issues/PRs References

Builds-on: https://github.com/ariel-os/ariel-os/pull/2125
Closes: https://github.com/ariel-os/ariel-os/issues/501

## Open Questions

* Users who don't have v6 networks or who are on a link where upstream doesn't offer PD (eg. when your router has no prefixes to give out and you're on a routed connection from your PC via USB Ethernet) will need to know their device's link-local addresses, as those are well usable, but can't easily find them because
  - there is no user-facing API in embassy-net to read the LL address, and
  - <del>ping responses are off by default so users can't just `ping ff02::1%ethusb0`</del> [edit: they can use that but it makes ugly READMEs]

## Changelog Entry

* SLAAC for configuring IPv6 addresses is now available as preview; it can be enabled using the newly introduced `network-config-ipv6-slaac`.

## Change Checklist

### DIFF
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -80,7 +80,7 @@ The configuration can be customized with the following environment variables:
 Support for IPv6 is not currently enabled by default, but can be enabled by selecting the `ipv6` [laze module](./build-system.md#laze-modules).
 IPv4 and IPv6 can both be enabled at the same time.
 
-IPv6 currently only supports static configuration, which is therefore enabled by default; it can also be explicitly selected with the `network-config-ipv6-static` [laze module](./build-system.md#laze-modules).
+IPv6 currently mainly supports static configuration, which is therefore enabled by default; it can also be explicitly selected with the `network-config-ipv6-static` [laze module](./build-system.md#laze-modules).
 
 The configuration must be customized with the following environment variables:
 
@@ -90,8 +90,12 @@ The configuration must be customized with the following environment variables:
 | `CONFIG_NET_IPV6_STATIC_CIDR_PREFIX_LEN` | `64`                        |
 | `CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS` | *No default, but mandatory* |
 
-> [!NOTE]
-> Non-static IPv6 address allocation will be supported in the future.
+There is experimental support for IPv6 configuration through SLAAC,
+which can be enabled using the `network-config-ipv6-slaac` [laze module].
+It is experimental due to shortcomings of the implementation (around DNS and privacy),
+and lack of documentation.
+Once the [associated issues](https://github.com/ariel-os/ariel-os/issues/2135) are resolved,
+it will become the default configuration for IPv6.
 
 #### Custom Configuration Provider
 
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1282,6 +1282,17 @@ modules:
         FEATURES:
           - ariel-os/network-config-ipv6-static
 
+  - name: network-config-ipv6-slaac
+    help: use SLAAC for IPv6 network configuration
+    selects:
+      - ipv6
+    provides_unique:
+      - network-config-ipv6
+    env:
+      global:
+        FEATURES:
+          - ariel-os/network-config-ipv6-slaac
+
   - name: ipv6
     selects:
       - network
--- a/src/ariel-os-embassy/Cargo.toml
+++ b/src/ariel-os-embassy/Cargo.toml
@@ -189,6 +189,7 @@ timer-generic-queue-64 = ["embassy-time?/generic-queue-64"]
 timer-generic-queue-128 = ["embassy-time?/generic-queue-128"]
 
 network-config-ipv4-static = []
+network-config-ipv6-slaac = ["embassy-net/slaac"]
 network-config-ipv6-static = []
 network-config-override = []
 override-usb-config = []
--- a/src/ariel-os-embassy/src/net.rs
+++ b/src/ariel-os-embassy/src/net.rs
@@ -253,5 +253,10 @@ fn ariel_os_network_config() -> embassy_net::Config {
         });
     }
 
+    #[cfg(feature = "network-config-ipv6-slaac")]
+    {
+        config.ipv6 = embassy_net::ConfigV6::Slaac;
+    }
+
     config
 }
--- a/src/ariel-os/Cargo.toml
+++ b/src/ariel-os/Cargo.toml
@@ -109,6 +109,8 @@ liboscore-provide-assert = ["ariel-os-coap/liboscore-provide-assert"]
 network-config-ipv4-static = ["ariel-os-embassy/network-config-ipv4-static"]
 # Selects static IPv6 configuration.
 network-config-ipv6-static = ["ariel-os-embassy/network-config-ipv6-static"]
+# Selects SLAAC based IPv6 configuration.
+network-config-ipv6-slaac = ["ariel-os-embassy/network-config-ipv6-slaac"]
 
 #! ## Serial communication
 ## Enables I2C support.


chrysn:
Rebased onto merge-ready #2125.

One concern that was fixed there was that pings were not responded to (they are now on by default again) -- but still, for a user it's unreasonably hard to find out which address to use. So some work will be required mostly for documentation.

The hardness is really twofold, depending on the setup:
- If upstream provides a subnet, even just a ULA, all is good.
- If upstream does not, there is no log message about the fe80::xxx link-local address that the device picks. And worse, but that's maybe on Linux: Then the user needs to know the interface name before they can even ping it.

This is exacerbated by the behavior of NetworkManager, when "shared" is selected (as is typical of downlink interfaces such as CDC-ECM or tap), to not pick up a ULA to announce. (This is unlike for example what OpenWRT does, which is to pick a ULA for a network unless there already is one).

kaspar030:
We discussed this morning whether we should merge in time for 0.5.0 (which hopefully gets out tomorrow). We'd like to test this more & document good, but also would like to make it available and get expsure. The conclusion was to not make it default, and document as "experimental".

laze-project.yml
```diff
@@ -1268,7 +1268,7 @@ modules:
   - name: network-config-ipv6-default
     help: use default network configuration method
     selects:
-      - ?network-config-ipv6-static
+      - ?network-config-ipv6-slaac
```
```suggestion
      - ?network-config-ipv6-static
```

Keep old default for now please.

kaspar030:
Please rebase! :slightly_smiling_face:

laze-project.yml
chrysn: Done.

chrysn:
Changed to not be default and pushed.

Switching off draft, but this will still need
- an issue that documents the be RC bugs in this (ie that
  - you can't find the address easily and
  - the docs have no good guidance on how to set up the network because
  - we don't know what to best recommend to NetworkManager users with legacy ISPs)
- documentation that it's not on by default due to issue TBD

(and I've just exhausted my time for today)

chrysn:
Also for that list:

- Check if privacy is on by default (I don't want Ariel to be one of the "IPv6 is bad bc of some bad apples' privacy" justifiers)

ROMemories (commented):
I tried this quickly:

- On `espressif-esp32-c6-devkitc-1`, the `udp-echo` works as expected: I'm echo to send message and get them echoed to me using the SLAAC-obtained IPv6 address. The IPv6 seems stable across reboots however. The `http-server` however fails wit the following error: `Error while sending an HTTP request: "Dns"`. I suppose that's because this requires obtaining DNS resolvers which are not currently obtained through NDP (unlike the gateway address), unlike what happens with DHCPv4? If this is correct, is the future plan to add support for these newer NDP advertisements or to add support for DHCPv6 (not for address assignment, but to obtain the list of DNS resolvers)?
- On `rpi-pico-w` and `rpi-pico2-w`, IPv6 apparently doesn't come up (but it seems I'm having the same issue on main so that might be unrelated).

bergzand:
> * On `espressif-esp32-c6-devkitc-1`, the `udp-echo` works as expected: I'm echo to send message and get them echoed to me using the SLAAC-obtained IPv6 address. The IPv6 seems stable across reboots however. The `http-server` however fails wit the following error: `Error while sending an HTTP request: "Dns"`. I suppose that's because this requires obtaining DNS resolvers which are not currently obtained through NDP (unlike the gateway address), unlike what happens with DHCPv4? If this is correct, is the future plan to add support for these newer NDP advertisements or to add support for DHCPv6 (not for address assignment, but to obtain the list of DNS resolvers)?

The [DNS part of SLAAC](https://datatracker.ietf.org/doc/html/rfc8106) is not yet supported in smoltcp, does that explain your dns resolver issue?

chrysn:
Added a commit that updates documentation, as well as testing steps in the PR docs. (But you already figured that out anyway.)

Sorry for the extra rebase, I was scrolled to an earlier "please rebase", and used that opportunity to fix missing DCOs.

kaspar030:
I tried this on a `rpi-pico-w` (`laze build -b rpi-pico-w -s network-config-ipv6-slaac run`), IPv6 stays `DOWN`. There is a prefix announced on that network, so I guess it should go ~~IP~~ UP at some point?

chrysn:
Tests I've done:
- on native with examples/coap-server: No ULA received, but found using mulitcast ping and `ip neigh`, and then pinged and aiocoap-client'ed.
- testing on other devices sent me into the rabbit hole of none of our devices being able to connect to my WiFi, hoping others can debug that better.
- (not recently tested: native with a radvd announced ULA)

chrysn:
On a USB link that I advertise, here's what I got:

```
[INFO ] IPv4: DOWN (embassy_net embassy-net/src/lib.rs:788)
[INFO ] IPv6: UP (embassy_net embassy-net/src/lib.rs:793)
[INFO ]    IP address:      fd91:abcd:ef55:0000:70c4:86ff:fe3d:0c9a/64 (embassy_net embassy-net/src/lib.rs:794)
[INFO ]    Default gateway: Some(fe80:0000:0000:0000:c988:8307:9d98:f671) (embassy_net embassy-net/src/lib.rs:795)
[INFO ] Starting up CoAP server (ariel_os_coap ariel-os-coap/src/transport_udp.rs:46)
[INFO ] Server is ready. (ariel_os_coap ariel-os-coap/src/transport_udp.rs:53)

[INFO ] IPv4: UP (embassy_net embassy-net/src/lib.rs:776)
[INFO ]    IP address:      10.42.1.175/24 (embassy_net embassy-net/src/lib.rs:777)
[INFO ]    Default gateway: Some(10.42.1.1) (embassy_net embassy-net/src/lib.rs:778)
[INFO ] IPv6: UP (embassy_net embassy-net/src/lib.rs:793)
[INFO ]    IP address:      fd91:abcd:ef55:0000:70c4:86ff:fe3d:0c9a/64 (embassy_net embassy-net/src/lib.rs:794)
[INFO ]    Default gateway: Some(fe80:0000:0000:0000:c988:8307:9d98:f671) (embassy_net embassy-net/src/lib.rs:795)
```

… but at a later run it did *not* work (but the IP/DOWN message came again??), so clearly, further testing is needed :'-(

[edit, taking some notes: might be a race condition; restarting radvd makes the IP UP/DOWN message show again, and if I time it right so that the RA comes before the DHCP, both come up. Aligns my older successful tests b/c there I didn't even have v4 on.]

kaspar030:
> I tried this on a `rpi-pico-w` (`laze build -b rpi-pico-w -s network-config-ipv6-slaac run`), IPv6 stays `DOWN`. There is a prefix announced on that network, so I guess it should go ~IP~ UP at some point?

I just re-tried this with `-s ipv4`, then it gets an address! So confirmed that it seems to race IPv4 network config.

book/src/networking.md
```diff
@@ -90,8 +90,12 @@ The configuration must be customized with the following environment variables:
 | `CONFIG_NET_IPV6_STATIC_CIDR_PREFIX_LEN` | `64`                        |
 | `CONFIG_NET_IPV6_STATIC_GATEWAY_ADDRESS` | *No default, but mandatory* |
 
-> [!NOTE]
-> Non-static IPv6 address allocation will be supported in the future.
+There is experimental support for IPv6 configuration through SLAAC,
+which can be enabled using the `network-config-ipv6-slaac` [laze module].
```
ROMemories, line 94: That link is otherwise broken:
```suggestion
which can be enabled using the `network-config-ipv6-slaac` [laze module][laze-modules-book].
```

kaspar030:
>> `laze build -b rpi-pico-w -s network-config-ipv6-slaac run`

> I just re-tried this with `-s ipv4`, then it gets an address! So confirmed that it seems to race IPv4 network config.

So the board does get an address now. `http-client` (as expected) fails at the DNS stage. `udp-echo` does not reply to pings, and does not respond via UDP.

kaspar030:
> `laze build -b nrf52840dk -s network-config-ipv6-slaac run`

ping & udp-echo does work on the nrf52840dk & usb-ethernet!

PR #2138, opened 2026-06-05
feat(ble): obtain the current address

### LINKED ISSUE #1666: BLE: add a way to access the current Address
### Description

Adding a function in Ariel OS that allows the application to retrieve the currently advertised BLE address.

We can either add modifications to `trouble` to allow that or manage it outside of `trouble` where we store the address we set separately.

### Motivation

The BLE address is used in BTHOME encryption process.

### Current Implementation

Current address is set at random, and it doesn't seem like there is a way to get it anywhere. `ReadBdAddr` hci command is not a solution, it only returns the public address if available.

Yes

nponsard:
# Description

This adds a way to retrieve the currently configured BLE address.

## Testing

```
laze -C examples/ble-advertiser build -b nrf52840dk
``` 

Should print the current random address at the start.

## Issues/PRs References

- Fixes #1666

## Changelog Entry

The current BLE address can now be accessed using `ariel_os::ble::current_address()`.

## Change Checklist

### DIFF
--- a/book/src/bluetooth.md
+++ b/book/src/bluetooth.md
@@ -23,7 +23,9 @@ The ability to configure which Bluetooth address is used and other capacity para
 >
 > Current implementation: the address is a static device address and is not rotated during execution.
 > This allows to use the BLE feature of Ariel OS on multiple devices in the same location.
-> We later plan to switch to private device addresses by default, which *are* rotated during execution.
+> We later plan to switch to private device addresses by default, which _are_ rotated during execution.
+>
+> To obtain the address at runtime, use `ariel_os::ble::current_address()`.
 
 ## Using the BLE Stack
 
--- a/book/src/bluetooth.md
+++ b/book/src/bluetooth.md
@@ -53,6 +53,8 @@ embassy_futures::join::join(host.runner.run(), async {
 }).await;
 ```
 
+To obtain the address at runtime, use [`ariel_os::ble::current_address()`](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/ble/fn.current_address.html).
+
 See [the Ariel OS examples](https://github.com/ariel-os/ariel-os/tree/main/examples) for more details.
 
 [laze-modules-book]: ./build-system.md#laze-modules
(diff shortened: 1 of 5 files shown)


book/src/bluetooth.md
```diff
@@ -23,7 +23,7 @@ The ability to configure which Bluetooth address is used and other capacity para
 >
 > Current implementation: the address is a static device address and is not rotated during execution.
 > This allows to use the BLE feature of Ariel OS on multiple devices in the same location.
-> We later plan to switch to private device addresses by default, which *are* rotated during execution.
+> We later plan to switch to private device addresses by default, which _are_ rotated during execution.
```
ROMemories: That seems unrelated, and we *mostly* use stars for italics instead of underscores.

src/ariel-os-embassy/src/ble.rs
```diff
@@ -20,6 +22,8 @@ use ariel_os_log::debug;
 // Must be async and return &trouble_host::Stack<'static, impl Controller>
 pub use crate::hal::ble::ble_stack;
 
+static CURRENT_ADDRESS: OnceLock<Address> = OnceLock::new();
```
ROMemories, line 25: I suppose that `OnceLock` will later be replaced by a mutex when we have Private Random Addresses, but the `static` is private so that's totally fine for now.

src/ariel-os-embassy/src/ble.rs
```diff
@@ -33,11 +37,19 @@ pub(crate) fn config() -> Config {
...
     };
 
+    let _ = CURRENT_ADDRESS.init(address);
+
     debug!("Setting random address: {:?}", address);
 
     Config { address }
 }
 
+/// Returns the BLE address of the BLE adapter.
```
ROMemories: I don't think that's the address of the *BLE adapter*, that would be the Public Address. The phrasing should also allow for the address in use to change over time (which the function name does allow); maybe:


```suggestion
/// Returns the BLE address currently in use.
///
/// Note that the BLE address may be rotated over time.
```

ROMemories (commented):

```sh
laze -C examples/ble-advertiser/ build -b espressif-esp32-c6-devkitc-1 run
```

I do get different address when running the application multiple times.

book/src/bluetooth.md
nponsard: Ah this shouldn't have been changed, seems like a formatting tool changed it.

ROMemories (approved):
LGTM, please squash and DCO!

PR #2147, opened 2026-06-16
fix: check if peripheral is already used

### LINKED ISSUE #2145: UART 'singleton' doesn't prevent from using the same peripherals twice concurently
### Summary

You shouldn't be able to use the same UART peripheral with two different pairs of RX/TX pins. Right now Ariel OS allows that.

Using this reproducer: 

```diff
diff --git a/tests/uart-loopback/src/main.rs b/tests/uart-loopback/src/main.rs
index 42045193..791a9d3d 100644
--- a/tests/uart-loopback/src/main.rs
+++ b/tests/uart-loopback/src/main.rs
@@ -35,16 +35,36 @@ async fn main(peripherals: pins::Peripherals) {
     )
     .expect("Invalid UART configuration");
 
+    let mut rx_buf2 = [0u8; 32];
+    let mut tx_buf2 = [0u8; 32];
+
+    let mut uart2 = pins::TestUart::new(
+        peripherals.uart_rx2,
+        peripherals.uart_tx2,
+        &mut rx_buf2,
+        &mut tx_buf2,
+        config,
+    )
+    .expect("Invalid UART configuration");
+
     const OUT: &str = "Test Message";
+    const OUT2: &str = "Test 22222";
+
     let mut input = [0u8; OUT.len()];
+    let mut input2 = [0u8; OUT2.len()];
 
     uart.write_all(OUT.as_bytes()).await.unwrap();
+    uart2.write_all(OUT2.as_bytes()).await.unwrap();
     uart.flush().await.unwrap();
+    uart2.flush().await.unwrap();
     info!("Wrote bytes");
+    let _ = with_timeout(Duration::from_secs(5), uart2.read_exact(&mut input2)).await;
     let _ = with_timeout(Duration::from_secs(5), uart.read_exact(&mut input)).await;

 
     info!("Got: {}", Hex(input));
     assert_eq!(OUT.as_bytes(), input);
+    info!("Got: {}", Hex(input2));
+    assert_eq!(OUT2.as_bytes(), input2);
     info!("Test passed!");
 
     exit(ExitCode::SUCCESS);
diff --git a/tests/uart-loopback/src/pins.rs b/tests/uart-loopback/src/pins.rs
index 9354ab73..87e33f66 100644
--- a/tests/uart-loopback/src/pins.rs
+++ b/tests/uart-loopback/src/pins.rs
@@ -30,6 +30,8 @@ pub type TestUart<'a> = uart::UARTE0<'a>;
 ariel_os::hal::define_peripherals!(Peripherals {
     uart_rx: P1_01,
     uart_tx: P1_02,
+    uart_rx2: P1_03,
+    uart_tx2: P1_04,
 });
 
 #[cfg(context = "nrf5340-app")]
```

Edit: read of the second UART needs to happen first to catch the error.

With the pin P1_01 bridged to P1_02 and P1_03 bridged to P1_04,
```
laze build -C tests/uart-loopback/ -b nrf52840dk run
```
Does compile and instead return a runtime error at the assert part (first UART pair appearing to have the results from the second pair).

### Relevant log output

```shell
[INFO ] Starting UART test (uart_loopback uart-loopback/src/main.rs:20)
[INFO ] Selected configuration: Config { baudrate: _115200, data_bits: Data8, stop_bits: Stop1, parity: None } (uart_loopback uart-loopback/src/main.rs:24)
[INFO ] Wrote bytes (uart_loopback uart-loopback/src/main.rs:60)
[INFO ] Got: [67, 65, 54, 65, 73, 74, 20, 32, 32, 32, 32, 32] (uart_loopback uart-loopback/src/main.rs:64)
panicked at tests/uart-loopback/src/main.rs:65:5:
assertion `left == right` failed
  left: [84, 101, 115, 116, 32, 77, 101, 115, 115, 97, 103, 101]
 right: [103, 101, 84, 101, 115, 116, 32, 50, 50, 50, 50, 50]
```

### Hardware/ Setup

nrf52840dk with the pin P1_01 bridged to P1_02 and P1_03 bridged to P1_04

### Possible Solution

_No response_

### Would you like to work on fixing this bug?

Maybe

nponsard:
# Description

Currently an application can create multiple instances of the same UART/I2C/SPI peripheral. This is because internally `::steal()` is used on the underlying HAL peripheral.

There is a safety comment claiming that the Ariel OS struct representing an UART/I2C/SPI peripheral works as a singleton, but this isn't actually enforced. This PR fixes that by panicking if another instance of the peripheral is already active.  

## Testing

[I created a branch](https://github.com/nponsard/ariel-os/tree/test/uart-reproducer) with reproducers for UART/SPI/I2C.

- `tests/uart-loopback`: panics in the assert because it didn't receive the correct message. Should now panic at initialization. 
- `tests/spi-loopback`:  panics in the assert because it didn't receive the correct message. Should now panic at initialization.
- `tests/i2c-controller`: (stm32u083c) manages to read the same WHOAMI from different pins (does it use both pairs of pins at once ?). Should now panic at initialization. 

Tested spi on `nrf52840dk`, `stm32u083c-dk`, `rpi-pico-w`, `seeedstudio-xiao-esp32c6`.
UART on `nrf52840dk`, `stm32u083c-dk`, `rpi-pico-w`, `seeedstudio-xiao-esp32c6`.
I2C on `stm32u083c-dk`.

## Issues/PRs References

- Fixes #2145 

## Changelog Entry

It is no longer possible to mistakenly create multiple driver instances for the same I2C/SPI/UART peripheral by providing different pins to them.

## Change Checklist

### DIFF
--- a/src/ariel-os-esp/src/i2c/controller/mod.rs
+++ b/src/ariel-os-esp/src/i2c/controller/mod.rs
@@ -2,6 +2,8 @@
 
 #![expect(unsafe_code)]
 
+use portable_atomic::{AtomicBool, Ordering};
+
 use ariel_os_embassy_common::impl_async_i2c_for_driver_enum;
 use esp_hal::{
     Async,
@@ -93,6 +95,11 @@ macro_rules! define_i2c_drivers {
                 twim: EspI2c<'static, Async>,
             }
 
+            // Ensure this peripheral has only one active Instance.
+            paste::paste! {
+                static [< ACTIVE_ $peripheral >]: AtomicBool = AtomicBool::new(false);
+            }
+
             impl $peripheral {
                 /// Returns a driver implementing [`embedded_hal_async::i2c::I2c`] for this
                 /// I2C peripheral.
@@ -121,9 +128,16 @@ macro_rules! define_i2c_drivers {
                             BusTimeout::Maximum
                             );
 
+                    // Check if we can initialize this peripheral (check if the value was previously false, set it to true).
+                    paste::paste! {
+                        if [< ACTIVE_ $peripheral >].swap(true, Ordering::AcqRel) {
+                            panic!("I2C peripheral already initialized")
+                        }
+                    }
+
                     // FIXME(safety): enforce that the init code indeed has run
-                    // SAFETY: this struct being a singleton prevents us from stealing the
-                    // peripheral multiple times.
+                    // SAFETY: We check with an AtomicBool that only one instance of this peripheral
+                    // is active at once.
                     let i2c_peripheral = unsafe { peripherals::$peripheral::steal() };
 
                     let twim = EspI2c::new(
@@ -136,6 +150,14 @@ macro_rules! define_i2c_drivers {
                         .with_scl(scl_pin.into_hal_peripheral());
 
                     I2c::$peripheral(Self { twim })
+                }
+            }
+
+            impl Drop for $peripheral {
+                fn drop(&mut self) {
+                    paste::paste! {
+                        [< ACTIVE_ $peripheral >].store(false, Ordering::Release);
+                    }
                 }
             }
         )*
--- a/src/ariel-os-esp/src/uart.rs
+++ b/src/ariel-os-esp/src/uart.rs
@@ -2,6 +2,8 @@
 
 #![expect(unsafe_code)]
 
+use portable_atomic::{AtomicBool, Ordering};
+
 use ariel_os_embassy_common::{impl_async_uart_for_driver_enum, uart::ConfigError};
 
 use esp_hal::{
@@ -182,6 +184,11 @@ macro_rules! define_uart_drivers {
                 static [<PREVENT_MULTIPLE_ $peripheral>]: () = ();
             }
 
+            // Ensure this peripheral has only one active Instance.
+            paste::paste! {
+                static [< ACTIVE_ $peripheral >]: AtomicBool = AtomicBool::new(false);
+            }
+
             impl<'d> $peripheral<'d> {
                 /// Returns a driver implementing embedded-io traits for this Uart
                 /// peripheral.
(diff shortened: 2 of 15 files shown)


src/ariel-os-esp/src/i2c/controller/mod.rs
```diff
@@ -121,9 +128,16 @@ macro_rules! define_i2c_drivers {
                             BusTimeout::Maximum
                             );
 
+                    // Check if we can initialize this peripheral (check if the value was previously false, set it to true).
+                    paste::paste! {
+                        if [< ACTIVE_ $peripheral >].swap(true, Ordering::SeqCst) {
```
ROMemories: Unless I missed something, I think these can simply use `AcqRel` ordering: we do need the load to use `Acquire` ordering and the storage to use `Release`, but we don't need [all the operations to all `ACTIVE_*` variables (not just this one) to have a single total order](https://mara.nl/atomics/memory-ordering.html#seqcst:~:text=This%20means%20that%20every%20single%20operation%20using%20SeqCst%20ordering%20within%20a%20program%20is%20part%20of%20a%20single%20total%20order%20that%20all%20threads%20agree%20on%2E):
```suggestion
                        if [< ACTIVE_ $peripheral >].swap(true, Ordering::AcqRel) {
```

src/ariel-os-nrf/src/uart.rs
```diff
@@ -243,28 +250,36 @@ macro_rules! define_uart_drivers {
                     tx_buffer: &'d mut [u8],
                     config: Config,
                 ) -> Result<Uart<'d>, ConfigError> {
+
```
ROMemories: optional:
```suggestion
```

src/ariel-os-rp/src/uart.rs
```diff
@@ -192,6 +199,7 @@ macro_rules! define_uart_drivers {
                     tx_buf: &mut [u8],
                     config: Config,
                 ) -> Result<Uart<'d>, ConfigError> {
+
```
ROMemories: optional:
```suggestion
```

src/ariel-os-esp/src/uart.rs
```diff
@@ -208,23 +215,44 @@ macro_rules! define_uart_drivers {
...
 
                     let uart = EspUart::new(
                         uart_peripheral,
                         uart_config
                     )
-                        .map_err(convert_error)?
+                        .map_err(|e| {
+                            // Turns out we didn't initialize it.
+                            paste::paste! {
+                                [< ACTIVE_ $peripheral >].store(false, Ordering::Release);
```
ROMemories, line 237: Indeed, it should be freed up in this case 👍

ROMemories (commented):
Successfully tested the following with the reproducers:

```sh
laze -C tests/uart-loopback/ build -b espressif-esp32-c6-devkitc-1 run
```

which panics (as expected) with:

```sh
panicked at src/ariel-os-esp/src/uart.rs:279:1:
UART peripheral already initialized
```

and

```sh
laze -C tests/i2c-controller/ build -b stm32u083c-dk run
```

which panics with:

```sh
panicked at src/ariel-os-stm32/src/i2c/controller/mod.rs:285:1:
I2C peripheral already initialized
```

(I've also tested this with `AcqRel` ordering instead of `SeqCst`)

I've also tested that the following still worked as expected:

```sh
```

Trusting your testing on the rest.

src/ariel-os-esp/src/i2c/controller/mod.rs
nponsard: Ah yes that would be better, thanks

ROMemories (approved):
LGTM, please squash and DCO!

PR #2148, opened 2026-06-16
docs(book): add a page about the optional global allocator

ROMemories:
# Description

Following #1422, this adds a book page about the global allocator, and in particular documents `heapsize_required`.

## Changelog Entry

Documentation about the global allocator has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/global-allocator.md
+++ b/book/src/global-allocator.md
@@ -1,0 +1,63 @@
+# Global Allocator
+
+Ariel OS provides an optional, disabled-by-default [global heap allocator][global-alloc-alloc-std-docs], which can be enabled with the `alloc` [laze module][laze-modules-book].
+
+Using a heap allocator is usually avoided in embedded applications since, by nature, it introduces memory fragmentation and results in runtime errors when running out of memory.
+Therefore it should only be enabled when actually necessary.
+Use cases include:
+
+- Using a dependency targeting embedded explicitly requiring a dynamic allocator
+- Using a [`no_std`][no-std-attr-rust], but not no-`alloc` (transitive) dependency that has not been written with embedded in mind
+- Experimenting with [`alloc`][alloc-std-docs]-provided types (e.g., [`Vec<T>`][vec-alloc-std-docs])
+
+Some functionality of Ariel OS may depend on a heap allocator, in which case it is automatically enabled.
+
+## Using the `alloc` Crate
+
+The global allocator allows using the [`alloc` crate][alloc-std-docs] (part of the standard library).
+It can be imported using the following:
+
+```rust
+extern crate alloc;
+```
+
+Its items can then be imported as usual.
+
+> [!TIP]
+> When possible, consider using the [`heapless` crate][heapless-docsrs] instead, which provides similar types without dynamic allocation, by statically allocating a memory pool (on the stack).
+
+## Sizing the Heap
+
+The heap is allocated in the space left at the end of the RAM.
+See [Ariel OS memory layout][memory-layout-book].
+
+Applications must adjust the value of the laze variable `heapsize_required` with the size (in bytes) that they require:
+
+```yaml
+apps:
+  - name: example-app
+    env:
+      global:
+        heapsize_required:
+          - "32768"
+```
+
+`heapsize_required` must be a YAML sequence of strings.
+The value given will be added to the variable, *increasing* the size of the heap by the given amount.
+The `heapsize_required` variable is initialized to a non-zero value, which might suffice to some applications; it is however strongly recommended to still specify the required heap size explicitly.
+Additionally, Ariel OS functionalities that require a larger heap may automatically increase the value of `heapsize_required` as needed.
+
+> [!TIP]
+> When the requested heap size does not fit in RAM, the build will fail at link time.
+
+Memory exhaustion results in the application being terminated.
+Depending on the allocator and on the item initiating the allocation (e.g., [`Vec::push()`][vec-push-alloc-std-docs]), the application will either panic or abort.
+
+[global-alloc-alloc-std-docs]: https://doc.rust-lang.org/stable/alloc/alloc/trait.GlobalAlloc.html
+[laze-modules-book]: ./build-system.md#laze-modules
+[no-std-attr-rust]: https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute
+[alloc-std-docs]: https://doc.rust-lang.org/stable/alloc/
+[vec-alloc-std-docs]: https://doc.rust-lang.org/stable/alloc/vec/struct.Vec.html
+[heapless-docsrs]: https://docs.rs/heapless/latest/heapless/
+[memory-layout-book]: ./memory-layout.md
+[vec-push-alloc-std-docs]: https://doc.rust-lang.org/stable/alloc/vec/struct.Vec.html#method.push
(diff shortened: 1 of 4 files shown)


src/ariel-os/Cargo.toml
```diff
@@ -45,7 +45,7 @@ static_cell = { workspace = true }
 default = ["ariel-os-rt/_panic-handler"]
 
 #! ## System functionality
-## Enables a global system allocator.
+# Enables a global system allocator.
```
ROMemories: This hides the Cargo feature from the docs as the laze module should be used instead, otherwise the environment variable sizing the heap is not injected.

book/src/memory-layout.md
```diff
@@ -19,7 +19,7 @@ Additionally, no stability guarantees are currently provided regarding the memor
 - The thread stacks, if [multithreading][multithreading-book] is enabled, are currently allocated as individual `static`s, anywhere in the `.bss` section.
 - If [multithreading][multithreading-book] is enabled, and if the MCU has two symmetrical cores and [`multi-core`][multicore-support-book] is enabled, the ISR stack of the second core is allocated as a `static`, anywhere in the `.bss` section.
 - Depending on the architecture, the uninitialized section is either called `.uninit` or `.noinit`.
-- The heap, if enabled, may take up to the remaining space.
+- The [heap][global-allocator-book], if enabled, may take up to the remaining space.
   `heapsize_required` ensures that there is enough space for the heap, at link time.
```
ROMemories, line 23: We might want to simply remove this sentence now, as it is also mentioned in the new page (but it's also fine to keep it).

book/src/global-allocator.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+The value given will be added to the variable, *increasing* the size of the heap by the given amount.
+The `heapsize_required` variable is initialized to a non-zero value, which might suffice to some applications; it is however strongly recommended to still specify the required heap size explicitly.
+Additionally, Ariel OS functionalities that require a larger heap may automatically increase the value of `heapsize_required` as needed.
+
+> [!TIP]
+> When the requested heap size does not fit in RAM, the build will fail at link time.
+
+Depending on the allocator and on the item initiating an allocation (e.g., [`Vec::push()`][vec-push-alloc-std-docs]), memory exhaustion may result in the application panicking or aborting.
```
ROMemories: Is there anything else to say about memory exhaustion?

book/src/global-allocator.md
```diff
@@ -0,0 +1,61 @@
...
+
```
ROMemories, line 55: We might also want to document which allocator implementations are currently used (without stability guarantees), for instance [the TLSF implementation of `embedded-alloc`](https://docs.rs/embedded-alloc/0.7.0/embedded_alloc/struct.TlsfHeap.html) (except on ESP32s), but this can be done in a follow-up (if at all).

book/src/global-allocator.md
```diff
@@ -0,0 +1,61 @@
...
+Ariel OS provides an optional, disabled-by-default [global heap allocator][global-alloc-alloc-std-docs], which can be enabled with the `alloc` [laze module][laze-modules-book].
+
+Using a heap allocator is usually avoided in embedded applications since, by nature, it introduces memory fragmentation and may result in runtime errors when running out of memory.
+Therefore it should only be enabled when actually necessary.
+Use cases include:
+
+- Using a dependency targeting embedded explicitly requiring a dynamic allocator
+- Using a [`no_std`][no-std-attr-rust], but not no-`alloc` (transitive) dependency that has not been written with embedded in mind
+- Experimenting with [`alloc`][alloc-std-docs]-provided types (e.g., [`Vec<T>`][vec-alloc-std-docs])
```
ROMemories, line 11: Any other use case I've missed?

src/ariel-os/Cargo.toml
ROMemories: This would actually be a breaking change (as the build would not fail if not env var is not provided, because of the default value), so I've removed this from the PR.

book/src/global-allocator.md
```diff
@@ -0,0 +1,61 @@
...
+
+> [!TIP]
+> When possible, consider using the [`heapless` crate][heapless-docsrs] instead, which provides similar types without dynamic allocation, by statically allocating a memory pool (on the stack).
+
+## Sizing the Heap
+
+The heap is allocated in the space left at the end of the RAM.
+See [Ariel OS memory layout][memory-layout-book].
+
+Applications must adjust the value of the laze variable `heapsize_required` with the size (in bytes) that they require:
```
kaspar030, line 34: This should mention that a laze list *must* be used.

book/src/global-allocator.md
kaspar030: maybe just that upstream Rust default is to panic on OOM.

book/src/memory-layout.md
kaspar030: I'd keep it.

book/src/global-allocator.md
kaspar030: fine list for now

book/src/global-allocator.md
kaspar030: follow-up

kaspar030 (changes requested):

book/src/global-allocator.md
ROMemories: I've added a sentence mentioning that a YAML sequence must be used (and that the value must be a string); I didn't want to introduce the concept of laze list as it's not part of either our documentation or laze's.

book/src/global-allocator.md
ROMemories: What do you mean by "upstream Rust"? My understanding is that behavior is allocator-dependent (I think [`GlobalAlloc` allows both](https://doc.rust-lang.org/stable/core/alloc/trait.GlobalAlloc.html)), and with `no_std` there is no default global allocator.

kaspar030: What I mean is that "Depending on the allocator and on the item initiating an allocation (e.g., [`Vec::push()`][vec-push-alloc-std-docs]), memory exhaustion may result in the application panicking or aborting." kinda makes it sound like this is some remote possiblity for some subset of the "items initiating an allocation", but the reality is that only a small subset of "items initiating an allocation" can actually handle OOMs, and that *memory exhaustion most likely causes panicking*. And that this is a deliberate upstream (`std::/alloc::`) design decision, because on non-embedded (desktop CPUs with gigs of RAM and MMU), memory is freely overcommitted, and OOMing applications gracefully die.
This is not something that Ariel OS can fix, and nothing that any `GlobalAlloc` can prevent, without basically also writing all alloc users with fallible allocation in mind.

kaspar030: "it introduces memory fragmentation and may result in runtime errors when running out of memory" -> OOM will almost certainly result in "runtime errors", "may" is just not strong enough

ROMemories: I've removed both "may"s, and rephrased this small paragraph to make this clearer. I think I meant that it may result in either a panic or an abort, not that termination was only unlikely.

> _memory exhaustion most likely causes panicking_

My understanding is that this isn't exactly true, it may also result in an abort: [allocators are *allowed* to themselves abort on memory exhaustion](https://doc.rust-lang.org/stable/alloc/alloc/trait.GlobalAlloc.html#:~:text=Implementations,exhaustion%2E%29).

kaspar030 (approved):

book/src/global-allocator.md
kaspar030: Yeah, abort or panic - in Ariel, that'll probably have the same result. anyhow, the new phrasing is fine.

PR #2149, opened 2026-06-17
docs(ariel-os)!: hide the `alloc` Cargo feature from the docs

ROMemories:
# Description

Following https://github.com/ariel-os/ariel-os/pull/2148#discussion_r3426113007, this hides the `alloc` Cargo feature from the docs, so the `alloc` laze module is used instead, otherwise the `CONFIG_HEAPSIZE` environment variable would not be provided to size the heap (even when the `heapsize_required` laze variable is set somewhere).

## Issues/PRs References

- Depends on #2148

## Changelog Entry

The `alloc` Cargo feature has been removed from the documentation: the `alloc` laze module should be used instead to enable the global allocator.

## Change Checklist

### DIFF
--- a/src/ariel-os/Cargo.toml
+++ b/src/ariel-os/Cargo.toml
@@ -45,7 +45,7 @@ static_cell = { workspace = true }
 default = ["ariel-os-rt/_panic-handler"]
 
 #! ## System functionality
-## Enables a global system allocator.
+# Enables a global system allocator.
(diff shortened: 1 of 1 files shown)


kaspar030 (approved):
LGTM.

kaspar030:
I wonder if we can fail at build time when cargo features are enabled that also do require laze machinery. Like, for those, set some `--cfg` and fail if the feature is enabled but the cfg is not set.

PR #2150, opened 2026-06-18
docs(book): add a page about USB support

ROMemories:
# Description

This introduces a book page about USB support. Currently only the USB peripheral role is supported. The documentation surfaces the dependency on `embassy-usb` as it is not an implementation detail, and links to our re-export as much as possible.

Other USB concepts (like endpoints) could be explained in that page as well, but this can be done (way) later.

## Changelog Entry

Documentation about USB support has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/application.md
+++ b/book/src/application.md
@@ -49,8 +49,6 @@ Functions can currently be registered as either `spawner`s or `task`s:
   They are provided with a [`Spawner`](https://docs.embassy.dev/embassy-executor/git/cortex-m/struct.Spawner.html) instance and can therefore be used to [`spawn`](https://docs.embassy.dev/embassy-executor/git/cortex-m/struct.Spawner.html#method.spawn) other `async` tasks.
 - [`task` functions][task-attr-docs] are `async` functions that are statically allocated at compile-time.
   They are especially useful for long-running, `async` tasks.
-  They must also be used to use *Ariel OS configuration hooks*, which can be requested with their associated macro parameter, and allow to provide configuration during boot.
-  Please refer to the documentation of [`task`][task-attr-docs] for a list of available hooks and to [Configuration Hooks](#configuration-hooks) to know more about hook usage.
 
 Both of these can be provided with an instance of an Ariel OS peripheral struct when needed, using the `peripherals` macro parameters (see the macros' documentation) and taking that Ariel OS peripheral struct as parameter.
 
@@ -72,10 +70,6 @@ async fn blinky(peripherals: pins::LedPeripherals) {
     }
 }
 ```
-
-## Configuration Hooks
-
-TODO
 
 [embassy-style-hals]: ./glossary.md#embassy-style-hals
 [ariel-os-hals]: ./glossary.md#ariel-os-hals
--- a/book/src/usb.md
+++ b/book/src/usb.md
@@ -1,0 +1,75 @@
+# Universal Serial Bus (USB)
+
+Ariel OS integrates support for USB peripherals built into many microcontrollers.
+
+## Hardware Support
+
+Many microcontrollers supported by Ariel OS include a USB peripheral that can be leveraged to build any USB device.
+At the time of writing, most of them support USB 2.0, but their signaling rate is often limited to that of Full-Speed USB (FS), i.e., 12 Mbit/s (now also known as Basic-Speed USB), even though some also do support High-Speed USB, i.e., 480 Mbit/s (marketed as Hi-Speed USB).
+Some also support USB On-The-Go (OTG), allowing the device to also behave as a USB Targeted Host on the same USB receptacle (by alternatively switching between the peripheral and host roles).
+Currently only the USB peripheral role is supported by Ariel OS.
+
+These USB microcontroller peripherals are "generic," in that they can be used to implement any USB device class.
+Some microcontrollers also feature USB microcontroller peripherals that only support a fixed set of USB classes: e.g., multiple ESP32 MCUs comprise a USB CDC-ACM/JTAG peripheral, which can only be used for the standard [USB CDC-ACM][usb-cdc-acm-book-glossary] device class or a vendor-specific device class implementing JTAG access over USB.
+Currently, only the "generic" USB peripherals are supported in Ariel OS applications.
+The others may still be integrated by Ariel OS to implement specific functionality, like [logging][logging-transports-book].
+
+> [!TIP]
+> Development kits that support USB often feature two USB receptacles: one for the onboard [debug probe][debug-probes-book] (if there is one), and the other connected to the USB peripheral of the microcontroller.
+> That second USB connection is usually called "user USB" to differentiate it from that of the debug probe and is the one that must be connected to the computer acting as USB host.
+
+## Software Integration
+
+Ariel OS provides support for the USB peripheral role through [`embassy-usb`][embassy-usb-docsrs], which provides a consistent API across the supported hardware.
+It can be enabled with the `usb` Cargo feature.
+
+An instance of [`embassy_usb::Builder`][ariel-os-usbbuilder-rustdoc] is created by Ariel OS, on which support for well-known USB device classes can be added using a dedicated [Ariel OS task hook][task-attr-docs]:
+
+```rust
+#[ariel_os::task(autostart, usb_builder_hook)]
+async fn main() {
+    let mut usb_class = USB_BUILDER_HOOK
+        .with(|builder| {
+            // USB class constructor that mutates the builder.
+        })
+        .await;
+}
+```
+
+Support for well-known USB device classes is provided by `embassy-usb`, which is re-exported as [`ariel_os::reexports::embassy_usb`][ariel-os-reexports-embassy-usb-rustdoc].
+Custom USB device classes can also be implemented.
+Additionally, multiple USB device classes can be added on the builder, to create a composite USB device.
+
+### USB Device Classes
+
+The table below lists some of the well-known USB device classes and how to use them in Ariel OS :
+
+| Device class                             | How to use                                                                                                                                    |
+| ---------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| [USB CDC-ACM][usb-cdc-acm-book-glossary] | Apply the [`CdcAcmClass`][ariel-os-embassy-usb-cdcacmclass-rustdoc] constructor on the builder                                                |
+| [USB CDC-NCM][usb-cdc-ncm-glossary-book] | Enable the [`usb-ethernet` laze module][usb-ethernet-laze-module-book]                                                                        |
+| USB HID                                  | Enable the `usb-hid` Cargo feature and apply the [`HidReaderWriter`][ariel-os-embassy-usb-hidreaderwriter-rustdoc] constructor on the builder |
+
+Other device classes are supported.
+
+### Device Configuration
+
+[Configuration for the USB device][ariel-os-embassy-usb-config-rustdoc] created can be provided using the [`#[ariel_os::config(usb)]`][config-attr-macro-rustdoc] attribute.
+In particular, it allows setting the Vendor ID (VID) and Product ID (PID), and the manufacturer and product names.
+
+Additionally, some environment variables are used by [`embassy-usb`][ariel-os-reexports-embassy-usb-rustdoc] for configuration.
+See its documentation for more.
+
+[usb-cdc-acm-book-glossary]: ./glossary.md#usb-cdc-acm
+[logging-transports-book]: ./logging.md#logging-transports
+[debug-probes-book]: ./flashing-debugging.md#debug-interfaces-protocols-and-probes
+[embassy-usb-docsrs]: https://docs.rs/embassy-usb/latest/embassy_usb/
+[ariel-os-usbbuilder-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/usb/type.UsbBuilder.html
+[task-attr-docs]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/attr.task.html
+[ariel-os-reexports-embassy-usb-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/reexports/embassy_usb/index.html
+[ariel-os-embassy-usb-cdcacmclass-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/reexports/embassy_usb/class/cdc_acm/struct.CdcAcmClass.html
+[ariel-os-embassy-usb-hidreaderwriter-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/reexports/embassy_usb/class/hid/struct.HidReaderWriter.html
+[usb-ethernet-laze-module-book]: ./networking.md#network-link-selection
+[usb-cdc-ncm-glossary-book]: ./glossary.md#usb-cdc-ncm
+[ariel-os-embassy-usb-config-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/reexports/embassy_usb/struct.Config.html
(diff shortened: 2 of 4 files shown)


book/src/application.md
```diff
@@ -49,8 +49,6 @@ Functions can currently be registered as either `spawner`s or `task`s:
   They are provided with a [`Spawner`](https://docs.embassy.dev/embassy-executor/git/cortex-m/struct.Spawner.html) instance and can therefore be used to [`spawn`](https://docs.embassy.dev/embassy-executor/git/cortex-m/struct.Spawner.html#method.spawn) other `async` tasks.
 - [`task` functions][task-attr-docs] are `async` functions that are statically allocated at compile-time.
   They are especially useful for long-running, `async` tasks.
-  They must also be used to use *Ariel OS configuration hooks*, which can be requested with their associated macro parameter, and allow to provide configuration during boot.
-  Please refer to the documentation of [`task`][task-attr-docs] for a list of available hooks and to [Configuration Hooks](#configuration-hooks) to know more about hook usage.
```
ROMemories, line 53: It was thought at the time that there would be more configuration hooks, but it now seems sufficient to introduce them ad hoc for USB setup in the dedicated page.

book/src/usb.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Hardware Support
+
+Many microcontrollers supported by Ariel OS include a USB peripheral that can be leveraged to build any USB device.
+At the time of writing, most of them support USB 2.0, but their signaling rate is often limited to that of Full-Speed USB (FS), i.e., 12 Mbit/s (now also known as Basic-Speed USB), even though some also do support High-Speed USB, i.e., 480 Mbit/s (marketed as Hi-Speed USB).
+Some also support USB On-The-Go (OTG), allowing the device to also behave as a USB Targeted Host on the same USB receptacle (by alternatively switching between the peripheral and host roles).
+Currently only the USB peripheral role is supported by Ariel OS.
+
+These USB microcontroller peripherals are "generic," in that they can be used to implement any USB device class.
+Some microcontrollers also feature USB microcontroller peripherals that only support a fixed set of USB classes: e.g., multiple ESP32 MCUs comprise a USB CDC-ACM/JTAG peripheral, which can only be used for the standard [USB CDC-ACM][usb-cdc-acm-book-glossary] device class or a vendor-specific device class implementing JTAG access over USB.
```
bugadani, line 25: > Some microcontrollers also feature USB microcontroller peripherals

I don't have a good alternative, but at this point the word "microcontroller" has lost its meaning to semantic saturation for me.

book/src/usb.md
```diff
@@ -0,0 +1,75 @@
+# Universal Serial Bus (USB)
+
+Ariel OS integrates support for USB peripherals built into many microcontrollers.
+
+## Hardware Support
+
+Many microcontrollers supported by Ariel OS include a USB peripheral that can be leveraged to build any USB device.
```
bugadani: Some devices include more than one USB peripheral. I'm not sure about the currently supported ones, but ESP32-P4 has one full-speed and one high-speed OTG block. I assume Ariel OS will default to exposing the HS port initially, but this particularity might be worth mentioning.

book/src/usb.md
ROMemories: It's definitely not perfect, it's in an attempt to distinguish MCU peripherals from the USB peripheral device role. If only the USB spec had used a more specific term…

bugadani: The specification uses host and device terminology, though, not peripheral 🤔

book/src/usb.md
ROMemories: > Some devices include more than one USB peripheral.

This page also doesn't introduce the concept of USB PHY and whether it supports an external one or not (for instance my understanding is that the ESP32-S3 can use both the "generic" USB peripheral and the USB CDC-ACM/JTAG USB peripheral at the same time if an external PHY is added on top of the internal one).

The current intent is to give an idea of the general picture, I think these peculiarities are better addressed in chip/board pages, and when we add support for it.

book/src/usb.md
ROMemories: It was my impression that the OTG specification used the term peripheral as opposed to Targeted Host:

> An OTG product is a portable device that uses a single Micro-AB receptacle (and no other USB receptacles) to operate at times as a USB Targeted Host and at times as a USB peripheral.

But its usage may not be very consistent in the specification itself.

bugadani: Okay I've only been reading the main USB2.0 spec, not the OTG additions. This doesn't make it any clearer:

From OTG 1_1a 1.3 Acronyms and Terms:

> Peripheral: A physical entity that is attached to a USB cable and is currently operating as a
“device” as defined in [USB2.0].

:(

book/src/usb.md
```diff
@@ -0,0 +1,75 @@
...
+At the time of writing, most of them support USB 2.0, but their signaling rate is often limited to that of Full-Speed USB (FS), i.e., 12 Mbit/s (now also known as Basic-Speed USB), even though some also do support High-Speed USB, i.e., 480 Mbit/s (marketed as Hi-Speed USB).
+Some also support USB On-The-Go (OTG), allowing the device to also behave as a USB Targeted Host on the same USB receptacle (by alternatively switching between the peripheral and host roles).
+Currently only the USB peripheral role is supported by Ariel OS.
+
+These USB microcontroller peripherals are "generic," in that they can be used to implement any USB device class.
```
nponsard:
```suggestion
These USB microcontroller peripherals are "generic" in that they can be used to implement any USB device class.
```

book/src/usb.md
```diff
@@ -0,0 +1,75 @@
...
+
+The table below lists some of the well-known USB device classes and how to use them in Ariel OS :
+
+| Device class                             | How to use                                                                                                                                    |
+| ---------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| [USB CDC-ACM][usb-cdc-acm-book-glossary] | Apply the [`CdcAcmClass`][ariel-os-embassy-usb-cdcacmclass-rustdoc] constructor on the builder                                                |
+| [USB CDC-NCM][usb-cdc-ncm-glossary-book] | Enable the [`usb-ethernet` laze module][usb-ethernet-laze-module-book]                                                                        |
+| USB HID                                  | Enable the `usb-hid` Cargo feature and apply the [`HidReaderWriter`][ariel-os-embassy-usb-hidreaderwriter-rustdoc] constructor on the builder |
+
+Other device classes are supported.
```
nponsard: It would be nice to have a link to the list of classes in `embassy_usb` and a reminder that custom classes can be implemented.

book/src/usb.md
ROMemories: A comma is required for rhythm, which is [placed inside the quotation marks in American style](https://en.wikipedia.org/wiki/Quotation_marks_in_English#American_style).

book/src/usb.md
ROMemories: I've added a link to the `embassy_usb::class` module with the list of supported well-known device classes, and a reminder for custom ones.

book/src/usb.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Device Configuration
+
+[Configuration for the USB device][ariel-os-embassy-usb-config-rustdoc] created can be provided using the [`#[ariel_os::config(usb)]`][config-attr-macro-rustdoc] attribute.
+In particular, it allows setting the Vendor ID (VID) and Product ID (PID), and the manufacturer and product names.
+
+Additionally, some environment variables are used by [`embassy-usb`][ariel-os-reexports-embassy-usb-rustdoc] for configuration.
+See its documentation for more.
+
+### Clock Configuration
```
ROMemories, line 74: Now that we have some documentation about clock configuration (#2165), I've added a section that explains some of the requirements. Something similar could also be done for the HWRNG, as it's a common source of confusion when porting new boards.

book/src/usb.md
nponsard: Oh didn't know about that 👍

book/src/usb.md
bergzand: We could clarify this in the first sentence and hopefully prevent further confusion.

> Currently only the USB peripheral (device) role is supported by Ariel OS. USB host role is not supported.

book/src/usb.md
```diff
@@ -0,0 +1,94 @@
+# Universal Serial Bus (USB)
+
+Ariel OS integrates support for USB peripherals built into many microcontrollers.
+
+## Hardware Support
+
+Many microcontrollers supported by Ariel OS include a USB peripheral that can be leveraged to build any USB device.
+At the time of writing, most of them support USB 2.0, but their signaling rate is often limited to that of Full-Speed USB (FS), i.e., 12 Mbit/s (now also known as Basic-Speed USB), even though some also do support High-Speed USB, i.e., 480 Mbit/s (marketed as Hi-Speed USB).
```
bergzand: This sentence is hard to read due to the extra clauses and parenthesis. My suggestion would be to split the signal rates into a list.

book/src/usb.md
```diff
@@ -0,0 +1,94 @@
...
+Some also support USB On-The-Go (OTG), allowing the device to also behave as a USB Targeted Host on the same USB receptacle (by alternatively switching between the peripheral and host roles).
+Currently only the USB peripheral role is supported by Ariel OS.
+
+These USB microcontroller peripherals are "generic," in that they can be used to implement any USB device class.
+Some microcontrollers also feature USB microcontroller peripherals that only support a fixed set of USB classes: e.g., multiple ESP32 MCUs comprise a USB CDC-ACM/JTAG peripheral, which can only be used for the standard [USB CDC-ACM][usb-cdc-acm-book-glossary] device class or a vendor-specific device class implementing JTAG access over USB.
+Currently, only the "generic" USB peripherals are supported in Ariel OS applications.
+The others may still be integrated by Ariel OS to implement specific functionality, like [logging][logging-transports-book].
```
bergzand: The point of this paragraph is to tell the reader that Ariel OS only supports "generic" USB peripherals. I'd start with that here (currently line 14 here) and then describe what that means and the alternatives.

book/src/usb.md
```diff
@@ -0,0 +1,94 @@
...
+| [USB CDC-ACM][usb-cdc-acm-book-glossary] | Apply the [`CdcAcmClass`][ariel-os-embassy-usb-cdcacmclass-rustdoc] constructor on the builder                                                |
+| [USB CDC-NCM][usb-cdc-ncm-glossary-book] | Enable the [`usb-ethernet` laze module][usb-ethernet-laze-module-book]                                                                        |
+| USB HID                                  | Enable the `usb-hid` Cargo feature and apply the [`HidReaderWriter`][ariel-os-embassy-usb-hidreaderwriter-rustdoc] constructor on the builder |
+
+[Other well-known device classes][ariel-os-embassy-usb-class-rustdoc] are supported, and it is also possible to implement custom ones.
+
+### Device Configuration
+
+[Configuration for the USB device][ariel-os-embassy-usb-config-rustdoc] created can be provided using the [`#[ariel_os::config(usb)]`][config-attr-macro-rustdoc] attribute.
+In particular, it allows setting the Vendor ID (VID) and Product ID (PID), and the manufacturer and product names.
```
bergzand, line 69: Both this section and the api docs have nothing on how to use that macro. How about a link to the USB examples here, those use the macro at this point.

book/src/usb.md
```diff
@@ -0,0 +1,94 @@
...
+Because USB requires accurate timings[^usb-timings-requirements], the clock source typically relies on a [crystal resonator][crystal-resonator-book] (or an [external crystal oscillator][external-crystal-oscillator]).
+When that is not the case, the microcontroller must feature a clock recovery system that is able to recover a clock from the USB Start Of Frame (SOF) packets (sent by the USB host every 1 ms for Full-Speed USB) and that trims an internal oscillator, keeping it in sync with the USB host and thus enabling crystal-less USB.
+Many STM32 MCUs feature such clock recovery system (CRS).
+
+To be able to use USB, the [clock configuration][clock-tree-configuration-book] must enable and configure the clock source required for the USB microcontroller peripheral.
+[The default clock configuration provided by Ariel OS][clock-tree-configuration-book] usually already configures it appropriately when made possible by the board.
+Otherwise, an appropriate clock configuration must be [provided in the application][clock-tree-configuration-book].
```
bergzand, line 84: This seems to be missing a

> Check the manual of your hardware to see if it supports the USB peripheral role.

bergzand (changes requested):

book/src/usb.md
ROMemories: I'm unsure of what you want emphasize with this: do you mean something along the lines of "Check that your board has the appropriate components to generate the required clock signals" or something else?

book/src/usb.md
ROMemories: I've added the "(device)" clarification and the extra sentence.

book/src/usb.md
ROMemories: I've restructured the paragraph, added a summary table, and added subheadings (hopefully they help, and at least they make anchors for linking to them).

book/src/usb.md
ROMemories: Agreed, I've restructured the paragraph.

book/src/usb.md
bergzand: Yes, and (assuming dev kits and not custom build hardware) usually the manual of the board contains some information on whether the USB peripheral role is supported. Pointing users in that direction might be helpful in addition to having them check all clock signals.

book/src/usb.md
bergzand: much clearer, thanks!

book/src/usb.md
ROMemories: In this case, I would rather not for now: I think none of us has yet done the research of the exact values that should be provided, and I can see this leading to cargo culting in applications; I'm also still a bit wary to point to the repo from the book (and I'd rather not have an example inline *for now* to avoid cargo culting). I agree this could be useful in the future, but I don't think it has to be in this PR, which at least mention that it is possible, if application authors know they need this.

book/src/usb.md
ROMemories: I'm still confused about the emphasis on the USB *peripheral role* here. I think the clock configuration and the board hardware are identical for both the host and peripheral roles (and whether "user USB" is supported, as opposed to just having a USB port for the debug probe, seems implied here). (I do agree about the having to check the manual for clock generation.)

bergzand: You're right, the role likely doesn't matter most of the time, just whether user USB is supported. Sorry for the confusion

book/src/usb.md
bergzand: Ok, my concern is that you're announcing a feature here, while we have zero documentation on how to use that feature, and it is not obvious from the api. If that's what you intend here, fine with me.

ROMemories: I think there's (some) documentation, the macro's docs do mention that you need to return an `embassy_usb::Config`, we just don't provide an example (in the book) of such a value yet, which is what I want to avoid for now.

book/src/usb.md
ROMemories: So do we need an extra note about it? There's already a "tip" admonition early in the page about the distinction between "user USB" and the debug probe's USB port. In the case of devkits, it can be safely assumed that if it has a "user USB" port, then it has the appropriate hardware to generate clock signals for USB (that'd be a really odd devkit otherwise), and in the case of custom boards, they should already be aware of what hardware is required for that (which I'm not against expanded on later, but this would require giving figures about resonators' and oscillators' accuracy, and I don't think this has to go in this PR).

bergzand: I think a simple pointer makes it a bit more friendly for the reader. Something such as: "Check the manual of the board whether a USB microcontroller peripheral is supported and how to set up the required clock signals."

ROMemories: I've unfortunately never seen a devkit manual giving hints about the clock configuration needed for USB, and I'd rather not send the users on a goose chase…

bergzand: Ok, then let's leave it out

bergzand (approved):
LGTM, please squash.

PR #2161, opened 2026-06-26
refactor(doc): move board/builder metadata from support_matrix.yml in…

dgtlrift:
Support matrix board and builder data is now sourced from ariel_doc sections in each BSP file under boards/. This eliminates the need to update two files when changing board metadata. The support_matrix.yml retains only global configuration (support_keys, functionalities, note_snippets, chips).

Depends on minor changes to sbd-gen ariel-os/sbd#159

# Description

As more and more BSP yaml files are created and included as part of the OTS support from Ariel-OS its going to get unruly and cumbersome to edit multiple files (the BSP yaml and the documentation file in matrix_support.)  There may be other references in the documentation that can references the BSPs so those can also be examined in this PR. Also, there may be a better organizational way to do this, but I wanted to get this in place for discussion and testing. 

## Testing

`doc/gen_book.rs matrix doc/support_matrix.yml book/src/support_matrix.html --template-path book/templates/support-matrix.html.tmpl
`
## Issues/PRs References

This depends on a PR in sbd-gen to ignore arial_doc tree.

## Open Questions

Where else are BSPs details repeated in documentation

## Changelog Entry

Currently experimental to understand if this is valuable to others... I personally believe it is as I've been going down a rabbit hole trying to add all the open source hardware boards available to the Meshtastic stack for users to have an alternative solution with Ariel-OS.

## Change Checklist


ROMemories:
My concerns with this are two-fold. This changes the *scope* of SBD files, which were originally intended to (only) describe the hardware (not to contain the support info of various OSes) and this muddies the *intent* of these description files: it was the hope that these could be reused as is by different projects; I think adding support info specific to Ariel OS would make this less clear and hinder this goal.

This also means that if we were to do this, we should be ready to also incorporate the support info of other projects that'd wish so.

dgtlrift:
> My concerns with this are two-fold. This changes the _scope_ of SBD files, which were originally intended to (only) describe the hardware (not to contain the support info of various OSes) and this muddies the _intent_ of these description files: it was the hope that these could be reused as is by different projects; I think adding support info specific to Ariel OS would make this less clear and hinder this goal.

I appreciate the concern and the original intent, however as potential popularity grows and BoM variants (never mind whole new board designs) increase organizing this information in a the single file defining the interfaces makes more sense.  That said, even with the expansion of new projects utilizing the yaml board files would still ignore the XXX_doc. Specifically, sbd-gen ignores ariel_doc as it's irrelevant to code generation, such would be the same for riot as well, and also  implying that there could also be a riot_doc section as well, but infrastructure support would be needed in the riot project to generate the documentation from the yaml file.

There is a potential opportunity for common objects or modules to be captured between the projects RIoT and Ariel with something like common_doc, but that's really an anecdotal point.

> 
> This also means that if we were to do this, we should be ready to also incorporate the support info of other projects that'd wish so.

Agree - and my next comment is out of scope of this merge (nevermind repo) ... currently, I'm aware that SBD code auto-generates BSP c-header files (board.h), pin definitions, initializers, interfaces the traditional and Makefile scaffolding that RIOT expects... This code generation seems to be interlaced within SBD generation with Ariel-OS... architecturally, it would make sense to split this out into backend crates that handle each language specifically and define a pattern for adding new language or RTOS support.

kaspar030:
IMO it is fine to have this extra information in sbd files. First and foremost this is supposed to be practical, and having all board info of Ariel OS in one file would be a huge improvement.

chrysn:
There are two kinds of information we don't pull yet from sbd: What is supported, and other metadata (website, full name, which builder corresponds to which board).

If getting more of that means that we temporarily pull some of the former to get the latter deduplicated, that's an improvement. I agree with @ROMemories that support status should ultimately not live in sbd, but where it's currently living isn't great either; finding a good spot for that will take work either way (but it'll also take time for us to figure out how to best do it, eg. whether we can get "Person (or tool) tested Feature on Board (and thus CPU) with Version" is something we can get from different sources).

chrysn:
Looking at some of the concrete changes, I'd like to suggest moving the name and url pieces right up to toplevel (or a top-level group that is not Ariel specific). But if that'd cause any trouble or complications, that can be left to a next round as well.

dgtlrift:
> Looking at some of the concrete changes, I'd like to suggest moving the name and url pieces right up to toplevel (or a top-level group that is not Ariel specific). But if that'd cause any trouble or complications, that can be left to a next round as well.

Do you mean I should move the name & url out of the ariel_doc and into a common_doc structure, and also these *_doc trees move in up to the top of each yaml doc? I wouldn't mind doing that, but I rather do it in another PR as moving text blocks around gets annoying with merge conflicts.

PR #2166, opened 2026-07-03
feat(ble): configure static address from env

nponsard:
# Description

This adds a way to set a static BLE address that persists across reboots for use cases like BLE Beacons.

## Testing

```sh
CONFIG_BLE_STATIC_ADDRESS=aa:aa:aa:aa:aa:aa  laze -C examples/ble-advertiser/ build -b nrf52840dk -s ble-config-static-address run   
```

And notice that the address advertised is `ea:aa:aa:aa:aa:aa` (two most significant bits set to 1).

## Changelog Entry

It is now possible to set a static BLE address that persists across reboots, for use cases like BLE beacons.

## Change Checklist

### DIFF
--- a/book/src/bluetooth.md
+++ b/book/src/bluetooth.md
@@ -14,16 +14,28 @@ If you want to learn more about BLE concepts, you can read the [TrouBLE document
 
 ## Configuring the BLE Stack
 
-The ability to configure which Bluetooth address is used and other capacity parameters like the MTU is planned in future updates.
+The ability to use a private address and other capacity parameters like the MTU is planned in future updates.
 
 > [!IMPORTANT]
 > For compatibility reasons the MTU is fixed at 27 bytes.
 >
-> The device address is randomly generated at boot and may be periodically rotated.
+> By default the device address is randomly generated at boot and may be periodically rotated.
 >
 > Current implementation: the address is a static device address and is not rotated during execution.
 > This allows to use the BLE feature of Ariel OS on multiple devices in the same location.
 > We later plan to switch to private device addresses by default, which *are* rotated during execution.
+
+### Using a Static Address
+
+> [!IMPORTANT]
+> It is not recommended to have personal devices advertise a static (and constant) address as this can easily be tracked and compromise the privacy of the user.
+>
+> The use case of advertising a static address is for static devices like BLE beacons.
+
+To set an address that will persist across reboots, use the `ble-config-static-address` [laze module][laze-modules-book] and set the address using the `CONFIG_BLE_STATIC_ADDRESS` environment variable, the expected format is `XX:XX:XX:XX:XX:XX` where X is a hex digit.
+
+> [!NOTE]
+> The address will be advertised as a [random static address](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/low-energy-controller/link-layer-specification.html#UUID-7edea27a-a47f-8436-4bd7-aedc1945c366_figure-idm4497995733171233616486354268) and the two most significant bits of the address will therefore be set to 1.
 
 ## Using the BLE Stack
 
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1456,6 +1456,15 @@ modules:
     selects:
       - doc-only
 
+  - name: ble-config-static-address
+    help: Use a static address for BLE, the two most significant bits will be set to 1 to indicate a static random address.
+    selects:
+      - ble
+    env:
+      global:
+        FEATURES:
+          - ariel-os/ble-config-static-address
+
   - name: has_nrf91_modem
     selects:
       - doc-only
--- a/src/ariel-os-embassy/src/ble.rs
+++ b/src/ariel-os-embassy/src/ble.rs
@@ -53,8 +53,19 @@ pub fn current_address() -> impl Future<Output = Address> {
 }
 
 /// Generates a random address.
+#[cfg(not(feature = "ble-config-static-address"))]
 fn get_random_addr() -> [u8; 6] {
     let mut addr = [0u8; 6];
     rand_core::RngCore::fill_bytes(&mut ariel_os_random::crypto_rng(), &mut addr);
     addr
+}
+
+/// Get random address from env.
+#[cfg(feature = "ble-config-static-address")]
+fn get_random_addr() -> [u8; 6] {
+    use ariel_os_utils::eui48_from_env;
+    eui48_from_env!(
+        "CONFIG_BLE_STATIC_ADDRESS",
+        "static address for BLE in format XX:XX:XX:XX:XX:XX",
+    )
 }
--- a/src/ariel-os-utils/src/env.rs
+++ b/src/ariel-os-utils/src/env.rs
@@ -232,3 +232,37 @@ macro_rules! ipv6_addr_from_env_or {
         }
     }};
 }
+
+/// Reads an eui48 at compile time from the given environment variable, produces an
+/// `[u8; 6]`, matched to the endianness of the target.
+///
+/// - The `$doc` parameter allows to provide a documentation string for this tunable (see
+///   [`str_from_env!`](str_from_env)).
+///
+/// # Errors
+///
+/// - Produces a compile-time error when [`option_env!`](option_env) does.
+/// - Produces a compile-time error when the environment variable cannot be parsed into a eui48
+#[macro_export]
+macro_rules! eui48_from_env {
+    // $doc is currently unused
+    ($env_var:literal, $doc:literal $(,)?) => {{
+        const {
+            const str_addr: &str = if let Some(str_value) = option_env!($env_var) {
+                str_value
+            } else {
+                $crate::env::const_panic::concat_panic!(
+                    "`",
+                    $env_var,
+                    "` environment variable was expected to provide the ",
+                    $doc,
+                );
+            };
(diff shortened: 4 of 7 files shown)


book/src/bluetooth.md
```diff
@@ -14,17 +14,29 @@ If you want to learn more about BLE concepts, you can read the [TrouBLE document
...
 > We later plan to switch to private device addresses by default, which *are* rotated during execution.
 
+### Static Address
+
+> [!IMPORTANT]
+> It is not recommended to have personal devices advertise a static (and constant) address as this can easily be tracked and compromise the privacy of the user.
+>
+> The use case of advertising a static address is for static devices like BLE beacons.
+
+To set an address that will persist across reboots, use the `ble-config-static-address` [laze module][laze-modules-book] and set the address using the `CONFIG_BLE_STATIC_ADDRESS` environment variable, format is `XX:XX:XX:XX:XX:XX` where X is a hex digit.
```
ROMemories:
```suggestion
To set an address that will persist across reboots, use the `ble-config-static-address` [laze module][laze-modules-book] and set the address using the `CONFIG_BLE_STATIC_ADDRESS` environment variable, the expected format is `XX:XX:XX:XX:XX:XX` where X is a hex digit.
```

book/src/bluetooth.md
```diff
@@ -14,17 +14,29 @@ If you want to learn more about BLE concepts, you can read the [TrouBLE document
...
+
+> [!NOTE]
+> The address will be advertised as a [random static address](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/low-energy-controller/link-layer-specification.html#UUID-7edea27a-a47f-8436-4bd7-aedc1945c366_figure-idm4497995733171233616486354268) and the two most significant bits of the address will be set to 1.
```
ROMemories: This clarifies that the MSbs are set to comply with the format.
```suggestion
> The address will be advertised as a [random static address](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/low-energy-controller/link-layer-specification.html#UUID-7edea27a-a47f-8436-4bd7-aedc1945c366_figure-idm4497995733171233616486354268) and the two most significant bits of the address will therefore be set to 1.
```

src/ariel-os-utils/src/env.rs
```diff
@@ -232,3 +232,37 @@ macro_rules! ipv6_addr_from_env_or {
...
+/// - The `$doc` parameter allows to provide a documentation string for this tunable (see
+///   [`str_from_env!`](str_from_env)).
+///
+/// # Errors
+///
+/// - Produces a compile-time error when [`option_env!`](option_env) does.
+/// - Produces a compile-time error when the environment variable cannot be parsed into a MAC
+///   address.
+#[macro_export]
+macro_rules! mac_addr_from_env {
```
ROMemories: This isn't always a MAC address, for instance it's not one for BLE; I think EUI-48 is more correct to just refer to the format (the macro's docs also need to be adjusted):
```suggestion
macro_rules! eui48_from_env {
```

laze-project.yml
```diff
@@ -1456,6 +1456,15 @@ modules:
     selects:
       - doc-only
 
+  - name: ble-config-static-address
+    help: Use a static MAC address for BLE, the two most significant bits will be set to 1 to indicate a static random address.
```
```suggestion
    help: Use a static address for BLE, the two most significant bits will be set to 1 to indicate a static random address.
```

src/ariel-os-embassy/src/ble.rs
```diff
@@ -53,8 +53,19 @@ pub fn current_address() -> impl Future<Output = Address> {
...
     addr
 }
+
+/// Get random address from env.
+#[cfg(feature = "ble-config-static-address")]
+fn get_random_addr() -> [u8; 6] {
+    use ariel_os_utils::mac_addr_from_env;
+    mac_addr_from_env!(
+        "CONFIG_BLE_STATIC_ADDRESS",
+        "Mac address for BLE stack in format XX:XX:XX:XX:XX:XX",
```
ROMemories:
```suggestion
        "static address for BLE in format XX:XX:XX:XX:XX:XX",
```

src/ariel-os-utils/src/env.rs
```diff
@@ -232,3 +232,37 @@ macro_rules! ipv6_addr_from_env_or {
...
+                str_value
+            } else {
+                $crate::env::const_panic::concat_panic!(
+                    "`",
+                    $env_var,
+                    "` environment variable was expected to provide the ",
+                    $doc,
+                );
+            };
+            // check length
```
ROMemories: Was this meant as a todo? This does fail as expected if the wrong number of bytes is provided in the env var.
```suggestion
```

book/src/bluetooth.md
```diff
@@ -14,17 +14,29 @@ If you want to learn more about BLE concepts, you can read the [TrouBLE document
...
 > For compatibility reasons the MTU is fixed at 27 bytes.
 >
-> The device address is randomly generated at boot and may be periodically rotated.
+> By default the device address is randomly generated at boot and may be periodically rotated.
 >
 > Current implementation: the address is a static device address and is not rotated during execution.
 > This allows to use the BLE feature of Ariel OS on multiple devices in the same location.
 > We later plan to switch to private device addresses by default, which *are* rotated during execution.
 
+### Static Address
```
ROMemories: For consistency with the other titles of the page:
```suggestion
### Using a Static Address
```

src/ariel-os/Cargo.toml
```diff
@@ -174,6 +174,8 @@ ble-esp = ["ariel-os-embassy/ble-esp"]
 ble-central = ["ariel-os-embassy/ble-central", "ble"]
 # Enables BLE to act as a peripheral. Can be used in conjunction with `ble-central`.
 ble-peripheral = ["ariel-os-embassy/ble-peripheral", "ble"]
+# Selects static BLE address configuration for setting a "random static" address at build.
```
```suggestion
# Selects static BLE address configuration for setting a static address at build.
```

src/ariel-os-embassy/src/ble.rs
```diff
@@ -53,8 +53,19 @@ pub fn current_address() -> impl Future<Output = Address> {
...
+#[cfg(not(feature = "ble-config-static-address"))]
 fn get_random_addr() -> [u8; 6] {
     let mut addr = [0u8; 6];
     rand_core::RngCore::fill_bytes(&mut ariel_os_random::crypto_rng(), &mut addr);
     addr
 }
+
+/// Get random address from env.
+#[cfg(feature = "ble-config-static-address")]
+fn get_random_addr() -> [u8; 6] {
```
ROMemories, line 65: I think there's an issue with the order of bytes here (or in the parsing logic): if I set `CONFIG_BLE_STATIC_ADDRESS='8a:aa:aa:aa:aa:11'`, `D1:AA:AA:AA:AA:8A` will be used as address (as shown when enabling debug logs), instead of `CA:AA:AA:AA:AA:11` as I would expect (maybe my expectations are wrong?).

ROMemories (commented):

```sh
CONFIG_BLE_STATIC_ADDRESS='aa:aa:aa:aa:aa:aa' laze -C examples/ble-advertiser/ build -s ble-config-static-address -b nrf52840dk run
```

and observed that the advertised address is indeed EA:AA:AA:AA:AA:AA.
I've also checked that the `ble-advertiser` example still works as expected otherwise, and still advertises a random address rotated at boot.

However, when using an address that's not symmetrical, the order of bytes seems to be reversed.

---

I think occurrences of "MAC addresses" should be removed, as my understanding is that this term is not used for BLE, and replaced with just "static address" in this PR or with "EUI-48" to refer to the identifier format if necessary.

src/ariel-os-embassy/src/ble.rs
nponsard: I forgot about that, the human readable way to display an EUI48 is to put the most significant byte first, but in storage it should be last.
Thanks for testing this 👍

src/ariel-os-utils/src/env.rs
nponsard: This should have been reworded or removed, I wanted to say here that since we store it in an array, the length is checked.

src/ariel-os-utils/src/env.rs
```diff
@@ -232,3 +232,38 @@ macro_rules! ipv6_addr_from_env_or {
...
+                    $env_var,
+                    "` environment variable was expected to provide the ",
+                    $doc,
+                );
+            };
+            let mut eui48: [u8; 6] =
+                $crate::const_str::hex!($crate::const_str::replace!(str_addr, ":", ""));
+            // Human readable eui48 is big endian.
+            #[cfg(target_endian = "little")]
+            eui48.reverse();
```
nponsard, line 264: I think this is the right way to do it ?
Eui48 is [made for networking that is big endian](https://www.rfc-editor.org/info/rfc7043/#section-3.1).

ROMemories: I don't think that's related to the target's endianness, that's just [how the parsing works](https://docs.rs/const-str/latest/const_str/macro.hex.html).
```suggestion
            eui48.reverse();
```

nponsard: Ah that makes more sense, I thought I was splitting the string into 2-character strings.

ROMemories (approved):

PR #2170, opened 2026-07-07
build: introduce a laze task to flash the binary to a SLICES-RI node

ROMemories:
# Description

This adds a laze task to make it a bit easier to flash/run a binary on an [IoT-LAB](https://www.iot-lab.info/)/[SLICES-FR](https://slices-fr.eu/)/[SLICES-RI](https://www.slices-ri.eu/) node. The laze task is made only available when the `logging-over-uart` laze module is enabled, so it's not possible to forget to select it (and have no output).

The laze task is not documented yet as its scope is not well defined yet (is it possible to make it start and stop the experiment?) and may still evolve quickly.

## Testing

Successfully tested using the following (requires an IoT-LAB account and prior authentication):

```sh
iotlab-experiment submit -d 3 -l 1,archi=nrf52840dk:multi+site=saclay
laze -C examples/hello-world build -s logging-over-uart -b nrf52840dk flash-slices-ri
```

## Change Checklist

### DIFF
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -1958,6 +1958,13 @@ modules:
       global:
         FEATURES:
           - ariel-os/logging-over-uart
+    tasks:
+      # This task is currently excluded from Ariel OS's compatibility guarantees.
+      flash-slices-ri:
+        help: flash the binary to a SLICES-RI node; requires a running experiment but will wait for it
+        cmd:
+          - iotlab-experiment wait --timeout 30 --cancel-on-timeout
+          - iotlab-node --flash ${out} $@
 
   # Temporary module for boards that have ad hoc support.
   - name: debug-uart


laze-project.yml
```diff
@@ -1967,6 +1958,13 @@ modules:
       global:
         FEATURES:
           - ariel-os/logging-over-uart
+    tasks:
+      # This task is currently excluded from Ariel OS's compatibility guarantees.
+      flash-slices-ri:
```
ROMemories, line 1963: @aabadie What do you think would be the best name for this task? I picked this for now as I think it should be generic enough so we don't have to rename it later, and I suppose it should be eventually possible to flash any SLICES-RI node. (The future documentation would point to what IoT-LAB/SLICES-FR/SLICES-RI are.)

aabadie: I have no strong opinion and naming things is always hard :)
Other names I could think of are `flash-iotlab`, `flash-slices` or `flash-slices-iot`.

laze-project.yml
```diff
@@ -1967,6 +1958,13 @@ modules:
...
+        help: flash the binary to a SLICES-RI node; requires a running experiment but will wait for it
+        cmd:
+          - iotlab-experiment wait --timeout 30 --cancel-on-timeout
+          - iotlab-node --flash ${out} $@
```
aabadie, line 1967: That seems super limited: it's not possible to select a specific node to flash and if the user has several experiments running, the commands will fail unless the `--id` (or `-i`) is passed with the experiment id value.
In RIOT make build system, it's possible to pass the IOTLAB_NODE env variable. This variable either contain the precise node fqdn or "auto" to select the first node in the list of nodes booked in the experiment. It's also possible to specify the experiment id in the `IOTLAB_EXP_ID` env var. See https://github.com/RIOT-OS/RIOT/blob/master/dist/testbed-support/makefile.iotlab.single.inc.mk
I don't know how feasible that is in a laze task.

PR #2171, opened 2026-07-09
chore: update sbd schema

anlavandier:
# Description

This PR applies the refactor of monocolor LEDs done in https://github.com/ariel-os/sbd/pull/163

## Testing

I have successfully tested compilation of `blinky` `gpio` and `gpio-reuse` examples on the `nrf528940dk`
`grep -e 'LedPeripherals` should return only instances of `MonocolorLedPeripherals`

## Issues/PRs References

Depends on ariel-os/sbd#163

## Changelog Entry

Ariel OS now uses the version 0.4.0 of the SBD schema.

## Change Checklist

- [n/a] I have made corresponding changes to the documentation.

### DIFF
--- a/boards/adafruit-feather-nrf52840-express.yaml
+++ b/boards/adafruit-feather-nrf52840-express.yaml
@@ -1,4 +1,4 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   # different names in use: "Adafruit Feather nRF52840 Express" on shop as of
   # 2025-12-14, whereas schematics dating from 2025-05-05 label it as "Adafruit
@@ -8,7 +8,7 @@ targets:
     chip: nrf52840
     flags:
       - has_usb_device_port
-    leds:
+    monocolorleds:
       # FIXME: led0/led1 names are purely synthetic to work with the current
       # sbd::ariel convention of naming them sequentially.
       led0:
--- a/boards/adafruit-feather-nrf52840-sense.yaml
+++ b/boards/adafruit-feather-nrf52840-sense.yaml
@@ -1,10 +1,10 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   adafruit-feather-nrf52840-sense:
     chip: nrf52840
     flags:
       - has_usb_device_port
-    leds:
+    monocolorleds:
       # FIXME: led0/led1 names are purely synthetic to work with the current
       # sbd::ariel convention of naming them sequentially.
       led0:
--- a/boards/arduino-uno-q.yaml
+++ b/boards/arduino-uno-q.yaml
@@ -1,4 +1,4 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   arduino-uno-q:
     chip: stm32u585ai
@@ -9,7 +9,7 @@ targets:
         CARGO_ENV:
           - CONFIG_VBUS_DETECTION=false
           - CONFIG_SWI=USART2
-    leds:
+    monocolorleds:
       led0:
         color: red
         pin: PH10
--- a/boards/ariel-chips.yaml
+++ b/boards/ariel-chips.yaml
@@ -1,6 +1,6 @@
 # This file contains Ariel OS specific information needed by sbd-gen for
 # generating an Ariel OS BSP crate.
-version: 0.3.0
+version: 0.4.0
 ariel:
   # This list contains all currently supported chips. The names must correspond
   # to an Ariel OS laze context.
--- a/boards/bbc-microbit-v1.yaml
+++ b/boards/bbc-microbit-v1.yaml
@@ -1,4 +1,4 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   bbc-microbit-v1:
     chip: nrf51822-xxaa
--- a/boards/bbc-microbit-v2.yaml
+++ b/boards/bbc-microbit-v2.yaml
@@ -1,4 +1,4 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   bbc-microbit-v2:
     chip: nrf52833
--- a/boards/dfrobot-firebeetle2-esp32-c6.yaml
+++ b/boards/dfrobot-firebeetle2-esp32-c6.yaml
@@ -1,10 +1,10 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   dfrobot-firebeetle2-esp32-c6:
     chip: esp32c6fx4
     flags:
       - has_usb_cdc_acm_port
-    leds:
+    monocolorleds:
       led0:
         pin: GPIO15
     buttons:
--- a/boards/dwm1001.yaml
+++ b/boards/dwm1001.yaml
@@ -1,8 +1,8 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   dwm1001:
     chip: nrf52832
-    leds:
+    monocolorleds:
       led0:
         pin: P0_14
         aliases:
--- a/boards/espressif-esp32-c3-devkit-rust-1.yaml
+++ b/boards/espressif-esp32-c3-devkit-rust-1.yaml
@@ -1,4 +1,4 @@
-version: 0.3.0
+version: 0.4.0
 targets:
   espressif-esp32-c3-devkit-rust-1:
     chip: esp32-c3-mini-1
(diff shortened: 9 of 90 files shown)


boards/adafruit-feather-nrf52840-express.yaml
```diff
@@ -8,7 +8,7 @@ targets:
     chip: nrf52840
     flags:
       - has_usb_device_port
-    leds:
+    monocolorleds:
```
ROMemories, line 11: Seeing this now, maybe there should be an underscore to separate the two words:
```suggestion
    monocolor_leds:
```

kaspar030: so we'll have multiple entries that don't necessarily sort well (monocolor_leds, bicolor_leds, rgb_leds, smart_leds)?
Why not make the led entries typed, defaulting to "monocolor"?

anlavandier: > Seeing this now, maybe there should be an underscore to separate the two words:

Sure, this means changing the schema, so I'll wait a bit for more feedback

anlavandier: > so we'll have multiple entries that don't necessarily sort well (monocolor_leds, bicolor_leds, rgb_leds, smart_leds)? Why not make the led entries typed, defaulting to "monocolor"?

That could be a cleaner way. I'm not sure on how this would be parsed. Maybe it could be a two step process ? First it's parsed into an `Led` struct with `led.type = Monocolor` and then `led.into()` turns into a `MonocolorLed` ? I'm thinking of something like this

```rust
#[derive(Deserialize)]
struct Led {
  type: LedType
  // ...
}

struct MonocolorLed {}

impl TryFrom<Led> for MonocolorLed {
  fn try_from(led: Led)-> Option<Self> {
    match led.type {
      LedType::Monocolor => Some(MonocolorLed { })
      _ => None
    }
  }
```

I'm against the default to monocolor (though I understand that this is probably for compatibiliy with existing files).

ROMemories: > Why not make the led entries typed

In my experience having a "kind" field makes either for manual and unpleasant error handling in the implementation, or unclear error messages for users as the relationships are not encoded in the schema itself.

In addition, we'll [have other constructs that involve LEDs](#152) (e.g., LED strips and matrices), and I'm not particularly concerned about the sorting of the keys.

kaspar030: > In my experience having a "kind" field makes either for manual and unpleasant error handling in the implementation, or unclear error messages for users as the relationships are not encoded in the schema itself.

Hm, tagged enums work ok-ish using serde, including errors. improving those is a matter of catching the error and making it nice, or using miette/...

kaspar030: > I'm against the default to monocolor (though I understand that this is probably for compatibiliy with existing files).

Why? 95% (guess) of the led's we'd define are monocolor.

ROMemories: > 95% (guess) of the led's we'd define are monocolor.

I don't think this holds true:

- All the Espressif devkits we currently support and that have some form of LED have a NeoPixel and no monocolor LED.
- The Unihiker K10 also has NeoPixels.
- The Thingy:91 X has RGB LEDs.
- The Arduino UNO Q has RGB LEDs and an LED matrix.

anlavandier: > > I'm against the default to monocolor (though I understand that this is probably for compatibiliy with existing files).
> 
> Why? 95% (guess) of the led's we'd define are monocolor.

Mostly because I dislike the fact that it's not explicit for one kind of LED and not the others and because it's not the only kind of LED that would declare only one pin, WS2812 and related pixel would be unrecognizable at a glance from monocolor leds

kaspar030: > I don't think this holds true:

Quick `rg led[0-9] | cat | wc -l` yields >80 LEDs, so the number is more like >80%. This is still *most of them* in my book.

kaspar030: > Mostly because I dislike the fact that it's not explicit for one kind of LED and not the others and because it's not the only kind of LED that would declare only one pin, WS2812 and related pixel would be unrecognizable at a glance from monocolor leds

I don't think *forgetting to set the WS2812 type* (or whatever) will be an issue, or something that would slip by reviews, or even the most basic blinky test?

Less churn/no schema change, and not throwing the "monocolor_led" naming into users' faces, and less noise in the board descriptions, are all better arguments for adding support for rgb/smart LEDs behind a `type:` tag.

```
type: monocolor
color: green
```

also adds redundancy.

anlavandier: > I don't think _forgetting to set the WS2812 type_ (or whatever) will be an issue, or something that would slip by reviews, or even the most basic blinky test?
> 

I'm thinking more in the context of out of tree board definitions where users could find themselves faced with unclear errors

kaspar030: > I'm thinking more in the context of out of tree board definitions where users could find themselves faced with unclear errors

Important to think about, but which errors specifically? ... we can customize error messages as we like.

anlavandier: Well since right now it's still unclear whether there would be a fundamental difference between the code generated for a monocolor led and smart pixel. Right now sbd only declares peripheral usage and doesn't not use the peripherals to initiate drivers. Even then i'm not sure of what the error would look like...

anlavandier: I thought a bit more about the enum based solution that Kaspar proposed. I think that overall I like it, notably because using a `#[non_exhaustive]` enum should mean that this part of the schema should be 'breaking-changes-free' for the most part. I haven't dived into how best to implement this with serde yet. 

I'm still not a fan of assuming that LEDs are monocolor by default instead of asking users for something explicit, especially since right now, the color field is optional and I think that it should probably stay optional.

I'd like an ack from you @ROMemories  to know if I should work on this and if we need more time to discuss on the implementation of the design

ROMemories: After thinking about this some more and prototyping a bit, I think it's worth exploring after all, mostly because that would make adding more types/kinds in the future more seamless, in a way that we could have many of them in a way that helps users and actually makes for better error messages.

For instance I suppose we could end up with something like this (adapted from `nordic-thingy-91-x-nrf5340-app`):

```yaml
leds:
  - type: rgb
    pins:
      - P0_14
      - P0_26
      - P0_15
    colors:
      - red
      - green
      - blue
```

(Whether we need the explicit colors in that case is TBD.)

LED matrices should still get a dedicated key and not be conflated into `leds`. 

> because using a `#[non_exhaustive]` enum should mean that this part of the schema should be 'breaking-changes-free' for the most part.

I don't think this makes a difference however, adding more `*_leds` keys would not be a breaking change either, even though a bit less seamless.

> I haven't dived into how best to implement this with serde yet.

You'll likely want to use an [internally-tagged enum](https://serde.rs/enum-representations.html#internally-tagged). We'll need to decide whether we want to use `type` or `kind` for the tag (`type` is a keyword in many languages, including Rust).

kaspar030: > I'm still not a fan of assuming that LEDs are monocolor by default instead of asking users for something explicit, especially since right now, the color field is optional and I think that it should probably stay optional.

The most basic case is "led on pin foo", and as measured, the vast majority will be regular leds. Let's keep that variant shortest. We can optimize the error messages later.

Implementation wise, you'll need a wrapper enum that is untagged, with the default as one field and the internally tagged enum (without the default) as swcond field.

`type` is fine, serde allows renaming, we already do that elsewhere. Other languages have similar methods.

anlavandier:
I'm closing this PR as the new ariel-os/sbd#163 no longer breaks existing files and nor does ariel-os/sbd#164

boards/adafruit-feather-nrf52840-express.yaml
anlavandier: I ended up using an untagged enum, see ariel-os/sbd#164

PR #2174, opened 2026-07-09
fix(rt): rerun build scripts on used env changed

kaspar030:
# Description

Debugging some lost memory on #1939, I realized that changing `CONFIG_ISR_STACKSIZE` does not rebuild the rt crate, which it needs to update the xtensa linker script.

Fixed by emitting the correct `cargo-rerun-if-env-changed` line.

~~I piggibacked the same fix for `CARGO_CFG_CTX`, not sure if that is redundant.~~ dropped

## Testing

1. build some xtensa binary
2. hard-code `CONFIG_ISR_STACKSIZE` to something large (`"100000000"`) in line 73 in `laze-project.yml`.
3. rebuild

this succeeds on main, properly fails with this PR.

## Changelog Entry

(Xtensa) The build system now correctly acts on changing `isr_stacksize_required`.

## Change Checklist

### DIFF
--- a/src/ariel-os-rt/build.rs
+++ b/src/ariel-os-rt/build.rs
@@ -54,6 +54,7 @@ fn main() {
             .replace("${ISR_STACKSIZE}", &isr_stacksize);
         std::fs::write(out.join("isr_stack_xtensa.x"), &template).unwrap();
         println!("cargo:rerun-if-changed=isr_stack_xtensa.ld.in");
+        println!("cargo:rerun-if-env-changed=CONFIG_ISR_STACKSIZE");
     }
 
     std::fs::copy("linkme.x", out.join("linkme.x")).unwrap();


src/ariel-os-rt/build.rs
```diff
@@ -67,6 +68,8 @@ fn main() {
     println!("cargo:rerun-if-changed=eheap.x");
     println!("cargo:rerun-if-changed=keep-stack-sizes.x");
 
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_CONTEXT");
+
```
ROMemories: The [documentation of `cargo::rerun-if-env-changed` says that this is not possible](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed:~:text=Note%20that%20the,script):

> Note that the environment variables here are intended for global environment variables like `CC` and such, it is not possible to use this for environment variables like `TARGET` that [Cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts). The environment variables in use are those received by `cargo` invocations, not those received by the executable of the build script.

and [`CARGO_CFG_CONTEXT` is an environment variable that Cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#:~:text=CARGO%5FCFG%5F%3Ccfg).

```suggestion
```

src/ariel-os-rt/build.rs
```diff
@@ -54,6 +54,7 @@ fn main() {
             .replace("${ISR_STACKSIZE}", &isr_stacksize);
         std::fs::write(out.join("isr_stack_xtensa.x"), &template).unwrap();
         println!("cargo:rerun-if-changed=isr_stack_xtensa.ld.in");
+        println!("cargo:rerun-if-env-changed=CONFIG_ISR_STACKSIZE");
```
ROMemories, line 57: Right, this is necessary because that variable is used in the build script *and* the code fragment that accesses it with `option_env!` (which should otherwise register the variable for rebuild) is `cfg`ed out for the `xtensa` context specifically.

ROMemories (approved):
Successfully tested and I agree this is necessary for this package (something similar might also be necessary in other packages but I haven't checked).

LGTM after removing the `cargo:rerun-if-env-changed=CARGO_CFG_CONTEXT` instruction.

PR #2176, opened 2026-07-10
feat(boards): add makerdiary nRF9151 Connect Kit

chrysn:
# Description

Generally it's a straightforward kit, similar to Nordic's (but with a CMSIS rather than a JLink bootloader).

A bit odd that its on-board LEDs are all only wired to the interface chip, but :shrug: 

## Testing

I've tested:
* examples/device-identity (both with the LTE firmware and with the DECT firmware; the latter needs `-s nrf-radiocore-firmware-dect` as is common on those chips)
* examples/gpio in a patched form that uses only inputs and prints instead of binking (so I could see the button go up and down)

## Issues/PRs References

Contributes-to: https://github.com/ariel-os/hophop/issues/49 (not closing; that'll need documentation updates in hophop)

## Changelog Entry

The Makerdiary nRF9151 Connect Kit board is now supported.

## Change Checklist


boards/makerdiary-nrf9151-connect-kit.yaml
```diff
@@ -0,0 +1,18 @@
...
+  makerdiary-nrf9151-connect-kit:
+    chip: nrf9151
+    # Pinout from https://wiki.makerdiary.com/nrf9151-connectkit/assets/attachments/nrf9151-connectkit-hardware-diagram_reva.pdf
+    leds:
+      # No user accessible LEDs: There is a green and an RGB LED, but both are
+      # only accessible from the interface chip.
+    buttons:
+      button0:
+        pin: P0_25
+        active: low # … and no built-in pull-up.
```
ROMemories, line 16: Indeed, and the DFU button is only accessible by the interface MCU.

doc/support_matrix.yml
```diff
@@ -1395,11 +1400,16 @@ boards:
     builders:
       - bbc-microbit-v2
 
-  - name: Makerdiary nRF52840 MDK USB Dongle
+  - name: makerdiary nRF52840 MDK USB Dongle
```
ROMemories, line 1403: Agreed, even though they're not *super* consistent about it: the [About Us page](https://web.archive.org/web/20260417155233/https://makerdiary.com/pages/about-us) starts with a capitalized version but the occurrence in the signature isn't….

ROMemories (approved):
I don't have the board, only checked the schematic.

LGTM!

PR #2178, opened 2026-07-13
refactor: move nrf-specific linking to ariel-os-nrf

bergzand:
# Description

Moves the nrf-specific linker generation into ariel-os-nrf

## Testing

Resulting binaries must not change in any way. Tested locally by comparing the output for examples/hello-world on the:
- nrf9151-dk
- nrf52840dk
- st-nucleo-f401re
- espressif-esp32-c3-devkit-rust-1
- rpi-pico

## Issues/PRs References

Depends on #2177

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2180, opened 2026-07-15
feat: generate nrf memory layout using memsolve

bergzand:
# Description

This migrates the memory section generation as used by the nrf to use [memsolve] to allow for dynamic generation of the memory sections. This paves the way for including bootloaders, configuration sections etc.

## Testing

Most nrf-based builds output identical binaries after and before this branch when building the `hello-world` example, except for boards based on the nrf9160 and nrf9151:
- nordic-thingy-91-x-nrf9151
- makerdiary-nrf9151-connect-kit
- nrf9151-dk
- nrf9160dk-nrf9160

The resulting binary size stays the same, but some objects are moved to different locations in the binary. These boards could use a test run to verify everything still works, but I don't have access to them 

## Issues/PRs References

~~Follow up to #2178~~

## Open Questions

Is this something we want to have mentioned in the changelog?

## Change Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -459,6 +459,7 @@ dependencies = [
  "esp-hal",
  "ld-memory",
  "linkme",
+ "memsolve",
  "portable-atomic",
  "riscv",
 ]
@@ -1002,6 +1003,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.20.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72f5acc6cb2ba439de613abc23857ec3d78374d8ed5ac84e9d11336e87da8649"
+
+[[package]]
 name = "bytemuck"
 version = "1.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1406,6 +1413,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "conv"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "78ff10625fd0ac447827aa30ea8b861fead473bb60aeb73af6c1c58caf0d1299"
+dependencies = [
+ "custom_derive",
+]
+
+[[package]]
 name = "convert_case"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1554,6 +1570,12 @@ dependencies = [
  "quote",
  "syn 2.0.114",
 ]
+
+[[package]]
+name = "custom_derive"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef8ae57c4978a2acd8b869ce6b9ca1dfe817bff704c220209fdef2c0b75a01b9"
 
 [[package]]
 name = "cyw43"
@@ -3912,6 +3934,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "js-sys"
+version = "0.3.77"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cfaf33c695fc6e08064efbc1f72ec937429614f25eef83af942d0e227c3a28f"
+dependencies = [
+ "once_cell",
+ "wasm-bindgen",
+]
+
+[[package]]
 name = "keccak"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4127,9 +4159,9 @@ dependencies = [
 
 [[package]]
 name = "log"
-version = "0.4.28"
+version = "0.4.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34080505efa8e45a4b816c349525ebe327ceaa8559756f0356cba97ef3bf7432"
+checksum = "f9f8bd3e56ce4dfc153cf470fffbfa98c7620958b312ca5c3a4b8d5181fd13c6"
 
 [[package]]
 name = "managed"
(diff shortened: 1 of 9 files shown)


src/ariel-os-nrf/build.rs
```diff
@@ -0,0 +1,96 @@
...
+const NRF91_MODEM_IPC_KB: u64 = 32;
+
+fn main() {
+    if !context("ariel-os") {
+        // Platform-independent tooling.
+        return;
+    }
+
+    let rom_start = parse_dec_or_hex(
+        &std::env::var("CHIP_ROM_START_ADDRESS").expect("CHIP_ROM_START_ADDRESS env var not set"),
```
kaspar030: don't forget to emit the cargo rerun statements ([docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed)).

bergzand: Convenience functions for this are now included in 72492e0c68ad5fc7e0be38a98147dcdd40a065a9

laze-project.yml
```diff
@@ -54,6 +54,9 @@ contexts:
           CARGO_TARGET_DIR=${relroot}/${build-dir}/bin/${builder}/cargo
           CONFIG_EXECUTOR_STACKSIZE=${CONFIG_EXECUTOR_STACKSIZE}
           CONFIG_ISR_STACKSIZE=${CONFIG_ISR_STACKSIZE}
+          CHIP_ROM_PAGE_COUNT=${CHIP_ROM_PAGE_COUNT}
+          CHIP_ROM_START_ADDRESS=${CHIP_ROM_START_ADDRESS}
+          CHIP_ROM_PAGE_SIZE_BYTES=${CHIP_ROM_PAGE_SIZE_BYTES}
```
ROMemories: Drive-by comment: I would much rather we didn't use "ROM" here: non-volatile memory/program memory has not been read-only for 35+ years and is now called that only for historical reasons; also many of our supported MCUs do have actual, manufacturer-written ROM (e.g., RPs have the bootrom and floats utility functions in ROM, STM32s have the bootloader, ESPs also have some bootloader I think).

Also, as flash is not the only technology of non-volatile memory anymore, as ReRAM is starting to be used in production (e.g., in the nRF54s or the [Baochip](https://github.com/baochip/baochip-1x)), I think it's important we really use the right terminology here, and refer to this as non-volatile memory (and later update the book to use "non-volatile memory" instead of "flash" where relevant, but that's orthogonal to this).

I'd thus prefer we use "NVM" instead of "ROM," that abbrevation is already well-established (e.g., nRF datasheets use "NVMC" for the peripheral allowing access to the flash and "NVM" to refer to ReRAM). #2207 proposes to add the term to the glossary.

bergzand: Makes sense to me, renamed in [bfb18eb](https://github.com/ariel-os/ariel-os/pull/2180/commits/bfb18eb57e6c4c5f275d4f1137f5f57451e58f20) and [7a5ea69](https://github.com/ariel-os/ariel-os/pull/2180/commits/7a5ea698e560e85506ca9da9313894b583b50945)

kaspar030:
please rebase & squash!

kaspar030:
(vetting needs fixing. exemptions fine.)

Cargo.lock
```diff
@@ -136,7 +136,7 @@ name = "ariel-os-bench"
 version = "0.5.0"
 dependencies = [
  "cortex-m",
- "defmt 1.1.1",
+ "defmt 1.1.0",
```
ROMemories: Why does this downgrade `defmt`?

bergzand: I guess that sneaked in during the rebase, let me fix that

bergzand: fixed!

laze-project.yml
```diff
@@ -238,6 +249,8 @@ contexts:
       - has_ble_nrf
     env:
       PROBE_RS_CHIP: nrf52832_xxAA
+      CHIP_NVM_PAGE_SIZE_BYTES: "512"
```
bergzand: The page size here should be `"4096"`, the chip has 512 byte blocks, with 8 of those in a  page, making them 4096 bytes

kaspar030:
@nponsard @anlavandier could you quickly test if the modem still works on these:

- [ ] nordic-thingy-91-x-nrf9151
- [ ] makerdiary-nrf9151-connect-kit
- [ ] nrf9151-dk
- [ ] nrf9160dk-nrf9160

Probably only one nrf9151 and one nrf9160 board needs testing.

nponsard:
Successfully tested with `examples/tcp-client` on `nordic-thingy-91-x-nrf9151` and `nrf9160dk-nrf9160`

bergzand:
Rebased this PR and gave memsolve the bump from 0.2.3 to 0.2.4.

Also fixed the linter issues that popped up in the mean time.

kaspar030 (changes requested):
So I went through the whole stack ([this](https://github.com/ariel-os/ariel-os/pull/2180), [rp](https://github.com/ariel-os/ariel-os/pull/2244), [esp](https://github.com/ariel-os/ariel-os/pull/2246), [stm32](https://github.com/ariel-os/ariel-os/pull/2247)), and I think it needs some rework on the metadata side.

- hard-coding RAM&flash sizes & addresses, what somewhat worked as hack for a handful of nrfs, is now multiplied, but per mcu stm32 data won't scale.
- Environment variables are used for another set of data (`CHIP_NVM_*`).
- "constant" memory.x stuff (RAM) is put together with the soon-to-be dynamic stuff (nvm)
- the per platform extra sections are constructed using individual rust functions
- every platform is special-cased, whereas they should be mostly the same

I think this would benefit alot of getting refactored so it:

1. gathers metadata (RAM base address & size, flash address & size & page size, any extra sections) => this could have `if(context("stm32"))` for now)
2. generates the memory.x based on that info

PR #2181, opened 2026-07-15
feat(uart): expose sbd defined UARTs

kaspar030:
# Description

This splits out the UART sbd handling code from the [slipmux PR](#1613).

Basically, it adds machinery to generate structs and types from the sbd generated `define_uarts!()` *call*.

Some board's host-facing UART pinouts are added so there's something to test with.

An example is added (`examples/uart-echo`).

~~This is a draft so we can discuss API et al.~~ -> API works, lets use it for now and iterate later.

**update 10.09.*

- dropped the `Assignment` trait, re-used @ROMemories idea of just implementing a `with_config()` on the generated type.

## Testing

`examples/uart-echo` hard-codes the "host facing uart".
I verified that replacing that with direct `uart0` still works:

```
-type UartPeripherals = ariel_os_boards::pins::HOST_FACING_UART;
+type UartPeripherals = ariel_os_boards::pins::uart0;
```

## Issues/PRs References

Split out of #1613.
~~Would be affected by https://github.com/ariel-os/sbd/pull/141 / https://github.com/ariel-os/ariel-os/pull/2123, and needs to either go in first or get adapted.~~ (rebased on this)

- Depends on #2123

## Changelog Entry

UARTs defined in sbd files are now exposed via macros.

## Change Checklist

### DIFF
--- a/boards/bbc-microbit-v2.yaml
+++ b/boards/bbc-microbit-v2.yaml
@@ -7,3 +7,12 @@ targets:
       - pin: P0_14
       # button1
       - pin: P0_23
+    uarts:
+      - aliases:
+          - UART_INT
+        # According to https://tech.microbit.org/hardware/schematic/, which
+        # explains that pins are labeled from the host point of view.
+        rx_pin: P1_08 # UART_INT_TX
+        tx_pin: P0_06 # UART_INT_RX
+        possible_peripherals: [UARTE0, UARTE1]
+        host_facing: true
--- a/boards/nrf52840dk.yaml
+++ b/boards/nrf52840dk.yaml
@@ -34,3 +34,12 @@ targets:
       # button3
       - pin: P0_25
         active: low
+    uarts:
+      - aliases:
+          - VCOM0
+        tx_pin: P0_06
+        rx_pin: P0_08
+        rts_pin: P0_05
+        cts_pin: P0_07
+        possible_peripherals: [UARTE0, UARTE1]
+        host_facing: true
--- a/boards/nrf9151-dk.yaml
+++ b/boards/nrf9151-dk.yaml
@@ -44,3 +44,23 @@ targets:
         active: low
         aliases:
           - Button 4
+    uarts:
+      - aliases:
+          - VCOM0
+        tx_pin: P0_27
+        rx_pin: P0_26
+        rts_pin: P0_14
+        cts_pin: P0_15
+        # FIXME: Is this exhaustive? (Seems those are all we have in Ariel right now)
+        possible_peripherals: [SERIAL3]
+        host_facing: true
+      # Pins are correct, but SERIAL3 is already taken
+      # - aliases:
+      #   - VCOM1
+      #   tx_pin: P0_29
+      #   rx_pin: P0_28
+      #   rts_pin: P0_16
+      #   cts_pin: P0_17
+      #   # FIXME: Is this exhaustive? (Seems those are all we have in Ariel right now)
+      #   possible_peripherals: [SERIAL3]
+      #   host_facing: true
--- a/examples/uart-echo/README.md
+++ b/examples/uart-echo/README.md
@@ -1,0 +1,13 @@
+# UART echo example
+
+## About
+
+This application shows how to use the UARTs that are defined in a board's SBD descriptions.
+
+## How to run
+
+In this directory, run:
+
+    laze build -b nrf52840dk run
+
+The example initializes the host-facing UART (at 115200baud 8N1) and echoes all incoming bytes.
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -163,8 +163,8 @@ macro_rules! define_uart {
 
         #[allow(nonstandard_style)]
         pub struct $name {
-            tx: $crate::hal::peripheral::Peri<'static, peripherals::$tx>,
-            rx: $crate::hal::peripheral::Peri<'static, peripherals::$rx>,
+            tx: $crate::__peripheral_ty!($tx),
+            rx: $crate::__peripheral_ty!($rx),
         }
 
         impl $crate::hal::TakePeripherals<$name> for &mut $crate::hal::OptionalPeripherals {
@@ -178,8 +178,8 @@ macro_rules! define_uart {
 
         impl $crate::uart::Assignment for $name {
             type Device<'a> = $crate::hal::uart::$device<'a>;
-            type Tx = $crate::hal::peripheral::Peri<'static, peripherals::$tx>;
-            type Rx = $crate::hal::peripheral::Peri<'static, peripherals::$rx>;
+            type Tx = $crate::__peripheral_ty!($tx);
+            type Rx = $crate::__peripheral_ty!($rx);
 
             fn into_pins(self) -> (Self::Tx, Self::Rx) {
                 (self.tx, self.rx)
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -145,15 +145,14 @@ macro_rules! define_uarts {
     ( $($_:tt)+ ) => {};
 }
 
-/// Creates a struct of the given `name`, containing everything needed to set up a UART.
+/// Creates a struct of the given `name`, containing all peripherals needed to set up a UART.
 ///
 /// It holds the TX and RX pins, and can be taken through the same mechanism as those defined using
 /// [`ariel_os::hal::define_peripherals!`] (that is, by using [`ariel_os::task(autostart,
 /// peripherals)`] or the underlying `TakePeripherals` trait).
 ///
-/// The struct also implement [`ariel_os::uart::Assignment`], and by that carries a type with the
-/// (not trait-backed) promise that it is something that can be initialized through the
-/// [`ariel_os::uart`] APIs.
+/// The struct also has a method [`with_config()`] that can be used to initialize an Ariel OS uart
+/// instance.
 #[cfg(feature = "uart")]
 #[macro_export]
 macro_rules! define_uart {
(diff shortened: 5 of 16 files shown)


examples/uart-echo/README.md
```diff
@@ -0,0 +1,13 @@
+# UART echo example
+
+## About
+
+This application hows how to use the UARTs that are defined in a board's SBD descriptions.
```
ROMemories, line 5:
```suggestion
This application shows how to use the UARTs that are defined in a board's SBD descriptions.
```

examples/uart-echo/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## How to run
+
+2. In this directory, run:
```
ROMemories, line 9:
```suggestion
In this directory, run:
```

boards/bbc-microbit-v2.yaml
```diff
@@ -7,3 +7,15 @@ targets:
       - pin: P0_14
       # button1
       - pin: P0_23
+    uarts:
+      - aliases:
+          - UART_INT
+        # According to https://tech.microbit.org/hardware/schematic/, which
+        # explains that pins are labeled from the host point of view. In the
+        # schematics of the 2.0 and 2.2 versions, the opposite is shown and
+        # claimed.
```
ROMemories, line 16: I'm not sure I follow this comment, especially the second sentence: I think this refers to [this comment from the documentation](https://tech.microbit.org/hardware/schematic/#:~:text=These,pin,-%2E):

> These names follow the schematic’s naming convention, labeled from the interface’s perspective, not the target MCU’s. So, UART_INT_TX corresponds to the MCU’s RX pin, and UART_INT_RX to the MCU’s TX pin.

However everything seems in order to me: P1.08 is indeed the RX pin for the (target) MCU and therefore the TX pin (i.e., `UART_INT_TX`) of the *interface* MCU (which is what's shown in the schematic). Am I missing something here? (If not I think the second sentence could just be removed, as I understood it as pointing at an inconsistency in the board's docs.)

boards/bbc-microbit-v2.yaml
```diff
@@ -7,3 +7,15 @@ targets:
...
+        rx_pin: P1_08 # UART_INT_TX
+        tx_pin: P0_06 # UART_INT_RX
+        # FIXME: Is this exhaustive? (Seems those are all we have in Ariel right now)
```
ROMemories, line 19: The nRF52840 only has two UART peripherals (and both are indeed supported):
```suggestion
```

boards/nrf52840dk.yaml
```diff
@@ -34,3 +34,14 @@ targets:
...
         active: low
+
+    uarts:
+      - aliases:
+          - VCOM0
+        tx_pin: P0_06
+        rx_pin: P0_08
+        rts_pin: P0_05
+        cts_pin: P0_07
+        # FIXME: Is this exhaustive? (Seems those are all we have in Ariel right now)
```
ROMemories, line 45: Same:
```suggestion
```

boards/nrf52840dk.yaml
```diff
@@ -34,3 +34,14 @@ targets:
       # button3
       - pin: P0_25
         active: low
+
```
ROMemories, line 37:
```suggestion
```

boards/nrf9151-dk.yaml
```diff
@@ -44,3 +44,22 @@ targets:
...
+          - VCOM0
+        tx_pin: P0_27
+        rx_pin: P0_26
+        rts_pin: P0_14
+        cts_pin: P0_15
+        # FIXME: Is this exhaustive? (Seems those are all we have in Ariel right now)
+        possible_peripherals: [SERIAL3]
+        host_facing: true
+      # Pins are correct, but SERIAL3 is already taken
+      # VCOM1:
```
ROMemories, line 58: Indentation might need some work:
```suggestion
      # - aliases:
      #    - VCOM1
```

boards/nrf9151-dk.yaml
```diff
@@ -44,3 +44,22 @@ targets:
...
           - Button 4
+    uarts:
+      - aliases:
+          - VCOM0
+        tx_pin: P0_27
+        rx_pin: P0_26
+        rts_pin: P0_14
+        cts_pin: P0_15
+        # FIXME: Is this exhaustive? (Seems those are all we have in Ariel right now)
+        possible_peripherals: [SERIAL3]
```
ROMemories, line 55: Looking at this again, we can't have `SERIAL3` as a name in SBD files, as this is not actually the name of a peripheral (the datasheet calls these `UARTEx` as for other nRFs), but apparently [the name of the peripheral's interrupt, which Embassy decided to use for its peripherals (likely for mutual exclusion)](https://redirect.github.com/embassy-rs/embassy/pull/1225).

I think we should consider introducing aliases for our I2C/SPI/UART drivers' peripherals (plus that can only make these names clearer and more consistent with the only nRFs), and/or rename this field or clarify its meaning.

src/ariel-os-hal/src/hal/define_peripherals.rs
ROMemories, line 1: All the doc links from the macros are broken.

boards/bbc-microbit-v2.yaml
ROMemories, line 1: Successfully tested the example on this board.

ROMemories (commented):
**I have only exhaustively reviewed the SBD files for now.** (The rest is just drive-by comments.)

boards/nrf9151-dk.yaml
kaspar030: offline discussion: we go with the embassy names for now

boards/bbc-microbit-v2.yaml
kaspar030: dropped the second sentence

src/ariel-os-hal/src/hal/define_peripherals.rs
kaspar030: fixed

examples/uart-echo/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This application shows how to use the UARTs that are defined in a board's SBD descriptions.
+
+## How to run
+
+In this directory, run:
+
+    laze build -b nrf52840dk run
+
+The example initializes the host-facing UART and echoes all incoming bytes.
```
ROMemories: This should specify the baudrate, as the serial monitor on the host should be configured with the correct one or it won't work.

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -108,6 +108,122 @@ macro_rules! group_peripherals {
     }
 }
 
+/// This macro defines a `uarts` module that gives access to UARTs by various accessors (currently:
+/// through accessor types, which can be used as peripherals in autostart tasks).
```
ROMemories: It doesn't define a module.

src/ariel-os-hal/src/hal/define_peripherals.rs
```diff
@@ -108,6 +108,122 @@ macro_rules! group_peripherals {
...
+        // FIXME: This can only be accessed by application that have some prior knowledge of
+        // multiple UARTs being available, and, more importantly, has negotiated between its
+        // components which one takes which. Once any future mechanism enables that negotiation,
+        // these types will need to be advertised to this mechanism.
+        pub type HOST_FACING_UART_2 = $name2;
+    };
+    ( $name:ty, $name2:ty $(, $($_:ty),+)? ) => {
+        compile_error!(
+            "Larger numbers of host facing UARTs can be supported by expanding the `_define_host_facing_uarts` macro of ariel-os-hal to more cases."
+            );
```
ROMemories: optional:
```suggestion
        );
```

src/ariel-os-hal/src/hal/define_peripherals.rs
ROMemories, line 1: Actually, do we need these macros to show up in the docs at all right now? I think we should hide them (at least for now); they're not actually enough for users to make sense of how to make use of the generated structs, and that should be easy to document in #2225 instead. (And this would give us more leeway until this really stabilizes.)

ROMemories (commented):
Successfully tested the example on nrf52840dk.

> `examples/uart-echo` hard-codes the "host facing uart".
>  I verified that replacing that with direct `uart0` still works:

It does, but that should be `Uart0`, following https://github.com/ariel-os/sbd/pull/184. Changing that would be a clear breaking change, unless we consider these explicitly hidden for the time being (but not for long, we'll need to document them in #2225 to make this usable).

boards/nrf9151-dk.yaml
ROMemories: We discussed to go with these for now (even if these are the interrupt names, not the hardware peripheral names), as a temporary path, and that @kaspar030 would clarify the expectations of the cross-project re-usability of the SBD files in the README of [sbd](https://github.com/ariel-os/sbd).

src/ariel-os-hal/src/hal/define_peripherals.rs
kaspar030: reworded

src/ariel-os-hal/src/hal/define_peripherals.rs
kaspar030: Yes, lets have them in.

kaspar030:
> It does, but that should be `Uart0`, following [ariel-os/sbd#184](https://github.com/ariel-os/sbd/pull/184).

Yes, but that capitalization is not part of the sbd-gen version *currently* used by Ariel OS main. The PR that bumps to the next sbd-gen version will have to update this.

ROMemories:
> Yes, but that capitalization is not part of the sbd-gen version _currently_ used by Ariel OS main. The PR that bumps to the next sbd-gen version will have to update this.

I know, but that would be an unneeded breaking change, just after landing the functionality, while we could just do this now as the rename seems like the only change in sbd-gen since the last release.

kaspar030:
> that would be an unneeded breaking change

Breaking changes only materialize at release time. If we do as you suggest, we materialize a breaking change in sbd-gen *now*.

ROMemories:
> Breaking changes only materialize at release time. If we do as you suggest, we materialize a breaking change in sbd-gen _now_.

The breaking change I'm concerned about is the one in Ariel OS (`ariel-os-boards` specifically). The change in `sbd-gen` is even only arguably breaking considering we don't use it and there is no documentation about what it generates anyway.

Is your plan to group this possible breaking *from sbd-gen* with other breaking changes there, and make sure to not cut a release of Ariel OS before the required (breaking) changes are propagated in Ariel OS?

ROMemories (approved):
LGTM, however we need to be careful not to cut a release before we release and use https://github.com/ariel-os/sbd/pull/184, otherwise this will result in an unneeded breaking change on the name of generated `uartx`/`Uartx` structs.

Issue #2090, opened 2026-05-13
feat: remote actionable ping-ish tools


chrysn:
### Description

I'd like to have a ping client as a component in Ariel OS, which can be added to an application to allow measuring network performance.

The [RIOT CoAP ping demo](https://codeberg.org/RIOT/rust-module-examples/src/branch/master/riot-coap-handler-demos/src/ping.rs) (usable in the [coap_through_embeddednal](https://codeberg.org/RIOT/rust-examples/src/branch/main/coap_through_embeddednal) example) can serve a starting point.

### Motivation

This would be useful for general applications ("how well have my devices been reachable"), and in particular for hophop.

### Requirements

(Requirements? More like suggestions of what it should do).

There is a spectrum of ping, tracert, multicast ping, CoAP ping / proxy trace, maybe port scanning, and I think even iperf, all revolving around the concept of having some fixed number of recent timings -- where all can be triggered, watched live, or be fetched in the end as a summary.

Common components are:

* Triggering follows the REST CRUD pattern (not really using the U): POST a job description, receive a URI, poll or observe (pending enhancements in the CoAP stack) that resource, see the final state, and eventually delete it.
  * The startup message contains details on what to do, in particular "where to", "how often" and "in which intervals".
* The operation runs as a task that regularly wakes up to send something.
* There's a sliding window of at least whether we've received responses back; events that get back are ticked off, compute their delta to transmission time (which may be in the table or may be implied from the table position) and include that data in a summary (min, max, total/count, maybe some preconfigured histogram if we're being fancy)
  * No we can't simplify here: While we could pack information to be round tripped in some cases (eg. encrypt the sending time stamp into the ping payload), maths will go wrong when duplicates arrive.
* Size of the window and whether there's data to be stored in the window is up to the configured mode -- can be just boolean bits, can be precise RTTs, can be other details.
  * Either there are fixed sizes so the task type becomes a union, or we make this configurable, dynamically sized and allocated (from Global or from a custom pool). In the former case, a valid POST succeeding depends on the availability of a slot, in the latter, on the availability of sufficient storage.
* Some modes report specific data into a configured number of slots; in multicast pings and traceroutes, that's the address that came back; in port scans, it's the port and maybe some part of the message.
[...]

Issue #2029, opened 2026-04-20
Buffered UART


nponsard:
### Description

Add `embedded_io_async::BufRead` implementation to UART drivers.

Currently it can be easily implemented on the MCUs that have an embassy HAL, ESP is missing the implementation.

Currently you have to edit the UART driver for the HAL you're using to add `impl_async_uart_bufread_for_driver_enum!(Uart, $( $peripheral ),*);` , see c0520d9e10579df0d2b488e26ef0c34e8d4a0aae

### Requirements

`embedded_io_async::BufRead` to be implemented on every available UART driver.

### Open questions

Should we make a laze context expressing the requirement for buffered UART so we can enable it on HALs that support it ?

### Are you planning to do it yourself in a pull request?

Maybe

PR #2184, opened 2026-07-15
feat(coap): add initial ping support POC

TallCheesecakeRust:
# Description

This PR adds an initial proof-of-concept ping component to `ariel-os-coap`.

The current implementation provides:
- ICMP ping execution using `embassy-net`
- a worker pool for asynchronous ping tasks
- CoAP request handling for submitting ping targets
- CBOR decoding of ping requests
- basic collection of ping timing statistics

This is intended as a first implementation to validate the approach and gather feedback on the final API/resource model.

## Testing

Tested by:
- building the affected crate with `cargo check`

## Issues/PRs References

[Related to #2029](https://github.com/ariel-os/ariel-os/issues/2090)

## Open Questions

The following design points are intentionally left open for discussion:

- Should ping requests become explicit job resources (`POST` → job URI → poll/observe)?
- What should the final `GET` semantics for statistics be?
- Should statistics represent current state, a history/ring buffer, or streamed events?
- How should ping failures and error codes be represented?
- Should CoAP Observe be used for live updates once supported?

## Changelog Entry

Adds an initial proof-of-concept CoAP ping component.

## Change Checklist

- [x] The commit history will be clean at the latest after applying autosquash.
- [x] I have followed the Coding Conventions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,6 +224,8 @@ dependencies = [
  "cboritem",
  "coap-handler",
  "coap-handler-implementations",
+ "coap-message",
+ "coap-message-utils",
  "coap-numbers",
  "coapcore",
  "critical-section",
@@ -237,6 +239,7 @@ dependencies = [
  "lakers",
  "lakers-crypto-rustcrypto",
  "minicbor",
+ "minicbor-derive",
  "serde",
  "serde_yaml",
  "static_cell",
@@ -1055,7 +1058,7 @@ dependencies = [
  "embedded-io 0.6.1",
  "embedded-io-async 0.6.1",
  "futures-intrusive",
- "heapless 0.8.0",
+ "heapless 0.9.1",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


chrysn:
Picking some simple questions:

> Should ping requests become explicit job resources (POST → job URI → poll/observe)?

Thanks to REST best practice that is a detail that can change; initially it can just return its own Location. What is definitely not needed is an `.at()` tree entry for each job – that can't be done at runtime anyway; a single handler added through `.below()` can process a remaining path component.

> How should ping failures and error codes be represented?

Without any good ideas on representing failures to send: All reports are successful on the CoAP level; an "every ping went through" report is just as 2.05 Content as an "I can't even send there because I have no route". (Only malformed POSTs would be a 4.00 Bad Request right away).

> Should CoAP Observe be used for live updates once supported?

That'd be great, but the current CoAP stack can't do that yet. But that changes little about the implementation, and will almost transparently work as soon as the stack improves.

TallCheesecakeRust:
Hey there, unfortunately due to some personal events I will not be able to do any work here

PR #2186, opened 2026-07-16
refactor(stm32): move the RCC configuration to a dedicated module

ROMemories:
# Description

This moves the default RCC configuration provided by Ariel OS for STM32 MCUs to a dedicated module. On top of slightly decoupling this a bit from the rest of the init, this is motivated by soon-to-be PRed changes to the way the clock configuration is managed on STM32, so it can be properly restored when waking-up from low-power stop modes.

## Testing

```sh
laze -C examples/gpio build -b stm32u083c-dk run
```

The `rcc-config-override` feature can be tested similarly to #1389.

## Change Checklist


src/ariel-os-stm32/src/lib.rs
```diff
@@ -104,346 +106,7 @@ pub fn init() -> OptionalPeripherals {
...
-        _ => {
-            config.rcc = rcc_config();
-        }
-    }
-}
-
-// TODO: find better place for this
-#[cfg_attr(feature = "rcc-config-override", expect(dead_code))]
-fn rcc_config() -> embassy_stm32::rcc::Config {
-    #[allow(unused_mut, reason = "conditional compilation")]
```
ROMemories, line 123: Everything below is moved as is.

src/ariel-os-stm32/src/rcc.rs
```diff
@@ -0,0 +1,339 @@
...
+            unsafe extern "Rust" {
+                fn __ariel_os_rcc_config() -> embassy_stm32::rcc::Config;
+            }
+            unsafe { __ariel_os_rcc_config() }
+        }
+        _ => default(),
+    }
+}
+
+#[cfg_attr(feature = "rcc-config-override", expect(dead_code))]
```
nponsard, line 13: Is there a reason why `#[cfg(not(feature = "rcc-config-override"))]` isn't used here ?

ROMemories: It's not part of this PR as it's really only intended as a pure refactor; I'll consider that in the next PRs that will touch this module (and there might be a few). The only cost is some extra codegen at most, so this doesn't seem like a problem for now.

nponsard (approved):
Looks good to me,
I was worried about a build warning (`unused import: `ariel_os_hal::hal::peripherals`) but this is caused by parent commits and fixed in main.

PR #2190, opened 2026-07-24
ci: check that only one changelog-x label is used per pr

### LINKED ISSUE #2185: CI check for `changelog:*` labels
#1757 had added a CI check to enforce that PR descriptions contain the magic comments used by changelog-harvester, unless the `changelog:skip` label is attached to the PR.

We should now have an extra check that enforces that *exactly one* `changelog:*` label be attached to every single PR.

anlavandier:
# Description

## Testing

Tested with: 
- [x] each individual changelog labels
- [x] changelog-skip + other labels
- [x] two non-skip labels
- [x] zero labels
## Issues/PRs References

Closes #2185 
## Changelog Entry

Test changelog

## Change Checklist

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -13,6 +13,11 @@ jobs:
   check-changelog:
     runs-on: ubuntu-24.04
     steps:
+      - name: "Check that a changelog label exists"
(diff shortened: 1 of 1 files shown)


ROMemories (approved):
The GH Action hasn't had commits in one year, but it seems to work well; the workflow has read-only access to content so this seems innocuous enough. Let's go with this new check for now, we can always change the implementation later.

PR #2191, opened 2026-07-24
refactor: BLE stack handling

nponsard:
# Description

This refactors how the BLE stack is managed in Ariel OS, drivers now have to provide the controller instead of the whole Stack.

This factors out the initialization of the stack, so its configuration can be done in one unique place in the code.

```
laze -C examples/ble-scanner/ build -b seeedstudio-xiao-esp32c6 run
laze -C examples/ble-scanner/ build -b rpi-pico-w run
laze -C examples/ble-scanner/ build -b nrf52840dk run
```

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/src/ble.rs
+++ b/src/ariel-os-embassy/src/ble.rs
@@ -9,20 +9,37 @@
 //!
 //! The address is not currently rotated during execution; however this behavior may not be relied upon.
 
-use embassy_sync::once_lock::OnceLock;
+use ariel_os_embassy_common::cell::SameExecutorCell;
+use embassy_executor::Spawner;
+use embassy_sync::{
+    blocking_mutex::raw::CriticalSectionRawMutex, mutex::Mutex, once_lock::OnceLock,
+};
 use futures_util::FutureExt as _;
+use static_cell::StaticCell;
 use trouble_host::{
-    Address,
-    prelude::{AddrKind, BdAddr},
+    Address, Stack,
+    prelude::{AddrKind, BdAddr, DefaultPacketPool},
 };
 
 use ariel_os_embassy_common::ble::Config;
 use ariel_os_log::debug;
 
-// Must be async and return &trouble_host::Stack<'static, impl Controller>
-pub use crate::hal::ble::ble_stack;
+use crate::hal::ble::BleController;
+
+/// The BLE Stack, representing access to this device's BLE controller.
+/// This type is returned by [`ble_stack()`].
+///
+/// It is a [`trouble_host::Stack`] configured with an appropriate BLE controller and the default
+/// packet pool.
+pub type BleStack = Stack<'static, BleController, DefaultPacketPool>;
 
 static CURRENT_ADDRESS: OnceLock<Address> = OnceLock::new();
+#[allow(dead_code)]
+static STACK: StaticCell<SameExecutorCell<BleStack>> = StaticCell::new();
+// The stack can effectively only be taken by a single application; once taken, the Option is None.
+static STACKREF: OnceLock<
+    Mutex<CriticalSectionRawMutex, Option<&'static mut SameExecutorCell<BleStack>>>,
+> = OnceLock::new();
 
 #[allow(dead_code, reason = "false positive during builds outside of laze")]
 pub(crate) fn config() -> Config {
@@ -68,4 +85,41 @@ fn get_random_addr() -> [u8; 6] {
         "CONFIG_BLE_STATIC_ADDRESS",
         "static address for BLE in format XX:XX:XX:XX:XX:XX",
     )
+}
+
+/// Returns the system BLE stack, [`BleStack`].
+///
+/// # Panics
+///
+/// For this function to succeed, it can only be called from the system executor, and only once:
+///
+/// - Panics if the stack was already taken.
+/// - Panics when not called from the system executor.
+#[doc(alias = "bluetooth_stack")]
+pub async fn ble_stack() -> &'static mut BleStack {
+    STACKREF
+        .get()
+        .await
+        .try_lock()
+        .expect("Two tasks racing for lock, one would fail the main-executor check")
+        .take()
+        .expect("Stack was already taken")
+        .get_mut_async()
+        .await
+        .expect("Stack needs to be taken from main executor")
+}
+
+#[allow(dead_code, reason = "false positive during builds outside of laze")]
+pub(crate) fn init_stack(controller: crate::hal::ble::BleController, spawner: Spawner) {
+    let config = config();
+    let mut rng = ariel_os_random::crypto_rng();
+
+    let resources = ariel_os_embassy_common::ble::get_ble_host_resources();
+
+    let stack = trouble_host::new(controller, resources)
+        .set_random_generator_seed(&mut rng)
+        .set_random_address(config.address);
+
+    let stackref = STACK.init(SameExecutorCell::new(stack, spawner));
+    let _ = STACKREF.init(Some(stackref).into());
 }
--- a/src/ariel-os-embassy/src/lib.rs
+++ b/src/ariel-os-embassy/src/lib.rs
@@ -333,12 +333,16 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     #[cfg(all(feature = "ble", not(any(context = "esp", context = "rp"))))]
     let ble_controller = hal::ble::build_controller(&mut peripherals, spawner);
 
-    #[cfg(all(feature = "ble-cyw43", not(feature = "wifi-cyw43")))]
-    let _ = hal::cyw43::device(&mut peripherals, &spawner, ble_config).await;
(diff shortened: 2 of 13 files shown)


src/ariel-os-embassy/src/ble.rs
```diff
@@ -69,3 +81,36 @@ fn get_random_addr() -> [u8; 6] {
         "static address for BLE in format XX:XX:XX:XX:XX:XX",
     )
 }
+
+/// Returns the system ble stack.
+///
+/// # Panics
+/// - panics if the stack was already taken
+/// - panics when not called from the main executor
```
ROMemories: This should keep the existing docs:

```suggestion
/// Returns the system BLE stack, [`trouble_host::Stack`].
///
/// # Panics
///
/// For this function to succeed, it can only be called from the system executor, and only once:
///
/// - Panics if the stack was already taken.
/// - Panics when not called from the system executor.
```

src/ariel-os-embassy/src/ble.rs
```diff
@@ -69,3 +81,36 @@ fn get_random_addr() -> [u8; 6] {
...
+        .expect("Stack was already taken")
+        .get_mut_async()
+        .await
+        .expect("Stack needs to be taken from main executor")
+}
+
+#[allow(dead_code, reason = "false positive during builds outside of laze")]
+pub(crate) fn init_stack(controller: crate::hal::ble::BleController, spawner: Spawner) {
+    let config = config();
+    let mut rng = ariel_os_random::crypto_rng();
```
ROMemories, line 115: If the dependency on an HWRNG is factor out here (as it should), the dependency on the `hwrng` laze module should also be moved from the `ble-*` modules to the `ble` laze module itself.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -340,15 +330,23 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
+    let ble_controller = hal::ble::build_controller(&mut peripherals, spawner);
+
+    #[cfg(any(feature = "ble-cyw43", feature = "wifi-cyw43"))]
+    let hal::cyw43::Cyw43Device {
+        #[cfg(feature = "ble-cyw43")]
+        ble_controller,
+        #[cfg(feature = "wifi-cyw43")]
+            net_device: device,
+        #[cfg(feature = "wifi-cyw43")]
+            net_control: control,
```
ROMemories, line 343:
```suggestion
        net_device: device,
        #[cfg(feature = "wifi-cyw43")]
        net_control: control,
```

src/ariel-os-embassy/src/lib.rs
```diff
@@ -237,11 +237,6 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
     #[cfg(all(feature = "usb", context = "nrf"))]
     hal::usb::init();
 
-    // Move out the peripherals required for drivers, so that tasks cannot mistakenly take them.
```
ROMemories, line 240: By removing this here I guess we lose this, but other network-related init functions don't do the same so I suppose that's fine. We can always consider moving them all before the tasks later.

src/ariel-os-embassy/src/ble.rs
```diff
@@ -9,20 +9,32 @@
...
 };
 
 use ariel_os_embassy_common::ble::Config;
 use ariel_os_log::debug;
 
-// Must be async and return &trouble_host::Stack<'static, impl Controller>
-pub use crate::hal::ble::ble_stack;
+use crate::hal::ble::BleController;
+
+pub type BleStack = Stack<'static, BleController, DefaultPacketPool>;
```
ROMemories, line 34: This now shows up in the docs.

src/ariel-os-esp/src/ble.rs
```diff
@@ -1,60 +1,16 @@
...
-> = OnceLock::new();
+pub type BleController = ExternalController<BleConnector<'static>, SLOTS>;
 
-pub async fn init(
+pub fn build_controller<'a>(
     peripherals: &mut esp_hal::peripherals::OptionalPeripherals,
-    config: &ariel_os_embassy_common::ble::Config,
-    spawner: Spawner,
-) {
+    _spawtner: Spawner,
```
ROMemories:
```suggestion
    _spawner: Spawner,
```

src/ariel-os-embassy/src/ble.rs
```diff
@@ -69,3 +81,36 @@ fn get_random_addr() -> [u8; 6] {
         "static address for BLE in format XX:XX:XX:XX:XX:XX",
     )
 }
+
+/// Returns the system ble stack.
+///
+/// # Panics
+/// - panics if the stack was already taken
+/// - panics when not called from the main executor
+pub async fn ble_stack() -> &'static mut BleStack {
```
ROMemories, line 99: We likely want to keep the doc alias on this:
```suggestion
#[doc(alias = "bluetooth_stack")]
pub async fn ble_stack() -> &'static mut BleStack {
```

src/ariel-os-hal/src/hal/dummy/ble.rs
```diff
@@ -3,30 +3,14 @@ use bt_hci::{
...
-///
-/// - Panics if the stack was already taken.
-/// - Panics when not called from the system executor.
-#[doc(alias = "bluetooth_stack")]
-pub async fn ble_stack() -> &'static Stack<'static, DummyController, DefaultPacketPool> {
-    async { unimplemented!() }.await
-}
-
-pub fn driver(_p: Peripherals, _spawner: Spawner, _config: ariel_os_embassy_common::ble::Config) {
+// This is what you have to implement, if it's tied to WiFi initialization, make a special case in `ariel-os-embassy`.
```
ROMemories: Who's "you?" When do you need to implement that?

ROMemories (commented):
Might be good to go after this one.

src/ariel-os-embassy/src/lib.rs
nponsard: Only the nRF implementation was doing this, the other ones were tied to the network setup so I moved everything close together.

src/ariel-os-embassy/src/ble.rs
nponsard: I think that's a good thing, is it ?

src/ariel-os-embassy/src/lib.rs
nponsard: `rustfmt` (`laze build fmt`) formats it like this.

src/ariel-os-hal/src/hal/dummy/ble.rs
nponsard: Reworded it to improve clarity. This is addressed to implementors.

src/ariel-os-embassy/src/lib.rs
ROMemories: Indeed, that's odd.

src/ariel-os-embassy/src/ble.rs
ROMemories: Then it should have documentation

nponsard: added documentation, should I also change the documentation of `ble_stack()` to link to this type alias ?

ROMemories: That wouldn't hurt, and maybe move the reference to `trouble_host` to the docs of `BleStack`.

ROMemories (approved):
Thanks for factoring this out, seems definitely clearer and more maintainable that way.

LGTM after ideally adding the backlinks to the docs.

src/ariel-os-embassy/src/ble.rs
```diff
@@ -9,20 +9,36 @@
...
 use ariel_os_log::debug;
 
-// Must be async and return &trouble_host::Stack<'static, impl Controller>
-pub use crate::hal::ble::ble_stack;
+use crate::hal::ble::BleController;
+
+/// The BLE Stack, representing access to this device's BLE controller.
+/// This type is returned by [`ble_stack()`].
+///
+/// It is setting a concrete controller type and packet pool to [`trouble_host::Stack`].
```
ROMemories: I'm not sure of what that implies: phrased like this it sounds to me like an implementation detail, which it may not be, but I'm not sure what to make of it as a user (I'm not sure of the meaning of "setting […] to" also).

nponsard: I want to express that it's a `trouble_host::Stack` but with the generic parameters already configured according to the driver. I can't find the right wording.

ROMemories: Would the following work? (I'm not sure what the packet pool does, maybe it should be "the" default packet pool.)

```suggestion
/// It is a [`trouble_host::Stack`] configured with an appropriate BLE controller and a default packet pool.
```

Edit: This -> It

nponsard: This would work thanks. Changing it to "the default packet pool" since it's a global packet pool.

ROMemories (approved):
LGTM, please autosquash!

PR #2193, opened 2026-07-29
[codex] Make wildcard exports explicit

### LINKED ISSUE #2099: Make wildcard exports explicit
### Description

Currently we have, in ariel_os:

```rust
pub use ariel_os_embassy::api::*;
```

As a consequence, users who use net without enabling it get:

```rust
error[E0433]: cannot find `net` in `ariel_os`
   --> src/bin/nordicmacdemo.rs:845:27
    |
845 |     let stack = ariel_os::net::network_stack();
    |                           ^^^ could not find `net` in `ariel_os`
```

If instead we had

```rust
#[cfg(feature = "net")]
pub use ariel_os_embassy::api::net;
```

in there explicitly, they'd get something like:

```
error[E0433]: cannot find `net` in `ariel_os`
   --> src/bin/nordicmacdemo.rs:845:27
    |
845 |     let stack = ariel_os::net::network_stack();
    |                           ^^^ could not find `net` in `ariel_os`
    |
note: found an item that was configured out
   --> build/imports/ariel-os/src/ariel-os/src/lib.rs:63:32
    |
 62 | #[cfg(feature = "net")]
    |       --------------- the item is gated behind the `net` feature
 63 | pub use ariel_os_embassy::api::net;
```

Yeah it's more work to keep those up to date, but also an opportunity to reduce indirections.

### Motivation

Ran into this because I hoped I'd get a sensible error message (rather than looking up whether I'd need to enable a feature in Cargo or somethin in laze).

### Current Implementation

Plain blanket imports about which for some reason no linter has complained too loudly yet.

Maybe but not soon

Jonesxq:
## Summary

- replace the wildcard re-exports from `ariel_os_hal::api` and `ariel_os_embassy::api` with explicit exports
- apply the matching Ariel OS feature gate to each optional API module
- preserve the unconditional exports for the always-available APIs

## Why

Wildcard re-exports hide the relevant `cfg` metadata from downstream compiler diagnostics. Putting the feature gates on the `ariel-os` facade exports lets Rust explain that a missing API was configured out and name the feature required to enable it.

## Validation

- `cargo fmt --check --all -- --config format_generated_files=false`
- positive consumer compile importing all 13 explicit exports with their required features enabled
- negative consumer compile without `net`, confirming the diagnostic says the item was configured out and is gated behind the `net` feature
- `git diff --check`

Closes #2099

### DIFF
--- a/src/ariel-os/src/lib.rs
+++ b/src/ariel-os/src/lib.rs
@@ -33,8 +33,14 @@ pub use ariel_os_buildinfo as buildinfo;
 pub use ariel_os_coap as coap;
 #[doc(inline)]
 pub use ariel_os_debug as debug;
+#[cfg(feature = "i2c")]
 #[doc(inline)]
-pub use ariel_os_hal::api::*;
+pub use ariel_os_hal::api::i2c;
+#[cfg(feature = "uart")]
+#[doc(inline)]
+pub use ariel_os_hal::api::uart;
+#[doc(inline)]
+pub use ariel_os_hal::api::{gpio, hal};
 #[doc(inline)]
 pub use ariel_os_identity as identity;
 #[doc(inline)]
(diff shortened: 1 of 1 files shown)


bergzand:
Hi, welcome to Ariel OS and thank you for your contribution.

We'll have time to review your PR in a bit. In the mean time, could you please use and fill out the pull request template we use for new PRs?

PR #2204, opened 2026-08-10
feat(esp32): Add PSRAM support

luukoudshoorn:
# Description
Some of the ESP32 boards ship with additional PSRAM. Esp-hal has support for this, we just need enable the feature, set one additional constant and initialize the PSRAM at the right moment.

One caveat is that either QSPI or OSPI has to be selected and this **must** match the actual board wiring. If the wrong type is selected, everything will compile, init, alloc and write without any issue, but reading beyond the internal buffer of the ESP32 will return garbage data (and confidently so, still no error). This makes the feature a little sensitive to proper board configuration. This PR describes this in the suggested documentation change.

## Testing
This was tested on a PCB that resembles the [Espressif ESP32-S3-DevKitC-1](https://ariel-os.github.io/ariel-os/dev/docs/book/boards/espressif-esp32-s3-devkitc-1.html), with a N16R8 memory configuration (16MiB additional flash, 8MiB additional PSRAM). This memory is wired up as OSPI.

To test this feature, alloc a substantial buffer (few MiB) and fill it with a known pattern. Then read it back and verify the pattern. If a QSPI chip is called with an OSPI driver or the other way around, data will be garbage (even though the alloc will succeed). If the right configuration is used, data wil read back as expected.

## Open Questions
This now only targets the board I can test here, but more candidates are available for this feature:
- unihiker-k10
- waveshare-esp32-s3-matrix
- espressif-esp32-s2-devkitc-1 (I think, though this is an older chip that doesn't even have octal mode, so unsure how to configure this properly)

Please advise on adding this now or only somehow noting this for future reference but leaving it unavailable until it can be confirmed.

## Changelog Entry

(ESP32) PSRAM support is available on

## Change Checklist

### DIFF
--- a/boards/espressif-esp32-s3-devkitc-1.yaml
+++ b/boards/espressif-esp32-s3-devkitc-1.yaml
@@ -5,6 +5,18 @@ targets:
     flags:
       - has_usb_cdc_acm_port
       - has_usb_device_port
+    ariel:
+      global_env:
+        CARGO_ENV:
+          # All current official ordering codes for this board (N8R8,
+          # N8R8U, N32R16V) ship an ESP32-S3-WROOM-1/-2 module with Octal
+          # SPI PSRAM. Other WROOM-1 variants (e.g. the older, no longer
+          # listed N8R2) have Quad SPI PSRAM instead, which needs
+          # `ESP_HAL_CONFIG_PSRAM_MODE=quad`: the two are electrically
+          # different and esp-hal cannot detect which one is present, so
+          # every board with PSRAM must set this explicitly to match its
+          # actual module variant.
+          - ESP_HAL_CONFIG_PSRAM_MODE=octal
     buttons:
       # button0
       - pin: GPIO0
--- a/src/ariel-os-esp/src/lib.rs
+++ b/src/ariel-os-esp/src/lib.rs
@@ -118,6 +118,13 @@ pub fn init() -> OptionalPeripherals {
         crate::time_driver::init(embassy_timer);
     }
 
+    #[cfg(feature = "psram")]
+    {
+        ariel_os_log::debug!("initializing psram");
+        esp_alloc::psram_allocator!(peripherals.PSRAM.take().unwrap(), esp_hal::psram);
+        ariel_os_log::debug!("psram: {} bytes free", esp_alloc::HEAP.free());
+    }
+
     peripherals
 }
 
--- a/src/ariel-os-esp/src/lib.rs
+++ b/src/ariel-os-esp/src/lib.rs
@@ -78,6 +78,11 @@ pub trait IntoPeripheral<'a, T> {
     fn into_hal_peripheral(self) -> T;
 }
 
+#[cfg(feature = "psram")]
+mod psram;
+#[cfg(feature = "psram")]
+pub use psram::PSRAM_HEAP;
+
 #[doc(hidden)]
 impl<T> IntoPeripheral<'_, T> for T {
     fn into_hal_peripheral(self) -> T {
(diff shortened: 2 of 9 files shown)


boards/espressif-esp32-s3-devkitc-1.yaml
```diff
@@ -5,6 +5,18 @@ targets:
...
+        CARGO_ENV:
+          # All current official ordering codes for this board (N8R8,
+          # N8R8U, N32R16V) ship an ESP32-S3-WROOM-1/-2 module with Octal
+          # SPI PSRAM. Other WROOM-1 variants (e.g. the older, no longer
+          # listed N8R2) have Quad SPI PSRAM instead, which needs
+          # `ESP_HAL_CONFIG_PSRAM_MODE=quad`: the two are electrically
+          # different and esp-hal cannot detect which one is present, so
+          # every board with PSRAM must set this explicitly to match its
+          # actual module variant.
+          - ESP_HAL_CONFIG_PSRAM_MODE=octal
```
bugadani, line 19: This excludes way too many devkits already out in the wild. The 2MB chips are still [available](https://hu.mouser.com/en/ProductDetail/Espressif-Systems/ESP32-S3-WROOM-1-N8R2?qs=Wj%2FVkw3K%252BMCC1Ha9x85iJA%3D%3D) to buy from at least one distributor.

luukoudshoorn: Agreed, but no distinction on specific board by amount of PSRAM is currently made in Ariel OS. We could split these at the board level, but I'm not sure that's necessary since it's only a default that can be overwritten by the user. How would you tackle this?

luukoudshoorn: New idea, untested: For boards where both quad and octal variants are available, don't set a default. Check during build if the env variable was set and if not, log an error explaining to the user that they have to set it themselves and how to find which they need. 

For now I'll leave this here in this state, I'm told @ROMemories will explore PSRAM support more broadly some time in the future.

src/ariel-os-esp/src/lib.rs
```diff
@@ -118,6 +118,13 @@ pub fn init() -> OptionalPeripherals {
         crate::time_driver::init(embassy_timer);
     }
 
+    #[cfg(feature = "psram")]
+    {
+        ariel_os_log::debug!("initializing psram");
+        esp_alloc::psram_allocator!(peripherals.PSRAM.take().unwrap(), esp_hal::psram);
```
ROMemories: The [docs of `psram_allocator!()`](https://docs.espressif.com/projects/rust/esp-alloc/0.10.0/esp_alloc/macro.psram_allocator.html#-limitations) mention that atomic operations cannot be used with memory locations in PSRAM on Xtensa-based ESPs (unfortunately the [docs of the `esp_hal::psram`](https://docs.espressif.com/projects/rust/esp-hal/1.1.0/esp32s3/esp_hal/psram/index.html) don't seem to mention this), so configuring this as a global allocator, and therefore in a way transparent to users, would very likely end up in UB. Because of this we're not ready to allow using PSRAM as a global allocator in Ariel OS.

What we can do however, if that is useful, is reduce the scope of this PR to only propagate the `psram` Cargo feature to `esp-hal`, through a `psram` Cargo feature as you're proposing.

This would then allow you to [instantiate a dedicated PSRAM allocator using `esp-alloc`](https://docs.espressif.com/projects/rust/esp-alloc/0.10.0/esp_alloc/index.html#using-this-with-the-nightly-allocator_api-feature) in your application (while making sure not to store atomics through it), with which you can allocate individual buffers. I don't think you should use `psram_allocator!` on the ESP32-S3 without being very careful on what gets stored on the heap, including in every dependency (which doesn't sound very practical).

As an aside, I'd be interested in knowing what kind of types/buffers you're looking to store in the PSRAM if you can share.

bugadani: Atomic operations in PSRAM work perfectly well on RISC-V chips. On Xtensas, they are language-illegal starting from 1.99, the core::atomic types will likely get (partially) disabled. Work is ongoing to work around this in portable-atomic.

luukoudshoorn: I was actually already changing the PR to address this, see my commit below. Already mentioned to trigger this UB so this is not just a theoretical danger.

If I'm the only one that's helped by this feature at the moment, I'll gladly move it into our not-upstreamed branch.

Our main thing that needs to be in PSRAM is a large rendering buffer.

PR #2207, opened 2026-08-10
docs(glossary): define some memory-related terms

ROMemories:
# Description

This introduces a few terms related to non-volatile memory. This is motivated by https://github.com/ariel-os/ariel-os/pull/2180/changes#r3750823906 and by the introduction of chips using ReRAM in production, especially the nRF54s and the [Baochip](https://github.com/baochip/baochip-1x).

## Future Work

The book could be updated to replace occurrences of "flash" with "non-volatile memory" where relevant (but this is better done later). The persistent storage API currently only supports flash anyway, as it is backed by `sequential-storage`.

## Change Checklist

### DIFF
--- a/book/src/glossary.md
+++ b/book/src/glossary.md
@@ -59,6 +59,9 @@
   <dt id="esp-hal"><a href="https://docs.espressif.com/projects/rust/esp-hal/latest/">esp-hal</a>:</dt>
   <dd>A HAL for ESP32 microcontrollers, developed by their manufacturer Espressif.</dd>
 
+  <dt id="flash-memory"><a href="https://en.wikipedia.org/wiki/Flash_memory">Flash memory</a>:</dt>
+  <dd>A technology of <a href="#non-volatile-memory">non-volatile memory</a>.</dd>
+
   <dt id="hal">HAL (Hardware Abstraction Layer):</dt>
   <dd>A software layer that makes specific pieces of hardware (e.g., microcontroller peripherals) easier to use, by hiding some of their details.</dd>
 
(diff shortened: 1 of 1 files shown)


book/src/glossary.md
```diff
@@ -78,9 +84,23 @@
...
+
   <dt id="risc-v"><a href="https://en.wikipedia.org/wiki/RISC-V">RISC-V</a>:</dt>
   <dd>An open, royalty-free instruction set architecture.</dd>
 
+  <!-- NOTE: For instance the RP2040 contains floating point utility functions. -->
+  <dt id="rom">ROM (Read-Only Memory):</dt>
+  <dd>
+    <a href="#non-volatile-memory">Non-volatile memory</a> that can never be written to by software.
+    Sometimes used by manufacturers to store bootloaders or utility functions.
+    Sometimes also incorrectly refers to program memory for historical reasons.
```
bergzand: To me this *sounds* opinionated or maybe even condescending.
I would just leave it out here, or omit the 'incorrectly' as it doesn't add anything for the reader. There's already a reference to the non-volatile memory, it should at least be clear that what we call non-volatile memory, might be called ROM by chip vendors (e.g. the espressif linker scripts call the external memory regions IROM and DROM).

book/src/glossary.md
```diff
@@ -68,6 +71,9 @@
     Contains a processor, memory, and peripherals to interact with the outside world.
   </dd>
 
+  <dt id="non-volatile-memory">Non-volatile memory (NVM):</dt>
+  <dd>Memory whose contents are retained even without power.</dd>
```
bergzand: This could use a reference to the flash-memory and to the reram items as examples

bergzand (changes requested):

book/src/glossary.md
ROMemories: I've added back links.

book/src/glossary.md
ROMemories: I've dropped "incorrectly", I suppose "for historical reasons" already does the job.

bergzand (approved):

PR #2218, opened 2026-08-11
refactor(log): use `cfg_select!()` to select `println` implementation

nponsard:
# Description

This makes the use of the new `cfg_select!` macro to select which println version to use and also makes the dependency on `ariel-os-debug` optional.

## Testing

Check if logging still works:
```
laze -C examples/hello-world build -b seeedstudio-xiao-esp32c6 -s log run
laze -C examples/hello-world build -b seeedstudio-xiao-esp32c6 -s defmt run
laze -C examples/hello-world build -b seeedstudio-xiao-esp32c6 -s defmt -s probe-rs run
```

## Change Checklist

### DIFF
--- a/src/ariel-os-log/Cargo.toml
+++ b/src/ariel-os-log/Cargo.toml
@@ -9,13 +9,17 @@ license.workspace = true
 [package.metadata.feature-groups]
 logging-facade.xor = { features = ["defmt", "log"] }
 logging-transport.xor = { features = [
-  "defmt-rtt",
+  "logging-over-debug-channel",
   "logging-over-uart",
   "logging-over-usb",
 ] }
 
+[package.metadata.feature-groups.features]
+defmt-rtt.requires = { features = ["logging-over-debug-channel"] }
+
+
 [dependencies]
-ariel-os-debug = { workspace = true }
+ariel-os-debug = { workspace = true, optional = true }
 ariel-os-utils = { workspace = true, optional = true }
 const-str = { workspace = true, optional = true }
 critical-section = { workspace = true, optional = true }
@@ -59,6 +63,8 @@ debug-uart = ["dep:embassy-sync", "logging-over-uart"]
 # `defmt-rtt` uses the debug channel as the logging transport, so it has to go
 # through this crate.
 defmt-rtt = ["ariel-os-debug/defmt-rtt"]
+# Using the debug channel as the transport
+logging-over-debug-channel = ["dep:ariel-os-debug"]
 logging-over-uart = ["esp-println?/uart"]
 logging-over-usb = ["esp-println?/jtag-serial"]
 std = []
--- a/src/ariel-os-log/src/lib.rs
+++ b/src/ariel-os-log/src/lib.rs
@@ -72,33 +72,33 @@ pub mod log {
     // adds/removes any items.
     pub use log::{debug, error, info, trace, warn};
 
-    #[cfg(all(
-        context = "ariel-os",
-        not(any(
-            feature = "esp-println",
-            feature = "logging-over-uart",
-            feature = "std"
-        ))
-    ))]
-    pub use ariel_os_debug::debug_channel_println as println;
-
-    #[cfg(feature = "esp-println")]
-    pub use esp_println::println;
-
-    #[cfg(feature = "std")]
-    pub use std::println;
-
-    #[cfg(feature = "debug-uart")]
-    pub use crate::uart_println as println;
-
     /// Prints to the logging output, with a newline.
     #[cfg(not(context = "ariel-os"))]
     #[macro_export]
     macro_rules! noop_println {
         ($($arg:tt)*) => {};
     }
-    #[cfg(not(context = "ariel-os"))]
-    pub use crate::noop_println as println;
+
+    cfg_select! {
+        feature = "logging-over-debug-channel" => {
+            pub use ariel_os_debug::debug_channel_println as println;
+        }
+        feature = "debug-uart" => {
+           pub use crate::uart_println as println;
+        }
+        feature = "esp-println" => {
+            pub use esp_println::println;
+        }
+        feature = "std" => {
+            pub use std::println;
+        }
+        not(context = "ariel-os") => {
+            pub use crate::noop_println as println;
+        }
+        _ => {
+            compile_error!("No logging transport selected!");
+        }
+    }
 }
 
 // NOTE: this module is used both for `log` and when no logging facades are enabled.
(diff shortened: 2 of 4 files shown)


src/ariel-os-log/src/lib.rs
```diff
@@ -72,33 +72,37 @@ pub mod log {
...
+            compile_error!("No logging transport available !");
+        }
     }
-    #[cfg(not(context = "ariel-os"))]
-    pub use crate::noop_println as println;
+}
+
+/// Prints to the logging output, with a newline.
+#[cfg(not(context = "ariel-os"))]
+#[doc(hidden)]
```
ROMemories: I don't think the macro definition should be moved out of the `log` module, which makes it not feature-gated on the `log` Cargo feature anymore. It indeed seems that we can't have it inside the `cfg_select!`, otherwise we get the following error:

```
error: macro-expanded `macro_export` macros from the current crate cannot be referred to by absolute paths
  --> src/ariel-os-log/src/lib.rs:99:21
   |
99 |             pub use crate::noop_println as println;
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the macro is defined here
  --> src/ariel-os-log/src/lib.rs:95:13
   |
95 | /             macro_rules! noop_println {
96 | |                 ($($arg:tt)*) => {};
97 | |             }
   | |_____________^
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #52234 <https://github.com/rust-lang/rust/issues/52234>
   = note: `#[deny(macro_expanded_macro_exports_accessed_by_absolute_paths)]` (part of `#[deny(future_incompatible)]`) on by default
```

However we can still have it inside the `log` module (I'd put it before the `cfg_select!`), which still satisfies the following command:

```sh
cargo clippy -p ariel-os-log --features log
```

and makes the `doc(hidden)` attribute unnecessary:

```suggestion
```

src/ariel-os-log/src/lib.rs
```diff
@@ -72,33 +72,37 @@ pub mod log {
...
+            pub use std::println;
+        }
+        feature = "custom-transport" => {
+            pub use crate::transport_println as println;
+        }
+        not(context = "ariel-os") => {
+            pub use crate::noop_println as println;
+        }
+        _ => {
+            compile_error!("No logging transport available !");
```
ROMemories:
```suggestion
            compile_error!("No logging transport selected!");
```

src/ariel-os/Cargo.toml
```diff
@@ -199,7 +199,7 @@ timer-generic-queue-128 = ["ariel-os-embassy/timer-generic-queue-128"]
 
 #! ## Development and debugging
 # Enables the debug channel.
-debug-channel = ["ariel-os-rt/debug-channel"]
+debug-channel = ["ariel-os-log/debug-channel", "ariel-os-rt/debug-channel"]
```
ROMemories: I think the Cargo feature should now be renamed to `logging-over-debug-channel` (and the laze module `logging-over-debug-channel` should enable this `logging-over-debug-channel` Cargo feature).

```suggestion
logging-over-debug-channel = ["ariel-os-log/debug-channel", "ariel-os-rt/debug-channel"]
```

The Cargo feature should then be moved with the other `logging-over-*` Cargo features below.

However `ariel-os-rt`'s Cargo feature should *still* be named `debug-channel`: the idea is that `ariel-os-rt` simply sets up the debug channel, while `ariel-os-log` uses it for the logging functionality specifically.

src/ariel-os-log/Cargo.toml
```diff
@@ -58,7 +58,9 @@ esp-println = ["dep:esp-println"]
 debug-uart = ["dep:embassy-sync", "logging-over-uart"]
 # `defmt-rtt` uses the debug channel as the logging transport, so it has to go
 # through this crate.
-defmt-rtt = ["ariel-os-debug/defmt-rtt"]
+defmt-rtt = ["dep:ariel-os-debug", "ariel-os-debug/defmt-rtt"]
+# Using the debug channel as the transport
+debug-channel = ["dep:ariel-os-debug"]
```
ROMemories: Same reasoning as in https://github.com/ariel-os/ariel-os/pull/2218/changes#r3764204526:
```suggestion
logging-over-debug-channel = ["dep:ariel-os-debug"]
```

src/ariel-os-log/Cargo.toml
ROMemories:
```suggestion
  "logging-over-debug-channel",
```

src/ariel-os-log/Cargo.toml
```diff
@@ -15,7 +15,7 @@ logging-transport.xor = { features = [
 ] }
```
ROMemories, line 16: This *transitively* makes `defmt-rtt` mutually exclusive with `logging-over-uart` and `logging-over-usb`, as `logging-over-debug-channel` must then be enabled when `defmt-rtt` is enabled:

```suggestion

[package.metadata.feature-groups.features]
defmt-rtt.requires = { features = ["logging-over-debug-channel"] }

```

src/ariel-os-log/Cargo.toml
```diff
@@ -58,7 +58,9 @@ esp-println = ["dep:esp-println"]
 debug-uart = ["dep:embassy-sync", "logging-over-uart"]
 # `defmt-rtt` uses the debug channel as the logging transport, so it has to go
 # through this crate.
-defmt-rtt = ["ariel-os-debug/defmt-rtt"]
+defmt-rtt = ["dep:ariel-os-debug", "ariel-os-debug/defmt-rtt"]
```
ROMemories: With https://github.com/ariel-os/ariel-os/pull/2218/changes#r3764220351, that shouldn't be necessary:
```suggestion
defmt-rtt = ["ariel-os-debug/defmt-rtt"]
```

src/ariel-os-log/src/lib.rs
```diff
@@ -72,33 +72,36 @@ pub mod log {
...
+        }
+        feature = "esp-println" => {
+            pub use esp_println::println;
+        }
+        feature = "std" => {
+            pub use std::println;
+        }
+        feature = "custom-transport" => {
+            pub use crate::transport_println as println;
+        }
```
ROMemories: Actually this doesn't exist yet 🙂
```suggestion
```

ROMemories (approved):
LGTM, *after removing the snippet from the future*.

ROMemories (approved):

PR #2223, opened 2026-08-14
docs(book): add a page about I2C

ROMemories:
# Description

This adds a page about I2C drivers to the book. The goals compared to just the API docs are to give an overview of what is and *isn't* supported and give some recommendations about when to (or not to) use specific functionality.

## Future Work

- Similar PRs for SPI and UART will be opened very soon.
- The end goal being to have a page about sensors in the book.

## Changelog Entry

A page about I2C drivers has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/i2c.md
+++ b/book/src/i2c.md
@@ -85,7 +85,7 @@ Some HALs allow enabling internal pull-up resistors on the I2C lines: these typi
 
 Some HALs also allow increasing the drive strength of the pins used for the I2C lines: this increases the current draw but helps achieve higher frequencies as well as accommodate a higher number of I2C targets on the bus (i.e., a higher bus capacitance).
 
-[^peripherals-free-cargo-feature]: When the Cargo feature is not enabled (including indirectly as a dependency of another Cargo feature), the peripherals and their interrupts are not managed by Ariel OS and can be used separately, including through their [Embassy-style HAL][embassy-style-hals].
+[^peripherals-free-cargo-feature]: When the Cargo feature is not enabled (including indirectly as a dependency of another Cargo feature), the peripherals and their interrupts are not managed by Ariel OS and can be used separately, including [through their Embassy-style HAL][using-third-party-hals-directly-book].
 
 [embedded-hal-async-i2c-i2c-docsrs]: https://docs.rs/embedded-hal-async/latest/embedded_hal_async/i2c/trait.I2c.html
 [embedded-hal-i2c-i2c-docsrs]: https://docs.rs/embedded-hal/latest/embedded_hal/i2c/trait.I2c.html
@@ -93,3 +93,4 @@ Some HALs also allow increasing the drive strength of the pins used for the I2C 
 [i2c-clock-stretching-wikipedia]: https://en.wikipedia.org/wiki/I2C#Clock_stretching_using_SCL
 [obtaining-peripheral-access-book]: ./application.md#obtaining-peripheral-access
 [embassy-style-hals]: ./glossary.md#embassy-style-hals
+[using-third-party-hals-directly-book]: ./application.md#using-the-third-party-hals-directly
--- a/book/src/i2c.md
+++ b/book/src/i2c.md
@@ -1,0 +1,95 @@
+# I2C
+
+## Capabilities
+
+Ariel OS provides portable I2C drivers, which implement [`embedded_hal_async::i2c:I2c`][embedded-hal-async-i2c-i2c-docsrs], so they can be used by higher-level drivers (e.g., sensor drivers).
+They can be enabled using the `i2c` Cargo feature[^peripherals-free-cargo-feature].
+Only the controller role is currently supported.
+
+<!-- NOTE: Some ESP32 MCUs support hardware timeouts. -->
+Ariel OS currently favors async interfaces, and the I2C drivers do not implement [`embedded_hal::i2c::I2c`][embedded-hal-i2c-i2c-docsrs], although the underlying driver may still block.
+This allows Ariel OS to implement a software timeout on I2C transactions, so that unresponsive I2C targets do not lead to an unresponsive application even on hardware that does not implement hardware timeouts.
+
+Even though the driver is called I2C (instead of TWI), [clock stretching][i2c-clock-stretching-wikipedia] is only supported if the microcontroller peripheral supports it.
+
+## Usage
+
+### Bus Sharing
+
+Like all buses, I2C allows connecting multiple devices (I2C targets) on a single bus.
+The driver for the I2C bus can be instantiated as follows (the configuration is [explained below](#bus-configuration)):
+
+```rust
+let i2c_bus = SensorI2c::new(peripherals.i2c_sda, peripherals.i2c_scl, i2c_config);
+let i2c_bus = Mutex::new(i2c_bus);
+```
+
+The bus driver determines which I2C peripheral is used: they are available in `ariel_os::hal::i2c::controller`; their names match that of the [Embassy-style HAL][embassy-style-hals] (in the example above, `SensorI2c` is a type alias for the driver selected).
+Pin peripherals can be [obtained through the usual mechanisms][obtaining-peripheral-access-book].
+Some microcontrollers allow using (almost) any pair of pins, but most require specific pins for each peripheral: this is automatically enforced at compile time.
+
+Then, a driver instance for each I2C target can be created using the following:
+
+```rust
+let mut i2c_device = ariel_os::i2c::controller::I2cDevice::new(&i2c_bus);
+```
+
+<!-- NOTE: `embassy_embedded_hal::shared_bus::asynch::i2c::I2cDeviceWithConfig` exists and support for it could be added later. -->
+Methods provided by [`embedded_hal_async::i2c::I2c`][embedded-hal-async-i2c-i2c-docsrs] can then be used to communicate with the I2C target.
+Configuration is applied to the whole bus, and it is not currently possible to use different configurations for individual I2C targets.
+
+> [!TIP]
+> If only a single I2C target is used on the bus, using the device driver is unnecessary as the bus driver also implements [`embedded_hal_async::i2c::I2c`][embedded-hal-async-i2c-i2c-docsrs].
+
+### Bus Configuration
+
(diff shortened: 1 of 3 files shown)


book/src/i2c.md
```diff
@@ -0,0 +1,95 @@
+# I2C
```
ROMemories, line 1: I'm fine renaming that to I²C if you prefer, "I2C" seems easier to search for.

book/src/i2c.md
```diff
@@ -0,0 +1,95 @@
...
+## Capabilities
+
+Ariel OS provides portable I2C drivers, which implement [`embedded_hal_async::i2c:I2c`][embedded-hal-async-i2c-i2c-docsrs], so they can be used by higher-level drivers (e.g., sensor drivers).
+They can be enabled using the `i2c` Cargo feature[^peripherals-free-cargo-feature].
+Only the controller role is currently supported.
+
+<!-- NOTE: Some ESP32 MCUs support hardware timeouts. -->
+Ariel OS currently favors async interfaces, and the I2C drivers do not implement [`embedded_hal::i2c::I2c`][embedded-hal-i2c-i2c-docsrs], although the underlying driver may still block.
+This allows Ariel OS to implement a software timeout on I2C transactions, so that unresponsive I2C targets do not lead to an unresponsive application even on hardware that does not implement hardware timeouts.
+
```
ROMemories, line 12: There should likely be a mention of DMA here (even though that will be more relevant to SPI and UART), but I'm not yet sure how to phrase a claim that is not too strong while still being useful, so I think it's better done a follow-up PR (that would take care of SPI and UART as well).

book/src/i2c.md
```diff
@@ -0,0 +1,95 @@
...
+In particular, Fast Mode Plus (i.e., 1 MHz) is generally not currently supported.
+
+#### Configuring Other Settings
+
+Other settings may be available depending on the HAL and the microcontroller.
+
+Some HALs allow enabling internal pull-up resistors on the I2C lines: these typically use the regular GPIO internal pull-ups, which have a higher resistance than usually recommended for I2C and will therefore limit the maximum achievable bus frequency.
+
+Some HALs also allow increasing the drive strength of the pins used for the I2C lines: this increases the current draw but helps achieve higher frequencies as well as accommodate a higher number of I2C targets on the bus (i.e., a higher bus capacitance).
+
```
ROMemories, line 87: There should also be some explanation of the clock configuration required (especially as many STM32s require using other PLL outputs than the one used for SYSCLK) in a dedicated section, similar to what we now have in the USB page, but I think this is better done in a later PR (that would also tackle SPI and UART at the same time).

book/src/i2c.md
bergzand: [Other](https://www.ti.com/product/TMP119) [parties](https://www.microchip.com/en-us/product/mcp9808) use I2C, I think it's fine.

book/src/i2c.md
```diff
@@ -0,0 +1,95 @@
...
+[^peripherals-free-cargo-feature]: When the Cargo feature is not enabled (including indirectly as a dependency of another Cargo feature), the peripherals and their interrupts are not managed by Ariel OS and can be used separately, including through their [Embassy-style HAL][embassy-style-hals].
```
ROMemories: This is allowed by #2234.

book/src/i2c.md
```diff
@@ -0,0 +1,95 @@
...
+
+Ariel OS provides portable I2C drivers, which implement [`embedded_hal_async::i2c:I2c`][embedded-hal-async-i2c-i2c-docsrs], so they can be used by higher-level drivers (e.g., sensor drivers).
+They can be enabled using the `i2c` Cargo feature[^peripherals-free-cargo-feature].
+Only the controller role is currently supported.
+
+<!-- NOTE: Some ESP32 MCUs support hardware timeouts. -->
+Ariel OS currently favors async interfaces, and the I2C drivers do not implement [`embedded_hal::i2c::I2c`][embedded-hal-i2c-i2c-docsrs], although the underlying driver may still block.
+This allows Ariel OS to implement a software timeout on I2C transactions, so that unresponsive I2C targets do not lead to an unresponsive application even on hardware that does not implement hardware timeouts.
+
+Even though the driver is called I2C (instead of TWI), [clock stretching][i2c-clock-stretching-wikipedia] is only supported if the microcontroller peripheral supports it.
```
bergzand: I don't think we support any microcontroller *without* clock stretching, but fine to leave this in.

book/src/i2c.md
```diff
@@ -0,0 +1,95 @@
...
+Like all buses, I2C allows connecting multiple devices (I2C targets) on a single bus.
+The driver for the I2C bus can be instantiated as follows (the configuration is [explained below](#bus-configuration)):
+
+```rust
+let i2c_bus = SensorI2c::new(peripherals.i2c_sda, peripherals.i2c_scl, i2c_config);
+let i2c_bus = Mutex::new(i2c_bus);
+```
+
+The bus driver determines which I2C peripheral is used: they are available in `ariel_os::hal::i2c::controller`; their names match that of the [Embassy-style HAL][embassy-style-hals] (in the example above, `SensorI2c` is a type alias for the driver selected).
+Pin peripherals can be [obtained through the usual mechanisms][obtaining-peripheral-access-book].
```
bergzand: Nitpick: I would make this concrete, and remove the 'usual' here.

```suggestion
Pin peripherals can be [obtained through the Ariel OS peripheral obtainment mechanisms][obtaining-peripheral-access-book].
```

book/src/i2c.md
ROMemories: I've added a backlink to the new section. Will do the same later for SPI and UART when we get to them.

book/src/i2c.md
```diff
@@ -0,0 +1,96 @@
...
+Some microcontrollers allow using (almost) any pair of pins, but most require specific pins for each peripheral: this is automatically enforced at compile time.
```
anlavandier, line 29: Do you have an example of an MCU where SDA/SCL pins are enforced at compile-time ? I've never encountered it

book/src/i2c.md
```diff
@@ -0,0 +1,96 @@
...
+> HAL-specific frequency settings should only be used when required, to keep the application as portable as possible.
+
+The maximum available frequency depends on the microcontroller, and may be further limited by the [Embassy-style HAL][embassy-style-hals] and by Ariel OS.
+In particular, Fast Mode Plus (i.e., 1 MHz) is generally not currently supported.
+
+#### Configuring Other Settings
+
+Other settings may be available depending on the HAL and the microcontroller.
+
+Some HALs allow enabling internal pull-up resistors on the I2C lines: these typically use the regular GPIO internal pull-ups, which have a higher resistance than usually recommended for I2C and will therefore limit the maximum achievable bus frequency.
```
anlavandier, line 84: Maybe add a little line on how this can be achieved (or not) within an Ariel-OS app

anlavandier (approved):
LGTM after addressing the comment below, and fixing CI

book/src/i2c.md
ROMemories: Yes, the RP and STM32 MCUs have that requirement, and it is enforced at compile-time. This can easily be tested with the `i2c-controller` test, by swapping the pair of pins defined in the `pins` module.

book/src/i2c.md
ROMemories: In this specific case I'd rather not: I think it's important to mention that it is *possible*, but considering it's mostly a bad idea (except maybe on nRFs), I would rather not give this too much space/importance. It's still pretty easy to discover as it is in the HAL-specific `Config` struct introduced by the section right above this, and can also be found by searching for "pullup" in the rustdoc docs.

anlavandier: Ok, I see your point.

anlavandier (approved):

book/src/i2c.md
```diff
@@ -0,0 +1,96 @@
...
+
+Some HALs also allow increasing the drive strength of the pins used for the I2C lines: this increases the current draw but helps achieve higher frequencies as well as accommodate a higher number of I2C targets on the bus (i.e., a higher bus capacitance).
```
anlavandier, line 86: `drive strength` meaning ? reducing the pull-up resistance I assume ? Either way unless you want to document how to do that I think that this wording is fine.

book/src/i2c.md
anlavandier: You mean tackling SPI and UART's relationship with (complex) clock trees I assume ?

book/src/i2c.md
```diff
@@ -0,0 +1,96 @@
+# I2C
+
+## Capabilities
+
+Ariel OS provides portable I2C drivers, which implement [`embedded_hal_async::i2c:I2c`][embedded-hal-async-i2c-i2c-docsrs], so they can be used by higher-level drivers (e.g., sensor drivers).
+They can be enabled using the `i2c` Cargo feature[^peripherals-free-cargo-feature].
+Only the controller role is currently supported.
+
+<!-- NOTE: Some ESP32 MCUs support hardware timeouts. -->
+Ariel OS currently favors async interfaces, and the I2C drivers do not implement [`embedded_hal::i2c::I2c`][embedded-hal-i2c-i2c-docsrs], although the underlying driver may still block.
```
anlavandier, line 10: Orthogonal to this piece of documentation but I think that it would be nice if they also implemented the blocking traits, probably as an opt-in feature,  as I've found myself needing to "async-ify" existing drivers because a lot of existing ones are based on the blocking traits.

anlavandier (approved):

book/src/i2c.md
ROMemories: The GPIO PR documents the drive strength better: I'll add a link from the I2C page to it in the GPIO PR <https://github.com/ariel-os/ariel-os/pull/2251#discussion_r4080349124>.

book/src/i2c.md
ROMemories: Yes

book/src/i2c.md
ROMemories: Sure, but that is indeed orthogonal; blocking APIs for I2C also risk the execution to just halt if a target does not acknowledge, if there is no hardware timeouts and the application is not using a preemptive scheduler.

PR #2224, opened 2026-08-14
docs(book): add a page about SPI

ROMemories:
# Description

This adds a page about SPI drivers to the book. The goals compared to just the API docs are to give an overview of what is and *isn't* supported and give some recommendations about when to (or not to) use specific functionality.

## Issues/PRs References

- Depends on #2223

## Changelog Entry

A page about SPI drivers has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/spi.md
+++ b/book/src/spi.md
@@ -1,0 +1,100 @@
+# SPI
+
+## Capabilities
+
+Ariel OS provides portable SPI drivers, which implement [`embedded_hal_async::spi:SpiBus`][embedded-hal-async-spi-spibus-docsrs] and [`embedded_hal_async::spi::SpiDevice`][embedded-hal-async-spi-spidevice-docsrs], so they can be used by higher-level drivers (e.g., sensor drivers).
+They can be enabled using the `spi` Cargo feature[^peripherals-free-cargo-feature].
+Only the main role is currently supported.
+
+Ariel OS currently favors async interfaces, and the SPI drivers do not implement [`embedded_hal::spi::SpiBus`][embedded-hal-spi-spibus-docsrs], although the underlying driver may still block.
+
+## Usage
+
+### Bus Sharing
+
+Like all buses, SPI allows connecting multiple devices (SPI subs) on a single bus.
+The driver for the SPI bus can be instantiated as follows (the configuration is [explained below](#bus-configuration)):
+
+```rust
+let spi_bus = SensorSpi::new(
+    peripherals.spi_sck,
+    peripherals.spi_miso,
+    peripherals.spi_mosi,
+    spi_config,
+);
+let spi_bus = Mutex::new(spi_bus);
+```
+
+The bus driver determines which SPI peripheral is used: they are available in `ariel_os::hal::spi::main`; their names match that of the [Embassy-style HAL][embassy-style-hals] (in the example above, `SensorSpi` is a type alias for the driver selected).
+Pin peripherals can be [obtained through the usual mechanisms][obtaining-peripheral-access-book].
+Some microcontrollers allow using (almost) any pair of pins, but most require specific pins for each peripheral: this is automatically enforced at compile time.
+
+> [!NOTE]
+> Currently, pins for both MISO and MOSI must always be provided.
+
+Then, a driver instance for each SPI sub can be created using the following:
+
+```rust
+let cs_output = ariel_os::gpio::Output::new(peripherals.spi_cs, gpio::Level::High);
+let mut spi_device = ariel_os::spi::main::SpiDevice::new(&spi_bus, cs_output);
+```
+
+The output must be set to high initially, as the chip select (CS) is an active-low signal.
+
+<!-- NOTE: `embassy_embedded_hal::shared_bus::asynch::spi::SpiDeviceWithConfig` exists and support for it could be added later. -->
+Methods provided by [`embedded_hal_async::spi::SpiDevice`][embedded-hal-async-spi-spidevice-docsrs] can then be used to communicate with the SPI sub.
+Configuration is applied to the whole bus, and it is not currently possible to use different configurations for individual SPI subs.
+
+### Bus Configuration
+
+SPI configuration is HAL-specific, and a default configuration can be obtained this way:
+
+```rust
+let mut spi_config = hal::spi::main::Config::default();
+```
+
+#### Selecting the Bus Frequency
+
+Ariel OS provides a [helper function `highest_freq_in()`][spi-main-highest-freq-in-rustdoc] that helps keep the application portable while allowing to use the bus at the highest frequency possible on the hardware:
+
+```rust
+spi_config.frequency = const { highest_freq_in(Kilohertz::kHz(1000)..=Kilohertz::kHz(2000)) };
+```
+
+Alternatively, if a specific frequency is required, enums are also available:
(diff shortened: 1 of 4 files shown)


book/src/spi.md
```diff
@@ -0,0 +1,100 @@
+# SPI
+
+## Capabilities
+
+Ariel OS provides portable SPI drivers, which implement [`embedded_hal_async::spi:SpiBus`][embedded-hal-async-spi-spibus-docsrs] and [`embedded_hal_async::spi::SpiDevice`][embedded-hal-async-spi-spidevice-docsrs], so they can be used by higher-level drivers (e.g., sensor drivers).
+They can be enabled using the `spi` Cargo feature[^peripherals-free-cargo-feature].
+Only the main role is currently supported.
+
+Ariel OS currently favors async drivers, and the SPI drivers do not implement [`embedded_hal::spi::SpiBus`][embedded-hal-spi-spibus-docsrs], although the underlying driver may still block.
+
```
ROMemories, line 10: Similarly to https://github.com/ariel-os/ariel-os/pull/2223#discussion_r3781907984, there should likely be a mention of DMA, which should be done a follow-up.

book/src/spi.md
```diff
@@ -0,0 +1,100 @@
...
+In particular, it is currently limited to that of the SPI peripheral supporting the lowest maximum frequency.
+
+#### Selecting the SPI Mode
+
+The SPI mode can be configured using [`Mode`][spi-mode-rustdoc]:
+
+```rust
+spi_config.mode = ariel_os::spi::Mode::Mode3;
+```
+
```
ROMemories, line 91: Similarly to https://github.com/ariel-os/ariel-os/pull/2223#discussion_r3781918783, there should be a section about the required clock configuration, done in a separate PR.

book/src/spi.md
```diff
@@ -0,0 +1,100 @@
...
+Alternatively, if a specific frequency is required, enums are also available:
+
+```rust
+spi_config.frequency = ariel_os::spi::main::Frequency::_500k;
+```
+
+It it also possible to use a HAL-specific frequency setting:
+
+```rust
+spi_config.frequency = ariel_os::hal::spi::main::Frequency::F(Kilohertz::kHz(250)); // Selects 250 kHz.
```
ROMemories, line 73: Seeing this again now, this could likely use a `From` implementation to make this readable, but better document what we currently have and update this later.

book/src/spi.md
```diff
@@ -0,0 +1,100 @@
...
+> Currently, pins for both MISO and MOSI must always be provided.
+
+Then, a driver instance for each SPI sub can be created using the following:
+
+```rust
+let cs_output = ariel_os::gpio::Output::new(peripherals.spi_cs, gpio::Level::High);
+let mut spi_device = ariel_os::spi::main::SpiDevice::new(&spi_bus, cs_output);
+```
+
+The output must be set to high initially, as the chip select (CS) is an active-low signal.
```
ROMemories, line 42: This is surprisingly not documented in [`embassy_embedded_hal::shared_bus::async::spi::SpiDevice`](https://docs.embassy.dev/embassy-embedded-hal/0.6.0/default/shared_bus/asynch/spi/struct.SpiDevice.html#method.new), nor in our documentation of `SpiDevice`, but I hope to improve that later.

book/templates/SUMMARY.md.tmpl
```diff
@@ -25,6 +25,8 @@
 - [Async Executors](./async-support.md)
 - [Memory Layout](./memory-layout.md)
 - [Clocks](./clocks.md)
+- [I2C](./i2c.md)
+- [SPI](./spi.md)
```
ROMemories, line 29: If after we introduce the UART page and other pages, we find the summary to have too many items, we could simply collapse them into a dedicated section (e.g., "Serial Protocols").

book/src/spi.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Selecting the SPI Mode
+
+The SPI mode can be configured using [`Mode`][spi-mode-rustdoc]:
+
+```rust
+spi_config.mode = ariel_os::spi::Mode::Mode3;
+```
+
+[^peripherals-free-cargo-feature]: When the Cargo feature is not enabled (including indirectly as a dependency of another Cargo feature), the peripherals and their interrupts are not managed by Ariel OS and can be used separately, including through their [Embassy-style HAL].
```
ROMemories, line 92: The link will need to be adjusted.

book/src/spi.md
```diff
@@ -0,0 +1,100 @@
...
+spi_config.frequency = ariel_os::spi::main::Frequency::_500k;
+```
+
+It it also possible to use a HAL-specific frequency setting:
+
+```rust
+spi_config.frequency = ariel_os::hal::spi::main::Frequency::F(Kilohertz::kHz(250)); // Selects 250 kHz.
+// or, if the microcontroller or the HAL does not support arbitrary frequencies:
+spi_config.frequency = ariel_os::hal::spi::main::Frequency::_500k; // Selects 500 kHz.
+```
```
anlavandier, line 77: Very minor point but these frequencies are very slow compared to examples of usage of SPI I've seen in the wild. We might want to up them.

book/src/spi.md
```diff
@@ -0,0 +1,100 @@
...
+
+> [!TIP]
+> HAL-specific frequency settings should only be used when required, to keep the application as portable as possible.
+
+The maximum available frequency depends on the microcontroller, and may be further limited by the [Embassy-style HAL][embassy-style-hals] and by Ariel OS.
+In particular, it is currently limited to that of the SPI peripheral supporting the lowest maximum frequency.
```
anlavandier: Is this an Embassy or an Ariel-OS limitation ?

book/src/spi.md
anlavandier, line 1: It might be a niche case but since it's one I've come across, I'll mention it. 
We might want to mention that we don't support (or at least I don't think that we do) [half-duplex SPI](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface#Three-wire).

anlavandier (approved):
LGTM after addressing comments. I only reviewed the SPI page and not the I2C.

book/src/spi.md
ROMemories: True, I've switched to 2 MHz, which at least is not the same ballpark as I2C.

book/src/spi.md
ROMemories: I've clarified that is indeed an Ariel OS limitation.

book/src/spi.md
ROMemories: I've thought about this when writing this initially but decided not to; I've now added a small sentence in the note that says that pins for both MISO and MOSI are required. With some assumptions it may be possible to implement three-wire, half-duplex SPI with an external tristate buffer, so I've clarified this was only a limitation at microcontroller level.

book/src/spi.md
ROMemories: Done in #2226.

anlavandier (approved):
LGTM, please rebase and squash once #2223 is merged.

book/src/spi.md
ROMemories: I've adjusted the link.

anlavandier (approved):
LGTM, please squash

PR #2225, opened 2026-08-14
docs(book): add a page about UART

ROMemories:
# Description

This adds a page about UART drivers to the book. The goals compared to just the API docs are to give an overview of what is and isn't supported and give some recommendations about when to (or not to) use specific functionality.

## Issues/PRs References

- Depends on #2224

## Changelog Entry

A page about UART drivers has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/uart.md
+++ b/book/src/uart.md
@@ -1,0 +1,92 @@
+# UART
+
+## Capabilities
+
+Ariel OS provides portable [UART][uart-glossary-book] drivers, which implement [`embedded_io_async::Read`][embedded-io-async-read-docsrs] and [`embedded_io_async::Write`][embedded-io-async-write-docsrs].
+They can be enabled using the `uart` Cargo feature[^peripherals-free-cargo-feature].
+
+Ariel OS currently favors async interfaces, and the UART drivers do not implement [`embedded_io::Read`][embedded-io-read-docsrs] or [`embedded_io::Write`][embedded-io-write-docsrs].
+
+TODO: a couple of sentences on the UART drivers being buffered, the benefits, the trade-offs, other implications (if any), and the usual "non-buffered UART drivers are not currently supported/provided" if that makes sense
+
+## Usage
+
+### Instantiating the Driver
+
+The driver for a buffered UART can be instantiated as follows (the configuration is [explained below](#configuration)):
+
+```rust
+let mut rx_buf = [0u8; 32];
+let mut tx_buf = [0u8; 32];
+
+let mut uart = TestUart::new(
+    peripherals.uart_rx,
+    peripherals.uart_tx,
+    &mut rx_buf,
+    &mut tx_buf,
+    uart_config,
+)
+.expect("Invalid UART configuration");
+```
+
+The UART driver determines which UART peripheral is used: they are available in `ariel_os::hal::uart`; their names match that of the [Embassy-style HAL][embassy-style-hals] (in the example above, `TestUart` is a type alias for the driver selected).
+Pin peripherals can be [obtained through the usual mechanisms][obtaining-peripheral-access-book].
+Some microcontrollers allow using (almost) any pair of pins, but most require specific pins for each peripheral: this is automatically enforced at compile time.
+
+> [!NOTE]
+> Currently, pins for both RX and TX must always be provided.
+
+TODO: how to determine the sizing of the buffers? The RX and TX buffers do not need to have the same size.
+
+### Configuration
+
+UART configuration is HAL-specific, and a default configuration can be obtained this way:
+
+```rust
+let mut uart_config = ariel_os::hal::uart::Config::default();
+```
+
+The default configuration is the following:
+
+- 8 data bits
+- 1 stop bit
+- No parity bit
+
+#### Selecting the Baud Rate
+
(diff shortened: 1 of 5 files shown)


book/src/uart.md
```diff
@@ -0,0 +1,92 @@
+# UART
+
+## Capabilities
+
+Ariel OS provides portable [UART][uart-glossary-book] drivers, which implement [`embedded_io_async::Read`][embedded-io-async-read-docsrs] and [`embedded_io_async::Write`][embedded-io-async-write-docsrs].
+They can be enabled using the `uart` Cargo feature[^peripherals-free-cargo-feature].
+
+Ariel OS currently favors async interfaces, and the UART drivers do not implement [`embedded_io::Read`][embedded-io-read-docsrs] or [`embedded_io::Write`][embedded-io-write-docsrs].
+
+TODO: a couple of sentences on the UART drivers being buffered, the benefits, the trade-offs, other implications (if any), and the usual "non-buffered UART drivers are not currently supported/provided" if that makes sense
```
ROMemories, line 10: @bergzand Can you fill in the blank?

book/src/uart.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+The UART driver determines which UART peripheral is used: they are available in `ariel_os::hal::uart`; their names match that of the [Embassy-style HAL][embassy-style-hals] (in the example above, `TestUart` is a type alias for the driver selected).
+Pin peripherals can be [obtained through the usual mechanisms][obtaining-peripheral-access-book].
+Some microcontrollers allow using (almost) any pair of pins, but most require specific pins for each peripheral: this is automatically enforced at compile time.
+
+> [!NOTE]
+> Currently, pins for both RX and TX must always be provided.
+
+TODO: how to determine the sizing of the buffers? The RX and TX buffers do not need to have the same size.
```
ROMemories, line 39: @bergzand It'd be useful if we could give some advice on how to size the buffers.

book/src/uart.md
```diff
@@ -0,0 +1,92 @@
...
+```rust
+uart_config.baudrate = ariel_os::uart::Baudrate::Hal(250_000.into())
+// or, if the microcontroller or the HAL does not support arbitrary baud rates:
+uart_config.baudrate = ariel_os::hal::uart::Baudrate::_1000000,
+```
+
+When providing an arbitrary baud rate, it is checked at runtime when the [driver is instantiated](#instantiating-the-driver).
+
+> [!IMPORTANT]
+> On the RP2040 and the RP235x MCUs, providing a baud rate too high instead uses the highest supported baud rate.
```
ROMemories, line 74: This is actually a serious limitation in the upstream API I've realized when writing this: the upstream constructor is not fallible and will just cap the baud rate (or silently use a baud rate too high if below the minimum value I think).

ROMemories, line 74: In the same vein, I've realized that passing a baud rate of 0 baud will panic on RP due to a division by zero. The STM32 constructors also panic due to a division by zero, but ESP's don't. I think we should also fix this upstream.

book/src/uart.md
```diff
@@ -0,0 +1,92 @@
...
+
+> [!TIP]
+> HAL-specific baud rates types should only be used when required, to keep the application as portable as possible.
+
+#### Configuring Other Settings
+
+The number of data and stop bits and whether a parity bit is used can also be configured, when supported by the microcontroller and the HAL.
```
ROMemories, line 81: We're just mentioning the existence of these settings here: I don't it's worth it giving examples as I expect this to be very rarely needed and the API is pretty straightforward.

book/src/uart.md
bergzand: See [the embassy API docs](https://docs.embassy.dev/embassy-nrf/0.11.0/nrf52840/uarte/index.html). The gist is that buffered uart prevents from losing data between the async reads, if it's not the only future being waited on.

book/src/uart.md
bergzand: afaik there's no recommendation from Embassy that we can pass on. In general buffers must be sufficient in size to bridge the time between the polls from the scheduler as it is receiving bytes on the uart size. A few bytes (4 - 8) is usually plenty, unless the other async wait in the select is blocking the scheduler for longer times.

book/src/uart.md
ROMemories, line 1: There should be a [backlink from the GPIO page](https://github.com/ariel-os/ariel-os/pull/2251/changes#r4080953425).

ROMemories: Done

book/src/uart.md
```diff
@@ -0,0 +1,96 @@
...
+> On the RP2040 and the RP235x MCUs, providing a baud rate too high instead uses the highest supported baud rate.
+
+> [!TIP]
+> HAL-specific baud rate types should only be used when required, to keep the application as portable as possible.
+
+#### Configuring Other Settings
+
+The number of data and stop bits and whether a parity bit is used can also be configured when supported by the microcontroller and the HAL.
+
+[^peripherals-free-cargo-feature]: When the Cargo feature is not enabled (including indirectly as a dependency of another Cargo feature), the peripherals and their interrupts are not managed by Ariel OS and can be used separately, including [through their Embassy-style HAL][using-third-party-hals-directly-book].
```
nponsard, line 85: I don't think this should be a footnote, but since it's already used in the merged I2C and SPI, it can stay like this.

nponsard (approved):
Should be good after squashing

PR #2226, opened 2026-08-14
docs(spi): clarify that `SpiDevice`'s CS output should be high initially

ROMemories:
# Description

As mentioned in https://github.com/ariel-os/ariel-os/pull/2224#discussion_r3781956694, this clarifies that CS should be set to high initially.

My reading of [`embassy_embedded_hal::shared_bus::async::spi::SpiDevice`'s implementation](https://crates.io/crates/embassy-embedded-hal/0.6.0/code/src/shared_bus/asynch/spi.rs) is that nothing sets CS to high explicitly (only to low when starting a transaction, and to high to end one or on Drop), so I think it's important it's set to the proper level, as that would otherwise likely not work properly when using multiple SPI subs and could result in a higher power draw overall even if a single sub is present. (I have not actually tested that on hardware however.)

## Changelog Entry

The documentation of `SpiDevice` has been clarified to mention that the CS output should be set to high initially.

## Change Checklist


anlavandier (approved):
LGTM

PR #2228, opened 2026-08-14
fix(hal): qualify the peripheral path in define_peripherals!() aliases

### LINKED ISSUE #2033: feat(peripherals): help users with `#[diagnostic::on_unknown]`
### Description

Doing just a `ariel_os::hal::define_peripherals!` without the right accompanying imports gets me:

```
error[E0433]: cannot find module or crate `peripherals` in this scope
 --> src/configured_cal.rs:4:45
  |
4 |       feature = "embedded-cal-stm32wba55" => {ariel_os::hal::define_peripherals!(Peripherals {
  |  _____________________________________________^
5 | |         hash: HASH,
6 | |         rcc: RCC,
7 | |     });}
  | |______^ use of unresolved module or unlinked crate `peripherals`
  |
  = help: if you wanted to use a crate named `peripherals`, use `cargo add peripherals` to add it to your `Cargo.toml`
  = note: this error originates in the macro `ariel_os::hal::define_peripherals` (in Nightly builds, run with -Z macro-backtrace for more info)
```

With `#[diagnostic::on_unknown]`, I think we could give a good error message.

### Motivation

Usability, and because `#[diagnostic::on_unknown]` is there

### Current Implementation

The default error doesn't point in the right direction.

Yes (but not now, so if anyone wants to do it faster, or follow the "good first issue" label…)

jvarias02:
# Description

Fixes the aliasing syntax from the `define_peripherals!()` macro by qualifying the path using a helper `__peripheral_alias_ty` macro.

## Testing

Checked the `examples/hello-world` example for multiple platforms (nrf52840dk, rpi-pico, st-nucleo-f042k6, espressif-esp32-c6-devkitc-1).

The feature itself is not being currently used anywhere in the repo. I've tested it building a simple example and using the alias feature without importing `hal::peripheral`.

```Rust
ariel_os::hal::define_peripherals!(LedPeripherals { led: P0_13 = Led });
```

## Issues/PRs References

Closes #2033 

## Changelog Entry

The aliasing syntax of the `define_peripherals!()` macro has been fixed. It originally comes from the [assign_resources! macro](https://crates.io/crates/assign-resources/0.5.0) , but it has not been used nor documented.

## Change Checklist

### DIFF
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -14,7 +14,7 @@ macro_rules! define_peripherals {
         $peripherals:ident {
             $(
                 $(#[$inner:meta])*
-                $peripheral_name:ident : $peripheral_field:ident $(=$peripheral_alias:ident)?
+                $peripheral_name:ident : $peripheral_field:ident
             ),*
             $(,)?
         }
@@ -27,11 +27,6 @@ macro_rules! define_peripherals {
                 pub $peripheral_name: $crate::__peripheral_ty!($peripheral_field),
             )*
         }
-
-        $($(
-            #[allow(missing_docs, non_camel_case_types)]
-            pub type $peripheral_alias = peripherals::$peripheral_field;
-        )?)*
 
         impl $crate::hal::TakePeripherals<$peripherals> for &mut $crate::hal::OptionalPeripherals {
             fn take_peripherals(&mut self) -> $peripherals {
--- a/src/ariel-os-hal/src/hal/define_peripherals.rs
+++ b/src/ariel-os-hal/src/hal/define_peripherals.rs
@@ -14,7 +14,7 @@ macro_rules! define_peripherals {
         $peripherals:ident {
             $(
                 $(#[$inner:meta])*
-                $peripheral_name:ident : $peripheral_field:ident
+                $peripheral_name:ident : $peripheral_field:ident $(=$peripheral_alias:ident)?
             ),*
             $(,)?
         }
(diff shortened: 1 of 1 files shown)


kaspar030:
hm, I had actually used this, mostly to unify between platforms.

kaspar030:
(close was accidental)

jvarias02:
> (close was accidental)

Oh, I opened the PR based on the @ROMemories suggestion of removing it. The alternative is as easy as qualifying the path in the alias to remove the bug.

So just let me know which solution you prefer and I'll push it.

kaspar030:
> Oh, I opened the PR based on the @ROMemories suggestion of removing it. The alternative is as easy as qualifying the path in the alias to remove the bug.
> 
> So just let me know which solution you prefer and I'll push it.

Please just fix this by qualifying the path!

jvarias02:
> > Oh, I opened the PR based on the @ROMemories suggestion of removing it. The alternative is as easy as qualifying the path in the alias to remove the bug.
> > So just let me know which solution you prefer and I'll push it.
> 
> Please just fix this by qualifying the path!

Should I add docs for the alias? And nothing in the repo currently uses the alias syntax, should I add it to an example?

kaspar030:
> Should I add docs for the alias? And nothing in the repo currently uses the alias syntax, should I add it to an example?

Both would be great! Not sure how to build an example though, as this is by definition pretty hardware dependent. Maybe build an example for just one board? ... It would be fine to split either or both into a follow-up PR though.

jvarias02:
> > Should I add docs for the alias? And nothing in the repo currently uses the alias syntax, should I add it to an example?
> 
> Both would be great! Not sure how to build an example though, as this is by definition pretty hardware dependent. Maybe build an example for just one board? ... It would be fine to split either or both into a follow-up PR though.

Okay, I'll do the docs after merging this one. And related to the example... I also thought of adding it to a test. For example, on `tests/gpio/src/pins.rs`, having all `pin_0` aliased and having it used on the test with something like this:

```Rust
#[cfg(not(context = "esp"))]
pub type Pin0Peri = ariel_os::hal::peripheral::Peri<'static, Pin0>;
#[cfg(context = "esp")]
pub type Pin0Peri = Pin0;

...

fn pull_up_input(pin: pins::Pin0Peri) -> Input<'static> {
    Input::new(pin, Pull::Up)
}
```

PR #2234, opened 2026-08-20
docs(book): document how underlying HALs can be used directly

ROMemories:
# Description

This addresses the long-standing open question of Ariel OS and its underlying HALs: this makes it explicit in the book that underlying, third-party HALs *may* be used in applications, and how and under which conditions this can be done.

This has two indirect structuring consequences for Ariel OS:

- This makes Ariel OS's underlying HALs *not* an implementation detail, which they already weren't de facto, even though that had been the hope in the past.
- This makes the (weak) promise that Ariel OS functionality feature-gates (Cargo features and laze modules) will continue to also feature-gate peripheral taking and interrupt binding, so that it keeps being possible to use the underlying HALs "alongside" the OS. Pieces of functionality should also document what leaving them *disabled* allows to do with third-party drivers, when relevant (e.g., see https://github.com/ariel-os/ariel-os/pull/2223#discussion_r3821315240).

## Changelog Entry

The relationship between Ariel OS and its underlying HALs has been clarified in the book, and instructions on how and when to use them directly have been added.

## Change Checklist

### DIFF
--- a/book/src/application.md
+++ b/book/src/application.md
@@ -71,6 +71,20 @@ async fn blinky(peripherals: pins::LedPeripherals) {
 }
 ```
 
+## Using the Third-Party HALs Directly
+
+<!-- NOTE: This implies that underlying HALs are not implementation details of Ariel OS. -->
+Ariel OS internally uses third-party HALs in [its HAL crates][ariel-os-hals], which currently all are [Embassy-style HALs][embassy-style-hals].
+[Ariel OS HAL crates][ariel-os-hals] provide portable drivers and APIs to abstract over the most common types of APIs and peripherals, so that applications can generally run just the same on boards featuring microcontrollers from different manufacturers.
+When Ariel OS does not (yet) provide a portable driver over a peripheral that *is* supported by the third-party HAL, it is still possible to directly use that driver.
+This can be done by simply adding the third-party HAL as a dependency of the application and passing the required peripheral ZSTs obtained through [Ariel OS mechanisms](#obtaining-peripheral-access) to instantiate the third-party driver in the application.
+
+> [!IMPORTANT]
+> When adding the third-party HAL as a dependency, the same version as that of Ariel OS *must* be used.
+
+This is made possible by the fact that Ariel OS takes peripherals and binds interrupts needed for a piece of functionality only when that functionality is enabled (through Cargo features or [laze modules][laze-modules-book]), unless it is a core functionality of the OS.
+For instance, leaving the `i2c` Cargo feature disabled allows drivers other than Ariel OS's to bind the interrupts required for the I2C peripherals, allowing to use I2C drivers from third-party HALs.
+
 [embassy-style-hals]: ./glossary.md#embassy-style-hals
 [ariel-os-hals]: ./glossary.md#ariel-os-hals
 [spawner-attr-docs]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/attr.spawner.html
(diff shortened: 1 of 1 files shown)


book/src/application.md
```diff
@@ -71,6 +71,18 @@ async fn blinky(peripherals: pins::LedPeripherals) {
...
+When Ariel OS does not (yet) provide a portable driver over a peripheral that *is* supported by the third-party HAL, it is still possible to directly use that driver, by simply adding the third-party HAL as a dependency of the application and passing the required peripheral ZSTs obtained through [Ariel OS mechanisms](#obtaining-peripheral-access) to instantiate the third-party driver in the application.
+
+> [!IMPORTANT]
+> When adding the third-party HAL as a dependency, the same version as that of Ariel OS *must* be used.
+
+This is made possible by the fact that Ariel OS takes peripherals and binds interrupts needed for a piece of functionality only when that functionality is enabled (through Cargo features or [laze modules][laze-modules-book]), unless it is a core functionality of the OS: for instance, leaving the `i2c` Cargo feature disabled allows drivers other than Ariel OS's to bind the interrupts required for the I2C peripherals, allowing to use I2C drivers from third-party HALs.
```
bergzand: Could you split this sentence up a bit to increase readability?

book/src/application.md
```diff
@@ -71,6 +71,18 @@ async fn blinky(peripherals: pins::LedPeripherals) {
 }
 ```
 
+## Using the Third-Party HALs Directly
+
+<!-- NOTE: This implies that underlying HALs are not implementation details of Ariel OS. -->
+Ariel OS internally uses third-party HALs in [its HAL crates][ariel-os-hals], which currently all are [Embassy-style HALs][embassy-style-hals].
+[Ariel OS HAL crates][ariel-os-hals] provide portable drivers and APIs to abstract over the most common types of APIs and peripherals, so that applications can generally run just the same on boards featuring microcontrollers from different manufacturers.
+When Ariel OS does not (yet) provide a portable driver over a peripheral that *is* supported by the third-party HAL, it is still possible to directly use that driver, by simply adding the third-party HAL as a dependency of the application and passing the required peripheral ZSTs obtained through [Ariel OS mechanisms](#obtaining-peripheral-access) to instantiate the third-party driver in the application.
```
bergzand: This sentence is rather long, but not impossibly long. Do you see an easy way to split it to improve readability? If not I'm fine with leaving it as is.

book/src/application.md
```diff
@@ -71,6 +71,18 @@ async fn blinky(peripherals: pins::LedPeripherals) {
...
+
+> [!IMPORTANT]
+> When adding the third-party HAL as a dependency, the same version as that of Ariel OS *must* be used.
```
bergzand, line 83: There's no easy way to guarantee this right?

bergzand (changes requested):

book/src/application.md
ROMemories: I've split the sentence by replacing the colon with a period.

book/src/application.md
ROMemories: I've split the sentence before "By simply", to first explain that it is possible, before giving instructions.

book/src/application.md
ROMemories: Unfortunately, Cargo doesn't currently allow referencing an other dependency's/package's version. We could mitigate that by having a quick note in our changelog each time we bump one of the underlying HAL as a reminder for users to also bump the HALs as direct dependencies (if they have used them), but at least forgetting to do that should never result in any unpredictable runtime behavior: it could just either not compile or, at worst, panic early in the startup phase when attempting to take peripherals.

bergzand (approved):
Ack, LGTM!

PR #2236, opened 2026-08-21
feat(esp)!: Add DeviceId implementation based on efuse base mac address

SimonIT:
# Description

I'm trying to add support for the ESP32-S3 ETH from Waveshare, which uses a WIZnet chip. However, to connect to the WIZnet, I need to provide a MAC address. Without this change, `ariel_os_identity::interface_eui48(0)` just returns zeros.

## Testing

Flashed it to my board and got `client_identifier: Some(42-04-01-7a-9f-b7)`

## Open Questions

- [read_base_mac_address](https://docs.rs/esp-hal/1.0.0/esp_hal/efuse/struct.Efuse.html#method.read_base_mac_address) or [mac_address](https://docs.rs/esp-hal/1.0.0/esp_hal/efuse/struct.Efuse.html#method.mac_address), I chose `read_base_mac_address` because it's changeable with `set_mac_address`, but I don't know if that's desired
- It seems that the latest version of ESP-HAL has changed the API.

## Changelog Entry

The device ID for ESP boards is now backed by the eFuse MAC address

## Change Checklist

### DIFF
--- a/src/ariel-os-esp/src/identity.rs
+++ b/src/ariel-os-esp/src/identity.rs
@@ -1,0 +1,20 @@
+use ariel_os_embassy_common::identity;
+
+/// The device's factory-programmed base MAC address, read from eFuse.
+pub struct DeviceId([u8; 6]);
+
+impl identity::DeviceId for DeviceId {
+    type Bytes = [u8; 6];
+
+    #[expect(
+        refining_impl_trait_reachable,
+        reason = "making this fallible would be a breaking API change for Ariel OS"
+    )]
+    fn get() -> Result<Self, core::convert::Infallible> {
+        Ok(Self(esp_hal::efuse::Efuse::read_base_mac_address()))
+    }
+
+    fn bytes(&self) -> Self::Bytes {
+        self.0
+    }
+}
--- a/laze-project.yml
+++ b/laze-project.yml
@@ -363,6 +363,7 @@ contexts:
     selects:
       - ?espflash
     provides:
+      - has_device_identity
       - has_hwrng
     env:
       esp_wifi_heapsize_required: $(72*1024)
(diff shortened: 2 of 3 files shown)


kaspar030:
Nice PR, thanks! ... does every ESP chip have this base mac address?

SimonIT:
Having read these documents: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/misc_system_api.html#mac-address and https://docs.espressif.com/projects/esp-idf/en/v6.0.2/esp32/api-reference/system/efuse.html, I would say so

src/ariel-os-esp/src/identity.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl identity::DeviceId for DeviceId {
+    type Bytes = [u8; 6];
+
+    #[expect(
+        refining_impl_trait_reachable,
+        reason = "making this fallible would be a breaking API change for Ariel OS"
+    )]
+    fn get() -> Result<Self, core::convert::Infallible> {
+        Ok(Self(esp_hal::efuse::Efuse::read_base_mac_address()))
```
ROMemories, line 14: It is indeed important to use `read_base_mac_address()` (from `esp-hal` *v1.0.0*) to read the MAC address set in eFuse bits, instead of one that can be changed in software.

ROMemories (approved):
Thanks a lot!

Successfully tested on an Espressif ESP32-C6-DevKitC-1 and a Unihiker K10.

That MAC address is indeed available in all currently supported ESPs (otherwise CI would be failing, unless `esp-hal` returned some dummy bytes instead of the real MAC address, which I don't believe it does).

I changed the PR title and changelog label to consider this a breaking change: this does return a different type and, even though it's not obvious this could really cause breakage, it's just safer to do so (and gives this some more visibility).

PR #2237, opened 2026-08-22
feat(net): Add generic support for WIZnet SPI Ethernet chips

SimonIT:
# Description

I'm trying to add support for the ESP32-S3 ETH from Waveshare, which uses a WIZnet chip. This wires up the embassy-net-wiznet driver. Later, I will make the PR for the board itself.

## Testing

I already tested it on my board and it works

## Issues/PRs References

This fix is needed for it to work: https://github.com/embassy-rs/embassy/commit/4aeb9c9966a9559a212e6690ccb625e2aada8e4a

For a unique MAC address on ESP ariel-os/ariel-os#2236 is needed

Would be cool to integrate with ariel-os/sbd#134, but it's almost half a year open

## Open Questions

Without an additional 'select' in the 'run' command (`laze build -b waveshare-esp32-s3-eth -s ethernet-wiznet-w5500 run`), it seems that Laze still uses/enables Wifi-Esp. I don't know if this should be included in the scope of this PR.

## Changelog Entry

WIZnet Ethernet chips connected via SPI are now supported

We'll skip the changelog for now, until we also land the docs (which we should only do after bumping `embassy-embedded-hal`).

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy/src/ethernet/wiznet.rs
+++ b/src/ariel-os-embassy/src/ethernet/wiznet.rs
@@ -1,0 +1,124 @@
+//! Provides support for WIZnet SPI Ethernet chips (W5500, W5100S, W6100, W6300).
+//!
+//! This backend is HAL-agnostic: it only relies on generic SPI/GPIO abstractions, so it works
+//! identically on any MCU family. Only the board-specific pin mapping needs to be added per board.
+
+cfg_select! {
+    // Add a `#[path = "wiznet/<board>.rs"] mod board;` arm here, gated on
+    // `context = "<board>"`, for each additional board that wires up a WIZnet chip.
+    context = "ariel-os" => {
+        compile_error!(
+            "no pin mapping for the WIZnet Ethernet driver is known for this board; add one to \
+             ariel-os-embassy/src/ethernet/wiznet.rs"
+        );
+    }
+    _ => {}
+}
+
+use embassy_executor::Spawner;
+use embassy_net_wiznet::{Runner, State};
+use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, mutex::Mutex};
+use static_cell::StaticCell;
+
+use ariel_os_hal::gpio;
+
+use crate::{
+    hal,
+    spi::main::{Kilohertz, SpiDevice, highest_freq_in},
+};
+
+/// Maximum SPI clock frequency to communicate with the WIZnet chip.
+///
+/// WIZnet chips are specified to support faster SPI clocks, but using the MCU's maximum
+/// available frequency (as [`hal::spi::main::Config::default()`] does) has been observed to
+/// corrupt reads (e.g. the chip version check failing) on at least the ESP32-S3. 20 MHz matches
+/// the clock speed used by Espressif's own W5500 examples and is a safe, well-tested choice.
+const MAX_WIZNET_SPI_FREQUENCY: Kilohertz = Kilohertz::MHz(20);
+
+cfg_select! {
+    feature = "wiznet-w5500" => {
+        pub use embassy_net_wiznet::chip::W5500 as WiznetChip;
+    }
+    feature = "wiznet-w5100s" => {
+        pub use embassy_net_wiznet::chip::W5100S as WiznetChip;
+    }
+    feature = "wiznet-w6100" => {
+        pub use embassy_net_wiznet::chip::W6100 as WiznetChip;
+    }
+    feature = "wiznet-w6300" => {
+        pub use embassy_net_wiznet::chip::W6300 as WiznetChip;
+    }
+    context = "ariel-os" => {
+        compile_error!(
+            "no WIZnet chip selected for the `ethernet-wiznet` backend; select exactly one of \
+             `wiznet-w5500`, `wiznet-w5100s`, `wiznet-w6100`, `wiznet-w6300`"
+        );
+    }
+    _ => {}
+}
+
+/// Number of packet buffers held by the driver in each direction.
+const N_RX: usize = 8;
+const N_TX: usize = 8;
+
+/// The [`embassy_net`](https://docs.rs/embassy-net) driver for this Ethernet chip.
+pub type NetworkDevice = embassy_net_wiznet::Device<'static>;
+
+#[embassy_executor::task]
+async fn ethernet_wiznet_task(
+    runner: Runner<
+        'static,
+        WiznetChip,
+        SpiDevice<'static>,
+        gpio::IntEnabledInput<'static>,
+        gpio::Output<'static>,
+    >,
+) -> ! {
+    runner.run().await
+}
+
+/// Initializes the WIZnet Ethernet chip and returns the [`NetworkDevice`] to be used by
+/// `embassy-net`.
+///
+/// # Panics
+///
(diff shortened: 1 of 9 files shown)


ROMemories:
Thanks a lot for this PR, this definitely sounds useful and something that makes sense to have in the OS (as it's part of the network stack setup)!

As you said, my understanding is that we are now waiting for a new release of `embassy-embedded-hal` (as your PR has now been merged upstream)?

Unfortunately, none of us currently has a recent *enough* WIZnet chip to test this in hardware, but this won't prevent me from reviewing the PR (I suppose it would make sense to wait for a bit for a release of `embassy-embedded-hal`). I've drafted the documentation for this in #2242 and #2243. In the meantime, feel free to have a look and comment if you think something should be mentioned in the docs.

> Without an additional 'select' in the 'run' command (`laze build -b waveshare-esp32-s3-eth -s ethernet-wiznet-w5500 run`), it seems that Laze still uses/enables Wifi-Esp.

Because that board supports both Wi-Fi and Ethernet, that is expected: currently the laze configuration arbitrarily favors Wi-Fi, so Ethernet has to be explicitly selected when needed, as you have figured out. If your application specifically requires Ethernet, you can select the laze module in your application's `laze.yml`.

SimonIT:
> As you said, my understanding is that we are now waiting for a new release of embassy-embedded-hal (as your PR has now been merged upstream)?

Yes, or you consider this fix important enough for a new entry in `[patch.crates-io]`

> currently the laze configuration arbitrarily favors Wi-Fi

I don't really mind it being this way. I just wanted to hear opinions so that I would be aware of them

> this won't prevent me from reviewing the PR

Great!

> feel free to have a look and comment

I will (but I suppose I have not enough deep knowledge)

src/ariel-os-embassy/src/ethernet/wiznet.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// Panics if the chip does not respond as expected over SPI (e.g., because of a wiring issue),
+/// or if launching the driver task fails.
+pub(crate) async fn device(
+    peripherals: &mut hal::OptionalPeripherals,
+    spawner: Spawner,
+) -> NetworkDevice {
+    let pins = board::take_pins(peripherals);
+
+    let mut spi_config = hal::spi::main::Config::default();
```
ROMemories, line 96: The SPI mode must be set explicitly; all of the four WIZnet chips support SPI mode 0 according to their datasheets:
```suggestion
    let mut spi_config = hal::spi::main::Config::default();
    spi_config.mode = ariel_os::spi::Mode::Mode0;
```

src/ariel-os-embassy/src/ethernet/wiznet.rs
```diff
@@ -0,0 +1,124 @@
...
+    spi_config.frequency = const { highest_freq_in(Kilohertz::kHz(1)..=MAX_WIZNET_SPI_FREQUENCY) };
```
ROMemories: I think the minimum frequency can be increased; we would want this to compile-time fail instead of doing SPI at 1 kHz I think (you might want to introduce a constant for this):
```suggestion
    spi_config.frequency = const { highest_freq_in(Kilohertz::MHz(10)..=MAX_WIZNET_SPI_FREQUENCY) };
```

src/ariel-os-embassy/src/ethernet/wiznet.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::{
+    hal,
+    spi::main::{Kilohertz, SpiDevice, highest_freq_in},
+};
+
+/// Maximum SPI clock frequency to communicate with the WIZnet chip.
+///
+/// WIZnet chips are specified to support faster SPI clocks, but using the MCU's maximum
+/// available frequency (as [`hal::spi::main::Config::default()`] does) has been observed to
+/// corrupt reads (e.g. the chip version check failing) on at least the ESP32-S3. 20 MHz matches
```
ROMemories: That sounds expected: all the currently supported ESPs have a max SPI frequency of 80 MHz, while the slowest WIZnet chip can only do 70/33.3 MHz according to their datasheets. 20 MHz seems fine for now (and indeed [seems to be used by Espressif](https://github.com/espressif/esp-eth-drivers/blob/400336ad705a2cd88c3d40a1d21bac1085af5865/w5500/test_apps/sdkconfig.defaults#L13)). I'll add that to the documentation in #2243.

I'd be interested in whether this can run at 30 MHz.

src/ariel-os-embassy/src/ethernet/wiznet.rs
```diff
@@ -0,0 +1,124 @@
...
+            "no WIZnet chip selected for the `ethernet-wiznet` backend; select exactly one of \
+             `wiznet-w5500`, `wiznet-w5100s`, `wiznet-w6100`, `wiznet-w6300`"
+        );
+    }
+    _ => {}
+}
+
+/// Number of packet buffers held by the driver in each direction.
+const N_RX: usize = 8;
+const N_TX: usize = 8;
```
ROMemories, line 65: That sounds like a lot, but matches [Embassy's example](https://github.com/embassy-rs/embassy/blob/73b44d7e867a371d0870a84624fea66a8b00f713/examples/rp/src/bin/ethernet_w5500_icmp.rs#L59). I'll add that to the documentation in #2243.

ROMemories (changes requested):
As mentioned previously, I don't have the hardware handy to test this, but I expect this to be good to go after this one. I do remember that we need the fix to `embassy-embedded-hal` for this to work properly.

src/ariel-os-embassy/src/ethernet/wiznet.rs
SimonIT: It runs fine on my board at 30 MHz. Should I increase this @ROMemories?

ROMemories: Based on reading the datasheets only, it seems safe to raise it, I think we can. (Do remember to update the comment when changing the value.)

ROMemories (commented):
I think we'll be able to land this after increasing the max SPI frequency: it's all feature-gated and doesn't add anything to the docs yet (that'll be part of #2243), so that seems safe to land even without the fix to `embassy-embedded-hal`.

src/ariel-os-embassy/src/ethernet/wiznet.rs
```diff
@@ -0,0 +1,129 @@
...
+    let reset = gpio::Output::new(pins.reset, gpio::Level::High);
+    let int = gpio::Input::builder(pins.int, gpio::Pull::Up)
+        .build_with_interrupt()
+        .unwrap();
+
+    static STATE: StaticCell<State<N_RX, N_TX>> = StaticCell::new();
+    let state = STATE.init(State::new());
+
+    let mac_addr = ariel_os_identity::interface_eui48(0)
+        .expect("Should provide a valid MAC address")
```
ROMemories, line 118: This could actually panic if used on a board that doesn't implement `ariel-os-identity` (like RP2040-based boards currently). I'll fix that in a follow-up that also factors out that MAC address logic with the STM32 Ethernet MAC support (likely by introducing a shared `ethernet` laze module and making it require `hw/device-identity`).

Edit: I've opened #2296 for this.

ROMemories (approved):
Thanks a lot for contributing that!

Issue #1780, opened 2026-02-06
Adding support for other RP235xx chips


ROMemories:
Now that these new chips are actually available for ordering, and that we are now able to document support for chips without an associated board in our tree, we should add support for the following chips:

- RP2350B (no in-package flash, more GPIOs)
- RP2354A (2 MiB of in-package flash)
- RP2354B (2 MiB of in-package flash, more GPIOs)

When doing this we should in particular revisit [how the size of the flash is specified for the persistent storage API](https://github.com/ariel-os/ariel-os/blob/841322f9078f287291deb6e7c856c0f646906228/src/ariel-os-rp/src/storage.rs#L6), and make it more specific for the RP2354x variants.

This may also require some changes to [our linker scripts for this HAL](https://github.com/ariel-os/ariel-os/tree/841322f9078f287291deb6e7c856c0f646906228/src/ariel-os-rp).

When this is done, it may be better to adjust the laze context the RPi Pico 2 builders use, and actually make it the `rp2350a` instead of `rp235xa`, and adjust the documentation accordingly (this *might* be a breaking change).

PR #2239, opened 2026-08-25
feat(identity)!: add support for reading the chip ID on RP235x

ROMemories:
# Description

This adds support for `identity::DeviceId` on RP235xA (because we don't [support the RP235xB](#1780) yet) by reading the unique chip ID stored in OTP.
My understanding is that reading the chip ID may fail when it is soft locked (the datasheet mentions that it cannot be hard locked), so this is the first example of a fallible implementation of `DeviceId`.

## Testing

```sh
laze -C examples/device-metadata/ build -b rpi-pico2-w
```

which does print a stable ID.

The following still returns "Device ID is unavailable." as expected:

```sh
laze -C examples/device-metadata/ build -b rpi-pico-w run
```

## Changelog Entry

(RP235xA) Support for obtaining a unique device identifier has been added, by retrieving the chip ID stored in OTP.

## Change Checklist

### DIFF
--- a/src/ariel-os-rp/src/identity.rs
+++ b/src/ariel-os-rp/src/identity.rs
@@ -1,0 +1,15 @@
+cfg_select! {
+    context = "rp2040" => {
+        mod identity_rp2040;
+
+        pub use identity_rp2040::*;
+    }
+    context = "rp235xa" => {
+        mod identity_rp235x;
+
+        pub use identity_rp235x::*;
+    }
+    _ => {
+        compile_error!("this RP chip is not supported");
+    }
+}
--- a/src/ariel-os-rp/src/identity/identity_rp235x.rs
+++ b/src/ariel-os-rp/src/identity/identity_rp235x.rs
@@ -1,0 +1,51 @@
+use embassy_rp::otp;
+
+pub struct DeviceId(u64);
+
+impl ariel_os_embassy_common::identity::DeviceId for DeviceId {
+    type Bytes = [u8; 8];
+
+    fn get() -> Result<Self, impl core::error::Error> {
+        // The chip ID is considered to be globally unique.
+        let res = embassy_rp::otp::get_chipid();
+
+        match res {
+            Ok(id) => Ok(Self(id)),
+            Err(err) => match err {
+                // OTP page 0, which contains the chip ID, is hard locked (i.e., locked in OTP) to
+                // read-only during manufacturing; however it still seems possible to soft lock it
+                // to "inaccessible" (see the datasheet's section about permissions on blank
+                // devices).
+                otp::Error::InvalidPermissions => Err(Error::InvalidPermissions),
+                // Not a write operation to OTP.
+                otp::Error::UnsupportedModification | otp::Error::Overflow => unreachable!(),
+                // OTP indices are hard-coded when reading the chip ID.
+                otp::Error::InvalidIndex => unreachable!(),
+                // Cannot happen when reading the chip ID.
+                otp::Error::UnexpectedFailure(_) => unreachable!(),
+            },
+        }
+    }
+
+    fn bytes(&self) -> Self::Bytes {
+        self.0.to_le_bytes()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
+    InvalidPermissions,
+}
+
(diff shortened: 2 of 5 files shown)


src/ariel-os-rp/src/identity.rs
```diff
@@ -0,0 +1,63 @@
+cfg_select! {
```
kaspar030, line 1: Hm, to my eyes having these huge `cfg_select!()` blocks doesn't parse well.
Could we maybe `cfg_select` the whole file?

src/ariel-os-rp/src/identity.rs
```diff
@@ -0,0 +1,63 @@
...
+    context = "rp2040" => {
+        use ariel_os_embassy_common::identity;
+
+        pub type DeviceId = identity::NoDeviceId<identity::NotImplemented>;
+    }
+    context = "rp235xa" => {
+        use embassy_rp::otp;
+
+        pub struct DeviceId(u64);
```
ROMemories: It seems safer to consider this change a breaking change as this now returns this type instead of `identity::NoDeviceId`; it's a bit unclear to me whether that really constitutes a breaking change wrt the API, but we already have breaking changes for the next release anyway.

src/ariel-os-rp/src/identity.rs
ROMemories: Done

src/ariel-os-rp/src/identity/identity_rp235x.rs
```diff
@@ -0,0 +1,51 @@
+use embassy_rp::otp;
+
+pub struct DeviceId(u64);
+
+impl ariel_os_embassy_common::identity::DeviceId for DeviceId {
+    type Bytes = [u8; 8];
+
+    fn get() -> Result<Self, impl core::error::Error> {
+        // The chip ID is considered to be globally unique.
+        let res = embassy_rp::otp::get_chipid();
```
ROMemories, line 10: The docs [don't really specify which errors can be returned](https://docs.embassy.dev/embassy-rp/0.8.0/rp235xa/otp/fn.get_chipid.html).

src/ariel-os-rp/src/identity/identity_rp235x.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    fn bytes(&self) -> Self::Bytes {
+        self.0.to_le_bytes()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Error {
```
ROMemories, line 37: @chrysn Can you check that this is how you intended the error interface of this API to be used? This is the first fallible implementation.

src/ariel-os-rp/src/identity/identity_rp235x.rs
```diff
@@ -0,0 +1,51 @@
...
+        let res = embassy_rp::otp::get_chipid();
+
+        match res {
+            Ok(id) => Ok(Self(id)),
+            Err(err) => match err {
+                // OTP page 0, which contains the chip ID, is hard locked (i.e., locked in OTP) to
+                // read-only during manufacturing; however it still seems possible to soft lock it
+                // to "inaccessible" (see the datasheet's section about permissions on blank
+                // devices).
+                otp::Error::InvalidPermissions => Err(Error::InvalidPermissions),
```
kaspar030, line 19: what does "soft lock" mean here, is that something that happens at boot before calling this, or is it irreversible? ... I'm wondering if it is sufficiently useful to actually catch this.

ROMemories: The datasheet defines it:

> The page permissions themselves are stored in OTP. Locking pages in this way is an irreversible operation, referred to as hard locking. The hardware also supports soft locking, where a page’s permissions are further restricted by writing to the relevant register in `SW_LOCK0` through `SW_LOCK63`; this restriction remains in effect until the next OTP reset. Resetting the OTP block also resets the processors, so soft locking can be used to restrict the availability of sensitive content like decryption keys to early boot stages.

It's a bit niche in this case, but it might be useful to prevent some NS code (when we support that) to identify the chip I suppose (after S code forbids access to the page)? I don't think propagating the error increases the complexity for applications authors in any way as they already have to handle the possibility of errors.

kaspar030: ok!

PR #2242, opened 2026-08-26
docs(book): add a page about Ethernet support

ROMemories:
# Description

This adds a book page about Ethernet support. The main motivation is to have a follow-up PR that documents support for external Ethernet MAC chips (see #2237). *This* PR only documents the current state, which is the support of the built-in Ethernet MAC on STM32 chips (with the RMII pinout being hard-coded).

## Future Work

This is only intended as a first version, the page could use some more background to be more beginner-friendly.

In addition, the following could be done in follow-up:

- Documenting typical supported Ethernet speeds
- Documenting the clock configuration required for the built-in Ethernet MAC on STM32
- Documenting power management considerations, especially Wake on LAN capabilities and [EEE](https://en.wikipedia.org/wiki/Energy-Efficient_Ethernet) (if relevant) 

## Changelog Entry

A page dedicated to Ethernet support has been added to the book.

## Change Checklist

### DIFF
--- a/book/src/ethernet.md
+++ b/book/src/ethernet.md
@@ -2,6 +2,11 @@
 
 Ariel OS supports Ethernet (IEEE 802.3) on specific microcontrollers and hardware configurations.
 
+## MAC Address Used for the Ethernet Link
+
+The MAC address used for the Ethernet link is derived from the device identity using [`if_index 0`][identity-interface-eui48-rustdoc].
+A different index should therefore be used to generate other EUI-48 identifiers.
+
 ## Using Microcontrollers With Built-in Ethernet MAC
 
 Some microcontrollers include an Ethernet MAC peripheral, which requires an external PHY chip for the board to support Ethernet.
@@ -14,6 +19,7 @@ Using Ethernet as the network link is enabled by selecting the [`ethernet-stm32`
 > [!NOTE]
 > The built-in Ethernet MAC can be used on other boards, as long as the same pinout is used.
 
+[identity-interface-eui48-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/identity/fn.interface_eui48.html
 [mii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface
 [rmii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface#Reduced_media-independent_interface
 [ethernet-stm32-networking-book]: ./networking.md#:~:text=ethernet%2Dstm32
--- a/book/src/ethernet.md
+++ b/book/src/ethernet.md
@@ -1,0 +1,20 @@
+# Ethernet
+
+Ariel OS supports Ethernet (IEEE 802.3) on specific microcontrollers and hardware configurations.
+
+## Using Microcontrollers With Built-in Ethernet MAC
+
+Some microcontrollers include an Ethernet MAC peripheral, which requires an external PHY chip for the board to support Ethernet.
+The MAC communicates with the external PHY using the [media-independent interface][mii-wikipedia] (MII) or the [reduced media-independent interface][rmii-wikipedia] (RMII), which is almost functionally identical to the MII but requires half the signals between the MAC and the PHY.
+
+Currently, among the set of MCUs supported by Ariel OS, only some STM32 microcontrollers feature a built-in Ethernet MAC.
+On these MCUs, only the RMII is currently supported: its pinout is currently fixed, and is the same as the one used by the manufacturer's development kit demonstrating Ethernet functionality.
+Using Ethernet as the network link is enabled by selecting the [`ethernet-stm32`][ethernet-stm32-networking-book] [laze-module][laze-modules-book].
+
+> [!NOTE]
+> The built-in Ethernet MAC can be used on other boards, as long as the same pinout is used.
+
+[mii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface
+[rmii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface#Reduced_media-independent_interface
+[ethernet-stm32-networking-book]: ./networking.md#:~:text=ethernet%2Dstm32
+[laze-modules-book]: ./build-system.md#laze-modules
--- a/book/src/networking.md
+++ b/book/src/networking.md
@@ -14,8 +14,7 @@ Boards may support all of them, only some of them, or none of them. However, cur
 Which link layer is used for networking is selected at compile time,
 through [laze modules][laze-modules-book].
 
-- `ethernet-stm32`: Selects Ethernet on STM32 chips.
-  Currently only the reduced media-independent interface (RMII) is supported.
+- `ethernet-stm32`: Selects [Ethernet on STM32 chips][ethernet-builtin-mac-book], through the built-in MAC.
 - `ltem-nrf-modem`: Selects LTE-M on nRF91 MCUs.
 - `usb-ethernet`: Selects Ethernet over USB (currently using USB CDC-NCM).
 - `wifi-cyw43`: Selects Wi-Fi using the CYW43 chip along an RP2040 or RP235x MCU (e.g., on the Raspberry Pi Pico W or Pico 2 W).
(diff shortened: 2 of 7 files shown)


doc/support_matrix.yml
```diff
@@ -780,8 +774,6 @@ chips:
           - removing items not supported
       wifi: not_available
       ble: not_available
-    note_snippets:
```
ROMemories, line 783: This is removed from the documentation of these MCUs (and moved to the book page) as this is not actually MCU-specific (not even STM32-specific), and we should factor out the generation of these MAC addresses before #2237.

book/src/ethernet.md
```diff
@@ -0,0 +1,26 @@
...
+On these MCUs, only the RMII is currently supported: its pinout is currently fixed, and is the same as the one used by the manufacturer's development kit demonstrating Ethernet functionality.
+Using Ethernet as the network link is enabled by selecting the [`ethernet-stm32`][ethernet-stm32-networking-book] [laze-module][laze-modules-book].
+
+> [!NOTE]
+> The built-in Ethernet MAC can be used on other boards, as long as the same pinout is used.
+
+[identity-interface-eui48-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/identity/fn.interface_eui48.html
+[mii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface
+[rmii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface#Reduced_media-independent_interface
+[ethernet-stm32-networking-book]: ./networking.md#:~:text=ethernet%2Dst32
```
ROMemories: The broken text fragment was just to test out #2245; it will be fixed before landing.

book/src/ethernet.md
```diff
@@ -0,0 +1,26 @@
+# Ethernet
+
+Ariel OS supports Ethernet (IEEE 802.3) on specific microcontrollers and hardware configurations.
+
+## MAC Address Used for the Ethernet Link
+
+The MAC address used for the Ethernet link is derived from the device identity using [`if_index 0`][identity-interface-eui48-rustdoc].
```
ROMemories, line 7: This calls for documentation about the device identity in the book, but this is out of scope of this PR.

ROMemories:
Odd that the text fragment check added in #2245 fails, but only on the print version. The link definitely seems valid and works as expected when I generate the print version locally and use my browser to check the linking (on the same page, as the print version generates a single page with the entire book).

Edit: ~~The issue seems to be that lychee complains because there is multiple occurrences of the text fragment target in the same page~~; because the print version is specifically a single page made of the entire book. I have not determined what the behavior expected by the text fragment specification is in that case.

Edit: The specification clearly [states that the *first* occurrence is the target text](https://wicg.github.io/scroll-to-text-fragment/#indication:~:text=If%20only%20start%20is%20specified%2C%20the%20first%20instance%20of%20this%20exact%20text%20string%20is%20the%20target%20text).

book/src/ethernet.md
ROMemories: There is now #2294 for this.

book/src/ethernet.md
```diff
@@ -0,0 +1,26 @@
...
+On these MCUs, only the RMII is currently supported: its pinout is currently fixed, and is the same as the one used by the manufacturer's development kit demonstrating Ethernet functionality.
+Using Ethernet as the network link is enabled by selecting the [`ethernet-stm32`][ethernet-stm32-networking-book] [laze-module][laze-modules-book].
+
+> [!NOTE]
+> The built-in Ethernet MAC can be used on other boards, as long as the same pinout is used.
+
+[identity-interface-eui48-rustdoc]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/identity/fn.interface_eui48.html
+[mii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface
+[rmii-wikipedia]: https://en.wikipedia.org/wiki/Media-independent_interface#Reduced_media-independent_interface
+[ethernet-stm32-networking-book]: ./networking.md#network-link-selection
```
ROMemories, line 25: I've removed the text fragment to `ethernet-stm32`: it's a bit of a shame, but this unlocks this PR and can still be made more specific later.

book/src/ethernet.md
```diff
@@ -0,0 +1,26 @@
...
+Ariel OS supports Ethernet (IEEE 802.3) on specific microcontrollers and hardware configurations.
+
+## MAC Address Used for the Ethernet Link
+
+The MAC address used for the Ethernet link is derived from the device identity using [`if_index 0`][identity-interface-eui48-rustdoc].
+A different index should therefore be used to generate other EUI-48 identifiers.
+
+## Using Microcontrollers With a Built-in Ethernet MAC
+
+Some microcontrollers include an Ethernet MAC peripheral, which requires an external PHY chip for the board to support Ethernet.
```
nponsard, line 12: It may be useful that the chips can be identified by the checkmark in the "support matrix" section of the chip view ?

It may also be needed to indicate that in the board support matrix, "ethernet" doesn't mean that the MAC is in the MCU.

book/src/networking.md
```diff
@@ -14,8 +14,7 @@ Boards may support all of them, only some of them, or none of them. However, cur
 Which link layer is used for networking is selected at compile time,
 through [laze modules][laze-modules-book].
 
-- `ethernet-stm32`: Selects Ethernet on STM32 chips.
-  Currently only the reduced media-independent interface (RMII) is supported.
+- `ethernet-stm32`: Selects [Ethernet on STM32 chips][ethernet-builtin-mac-book], through the built-in MAC.
```
nponsard: This feels like the linked page is only about STM32 chips, but from what I understand the page is meant to be generic.


```suggestion
- `ethernet-stm32`: Selects Ethernet on STM32 chips, through the [built-in MAC][ethernet-builtin-mac-book].
```

book/src/ethernet.md
```diff
@@ -0,0 +1,26 @@
...
+The MAC communicates with the external PHY using the [media-independent interface][mii-wikipedia] (MII) or the [reduced media-independent interface][rmii-wikipedia] (RMII), which is almost functionally identical to the MII but requires half the signals between the MAC and the PHY.
+
+Currently, among the set of MCUs supported by Ariel OS, only some STM32 microcontrollers feature a built-in Ethernet MAC.
+On these MCUs, only the RMII is currently supported: its pinout is currently fixed, and is the same as the one used by the manufacturer's development kit demonstrating Ethernet functionality.
+Using Ethernet as the network link is enabled by selecting the [`ethernet-stm32`][ethernet-stm32-networking-book] [laze-module][laze-modules-book].
```
nponsard:
```suggestion
Using Ethernet as the network link on supported STM32 MCUs is enabled by selecting the [`ethernet-stm32`][ethernet-stm32-networking-book] [laze-module][laze-modules-book].
```

nponsard (changes requested):

book/src/networking.md
ROMemories: Agreed, done

book/src/ethernet.md
ROMemories: Done

book/src/ethernet.md
ROMemories: Currently "Ethernet" in the support matrix means, for boards, that the one RJ45/8P8C socket is usable for Ethernet (through whichever implementation) and, for chips, that the internal Ethernet MAC is supported (at least for some hardware setup). We don't currently have boards with an MCU that features an internal MAC but where the board still comprises an external MAC used for Ethernet on the board, so I think this is never misleading. I think if this happens in the  future, this is better clarified as a footnote in the support info, as I can see this being slightly confusing for now if we try to articulate this here.

nponsard: What I meant here is that there should be an explanation to find which MCUs this page applies to.

ROMemories: I don't really think this needs anything else in this page: the introduction already clarifies that this is only available on *some* MCUs, and the built-in section reiterates that, referring to the Ethernet MAC peripheral (whose presence would be listed in the first two pages of datasheets), and further narrows it down to STM32s. I don't think this is any different from other pieces of functionality.

nponsard (approved):
Looks good to me, can be merged after squashing.

PR #2248, opened 2026-08-27
feat(rp): allow enabling internal pull-ups for the I2C controller

ROMemories:
# Description

This adds the couple of I2C configuration settings that allow enabling the internal pull-ups. That wasn't done initially because they were not exposed by `embassy-rp` (only [added in `embassy-rp@0.8.0`](https://docs.embassy.dev/embassy-rp/0.8.0/rp2040/i2c/struct.Config.html)).

As mentioned in #2223, internal pull-ups are rarely enough for I2C (especially as those on RP are pretty weak; between ~30 kΩ and ~80 kΩ on the RP2040 and RP235x at 3.3 V), but this makes the API consistent with that of `ariel-os-nrf` and `ariel-os-stm32`.

## Testing

I've only tested that enabling these still allowed I2C to work (by writing a quick example for a sensor); I didn't check that these pull-ups were enough as I don't have a sensor board *without* onboard pull-ups handy.

## Issues/PRs References

- Follow-up to #1328

## Changelog Entry

(RP) It is now possible to enable the internal pull-ups on I2C pins if necessary.

## Change Checklist

### DIFF
--- a/src/ariel-os-rp/src/i2c/controller/mod.rs
+++ b/src/ariel-os-rp/src/i2c/controller/mod.rs
@@ -21,12 +21,18 @@ const KHZ_TO_HZ: u32 = 1000;
 pub struct Config {
     /// The frequency at which the bus should operate.
     pub frequency: Frequency,
+    /// Whether to enable the internal pull-up resistor on the SDA pin.
+    pub sda_pullup: bool,
+    /// Whether to enable the internal pull-up resistor on the SCL pin.
+    pub scl_pullup: bool,
 }
 
 impl Default for Config {
     fn default() -> Self {
         Self {
             frequency: Frequency::UpTo100k(Kilohertz::kHz(100)),
+            sda_pullup: false,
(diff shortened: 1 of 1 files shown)


src/ariel-os-rp/src/i2c/controller/mod.rs
```diff
@@ -21,12 +21,18 @@ const KHZ_TO_HZ: u32 = 1000;
 pub struct Config {
     /// The frequency at which the bus should operate.
     pub frequency: Frequency,
+    /// Whether to enable the internal pull-up resistor on the SDA pin.
+    pub sda_pullup: bool,
+    /// Whether to enable the internal pull-up resistor on the SCL pin.
+    pub scl_pullup: bool,
```
ROMemories, line 27: These are the same doc comments as for the other HALs.

nponsard (approved):
Looks good to me, tested with an MPU6050 connected to a rpi-pico-w.

PR #2250, opened 2026-08-27
feat(log): add custom logging transport driver option

nponsard:
# Description

This adds the ability to send logs through a custom transport that can be defined in the application.

## Testing

```sh 
laze -C examples/usb-serial-logger build -b nrf52840dk run 
```

Read the logs through the usb serial device.

## Issues/PRs References

- Depends on #2272 

## Changelog Entry

A custom logging transport can now be defined from the application, see the [logging book page](https://ariel-os.github.io/ariel-os/dev/docs/book/logging.html#custom-logging-transports).

## Change Checklist

### DIFF
--- a/book/src/logging.md
+++ b/book/src/logging.md
@@ -106,6 +106,26 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 > - Other logging transports will later be supported, including UART and USB CDC-ACM on non-ESP32 devices.
 > - Using multiple transports at the same time may be supported in the future.
 
+### Custom Logging Transports
+
+An application can connect a custom logging transport to send the logs through.
+
+This is enabled using the `custom-logging-transport` [laze module][laze-modules-book]. This will 
+disable all other logging transports and gives access to [`register_custom_transport()`][register_custom_transport_fn].
+This functions takes two arguments: 
+ - `write_bytes`: this function should take a reference to a slice of bytes (`&[u8]`) as input and
+ send it through the transport.
+ - `flush`: this function should flush the data through the transport when available for
+the transport.
+
+> [!IMPORTANT]
+> The functions given to `register_custom_transport` must not wait on interrupts they may be called
+> in the context of a critical section and should not make calls to the logging facade.
+
+> [!NOTE]
+> The logs triggered before setting the logging transport functions using `register_custom_transport` 
+> may be lost.
+
 ## Fetching and Displaying the Logging Output on the Host
 
 When the [flashing][flashing-board-book] transport allows it, the `run` laze task not only flashes the board but also fetches and displays the logging output on the host.
@@ -145,3 +165,4 @@ When multiple tasks of the same name exist, which variant is used depends on whi
 [semihosting-book]: ./flashing-debugging.md#semihosting
 [laze-tasks-book]: ./build-system.md#laze-tasks
 [rtt-book]: ./flashing-debugging.md#real-time-transfer-rtt
+[register_custom_transport_fn]: https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/log/transport/fn.register_custom_transport.html
--- a/examples/usb-serial-logger/Cargo.toml
+++ b/examples/usb-serial-logger/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "example-usb-serial-logger"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = ["time"] }
+ariel-os-boards = { path = "../../src/ariel-os-boards" }
+embassy-futures = { workspace = true }
+embassy-sync = { workspace = true }
+embedded-io-async = { version = "0.7" }
+
+[lints]
+workspace = true
--- a/examples/usb-serial-logger/README.md
+++ b/examples/usb-serial-logger/README.md
@@ -1,0 +1,14 @@
+# usb-serial-logger
+
+## About
+
+This application is showing basic logging over _USB serial_ with Ariel OS.
+
+## How to run
+
+In this directory, run
+
+    laze build -b nrf52840dk run
+
+With the device USB cable connected, a USB serial port should show up on your computer.
+It will print logs at regular intervals.
--- a/examples/usb-serial-logger/laze.yml
+++ b/examples/usb-serial-logger/laze.yml
@@ -1,0 +1,6 @@
+apps:
+  - name: example-usb-serial-logger
+    selects:
+      - custom-logging-transport # Has to be selected before `usb` is selected
+      - log
+      - usb
--- a/examples/usb-serial-logger/src/main.rs
+++ b/examples/usb-serial-logger/src/main.rs
@@ -1,0 +1,106 @@
+#![no_std]
+#![no_main]
+
+use ariel_os::{
+    cell::StaticCell,
+    log::{debug, error, info, println, trace, transport::register_custom_transport, warn},
+    reexports::embassy_usb::{self, class::cdc_acm::CdcAcmError},
+    time::Timer,
+};
+use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, pipe::Pipe};
+use embassy_usb::{
+    class::cdc_acm::{CdcAcmClass, State},
+    driver::EndpointError,
+};
+use embedded_io_async::Write;
+
+const MAX_FULL_SPEED_PACKET_SIZE: u8 = 64;
+
+const LOG_BUFFER_SIZE: usize = 1024;
+static LOG_BUFFER: Pipe<CriticalSectionRawMutex, LOG_BUFFER_SIZE> = Pipe::new();
+
+#[ariel_os::task(autostart)]
+async fn logging() {
+    loop {
+        Timer::after_secs(1).await;
+        info!("Testing USB CDC ACM logging");
+        println!("-- this is printed via `println!()`");
+        trace!("-- trace log level enabled");
+        debug!("-- debug log level enabled");
+        info!("-- info log level enabled");
+        warn!("-- warn log level enabled");
+        error!("-- error log level enabled (just testing)");
+    }
+}
+
+#[ariel_os::task(autostart, usb_builder_hook)]
+async fn main() {
+    register_custom_transport(write_bytes, flush);
+
+    info!("Hello World!");
+
+    static STATE: StaticCell<State<'_>> = StaticCell::new();
+
+    // Create and inject the USB class on the system USB builder.
+    let class = USB_BUILDER_HOOK
+        .with(|builder| {
+            CdcAcmClass::new(
+                builder,
+                STATE.init_with(State::new),
+                MAX_FULL_SPEED_PACKET_SIZE.into(),
+            )
+        })
+        .await;
+
+    let (mut sender, mut receiver) = class.split();
+    loop {
+        let sender_fut = async {
+            sender.wait_connection().await;
+            let mut buffer = [0; LOG_BUFFER_SIZE as usize];
+            loop {
+                let len = LOG_BUFFER.read(&mut buffer).await;
+
+                if matches!(
+                    sender.write_all(&buffer[..len]).await,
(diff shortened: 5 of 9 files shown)


examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+        let n = match USB_LOG_PIPE.try_write(&bytes[total..end]) {
+            Ok(n) => n,
+            // Pipe full, drop the data.
+            Err(_) => return,
+        };
+        total += n;
+    }
+}
+
+// No-op, flash isn't possible with this setup.
```
ROMemories:
```suggestion
// No-op, flushing isn't possible with this setup.
```

examples/usb-serial-logger/README.md
```diff
@@ -0,0 +1,16 @@
+# usb-serial-logger
+
+## About
+
+This application is showing basic logging through
+[Embassy](https://github.com/embassy-rs/embassy) _USB serial_ with Ariel OS.
```
ROMemories:
```suggestion
This application is showing basic logging over _USB serial_ with Ariel OS.
```

examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
+#![no_std]
+#![no_main]
+
+use ariel_os::{
+    cell::StaticCell,
+    log::{Hex, debug, error, info, println, trace, transport::register_custom_transport, warn},
+    reexports::embassy_usb::{self, class::cdc_acm::CdcAcmError},
+    time::Timer,
+    usb::UsbDriver,
```
ROMemories: Apparently unused:
```suggestion
```

examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+    config.max_packet_size_0 = MAX_FULL_SPEED_PACKET_SIZE;
+
+    // Required for Windows support.
+    config.composite_with_iads = true;
+    config.device_class = 0xEF;
+    config.device_sub_class = 0x02;
+    config.device_protocol = 0x01;
+    config
+};
+
```
ROMemories: I think I'd rather we remove this from this example for now. I know the product value will show say "USB-Ethernet example", but the USB *configuration* is outside the scope of this example and I fear this gets cargo-culted (on top of adding noise to the example). We'd likely want to revisit the default `product` value, but that's outside the scope of this PR:


```suggestion
```

examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+static USB_LOG_PIPE: Pipe<CriticalSectionRawMutex, USB_PIPE_SIZE> = Pipe::new();
```
ROMemories: That's unfortunate terminology here as the USB spec also defines a concept of "USB pipe" (section 5.3.2 of the USB 2.0 spec). I thought `USB_PIPE_SIZE` was the size of a USB endpoint at first. It would be better to find something else (even just "buffer" if nothing else works).

examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+};
+use embassy_sync::{blocking_mutex::raw::CriticalSectionRawMutex, pipe::Pipe};
+use embassy_usb::{
+    class::cdc_acm::{CdcAcmClass, State},
+    driver::EndpointError,
+};
+use embedded_io_async::Write;
+
+const USB_PIPE_SIZE: usize = 1024;
+const MAX_FULL_SPEED_PACKET_SIZE: u8 = 64;
```
ROMemories, line 17: I wish that constant was part of `embassy-usb`…

examples/usb-serial-logger/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "example-usb-serial-logger"
+edition.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+ariel-os = { path = "../../src/ariel-os", features = [
+  "override-usb-config",
```
ROMemories: Can likely be removed with the in-app config above:
```suggestion
```

examples/usb-serial-logger/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## How to run
+
+In this directory, run
+
+    laze build -b nrf52840dk run
+
+With the device USB cable connected, a USB ACM serial port should show up on
```
ROMemories:
```suggestion
With the device USB cable connected, a USB serial port should show up on
```

examples/usb-serial-logger/laze.yml
```diff
@@ -0,0 +1,6 @@
+apps:
+  - name: example-usb-serial-logger
+    selects:
+      - log
+      - custom-logging-transport
```
ROMemories: I think these should be kept in lexicographical order: it does seem that forcing `log` is currently needed, but at least it seems it doesn't *have* to come before `custom-logging-transport`:
```suggestion
      - custom-logging-transport
      - log
```

examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+
+                        info!("Received from USB serial: {}", Hex(data));
+                    }
+                    Err(e) => {
+                        error!("USB error: {:?}", e);
+                    }
+                };
+            }
+        };
+
```
ROMemories, line 86: Is this really necessary? This doesn't seem related to demonstrating logging itself and the echoing behavior isn't documented in the example's README.

book/src/logging.md
ROMemories, line 1: I wouldn't be against splitting this piece of docs, so that we can land #2272 as soon as we land the example.

book/src/logging.md
```diff
@@ -106,6 +106,25 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 > - Other logging transports will later be supported, including UART and USB CDC-ACM on non-ESP32 devices.
 > - Using multiple transports at the same time may be supported in the future.
 
+### Custom logging transports
```
ROMemories: The rest of the page uses title case:
```suggestion
### Custom Logging Transports
```

book/src/logging.md
```diff
@@ -106,6 +106,25 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
...
+
+An application can connect a custom logging transport to send the logs through.
+
+This is enabled using the `custom-logging-transport` [laze module][laze-modules-book], this will disable 
+all other logging transports and gives access to [`register_custom_transport()`](https://ariel-os.github.io/ariel-os/dev/docs/api/ariel_os/log/transport/fn.register_custom_transport.html).
```
ROMemories: Can you define the link at the bottom of the file as done for the other links?

book/src/logging.md
```diff
@@ -106,6 +106,25 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
...
+This functions takes two arguments: 
+ - `write_bytes`: this function should take a reference to a slice of bytes (`&[u8]`) as input and
+ send it through the transport.
+ - `flush`: this function should flush the data through the transport if possible.
```
ROMemories: I'm not sure I understand "if possible" here: is it not a requirement of *the transport* (or is instead a requirement of the logger or logging facade)? In that case shouldn't be something like "if required by the transport"?

book/src/logging.md
```diff
@@ -106,6 +106,25 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
...
+
+> [!IMPORTANT]
+> The functions given to `register_transport_functions` must not wait on interrupts they may be called
+> in the context of a critical section and should not make calls to the logging facade.
+
+> [!NOTE]
+> The logs triggered before setting the logging transport functions using `register_transport_functions`
```
ROMemories:
```suggestion
> The logs triggered before setting the logging transport functions using `register_custom_transport` 
```

book/src/logging.md
```diff
@@ -106,6 +106,25 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
 > - Other logging transports will later be supported, including UART and USB CDC-ACM on non-ESP32 devices.
 > - Using multiple transports at the same time may be supported in the future.
 
+### Custom logging transports
+
+An application can connect a custom logging transport to send the logs through.
+
+This is enabled using the `custom-logging-transport` [laze module][laze-modules-book], this will disable
```
ROMemories: This could be split into two sentences at the comma.

examples/usb-serial-logger/src/main.rs
```diff
@@ -0,0 +1,127 @@
...
+            sender.wait_connection().await;
+            let mut buffer = [0; USB_PIPE_SIZE as usize];
+            loop {
+                let len = USB_LOG_PIPE.read(&mut buffer).await;
+
+                if matches!(
+                    sender.write_all(&buffer[..len]).await,
+                    Err(CdcAcmError::NotConnected)
+                ) {
+                    sender.wait_connection().await;
```
ROMemories, line 68: I suppose this can happen if the connection first got established then got disconnected, and we're hoping it gets connected again? Is there an easy of triggering that (I do agree this should handled even if not)? Can we get a comment about when this can happen (even if it's just something "wait for reconnection after disconnection")?

ROMemories (commented):
Successfully tested on the nrf52840dk. This might need another round for me to have a deeper look at the implementation, but maybe not, in which case it should be good to go after this one.

book/src/logging.md
nponsard: The idea is that if the transport can be flushed it should be flushed when calling this function, otherwise it's ok to give a no-op function.

Also this is never called when using the `log` facade, should `flush` be called somewhere in `transport_print()` ?

book/src/logging.md
nponsard: I'm not sure to understand what you want to split and how. Do you mean that the commit changing the book should be in a different PR that will be merged later ?

examples/usb-serial-logger/src/main.rs
nponsard: thought about renaming it to "LOG_QUEUE" but "LOG_BUFFER" sounds nicer.

book/src/logging.md
ROMemories: Still needs updating

nponsard: I was looking at the wrong line :facepalm: 
Thanks

book/src/logging.md
ROMemories: That's indeed what I meant but that shouldn't be needed after all.

book/src/logging.md
```diff
@@ -106,6 +106,26 @@ When `espflash` is selected at the time of compilation, `logging-over-debug-chan
...
+### Custom Logging Transports
+
+An application can connect a custom logging transport to send the logs through.
+
+This is enabled using the `custom-logging-transport` [laze module][laze-modules-book]. This will 
+disable all other logging transports and gives access to [`register_custom_transport()`][register_custom_transport_fn].
+This functions takes two arguments: 
+ - `write_bytes`: this function should take a reference to a slice of bytes (`&[u8]`) as input and
+ send it through the transport.
+ - `flush`: this function should flush the data through the transport when available for
```
ROMemories, line 118: Oddly enough the best explanation I could find on what flushing is supposed to do (at least for defmt) is in [defmt migration guide to 0.3](https://defmt.ferrous-systems.com/migration-02-03#flush): ideally the function would make sure the logs have been read/received on the other side. I suppose that's good enough for now, we may revisit/clarify this later.

book/src/logging.md
ROMemories, line 85: The custom transport should also be added to this table, but I'll do that in a follow-up.

ROMemories (approved):
LGTM, tested again on an nrf52840dk.

book/src/logging.md
ROMemories: > Also this is never called when using the `log` facade, should `flush` be called somewhere in `transport_print()` ?

The contract of [`Log::flush()`](https://docs.rs/log/0.4.34/log/trait.Log.html#tymethod.flush) isn't super clear to me, I think it's fine to leave it as is for now.

PR #2257, opened 2026-09-01
feat(esp): add Seeed Studio Xiao ESP32-S3 board support

shreyashsaitwal:
# Description

This PR adds support for the `seeedstudio-xiao-esp32s3` board based on the `esp32s3r8` chip.

## Testing

I've tested the examples on an actual Seeed Studio Xiao ESP32-S3.

## Changelog Entry

The ESP32-S3R8 MCU and the Seeed Studio XIAO ESP32-S3 board is now supported.

## Change Checklist


doc/support_matrix.yml
```diff
@@ -1215,6 +1237,11 @@ builders:
       user_usb: needs_testing
       ethernet_over_usb: needs_testing
 
+  seeedstudio-xiao-esp2s3:
```
nponsard: Just a drive-by review, found a typo:
```suggestion
  seeedstudio-xiao-esp32s3:
```

boards/seeedstudio-xiao-esp32-s3.yaml
```diff
@@ -0,0 +1,18 @@
+version: 0.4.0
+targets:
+  seeedstudio-xiao-esp32s3:
+    chip: esp32s3r8
+    flags:
+      - has_usb_cdc_acm_port
```
ROMemories: The ESP32-S3 has a "generic USB peripheral" (unlike most other ESPs):
```suggestion
      - has_usb_cdc_acm_port
      - has_usb_device_port
```

boards/seeedstudio-xiao-esp32s3.yaml
```diff
@@ -0,0 +1,18 @@
+version: 0.4.0
+targets:
+  seeedstudio-xiao-esp32s3:
```
ROMemories: The product name does have a dash here (unlike the Xiao ESP32C6, mostly); we try to match upstream names as closely as possible for these:
```suggestion
  seeedstudio-xiao-esp32-s3:
```
The file also has to be renamed.

boards/seeedstudio-xiao-esp32-s3.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - has_ble_esp
+    leds:
+      - pin: GPIO21
```
ROMemories, line 10: 👍

boards/seeedstudio-xiao-esp32-s3.yaml
```diff
@@ -0,0 +1,18 @@
...
+        color: yellow
+        active: low
+        aliases:
+          - USER_LED
+    buttons:
+      - pin: GPIO0
```
ROMemories, line 16: 👍

laze-project.yml
```diff
@@ -922,6 +925,9 @@ modules:
           # and the second $ escapes for laze's internal ninja, resulting in
           # the shell to see just ${ESPUP_...
           - . $\\${ESPUP_EXPORT_FILE:-~/export-esp.sh}
+        CARGO_ENV:
+          # liboscore's bindgen needs the Xtensa sysroot to find relevant headers
+          - BINDGEN_EXTRA_CLANG_ARGS=--sysroot=$$(${CC} -print-sysroot)
```
ROMemories: Please split this out of this PR; that way we'll be able to review that separately.

doc/support_matrix.yml
```diff
@@ -608,6 +608,28 @@ chips:
       ble: supported
       ethernet: not_available
 
+  # Currently identical to esp32s3.
+  esp32s3r8:
```
ROMemories, line 612: 👍

doc/support_matrix.yml
ROMemories: An [entry in `boards` also has to be added](https://ariel-os.github.io/ariel-os/0.5.0/docs/book/adding-board-support.html#:~:text=Add%20an%20entry%20under%20boards%2C%20that%20references%20that%20new%20builder%2E%20Include%20a%20link%20to%20a%20web%2Earchive%2Eorg%20snapshot%20that%20describes%20the%20board) below.

ROMemories (changes requested):
Thank your for the PR! This should be good to go after the couple of changes highlighted in comments.

The comments with 👍 are just things I've checked and don't require any action from your side.

shreyashsaitwal:
@ROMemories I've fixed the name of the board, split the bindgen fix into #2260, and applied other suggestions.

BTW, are the board flags like `has_usb_device_port` documented somewhere? I suspected I might have to add that initially but assumed `has_usb_cdc_acm_port` would be enough as there was no where I'd find more information about the flags and their purpose.

ROMemories (approved):
Thank your for splitting and applying the suggestions! I've taken the liberty to make the changes necessary to make the CI happy: removed some trailing spaces and re-generated `ariel-os-boards` with `sbg-gen`. I've also moved some items around to preserve the lexicographical order.

ROMemories:
> BTW, are the board flags like `has_usb_device_port` documented somewhere? I suspected I might have to add that initially but assumed `has_usb_cdc_acm_port` would be enough as there was no where I'd find more information about the flags and their purpose.

Unfortunately not currently: `has_usb_cdc_acm_port` means that the board has a USB port connected to the ESP's USB Serial/JTAG peripheral, and `has_usb_device_port` means that there is a USB port connected to the MCU's "generic USB peripheral" (which only the ESP32-S3 has among our currently supported set of ESPs); they're also not mutually exclusive as the MCU's USB PHY is shared between the two peripherals.

PR #2261, opened 2026-09-02
refactor(wifi): unify WiFi configuration and connection

nponsard:
# Description

This refactors the WiFi configuration and join mechanic to be unified between ESP and RP.
This work is the first step towards having WiFi AP in Ariel. 

## Testing

```
laze -C examples/tcp-client/ build -b espressif-esp32-s3-devkitc-1 run
laze -C examples/tcp-client/ build -b rpi-pico-w run
```

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy-common/src/wifi.rs
+++ b/src/ariel-os-embassy-common/src/wifi.rs
@@ -1,10 +1,10 @@
-//! Common WiFi configuration for Ariel OS.
+//! Common Wi-Fi configuration for Ariel OS.
 //!
 //! This module contains the types and [`WIFI_CONFIG`] constant containing the build-time configuration.
 
 use ariel_os_utils::str_from_env;
 
-/// Current wifi configuration.
+/// Current Wi-Fi configuration.
 // TODO: Feature-gate this when AP mode is implemented.
 pub const WIFI_CONFIG: StationConfig = {
     const WIFI_NETWORK: &str = str_from_env!("CONFIG_WIFI_NETWORK", "Wi-Fi SSID (network name)");
@@ -15,10 +15,10 @@ pub const WIFI_CONFIG: StationConfig = {
     }
 };
 
-/// WiFi configuration in station mode.
+/// Wi-Fi configuration in station mode.
 pub struct StationConfig {
-    /// WiFi access point SSID.
+    /// Wi-Fi access point SSID.
     pub ssid: &'static str,
-    /// WiFi access point password.
+    /// Wi-Fi access point password.
     pub password: &'static str,
 }
--- a/src/ariel-os-embassy-common/src/wifi.rs
+++ b/src/ariel-os-embassy-common/src/wifi.rs
@@ -1,0 +1,24 @@
+//! Common WiFi configuration for Ariel OS.
+//!
+//! This module contains the types and [`WIFI_CONFIG`] constant containing the build-time configuration.
+
+use ariel_os_utils::str_from_env;
+
+/// Current wifi configuration.
+// TODO: Feature-gate this when AP mode is implemented.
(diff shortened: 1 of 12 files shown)


src/ariel-os-rp/src/cyw43.rs
```diff
@@ -23,15 +26,13 @@ pub type NetworkDevice = cyw43::NetDriver<'static>;
 static STATE: StaticCell<cyw43::State> = StaticCell::new();
 
 #[cfg(feature = "wifi")]
-pub async fn join(mut control: cyw43::Control<'static>) {
+#[embassy_executor::task]
```
nponsard: Making it a task matches how it's done on ESP.
This also allows the board to fully boot without having to wait for the join function to finish.

src/ariel-os-esp/src/wifi/esp_wifi.rs
```diff
@@ -5,21 +5,24 @@ use esp_radio::wifi::{
...
+pub fn init(
+    peripherals: &mut crate::OptionalPeripherals,
+    spawner: Spawner,
+) -> (NetworkDevice, WifiController<'static>) {
     let config = Config::default();
     let wifi = peripherals.WIFI.take().unwrap();
 
     let (controller, interfaces) = esp_radio::wifi::new(wifi, config).unwrap();
 
-    spawner.spawn(connection(controller)).ok();
```
nponsard, line 16: This is now started by ariel-os-embassy to decouple the wifi modem initialization and station operation.

src/ariel-os-embassy/src/lib.rs
```diff
@@ -427,9 +427,21 @@ async fn init_task(mut peripherals: hal::OptionalPeripherals) {
...
+            .spawn(hal::cyw43::join(control, wifi_config))
+            .expect("start wifi join task");
+    };
+
+    #[cfg(feature = "wifi-esp")]
+    {
+        spawner
+            .spawn(hal::wifi::esp_wifi::join(control, wifi_config))
+            .expect("start wifi join task");
     };
```
nponsard, line 444: Should this be unified into a single call ? 

```rust
    #[cfg(feature = "wifi")]
    {
        spawner
            .spawn(hal::wifi::join(control, wifi_config))
            .expect("start wifi join task");
    };
```

(this means we have to make aliases for the join function)

src/ariel-os-embassy-common/src/wifi.rs
```diff
@@ -0,0 +1,24 @@
+//! Common WiFi configuration for Ariel OS.
```
ROMemories: Same for the other occurrences:
```suggestion
//! Common Wi-Fi configuration for Ariel OS.
```

ROMemories (commented):
Partial review: this increases the `text` size of `udp-echo` on rpi-pico-w by ~2.5 kB and the `bss` size by 800 bytes. This may be because of the new task, but I don't think we can afford that.

If you split out the factoring of the configuration constants, I think we can land that quickly however.

nponsard:
The embassy task change can easily be reverted so I did that.

src/ariel-os-embassy/src/lib.rs
ROMemories: That does sound better; also no need for this expect message I think, and you can just use `WIFI_CONFIG` directly when doing that.

ROMemories (commented):
I'm still to test this in hardware, but I expect this to be good to go after the next round.

src/ariel-os-embassy/src/lib.rs
nponsard: Sadly this can't be done without the cyw43 `join` being a task.

PR #2264, opened 2026-09-03
fix(threads): reduce overhead of multi-core scheduling with core-affinities

elenaf9:
# Description

Enabling the `core-affinity` feature for multi-core is currently quite expensive. 
This is due to two parts;
1. The selection of the next runnable thread is more complex
2. The `yield_now` function needs to check for more edge cases

## Regarding 1.

Instead of just popping[^1] the head of the runqueue, with core-affinities we first need to iterate over the runqueue to find the first thread with matching core affinity, and then delete the thread from the runqueue [^2]. 
Deleting a thread from the runqueue is quite expensive because we need to again find the thread's position in the runqueue to update the links from it's predecessor and potentially also the runqueue tail.  
This PR optimizes the deletion process by using the already present info from the first iteration for deleting with `O(1)` instead of `O(n)`.

## Regarding 2.

With core affinities, we have the unique case that it is possible that a lower priority thread is running, even though a higher prio thread is waiting in the runqueue. This is because the higher prio thread might be pinned to another core. 
When a thread yields, we therefore not only need to check if there is a another thread waiting, but also if the yielding thread can migrate to another core.
The PR optimizes this case a bit. We only check if a thread can migrate (the check is quite expensive) if the next waiting thread is actually pinned.

[^1]: the currently running thread is not in the runqueue on multicore
[^2]: This needs to happen in two steps because we otherwise run into conflicting mutable/immutable accesses.

## Testing

Benchmarked with the `bench_sched_yield` benchmark with 3 and 4 threads. Tested 4 cases:
1. `core-affinity` feature disabled
2. `core-affinity` feature enabled but not used
4. `core-affinity` feature enabled and the benchmark thread is pinned to `Core 0`
5. `core-affinity` feature enabled and the benchmark thread is pinned to `Core 1`

|Test | disabled | enabled | Pinned on Core 0 | Pinned on Core 1
|- | -| - | - | -
main (t=3) | 1397 | 2998 | 3585 | 3583 
this PR (t=3) |1397 | 1905 | 2716 | 2715
main (t=4) | 1845  | 4028 | 3773 | 3770 
this PR (t=4) | 1845 | 2537 |  2447 | 2699 

The numbers are still not ideal, but at least better. Also, I noticed that **2.** (handling the edge-case when yielding) is more expensive that **1.**. I'd assume that this function is used rather rarely, so generally the overhead of the `core-affinity` feature is acceptable once this PR lands. 

## Issues/PRs References

Core-affinities are needed for a quick-fix to re-enable multi-core for the esp32-s3, #1716.

## Changelog Entry

Improve scheduler performance under multi-core with the `core-affinity` feature enabled.

## Change Checklist

### DIFF
--- a/src/ariel-os-runqueue/src/runqueue.rs
+++ b/src/ariel-os-runqueue/src/runqueue.rs
@@ -115,6 +115,16 @@ impl<const N_QUEUES: usize, const N_THREADS: usize> RunQueue<{ N_QUEUES }, { N_T
         }
     }
 
+    /// Removes thread with tid `n`.
+    ///
+    /// The function is an optimized version of [`del`] with O(1) for when
+    /// the thread's predecessor in the runqueue and its runqueue ID is known.
+    pub fn del_opt(&mut self, n: ThreadId, prev: ThreadId, rq: RunqueueId) {
+        if self.queues.del_opt(n.0, prev.0, rq.0) {
+            self.bitcache &= !(1 << rq.0);
+        }
+    }
+
     /// Returns the tid that should run next.
     ///
     /// Returns the next runnable thread of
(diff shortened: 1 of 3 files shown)


src/ariel-os-runqueue/src/runqueue.rs
```diff
@@ -207,45 +229,6 @@ fn ffs(val: usize) -> u32 {
...
-    prev: u8,
-    // Head of the currently iterated runqueue.
-    rq_head: Option<u8>,
-    // Bitcache with the remaining queues that have to be iterated.
-    bitcache: usize,
-}
-
-impl<const N_QUEUES: usize, const N_THREADS: usize> Iterator
-    for RunQueueIter<'_, { N_QUEUES }, { N_THREADS }>
-{
```
elenaf9, line 227: Removed because we need to remember and return some intermediate state from the iteration, which doesn't fit well with the iterator trait.  Solved as a loop in `get_next_filter` instead.

PR #2269, opened 2026-09-07
fix: remove paths deps in favor of workspace deps when possible

anlavandier:
# Description
This PR changes many paths deps (and some non paths deps) to workspace deps.

This PR also adds `ariel-os-native`  and `ariel-os-stm32-mapping` to the workspace dependencies.

## Open Questions
- `ariel-os-native` has a dep on `get_random` with `v0.2`, whereas `ariel-os-random` has a dep on `v0.3.3`. And `tests/random-getrandom` has a dep on `v0.3` ...
- `embedd-io-async` is a workspace deps with `v0.6.1` and is used as a direct dep with `v.0.7.0` in the examples `tcp-client` and `tcp-echo`

## Change Checklist

### DIFF
--- a/examples/ble-scanner/Cargo.toml
+++ b/examples/ble-scanner/Cargo.toml
@@ -8,8 +8,8 @@ publish = false
 ariel-os = { path = "../../src/ariel-os", features = ["time"] }
 ariel-os-boards = { path = "../../src/ariel-os-boards" }
 
-embassy-futures = { version = "0.1.1" }
-heapless = "0.8"
+embassy-futures = { workspace = true }
+heapless = { workspace = true }
 trouble-host = { workspace = true, features = ["derive", "scan"] }
 
 [lints]
--- a/src/lib/coapcore/Cargo.toml
+++ b/src/lib/coapcore/Cargo.toml
@@ -25,7 +25,8 @@ liboscore = { version = "0.2.4", default-features = false }
 
 defmt = { workspace = true, optional = true }
 defmt-or-log = { version = "0.2.2", default-features = false }
-heapless = { workspace = true }
+# Decoupled from the workspace, as it is used in the public API.
+heapless = "0.9.0"
 log = { workspace = true, optional = true }
 minicbor = { version = "2", features = ["derive"] }
 minicbor-adapters = "0.0.8"
--- a/src/lib/coapcore/Cargo.toml
+++ b/src/lib/coapcore/Cargo.toml
@@ -25,8 +25,8 @@ liboscore = { version = "0.2.4", default-features = false }
 
 defmt = { workspace = true, optional = true }
 defmt-or-log = { version = "0.2.2", default-features = false }
-heapless = "0.9.0"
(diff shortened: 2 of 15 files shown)


examples/ble-scanner/Cargo.toml
```diff
@@ -5,11 +5,11 @@ license.workspace = true
...
 [dependencies]
-ariel-os = { path = "../../src/ariel-os", features = ["time"] }
-ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os = { workspace = true, features = ["time"] }
+ariel-os-boards = { workspace = true }
 
-embassy-futures = { version = "0.1.1" }
-heapless = "0.8"
+embassy-futures = { workspace = true }
+heapless = { workspace = true }
```
anlavandier, line 12: This is what cause the minor lockfile update

examples/alloc/Cargo.toml
```diff
@@ -5,8 +5,8 @@ license.workspace = true
 publish = false
 
 [dependencies]
-ariel-os = { path = "../../src/ariel-os" }
-ariel-os-boards = { path = "../../src/ariel-os-boards" }
+ariel-os = { workspace = true, default-features = true }
+ariel-os-boards = { workspace = true }
```
ROMemories: I think we want to keep `ariel-os` and `ariel-os-boards` as path dependencies in tests and examples, because this is how they have to be used in OoT applications. We only want to use the workspace for the other `ariel-os-*` packages. This also avoids having to awkwardly re-enable the default features in each application, or make an exception for not disabling them in the workspace manifest.

ROMemories (commented):
I'll do the rest of the review in parallel.

src/lib/coapcore/Cargo.toml
```diff
@@ -25,8 +25,8 @@ liboscore = { version = "0.2.4", default-features = false }
 
 defmt = { workspace = true, optional = true }
 defmt-or-log = { version = "0.2.2", default-features = false }
-heapless = "0.9.0"
-log = { version = "0.4", optional = true }
+heapless = { workspace = true }
```
ROMemories: I think we should keep it separate from the workspace, as it shows up [in the public API](https://ariel-os.github.io/ariel-os/0.5.0/docs/api/coapcore/seccfg/struct.ConfigBuilder.html#method.with_aif_asymmetric_es256), and this crate is managed separately from Ariel OS:
```suggestion
# Decoupled from the workspace, as it is used in the public API.
heapless = "0.9.0"
```

ROMemories (commented):

anlavandier:
> Should be good to go after this one.

You did not seem to address the open questions I put in the PR description. Do we keep the duplicate get_random and embedded-io-async versions ?

ROMemories (approved):
> You did not seem to address the open questions I put in the PR description. Do we keep the duplicate get_random and embedded-io-async versions ?

Yes at least for now, this is orthogonal to this PR.

PR #2270, opened 2026-09-07
chore: remove unused workspace deps

anlavandier:
# Description
Fixes the [`unused_workspace_dependencies`](https://doc.rust-lang.org/cargo/reference/lints.html#unused_workspace_dependencies) lint. 

## Change Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,6 @@ embedded-hal-async = { version = "1.0.0", default-features = false }
 embedded-io = { version = "0.6.1", default-features = false }
 embedded-io-async = { version = "0.6.1", default-features = false }
 embedded-nal-coap = "=0.1.0-alpha.5"
-embedded-storage = { version = "0.3.1" }
 embedded-storage-async = { version = "0.4.1" }
 embedded-test = { version = "0.7.1", default-features = false, features = [
   "ariel-os-09",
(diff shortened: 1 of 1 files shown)


Cargo.toml
ROMemories, line 1: Can you split this out and prune the vetting?

anlavandier: Split between lints ?

anlavandier: I removed everything but the unused workspace deps from this PR.

anlavandier: Regarding the vetting there is nothing to prune here because
- esp-wifi-sys and embedded-storage-async are still in our tree and lockfile and thus require exemptions.
- esp-storage was not in the lockfile or in the exemptions.

ROMemories: Ah indeed, I had missed the dependency between `embedded-storage-async` and `embedded-storage`. `esp-wifi-sys` was not in out exemptions.

ROMemories (approved):

ROMemories:
> `[missing_lints_inheritance](https://doc.rust-lang.org/cargo/reference/lints.html#missing_lints_inheritance)` still raises a warning on `ariel-os-boards` which is auto-generated so to properly silence this warning we would need to also update `sbd-gen`.

Not sure we should do this in `sbd-gen`, but we don't have to do anything about it right now anyway.

PR #2272, opened 2026-09-07
docs(log): make transport module visible

nponsard:
# Description

This makes the transport documentation visible from the rendered documentation.

## Testing

Look at the published documentation and check if the `transport` module is shown in `ariel-os-log` 

## Changelog Entry

Skipping changelog, the real feature will be added with #2250.

## Change Checklist

### DIFF
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -70,12 +70,12 @@ jobs:
               -p coapcore \
               --features "
                   ariel-os-sensors/max-sample-min-count-12,
+                  ariel-os-log/custom-transport-driver,
                   bench,
                   ble,
                   coap,
                   core-affinity,
                   csprng,
-                  defmt,
                   dns,
                   executor-thread,
                   external-interrupts,
--- a/src/ariel-os-log/src/lib.rs
+++ b/src/ariel-os-log/src/lib.rs
@@ -22,7 +22,10 @@
 mod _featurecomb {}
 
 #[doc(hidden)]
-#[cfg(feature = "_pluggable-transport")]
+#[cfg(feature = "internal-transport-driver")]
+pub mod transport;
+
+#[cfg(feature = "custom-transport-driver")]
 pub mod transport;
 
 #[allow(unused, reason = "conditional compilation")]
--- a/src/ariel-os-log/src/lib.rs
+++ b/src/ariel-os-log/src/lib.rs
@@ -99,7 +99,7 @@ pub mod log {
         feature = "std" => {
             pub use std::println;
         }
-        feature = "custom-transport" => {
+        any(feature = "internal-transport-driver", feature = "custom-transport-driver") => {
             pub use crate::transport_println as println;
         }
         not(context = "ariel-os") => {
--- a/src/ariel-os-log/src/transport.rs
+++ b/src/ariel-os-log/src/transport.rs
@@ -69,6 +69,6 @@ pub fn _print(args: core::fmt::Arguments<'_>) {
 macro_rules! transport_println {
     ($($arg:tt)*) => {{
         #[expect(clippy::used_underscore_items, reason = "consistency with std::println")]
-        $crate::custom_transport::_print(format_args!("{}\n", format_args!($($arg)*)));
+        $crate::transport::_print(format_args!("{}\n", format_args!($($arg)*)));
     }};
 }
--- a/src/ariel-os-log/src/transport.rs
+++ b/src/ariel-os-log/src/transport.rs
@@ -1,15 +1,10 @@
-//! Connector to logging transports.
-//!
-//! ## For implementors
-//!
-//! Log transport implementations outside of this crate have to register their `write_bytes` and `flush`
-//! functions using [`register_transport_functions`].
-//!
-//! Those functions should not wait for interrupts when called as they may be ran in a
-//! critical section context. The execution of those functions should not trigger calls to the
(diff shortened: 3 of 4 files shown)


.github/workflows/build-deploy-docs.yml
```diff
@@ -70,12 +70,12 @@ jobs:
...
               --features "
                   ariel-os-sensors/max-sample-min-count-12,
+                  ariel-os-log/custom-transport-driver,
                   bench,
                   ble,
                   coap,
                   core-affinity,
                   csprng,
-                  defmt,
                   dns,
```
nponsard, line 79: Since there's no `defmt` logger for custom logging transports, defmt has to be disabled here, is it an issue ?

src/ariel-os-log/src/lib.rs
```diff
@@ -21,7 +21,6 @@
 #[featurecomb::comb]
 mod _featurecomb {}
 
-#[doc(hidden)]
 #[cfg(feature = "_pluggable-transport")]
```
nponsard: This shows as gated behind `_pluggable-transport` in the rendered doc, should I do something about it ?

ROMemories: Yes, would it make sense to have something like that, since `internal-transport-driver` and `custom-transport-driver` are (at least currently) mutually exclusive:

```suggestion
#[doc(hidden)]
#[cfg(feature = "internal-transport-driver")]
pub mod transport;

#[cfg(feature = "custom-transport-driver")]
pub mod transport;
```

src/ariel-os-log/src/transport.rs
```diff
@@ -1,5 +1,7 @@
 //! Connector to logging transports.
 //!
+//! This is enabled by the `internal-transport-driver` or `custom-transport-driver` features.
+//!
```
ROMemories: That shouldn't be necessary (at least with the other suggestion), and `internal-transport-driver` is an implementation detail:
```suggestion
```

.github/workflows/build-deploy-docs.yml
ROMemories: That shouldn't be a problem for the docs.

ROMemories (commented):
I'll do another pass later to look at the now-public docs content.

src/ariel-os-log/src/transport.rs
ROMemories: I don't think this should show up in the public docs (and it's redundant with the docs of `register_transport_functions`); we don't really what other items could end up in this module later. I think you mean to address Ariel OS developers willing to add a new *internal* logging transport, which is a good idea, but should be done in a regular comment (something like: *internal* logging transports use the same infrastructure as custom logging transports, which are therefore currently mutually exclusive).

ROMemories: Seeing this now in the public docs, therefore aimed at users willing to add a custom transport from their applications, I think this should be renamed:
```suggestion
pub fn register_custom_transport(write_bytes_fn: fn(&[u8]), flush_fn: fn()) {
```

I understand this infrastructure is shared with the *internal* logging transports, but this should be clarified to the attention of Ariel OS developers in a comment (see https://github.com/ariel-os/ariel-os/pull/2272/changes#r3949942093), no need for the function name to cater to both.

ROMemories:
```suggestion
/// Registers a custom transport.
///
/// Registering a custom transport requires providing two functions:
```

ROMemories:
```suggestion
/// ## Important note
```

ROMemories:
```suggestion
/// They should therefore not wait for interrupts and should not trigger calls to the logging facade.
```

ROMemories: I'm not sure this is needed; in which case would users need to be aware of that?

```suggestion
/// - `flush_fn` is used to flush the data in the transport.
```

nponsard:
Is there a point in keeping `_pluggable-transport` as a feature ?

ROMemories:
> Is there a point in keeping `_pluggable-transport` as a feature ?

I don't think so indeed; you can remove it for now, we can always bring it back later if needed.

src/ariel-os-log/src/transport.rs
```diff
@@ -3,13 +3,12 @@
 //! ## For implementors
 //!
 //! Log transport implementations outside of this crate have to register their `write_bytes` and `flush`
-//! functions using [`register_transport_functions`].
-//!
-//! Those functions should not wait for interrupts when called as they may be ran in a
-//! critical section context. The execution of those functions should not trigger calls to the
-//! logging facade as that would create an infinite logging cycle (or a crash in the case of defmt).
+//! functions using [`register_custom_transport`].
```
ROMemories: I still don't think this should show up in the docs, users have no idea what "this crate" refers to for instance.

It may be useful instead to add the following (might need wrapping):

```suggestion
//! This module provides facilities for registering a custom logging transport from within an application.
```

nponsard: Ah I didn't understand that you wanted the whole section deleted. I now removed the entire module documentation section "for implementors".

ROMemories (commented):
Successfully tested the `log` example *without a custom transport* on native on native and nrf52840dk. I think this is in good state now, but I'd prefer waiting until we have an example that can build in CI to exercise the custom transport setup to land this.

ROMemories (approved):
Now that #2250 has been reviewed, I think we can merge this.

PR #2196, opened 2026-07-30
feat(log): add usb_cdc_acm_esp transport

nponsard:
# Description

This adds a custom log transport using the JTAG/Serial USB driver, available on esp32c3,esp32c6,esp32s3.

## Testing

```
laze -C examples/log/ build -b seeedstudio-xiao-esp32c6 -s log run
laze -C examples/log/ build -b seeedstudio-xiao-esp32c6 run
```

Check that `usb_cdc_acm_esp` is correctly selected:
```
laze -C examples/log/ build -b seeedstudio-xiao-esp32c6 -s log info-modules 
laze -C examples/log/ build -b seeedstudio-xiao-esp32c6 info-modules 
```

## Issues/PRs References

- Depends on #2227 
- Depends on #2195 

## Open Questions

`esp-println` uses ROM functions to send logs before `esp-hal` has been initialized. Do we need to go that route ? Current implementation only starts logging after `esp-hal` has been initialized.

## Changelog Entry

Added a custom JTAG/Serial log transport for supported ESP boards.

## Change Checklist

- [ ] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [ ] I have followed the [Coding Conventions][coding-conventions].
- [ ] I have tested and performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2275, opened 2026-09-08
chore: disable GitHub's auto folding of Cargo's lockfile

ROMemories:
# Description

Auto-folding of lockfiles considered harmful. We are of the opinion that lockfiles should not be folded by default, as their contents is not entirely determined by the rest of the repository, but also depends on the registry's content at some point in time, and because changes to the lockfile should be done with intent.

## Testing

A nonsensical change to the lockfile is done initially in this PR, to check that its diff does *not* get folded anymore.

## Change Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,6 @@
+# Lockfiles should not be folded by default.
+Cargo.lock linguist-generated=false
+
 book/src/SUMMARY.md linguist-generated=true
 book/src/boards/*.md linguist-generated=true
 book/src/chips/*.md linguist-generated=true


.gitattributes
```diff
@@ -1,3 +1,6 @@
+# Lockfiles should not be folded by default.
+Cargo.lock linguist-generated=false
```
ROMemories, line 2: [Relevant documentation](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)

Cargo.lock
ROMemories, line 1: The diff of this file is indeed not folded anymore, as expected.

nponsard (approved):
I confirm that the lockfile is unfolded by default. 
I also tested on #2196 that putting a comment on a file doesn't unfold it.

LGTM ! (After reverting the lockfile)

PR #2278, opened 2026-09-08
docs(usb): correct how to enable USB device support

ROMemories:
# Description

This clarifies how to properly enable USB device support, as #2150 mistakenly documented it as requiring the activation of the `usb` *Cargo feature* instead of the laze module of the same name. It also explains when the laze module is (not) available.

## Changelog Entry

No need for an entry as the USB page was only added in #2150.

## Change Checklist


nponsard (approved):
LGTM

PR #2296, opened 2026-09-11
refactor(ethernet): factor out how the MAC address is obtained

ROMemories:
# Description

Following https://github.com/ariel-os/ariel-os/pull/2237#discussion_r3969537565, this attempts to factor out as much as possible how the MAC address is obtained for the (first) Ethernet interface. When doing that, this introduces a (private) `ethernet` laze module that factors out the dependency on `hw/device-identity`, which is required to generate the MAC address.

The scope of this PR might change a bit if we find a way to factor this out a bit more.

## Issues/PRs References

- Documenting the dependency of Ethernet on the device identity is done in #2242.

## Change Checklist

### DIFF
--- a/src/ariel-os-embassy-common/src/lib.rs
+++ b/src/ariel-os-embassy-common/src/lib.rs
@@ -5,6 +5,7 @@
 #![deny(missing_docs)]
 
 pub mod cell;
+pub mod ethernet;
 pub mod gpio;
 
 #[cfg(context = "cortex-m")]
--- a/src/ariel-os-embassy/src/ethernet.rs
+++ b/src/ariel-os-embassy/src/ethernet.rs
@@ -5,3 +5,24 @@ pub(crate) use crate::hal::ethernet::NetworkDevice;
 pub(crate) mod wiznet;
 #[cfg(feature = "ethernet-wiznet")]
 pub(crate) use wiznet::NetworkDevice;
+
+/// Returns a stable MAC address based on the device identity.
+///
+/// # Panics
+///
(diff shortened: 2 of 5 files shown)


src/ariel-os-embassy/src/ethernet.rs
```diff
@@ -5,3 +5,24 @@ pub(crate) use crate::hal::ethernet::NetworkDevice;
...
+
+/// Returns a stable MAC address based on the device identity.
+///
+/// # Panics
+///
+/// Panics in case of transient issue when obtaining the device identity.
+/// This function only used when enforcing that an implementation of `ariel-os-identity` exists, so
+/// this will not unconditionally panic.
+// NOTE: Keep in sync with the implementation for the STM32 Ethernet MAC.
+// NOTE: This cannot be moved to `ariel-os-embassy-common` because a concrete implementation of
```
ROMemories, line 17: I initially thought the whole function could be moved into `ariel-os-embassy-common`, but that unfortunately doesn't seem possible, so only the interface index is factored out.

src/ariel-os-embassy-common/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![deny(missing_docs)]
 
 pub mod cell;
+pub mod ethernet;
```
ROMemories, line 8: The module is not feature-gated, as there isn't an `ethernet` Cargo feature in `ariel-os-embassy-common` yet, and it doesn't seem worth it introducing one right now.

PR #2297, opened 2026-09-11
docs(i2c-scanner): fix scanning on RP MCUs

ROMemories:
# Description

This fixes an issue with the `i2c-scanner`, which is unable to scan properly for I2C targets on RP MCUs (it always fails to find any targets).

## Testing

Tested the `i2c-scanner` example on the following:

- bbc-microbit-v2
- an RP2040 board with sensors
- an espressif-esp32-c6-devkitc-1 with an attached sensor
- st-steval-mkboxpro

## Changelog Entry

The `i2c-scanner` now succeeds in scanning when used on an RP MCU.

## Change Checklist

### DIFF
--- a/examples/i2c-scanner/src/main.rs
+++ b/examples/i2c-scanner/src/main.rs
@@ -21,7 +21,8 @@ async fn i2c_scanner(peripherals: pins::Peripherals) {
     info!("Checking for I2C devices on the bus...");
 
     for addr in 1..=127 {
-        if i2c_bus.write(addr, &[]).await.is_ok() {
+        // The byte is unnecessary, but this otherwise always fails on RP.
+        if i2c_bus.read(addr, &mut [0]).await.is_ok() {
             info!("Found device at address 0x{:x}", addr);
         }
     }
--- a/examples/i2c-scanner/src/main.rs
+++ b/examples/i2c-scanner/src/main.rs
@@ -4,6 +4,7 @@
 mod pins;
 
 use ariel_os::{
+    debug::{ExitCode, exit},
     hal,
     i2c::controller::{Kilohertz, highest_freq_in},
     log::info,
@@ -28,4 +29,5 @@ async fn i2c_scanner(peripherals: pins::Peripherals) {
     }
 
     info!("Done checking. Have a great day!");
+    exit(ExitCode::SUCCESS);
 }


examples/i2c-scanner/src/main.rs
```diff
@@ -21,10 +22,12 @@ async fn i2c_scanner(peripherals: pins::Peripherals) {
...
     for addr in 1..=127 {
-        if i2c_bus.write(addr, &[]).await.is_ok() {
+        // The byte is unnecessary, but this otherwise always fails on RP.
+        if i2c_bus.read(addr, &mut [0]).await.is_ok() {
             info!("Found device at address 0x{:x}", addr);
         }
     }
 
     info!("Done checking. Have a great day!");
+    exit(ExitCode::SUCCESS);
```
ROMemories, line 32: This is technically unrelated, but it bugs me that this doesn't exit when done.

examples/i2c-scanner/src/main.rs
```diff
@@ -21,10 +22,12 @@ async fn i2c_scanner(peripherals: pins::Peripherals) {
     info!("Checking for I2C devices on the bus...");
 
     for addr in 1..=127 {
-        if i2c_bus.write(addr, &[]).await.is_ok() {
+        // The byte is unnecessary, but this otherwise always fails on RP.
+        if i2c_bus.read(addr, &mut [0]).await.is_ok() {
```
ROMemories, line 26: It's unclear to me why this is necessary, it seems to only be needed on RP MCUs (and doesn't seem to cause issues on other MCU families in my testing); I haven't checked yet whether that was an bug in the I2C driver or just how the hardware works, but I've checked that the issue was still there on #1939 so I think we need the fix either way.

ROMemories, line 26: This also changes from a `write` to a `read` as attempting a write seemed unnecessary (not that reads are necessarily side-effects free).

nponsard: I would be more comfortable using a mutable buffer declared earlier to store the read result.

As this is the only i2c example, this may confuse people on how reading from i2c devices works.

nponsard (approved):
I was able to reproduce the issue with an external sensor connected to a rp-pico-w. 
Using this modification fixes it. 

Also successfully tested on stm32u083c-dk with its internal devices and on seeedstudio-xiao-esp32c6, nrf52840dk using and external device.

examples/i2c-scanner/src/main.rs
ROMemories: I've split out the the buffer definition (and tested again on an RP2040-based board).

nponsard (approved):
Looks good to me !

Issue #2308, opened 2026-09-14
OTA: ariel-os-update-storage


kaspar030:
This defines traits and implementations for firmware (or payload) read/write mechanisms. Must allow users to supply their own implementations for custom backends (might be off-chip). 

- Trait implementations for NVM and RAM
  - Write to NVM must be supported
  - Write to RAM must be supported (trevm use case)

bergzand:
On the [dress-up side](https://docs.rs/dress-up/0.1.1/dress_up/operatinghooks/trait.OperatingHooks.html), it requires implementations to read and write slices into a storage location / slot. It also needs a way to get the size and capacity of the component. Here size is the size of what's currently in the component and capacity is the maximum size of what can be written in the component. (Same as Rust's Vecs)

PR #2315, opened 2026-09-16
feat(ota): introduce `ariel-os-update-storage`

ROMemories:
# Description

This is a very WIP PR, part of the ongoing OTA+SUIT sprint, so this can easily be used by dependents. There is a possibility this simply gets closed, but otherwise this should be a valid unit for reviewing.

## Issues/PRs References

- Related to #2308

## Change Checklist


src/ariel-os-update-storage/src/rt.rs
ROMemories, line 1: This should eventually be part of `ariel-os-rt`.

src/ariel-os-update-storage/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+    flash.read(address, buf).await.map_err(|_| Error::Storage)
+}
+
+/// Writes the given payload to an update slot, at the given offset
+/// within that slot.
+// NOTE: `bytes` is a subslice of a SUIT payload.
+pub async fn write(bytes: &[u8], update_slot: UpdateSlot<'_>, offset: u32) -> Result<(), Error> {
+    let len = u32::try_from(bytes.len()).unwrap();
+
+    if capacity(update_slot)? < offset + len {
```
bergzand, line 37: Does it make sense to use a [`checked_add`](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add) here and also catch a potential offset + len overflow, while you're checking for errors anyway.

I don't think the overflow should occur in any sensible application though 🤷‍♂️

PR #2268, opened 2026-09-04
feat(esp32): Add support to ESP32-C2 and ESP32-H2

bugadani:
# Description

Adds support for the chips esp-hal 1.0 supports, but Ariel OS doesn't. To do this, I had to reshuffle `has_wifi_esp` a bit, following how BLE support is already done - the H2 is the first ESP32 chip family without Wi-Fi.

- ESP32-C2 is also known as ESP8684. One of the few chips without USB Serial/JTAG (CDC-ACM).
- ESP32-H2 is a BLE + IEEE802.15.4 SoC with an architecture focusing on low power consumption.

Assisted by Grok 4.6/Cursor.

Requires and old enough compiler (e.g. 1.96.1) because https://github.com/esp-rs/esp-hal/pull/6022 hadn't been backported to esp-hal 1.0.

## Testing

- C2: Ran the hello-world, ble-scanner, tcp-client and blinky examples. Tried running http-client but the chip doesn't have enough RAM for it.
- H2: Ran the hello-world and ble-scanner examples.

## Changelog Entry

The Espressif ESP8684-DevKitC-02 (ESP32-C2) and ESP32-H2-DevKitM-1 boards are now supported.

## Change Checklist

- [x] The [commit history][conventional-commits] will be clean at the latest after applying [autosquash].
- [x] I have followed the [Coding Conventions][coding-conventions].
- [x] I have tested and performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.

[conventional-commits]: https://www.conventionalcommits.org/en/v1.0.0/
[coding-conventions]: https://ariel-os.github.io/ariel-os/dev/docs/book/coding-conventions.html
[autosquash]: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

PR #2325, opened 2026-09-19
fix(laze): move `wifi-esp` back into esp context

kaspar030:
# Description

#2268 made `wifi-esp` availability dependent on `has_wifi_esp`, which is more consistent with how sbd passes through other hardware capabilities, and was necessary as there's now an esp chip without wifi. The `context: esp` was thus dropped.

Due to the way laze resolves modules, this changed evaluation order, making esps with wifi default to usb-ethernet.

laze collects candidates for "providers" starting at the topmost context, then by definition order. With that `context: esp`, `wifi-esp` is the first provider for `network_device` in all contexts that have `esp` as ancestor. Without that, it sorts behind `usb-ethernet`.

## Testing

`laze -Cexamples/http-client build -r context::esp -r usb-ethernet` should not build anything, as all usb capable esp have wifi.

## Issues/PRs References

Missed in #2268 review.

## Change Checklist


kaspar030:
self-ACK, fixing main.

